PR #1, opened 2015-06-12
Plugin template with sensible data

ares:
@inecas I'll continue pushing changes in here, when you're ready feel free to merge so we all build on the same base


app/models/execution_provider.rb
```diff
@@ -0,0 +1,13 @@
+class ExecutionProvider
```
iNecas: RemoteExecutionProvider? Especially since we're not namespacing…

ares:
@iNecas renamed, could you rereview?

iNecas:
Good start, let's continue with changes with other PRs, merging this one

PR #2, opened 2015-06-17
Add TemplateInput models and form

ares:
based on #1


lib/foreman_remote_execution/engine.rb
```diff
@@ -0,0 +1,93 @@
...
+        'template_input.js',
+      )
+    end
+
+    initializer 'foreman_remote_execution.configure_assets', :group => :assets do
+      SETTINGS[:foreman_remote_execution] =
+        {:assets => {:precompile => [
+          'template_input.js',
+        ]}}
+    end
```
ehelms: Do you need both this declaration for generating plugin assets and the declaration on line 14?

lib/foreman_remote_execution/engine.rb
```diff
@@ -0,0 +1,93 @@
...
+      begin
+        # Host::Managed.send(:include, ForemanRemoteExecution::HostExtensions)
+        Template.send(:include, ForemanRemoteExecution::TemplateExtensions)
+      rescue => e
+        Rails.logger.warn "ForemanRemoteExecution: skipping engine hook (#{e})"
+      end
+    end
+
+    rake_tasks do
+      Rake::Task['db:seed'].enhance do
```
ehelms: Consider taking advantage of using `seeds.d` (https://github.com/theforeman/foreman/blob/develop/db/seeds.rb#L53)

ares:
@ehelms thanks for reviews, could you continue in #1? This PR should add just the last commit, but I didn't want to wait until it's merged... Anyway I based this on foreman plugin template so if you think some of these should be changed, it would be a good idea to modify the template as well.

ehelms:
Good idea. I'll start with updates to the template.

lib/foreman_remote_execution/engine.rb
iNecas: I think we can just remove this line until the seeds are needed

app/models/template_input.rb
```diff
@@ -0,0 +1,10 @@
+class TemplateInput < ActiveRecord::Base
+  TYPES = { :user => N_('User input'), :fact => N_('Fact value'), :variable => N_('Smart variable') }
+
+  attr_accessible :name, :required, :input_type, :fact_name, :smart_variable_name, :description, :job_template_id
+
+  belongs_to :job_template
+
+  validates :name, :presence => true
+  validates :input_type, :presence => true
```
jlsherrill: maybe add an inclusion in TYPES validation?

jlsherrill:
I can't seem to create a Job template with this change.  It always flashes to the 'Template Input' section and says that "Name can't be blank".  (regardless of whether i've added an intput or not, if i have added an input it 'adds' another empty input section and complains about that).

Also for some reason when creating a new template the input type is 'stuck' with User Input.  I can't seem to change it.

ares:
argh sorry I forgot to mention we need this https://github.com/theforeman/foreman/pull/2468 in Foreman (not sure if it also solves the second issue)

jlsherrill:
aaaaahhh, yes that seemed to fix the first issue, but not the 'locked' selection on input type.  It seems to only happen for 'new inputs' on the job create & edit screen.  If i create the job with an input of 'user input' type, i can then change it later on.

ares:
I moved the validation from my WIP branch to this commit. Also I've added 4th input type since we have to distinguish between smart class parameters and global parameters (will be merged with smart variables). The first one needs specifying a smart class name. Also added more validations and fixed JS to display correct fields on validation errors.

The issue with select you found was introduced in Foreman nightly (they merged select2 PR) so I had to create another fix in Foreman, please test with https://github.com/theforeman/foreman/pull/2482

Good news is that theforeman/foreman#2468 got in develop already so you just need to update your Foreman.

Could you give it another rounds please?

jlsherrill:
All looks good to me, ACK

ares:
Thanks for the review, merging!

PR #4, opened 2015-06-25
fixes #10890 - Adding invocation models

jlsherrill:


jlsherrill:
Opening for early review.  

Todo:
- [ ] Testing
- [ ] Handle User deletion
- [ ] Handle bookmark deletion
- [ ]  Add targeting resolution

db/migrate/20150624004126_add_invocation.rb
```diff
@@ -0,0 +1,49 @@
+class AddInvocation< ActiveRecord::Migration
+  def change
+    create_table :targetings do |t|
+      t.string :search_query
+      t.references :bookmark
+      t.references :user
+      t.timestamps
+    end
+
+    add_index :targetings, [:bookmark_id, :user_id], :name => "targetings_bookmark_user_ids"
```
iNecas: Shouldn't this be two separate indexes, since we have the foreign keys on them separately?

iNecas:
Looks very nice so far

app/models/targeting.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  belongs_to :user
+  belongs_to :bookmark
+
+  has_many :targeting_hosts, :dependent => :destroy
+  has_many :hosts, :through => :targeting_hosts
+  has_many :template_invocations, :dependent => :destroy
+
+  validates :bookmark_id, :presence => {:message => N_("must be specified if search query is blank")},
+            :if => Proc.new { |a| a.search_query.blank? }
```
ohadlevy: maybe it makes sense to always create a bookmark, and flag it maniacally, vs storing a search_query? this will allow to give a user a history of searches with possible dates? (maybe save the last used time, and how many times it was used)

app/models/targeting.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  validate :bookmark_or_query
```
ohadlevy: a little be overlapping with the above validation? maybe unify it in one custom validation?

app/models/targeting.rb
```diff
@@ -0,0 +1,28 @@
...
+
```
ohadlevy: nitpick: extra whitespace

app/models/targeting.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def resolve_hosts!
+      query = self.bookmark ? self.bookmark.query : self.search_query
```
ohadlevy: nitpick: self is not needed.

ohadlevy: you can also write:
```ruby
query = bookmark.try(:query) || search_query
```

ohadlevy: and in generalize, if you keep the custom query in this model, I would extract it to a separate method.

app/models/targeting.rb
```diff
@@ -0,0 +1,28 @@
...
+      self.hosts = User.as(self.user.login) { Host.authorized('edit_hosts', Host).search_for(query) }
+  end
+
+  private
+
+  def bookmark_or_query
+    if self.bookmark && self.search_query
```
ohadlevy: nitpick, self is not needed.

app/models/template_invocation.rb
```diff
@@ -0,0 +1,9 @@
+class TemplateInvocation < ActiveRecord::Base
+
+  belongs_to :template, :class_name => "JobTemplate", :foreign_key => "template_id"
+  belongs_to :job_invocation, :inverse_of => :template_invocations
+  has_many :template_invocation_input_values
+
+  alias_method :input_values, :template_invocation_input_values
```
ohadlevy: i assume this is just to make it shorter, would we ever need to use the setter as well (e.g. input_values=)?

db/migrate/20150624004126_add_invocation.rb
```diff
@@ -0,0 +1,49 @@
...
+    add_foreign_key :targetings, :bookmarks, :name => "targetings_bookmark_id", :column => 'bookmark_id'
+    add_foreign_key :targetings, :user, :name => "targetings_user_id", :column => 'user_id'
+
+    create_table :targeting_hosts do |t|
```
ohadlevy: nitpick: I find it bad experience to create many operations in one migration, its usually harder to fix issues if it worked half way ( I know its wrapped in a transaction) but having smaller migrations makes imho sense.

app/models/targeting.rb
jlsherrill: Currently these targetings are one time use only. When you need to re-use a targeting it is simply cloned to a new object.  @ares correct me if i misunderstood the design.

I'm not sure what you mean by flagging the bookmark 'maniacally'?

app/models/targeting.rb
jlsherrill: This may be my personal style, but I prefer using self personally as its more explicit.  Using .try() is more elegant in this case, will update

jlsherrill: but i see rubocop is no enforcing no 'redundant self's'  so oh well :)

app/models/targeting.rb
```diff
@@ -0,0 +1,29 @@
+class Targeting < ActiveRecord::Base
+
+  belongs_to :user
+  belongs_to :bookmark
+
+  has_many :targeting_hosts, :dependent => :destroy
+  has_many :hosts, :through => :targeting_hosts
+  has_many :template_invocations, :dependent => :destroy
```
ares: should this be `has_many`? I think `has_one` is more appropriate, if you want to reuse same collection we should reuse bookmark/host collection I think and always copy out the query from it

app/models/targeting.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  validate :bookmark_or_query
+
+  def resolve_hosts!
+    fail _("Cannot resolve hosts without a user") if user.nil?
```
ares: we use raise in Foreman core everywhere, I know Katello prefers fail, I wonder with what we should be consistent here or if we allow using both

anyway whatever we choose I think we should raise/fail with some custom exception

ares: small nitpick, we should try to use single quotes for strings if we don't do any interpolation

app/models/targeting.rb
```diff
@@ -0,0 +1,29 @@
...
+    fail _("Cannot resolve hosts without a bookmark or search query") if bookmark.nil? && search_query.blank?
+    query = bookmark.try(:query) || search_query
```
ares: I'm not sure if we should fallback to bookmark, `self.query` should be copy of `bookmark.query` in time of invocation creation IIRC

app/models/template_invocation.rb
```diff
@@ -0,0 +1,9 @@
+class TemplateInvocation < ActiveRecord::Base
+
+  belongs_to :template, :class_name => "JobTemplate", :foreign_key => "template_id"
+  belongs_to :job_invocation, :inverse_of => :template_invocations
+  has_many :template_invocation_input_values
+
+  alias_attribute :input_values, :template_invocation_input_values
```
ares: just out of curiosity, is there a benefit over specifying custom `:class name` in `has_many` few lines above? I like `input_values` and I think we don't need the longer variant

test/unit/job_invocation_test.rb
```diff
@@ -0,0 +1,44 @@
+require 'test_plugin_helper'
+
+describe JobInvocation do
+
+  let(:job_invocation) { FactoryGirl.build(:job_invocation) }
+  let(:template) { FactoryGirl.build(:job_template) }
+
+  context 'Able to be created' do
+    it { assert job_invocation.save }
```
ares: what do we test here? if it's just validation, you can use assert_valid from Foreman core test_helper.rb which doesn't trigger DB query

test/unit/job_invocation_test.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  context 'Requires targeting' do
+    before { job_invocation.targeting = nil }
+
+    it { refute job_invocation.save }
```
ares: `it { refute job_invocation.valid? }` should be enough, if you wan't to be extra cautious, you can verify `job_invocation.errors` on `:targeting` key

test/unit/job_invocation_test.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  context 'has template invocations' do
+    before do
+      job_invocation.template_invocations << TemplateInvocation.new(:template => template)
+      job_invocation.save!
+    end
+
+    it { refute job_invocation.reload.template_invocations.empty? }
```
ares: this test seems like testing of ActiveRecord functionality, if you insist on this test, could you use FactoryGirl for TemplateInvocation building?

test/unit/job_invocation_test.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  context 'has template input values' do
+    let(:template) { FactoryGirl.create(:job_template, :with_input) }
+    let(:input_value) { FactoryGirl.build(:template_invocation_input_value) }
+
+    before do
+      job_invocation.template_invocations << TemplateInvocation.new(:template => template)
+      job_invocation.save!
```
ares: these two lines seems same as in previous context, I think it would be better to nest it in it and just define input values (or if you disagree, use FactoryGirl for building invocation)

test/unit/targeting_test.rb
```diff
@@ -0,0 +1,57 @@
+require 'test_plugin_helper'
+
+describe Targeting do
+  let(:targeting) { FactoryGirl.build(:targeting) }
+  let(:bookmark) { bookmarks(:one) }
+  let(:host) { FactoryGirl.create(:host) }
+
+  context 'Able to be created with search term' do
+    before { targeting.search_query = "name = foo" }
+    it { assert targeting.save }
```
ares: same comments for the whole file, I'd prefer using `assert_valid` and `.valid?` instead of doing real save

app/models/targeting.rb
jlsherrill: I'm not sure i follow what you mean?  Multiple templates with different inputs will be used, so you would need multiple template_invocations.

app/models/targeting.rb
jlsherrill: I thought we had discussed that a user could provide a query or select a bookmark.  I guess we could store the bookmark's query as part of this (if they selected a bookmark).  is that what you meant?

app/models/template_invocation.rb
jlsherrill: was just being specific, but that would be better, will change

test/unit/job_invocation_test.rb
jlsherrill: I prefer at least one real db save otherwise we're just assuming the schema is valid which it may not be.

app/models/targeting.rb
ares: but each such template_invocation will have it's own targeting I thought

app/models/targeting.rb
ares: yes, the copy was important since bookmark can change before the invocation really happen, that's why fallback to bookmark query might be dangerous

app/models/targeting.rb
jlsherrill: That wasn't my impression.  There would be a single targeting and the appropriate template would be used at invocation time for each system.  Specifying a targeting per template seems a tad complex, as I am unlikely to create my bookmarks for a single remote execution provider type.

app/models/targeting.rb
jlsherrill: so i guess one question that is not answered is 'when does invocation happen?'

If i create a targeting and point it at a bookmark, would i want that query saved right then or in the future (assuming its scheduled to run in the future).  I would think by default I would want it resolved in the future with whatever that bookmark is set to at that time.  So that would just mean saving the query when the host resolution happens.

app/models/targeting.rb
ares: sorry, got confused, has_many is correct :+1:

app/models/concerns/foreman_remote_execution/user_extensions.rb
```diff
@@ -0,0 +1,9 @@
+module ForemanRemoteExecution
+  module UserExtensions
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :targetings, :dependent => :nullify
```
ohadlevy: do we care about user groups too?

app/models/targeting.rb
```diff
@@ -0,0 +1,29 @@
...
+  belongs_to :bookmark
+
+  has_many :targeting_hosts, :dependent => :destroy
+  has_many :hosts, :through => :targeting_hosts
+  has_many :template_invocations, :dependent => :destroy
+
+  validate :bookmark_or_query
+
+  def resolve_hosts!
+    raise _('Cannot resolve hosts without a user') if user.nil?
```
ohadlevy: foreman::exception?

app/models/concerns/foreman_remote_execution/user_extensions.rb
jlsherrill: I assume you mean tying targetings to user groups rather than users?  I might could see that as an enhancement, but its not in the initial design

test/unit/job_invocation_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    before { job_invocation.targeting = nil }
+
+    it { refute_valid job_invocation }
+  end
+
+  context 'has template invocations' do
+    let(:input_value) { FactoryGirl.build(:template_invocation_input_value) }
+
+    before do
+      job_invocation.template_invocations << TemplateInvocation.new(:template => template)
```
ares: nit: use FactoryGirl here?

app/models/targeting.rb
```diff
@@ -0,0 +1,41 @@
...
+  belongs_to :bookmark
+
+  has_many :targeting_hosts, :dependent => :destroy
+  has_many :hosts, :through => :targeting_hosts
+  has_many :template_invocations, :dependent => :destroy
+
+  validates :targeting_type, :presence => true, :inclusion => Targeting::TYPES.keys
+  validates :bookmark, :presence => true, :if => "dynamic?"
+
+  before_create :assign_search_query, :if => :static?
```
ares: So with this, let's say I create invocation with dynamic targeting using bookmark "hostname ~ user*" and schedule it for tomorrow, one hour after me someone modifies the bookmark to "hostname ~ admin*", the invocation will run the command against admin machines right? IIRC we said that it's OK and we should display warning when editing bookmark that there is some targeting that will be affected. Would it make sense to copy the search query even for dynamic so the bookmark can be changed but targeting remains untouched? Then `resolve_hosts!` shouldn't distinguish between static and dynamic and the only difference would be, when is this method called, either during invocation creation or invocation execution.

ohadlevy: I tend to prefer to display a warning at this time, the main reason is that
if someone schedule something for the future and wants to change it, it
becomes a simple change vs changing all pre-generated targets in bulk (and
heavy updates in the db).

maybe we can add a state, like planning or scheduled, vs locked/deployed
where you can edit only while you are in the planning phase and not in the
locked/executed phase?

test/unit/job_invocation_test.rb
jlsherrill: I attempted this for some time and failed and figured my time was better spent elsewhere.  But if you really want this done in factory girl, i can try some more

jlsherrill:
I think there are a few use cases here, lets spell them out to make it clear:

A) I want to execute a command against a bookmark at some point in the future using the current query of the bookmark
B) I want to execute a command against a bookmark at some point in the future using the future query of the bookmark
C) I want to execute a command against a bookmark at some point in the future using the hosts that the bookmark resolves right now

(I'm excluding execute NOW as that is more simple).    Which of these do we want to support?  With the current implementation only A & B are supported.   It sounds like @ares you may be suggesting we support C) ?

app/models/targeting.rb
jlsherrill: I think there are a few use cases here, lets spell them out to make it clear:

A) I want to execute a command against a bookmark at some point in the future using the current query of the bookmark
B) I want to execute a command against a bookmark at some point in the future using the future query of the bookmark
C) I want to execute a command against a bookmark at some point in the future using the hosts that the bookmark resolves right now

(I'm excluding execute NOW as that is more simple).    Which of these do we want to support?  With the current implementation only A & B are supported.   It sounds like @ares you may be suggesting we support C) ?

ares: Yeah, my suggestion was A) and C) I think but I agree that we need a way to change targets of multiple invocations on one place. I remember I raised this few months ago and you persuaded me with same arguments :-) Is it too complex and confusing to add third targeting type for C)? UI will have to have inline help anyway. Also I'm fine with answer it's easy to add later (which I think is true) :-)

jlsherrill: Yeah, i think we should add it later and really once host collections (or whatever it will be called), is in core that may be a better solution for a static list?

ares: hopefully we'll have just one object for that but I don't care too much which one :-) anyway consider this as resolved, thanks

ares:
I think it's ready to merge (pending tests)

jlsherrill:
Updated to add Job Name to the job invocatio model itself, while playing with the controller, it seemed fairly important

ares:
Could you please rebase?

jlsherrill:
rebased!

ares:
:+1: to squash and merge

jlsherrill:
squashed!

PR #5, opened 2015-06-30
Make tests pass

stbenjam:


ares:
I think lock permission might be missing

stbenjam:
@ares Ah, thanks, you're right. Will fix shortly, want to make sure this is all that needs doing (http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/956/ is running).

One thing I'm wondering is if we shouldn't have separate permissions? Or maybe allow restricting a user based on the STI 'type' in the permission filter?

stbenjam:
Oh nevermind, there are already separate permissions for job vs. provisioning template.

stbenjam:
Ah yes, you're right, added.  Disregard the other comment I left, it makes sense now :)

Tests running: http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/957/

ares:
Right, you found out sooner than I was able to reply :-)

iNecas:
Rerunning the tests http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/980/ I will merge that if they are green

iNecas:
@stbenjam some rubocop tests were still failling when merged with master, I've rebased and fixed additional rubocop complains https://github.com/theforeman/foreman_remote_execution/pull/8

stbenjam:
Should've just done this in one PR.  Why were there more? I ran rubocop --auto-gen-config...

iNecas:
It was due to fact, that this PR was not rebased against the latest master, I've opened new PR as I wanted to see if that really fixes the issues on CI

stbenjam:
Yea that's fine, I just meant to close this one

PR #6, opened 2015-07-01
Rendering templates with inputs

ares:
This PR adds InputTemplateRenderer object which can render a template
that contains input("name") macros. It also allows previewing of such
templates which expands macros to placeholders if values are not ready
yet. Currently it supports fact, variable and puppet parameter input
types, user type is ready but can be implemented once we have job
invocation objects in place.

### DIFF
--- a/app/models/input_template_renderer.rb
+++ b/app/models/input_template_renderer.rb
@@ -1,0 +1,39 @@
+class InputTemplateRenderer
+  class UndefinedInput < ::Foreman::Exception
+  end
+
+  include Foreman::Renderer
+
+  attr_accessor :template, :host, :invocation, :error_message
+
+  def initialize(template, host = nil, invocation = nil)
+    @host = host
+    @invocation = invocation
+    @template = template
+  end
+
+  def render
+    render_safe(@template.template, [ :input ])
+  rescue => e
+    self.error_message ||= _('error during rendering: %s') % e.message
+    Rails.logger.debug e.to_s + "\n" + e.backtrace.join("\n")
+    return false
+  end
+
+  def preview
+    @preview = true
+    output = render
+    @preview = false
+    output
+  end
+
+  def input(name)
+    input = @template.template_inputs.where(:name => name.to_s).first || @template.template_inputs.detect { |i| i.name == name.to_s }
+    if input
+      @preview ? input.preview(self) : input.value(self)
+    else
+      self.error_message = _('input macro with name \'%s\' used, but no input with such name defined for this template') % name
+      raise UndefinedInput, "Rendering failed, no input with name #{name} for input macro found"
+    end
+  end
+end
--- a/app/models/template_input.rb
+++ b/app/models/template_input.rb
@@ -1,13 +1,16 @@
 class TemplateInput < ActiveRecord::Base
+  class ValueNotReady < ::Foreman::Exception
+  end
+
   TYPES = { :user => N_('User input'), :fact => N_('Fact value'), :variable => N_('Variable'),
             :puppet_parameter => N_('Puppet parameter') }.with_indifferent_access
 
   attr_accessible :name, :required, :input_type, :fact_name, :variable_name,
                   :puppet_class_name, :puppet_parameter_name, :description, :job_template_id
 
-  belongs_to :job_template
+  belongs_to :template
 
-  validates :name, :presence => true
+  validates :name, :presence => true, :uniqueness => { :scope => 'template_id' }
   validates :input_type, :presence => true, :inclusion => TemplateInput::TYPES.keys
 
   validates :fact_name, :presence => { :if => :fact_template_input? }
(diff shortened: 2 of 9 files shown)


app/models/input_template_renderer.rb
```diff
@@ -0,0 +1,39 @@
...
+  def initialize(template, host = nil, invocation = nil)
+    @host = host
+    @invocation = invocation
+    @template = template
+  end
+
+  def render
+    render_safe(@template.template, [ :input ])
+  rescue => e
+    self.error_message = _('error during rendering: %s') % e.message if error_message.nil?
```
ohadlevy: ||= ?

app/models/input_template_renderer.rb
```diff
@@ -0,0 +1,39 @@
...
+    Rails.logger.debug e.to_s + "\n" + e.backtrace.join("\n")
```
ohadlevy: I assume you use Rails.logger because you want to keep compatibility with older foreman versions?

app/models/input_template_renderer.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def preview
+    @preview = true
+    output = render
+    @preview = false
+    output
+  end
+
+  def input(name)
+    input = @template.template_inputs.where(:name => name.to_s).first || @template.template_inputs.detect { |i| i.name == name.to_s}
```
ohadlevy: can input not be a string?

app/models/input_template_renderer.rb
```diff
@@ -0,0 +1,39 @@
...
+    if input
+      @preview ? input.preview(self) : input.value(self)
+    else
+      self.error_message = _('input macro with name \'%s\' used, but no input with such name defined for this template') % name
+      raise UndefinedInput, "Rendering failed, no input with name #{name} for input macro found"
+    end
+  end
+
+  class UndefinedInput < ::Foreman::Exception
```
ohadlevy: nitpick: not sure if it matters, but maybe best to declare the exception class at the top before using it

app/models/template_input.rb
```diff
@@ -29,4 +29,120 @@ def variable_template_input?
...
+  end
+
+  class InputResolver
+    def initialize(input, renderer)
+      @input = input
+      @renderer = renderer
+    end
+
+    def preview
+      self.ready? ? self.resolved_value : self.preview_value
```
ohadlevy: nitpick: self is not needed.

app/models/template_input.rb
```diff
@@ -29,4 +29,120 @@ def variable_template_input?
...
+    end
+
+    def value
+      self.ready? ? self.resolved_value : raise(ValueNotReady, "Input '#{@input.name}' is not ready for rendering")
```
ohadlevy: nitpick: self is not needed.

app/models/template_input.rb
```diff
@@ -29,4 +29,120 @@ def variable_template_input?
...
+      raise NotImplementedError
+    end
+
+    # should be defined in descendants
+    def resolved_value
+      raise NotImplementedError
+    end
+  end
+
+  class UserInputResolver < InputResolver
```
ohadlevy: one class per file please

app/models/template_input.rb
```diff
@@ -29,4 +29,120 @@ def variable_template_input?
...
+      false
+    end
+
+    def resolved_value
+      # TODO based on job invocation
+      raise StandardError
+    end
+  end
+
+  class FactInputResolver < InputResolver
```
ohadlevy: ditto

app/models/input_template_renderer.rb
ares: Not sure I follow, what logger should be used? We don't have plugin specific logger so we're using Rails global logger. We can't call `logger`, if that's what you mean, since we are in pure Ruby object.

app/models/input_template_renderer.rb
ares: in theory users could write templates like this `<%= snippet(1) %>` or `<%= snippet(:name) %>`, as rubyist I'd expect the later to work

app/models/input_template_renderer.rb
ares: should not matter unless used on class level but I agree, it's better on top, moved

app/models/template_input.rb
ares: I'd like to keep it this way, `InputResolver` is internal object that should not be used out of `TemplateInput` scope. Maybe extracting `InputResolver` with all it's children to extra file could be a compromise but I don't see a benefit right now. (note that in Foreman small classes like exceptions, safe mode jails, validators are also in one file)

app/models/template_input.rb
ares: see previous comment

ares:
:+1: / :-1: ?

iNecas:
ACK

ares:
Thanks Ivan, merging.

PR #9, opened 2015-07-09
Template rendering with preview

ares:
based on https://github.com/theforeman/foreman_remote_execution/pull/6

### DIFF
--- a/app/controllers/job_templates_controller.rb
+++ b/app/controllers/job_templates_controller.rb
@@ -7,7 +7,19 @@ class JobTemplatesController < ::TemplatesController
   end
 
   def auto_complete_job_name
-    @job_names = resource_base.where(['job_name LIKE ?', "%#{params[:search]}%"]).pluck(:job_name)
+    @job_names = resource_base.where(['job_name LIKE ?', "%#{params[:search]}%"]).pluck(:job_name).uniq
+  end
+
+  def preview
+    base = Host.authorized(:view_hosts)
+    host = params[:preview_host_id].present? ? base.find(params[:preview_host_id]) : base.first
+    @template.template = params[:template]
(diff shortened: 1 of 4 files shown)


app/controllers/job_templates_controller.rb
```diff
@@ -7,7 +7,19 @@ def load_vars_from_template
   end
 
   def auto_complete_job_name
-    @job_names = resource_base.where(['job_name LIKE ?', "%#{params[:search]}%"]).pluck(:job_name)
+    @job_names = resource_base.where(['job_name LIKE ?', "%#{params[:search]}%"]).pluck(:job_name).uniq
```
ohadlevy: afair quoting around params is not needed, as where will do it for you.

ares: not sure I follow, I need to have % as part of that string that's injected

ares:
For testing you need to apply https://github.com/theforeman/foreman/pull/2524 to your Foreman instance

ares:
After core PR is merged I'll check compatibility with 1.9 and potentially add some version detection so this is active only in 1.10 if possible.

ares:
It seems compatible with 1.9 (just the preview button is missing) so we can safely merge to master branch.

ares:
Works for me with lastest 1.9-stable and is required for compatibility with 1.10 (otherwise job template form fails because of unknown preview route) so this should get in ASAP. @iNecas could you review?

iNecas:
I've tested this with https://github.com/theforeman/foreman_remote_execution/pull/12 against 1.9, as well as foreman develop (with https://github.com/theforeman/foreman/pull/2591), and worked as expected, one thing that I've noticed is that we don't accept the `@host` variable, therefore we can't use the things like hostname in the template, but I will open separate PR to fix that. ACK

PR #10, opened 2015-07-15
Make API controller test consistent

stbenjam:


stbenjam:
http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/1013/

jlsherrill:
Is this still needed if we start testing against 1.9?

stbenjam:
It's needed eventually, and its compatible with 1.9.

I botched the parameters on the one above, running again here http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/1014/

jlsherrill:
ACK

stbenjam:
We will actually need another PR once we switch to develop, as develop will required `assert :created` on the create tests but at least this made the test consistent in terms of quotes and assertions.

stbenjam:
Thanks @jlsherrill

PR #11, opened 2015-07-16
Initial support for triggering commands on smart proxy

iNecas:
TODO:
- [x] tests

Requires:

* on server side:
  * https://github.com/theforeman/foreman-tasks/pull/125 on server

* on smart proxy side
  * https://github.com/theforeman/smart_proxy_dynflow/pull/1
  * https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/1

There is still work to be done (etc. better edge case handling), but I would preffer
filing sub-tasks for that and finding the minimal viable patch to get in now.


ares:
LGTM so far, what's the plan to keep track of tasks for a particular invocation? locks somehow? I guess we'll need a way to cancel the whole batch...

iNecas:
The bulk actions is the thing that should be usable for that https://github.com/theforeman/foreman-tasks/blob/master/app/lib/actions/bulk_action.rb: this is another task that tracks the sub-tasks, and should allows distributing the cancel events to handle situations like that.

ares:
sounds good

iNecas:
@ares I've rebased the code and connected the actions with job invocation, so now, after you hit 'run job', the task will be created, rendering and passing the commands to the proxy, showing the progress in the meantime.

For testing, you might need to re-save the templates you use, as I needed to use 'Ssh', instead of 'ssh' in the provider name to match it with the 'Ssh' proxy feature

iNecas:
I've merged invocation details updates by @ares and added ability to cancel the tasks, requires latest updates from https://github.com/Dynflow/dynflow/pull/162 and https://github.com/theforeman/foreman-tasks/pull/125

ares:
ACK pending fixed tests, thanks for answers

ares:
:+1: except two nits

iNecas:
dynflow-0.8.3, foreman-tasks-0.7.3, smart_proxy_dynflow-0.0.1, smart_proxy_remote_execution_ssh-0.0.1 released, waiting for https://github.com/iNecas/foreman_remote_execution/pull/2 and some tests for RunHostJob

PR #12, opened 2015-08-04
Fix templates editing

ares:
Two issues I found recently


iNecas:
ACK tested and it proved to fix the issue

ares:
Tests failed because of timeout, not related, merging.

PR #13, opened 2015-08-04
Job invocations creation UI

ares:


ares:
if you prefer select instead of radio buttons, see my branch https://github.com/ares/foreman_remote_execution/tree/job_invocation_ui_selects_rebased

app/assets/javascripts/template_invocation.js
```diff
@@ -0,0 +1,32 @@
+$(function() { job_invocation_form_binds() });
+
+function refresh_execution_form() {
```
iNecas: js functions shouldBeNamedCamelCase

app/assets/javascripts/template_invocation.js
```diff
@@ -0,0 +1,32 @@
...
+    password_caps_lock_hint();
+    form.find('a[rel="popover-modal"]').popover({html: true});
+    form.find('select').select2({allowClear: true});
+    job_invocation_form_binds();
+  });
+}
+
+function job_invocation_form_binds() {
+  $('input.job_template_selector').on('click', function () {
+    parent_fieldset = $(this).closest('fieldset');
```
iNecas: also the variableNames

app/controllers/job_invocations_controller.rb
```diff
@@ -0,0 +1,249 @@
+class JobInvocationsController < ApplicationController
+  def new
+    @composer = JobInvocationComposer.new(JobInvocation.new,
+                                          :host_ids => params[:host_ids],
+                                          :targeting => {
+                                            :targeting_type => Targeting::STATIC_TYPE,
```
iNecas: Shouldn't this param be taken from the params as well, otherwise we would loose the choice on validation error

app/assets/javascripts/template_invocation.js
ares: :sob:

app/assets/javascripts/template_invocation.js
ares: :sob::sob:

app/controllers/job_invocations_controller.rb
ares: I don't think so, these are default attributes for new invocation (well we lose the ability to preselect targeting type, so we should fall back to static if params do not specify), when validation fails, you go through create action and only render `new` view.

iNecas: Ah, right, sry

app/views/job_invocations/_form.html.erb
```diff
@@ -0,0 +1,58 @@
+<%= form_for @composer.job_invocation, :html => { 'data-refresh-url' => refresh_job_invocations_path, :id => 'job_invocation_form' } do |f| %>
+  Assign host:
```
iNecas: Translation wrapping?

app/views/job_invocations/_form.html.erb
```diff
@@ -0,0 +1,58 @@
...
+  <hr >
+
+  <%= selectable_f f, :job_name, @composer.available_job_names %>
+
+  <% @composer.displayed_provider_types.each do |provider_type| %>
+    <fieldset id="provider_<%= provider_type %>" class="provider_form">
+      <% if @composer.needs_provider_type_selection? %>
+        <legend><%= _('Provider') + ': ' + provider_type %></legend>
+      <% end %>
+      <%= f.fields_for 'providers' do |p| %>
```
iNecas: i-am-not-a-big-fan_o_25018 of the 1-2 long vars, although the `f` is quite common, but `pt`, `jt`,`j`: it's getting a bit unreadable.

app/views/job_invocations/show.html.erb
```diff
@@ -0,0 +1,11 @@
+<h2><%= @job_invocation.job_name %></h2>
+
+Template invocations:
```
iNecas: Translation

test/unit/job_invocation_composer.rb
```diff
@@ -0,0 +1,256 @@
...
+RemoteExecutionProvider.register(:ansible, OpenStruct)
+RemoteExecutionProvider.register(:mcollective, OpenStruct)
+
+describe JobInvocationComposer do
+  before do
+    permission = FactoryGirl.create(:permission, :name => 'view_job_templates', :resource_type => 'JobTemplate')
+    filter = FactoryGirl.build(:filter, :permissions => [permission], :search => 'name ~ testing*')
+    filter.save
+    role = FactoryGirl.build(:role)
+    role.filters<< filter
```
iNecas: space ` <<`

app/models/job_invocation_composer.rb
```diff
@@ -0,0 +1,218 @@
...
+    @search_query = targeting_base[:search_query]
+
+    job_invocation.job_name = validate_job_name(job_invocation_base[:job_name]) || available_job_names.first
+    job_invocation.targeting = build_targeting
+
+    @job_template_ids = validate_job_template_ids(job_templates_base.keys.compact)
+  end
+
+  def valid?
+    targeting.valid? & job_invocation.valid? & !template_invocations.map(&:valid?).include?(false)
```
adamruzicka: Are you using `bitwise and` on purpose? Short circuiting is not desired here?

app/models/targeting.rb
```diff
@@ -1,18 +1,20 @@
 class Targeting < ActiveRecord::Base
 
   STATIC_TYPE = 'static_query'
-  DYNAMIC_TYPE = 'dynamic_bookmark'
-  TYPES = {STATIC_TYPE => N_('Static Query'), DYNAMIC_TYPE => N_('Dynamic Bookmark')}
+  DYNAMIC_TYPE = 'dynamic_query'
+  TYPES = {STATIC_TYPE => N_('Static Query'), DYNAMIC_TYPE => N_('Dynamic Query')}
```
adamruzicka: Shouldn't there be spaces as first and last character in the hash?

test/unit/job_invocation_composer.rb
```diff
@@ -0,0 +1,256 @@
...
+          let(:params) { { :job_invocation => providers_params }.with_indifferent_access }
+
+          it 'returns false because available template was selected' do
+            refute composer.preselect_disabled_for_provider('ssh')
+          end
+        end
+      end
+
+      describe '#template_invocations' do
+        let(:ssh_params) { { :job_template_id => testing_job_template_1.id.to_s, :job_templates => { testing_job_template_1.id.to_s => {
```
adamruzicka: Shouldn't `{}` be used only for one line blocks?

app/models/job_invocation_composer.rb
ares: Yes, I want AR to register validation errors on all objects even if targeting (first one) is invalid

app/models/targeting.rb
ares: I'll fix that (even thought it's not my code :-)

app/models/job_invocation_composer.rb
adamruzicka: Ok, makes sense

app/assets/javascripts/template_invocation.js
ares: luckily there's no such convention in Foreman - https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/application.js

could we keep it this way?

app/assets/javascripts/template_invocation.js
ares: same as above, no such convention in Foreman, we have mixed style, could we keep it similar to Ruby in this plugin?

app/views/job_invocations/show.html.erb
ares: This string won't be here, no user would understand "Template invocation" term, the whole page is just debugging output for now.

app/views/job_invocations/_form.html.erb
ares: see if it's more readable now, I don't care too much about this, in both cases I have to find out what the variable means...

app/models/job_invocation_composer.rb
```diff
@@ -0,0 +1,217 @@
+class JobInvocationComposer
+  attr_accessor :params, :job_invocation, :host_ids, :search_query
+  attr_reader :job_template_ids
+
+  # TODO aka initialize_from_params
+  def initialize(job_invocation, params)
```
jlsherrill: since this class is taking in params, would it make more sense for this to live under helpers or controllers?  I can see an argument either way, just wanted to bring it up.

app/models/job_invocation_composer.rb
```diff
@@ -0,0 +1,217 @@
...
+    @job_invocation = job_invocation
+    @params = params
+
+    @host_ids = validate_host_ids(params[:host_ids])
+    @search_query = targeting_base[:search_query]
+
+    job_invocation.job_name = validate_job_name(job_invocation_base[:job_name]) || available_job_names.first
```
jlsherrill: I might be misunderstanding this line but is this saying use the specified job_name if available or just grab the first one the user can read?   I guess this makes sense for the 'new' case, but seems odd if there's an 'update' that goes wrong

app/views/job_invocations/show.html.erb
iNecas: Ah, right. NM then

iNecas:
I've sent the JS hack to open the invocation from host index in new page here https://github.com/ares/foreman_remote_execution/pull/1, I've tested the code extensively (and used it in the demo already), so ACK from my side

app/models/job_invocation_composer.rb
ares: You're right, I haven't decided yet, but that's why I put TODO on line 5, we might need two different initializers, one that you use when you initialize the object from params, one when existing invocation is loaded from DB. Hopefully no persisted invocation will have job_name set to nil so specifically this case should be OK, but I expect other issues.

app/models/job_invocation_composer.rb
ares: I was thinking about creating `app/services` or something similar but it seems overcomplicating for one file... Since I don't think it should be in helpers (usually contains modules) or controllers (these should only take inputs and return outputs from other objects) so remaining option is models which usually holds the logic. Since this object cooperates with other models I think models/ is OK. Also you can view having same params as in controller being lucky coincidence, when we receive parameters from API, API controller should convert those to composer accepted format first. Does it make sense?

iNecas:
ACK

ares:
ready to merge pending tests

PR #15, opened 2015-08-07
Rendering user input when template invocation is provided

iNecas:


test/unit/input_template_renderer_test.rb
```diff
@@ -57,13 +65,17 @@
         end
 
         context 'with invocation specified' do
-          let(:result) do
-            renderer.invocation = nil # TODO
-            renderer.render
+
+          before do
+            @input_value = FactoryGirl.create(:template_invocation_input_value,
```
ares: where is `@input_value` used?

test/unit/input_template_renderer_test.rb
```diff
@@ -57,13 +65,17 @@
...
+                                              :template_invocation =>  template_invocation,
+                                              :template_input      => template.template_inputs.first,
```
ares: please don't indent hashes, this is not puppet \o/ (both lines above have extra spaces)

ares:
Two nits, then good to go.

test/unit/input_template_renderer_test.rb
iNecas: Thank God it's not puppet! :) Both updated, will merge after travis

PR #16, opened 2015-08-18
Include documentation with gem

stbenjam:
From a discussion on the packaging PR about the doc/ directory, it's there so should we include it with the gem?


stbenjam:
Discussion: https://github.com/theforeman/foreman-packaging/pull/735#issuecomment-132095035

ares:
ACK

PR #17, opened 2015-08-19
Fixes #11616 - Adding job invocation api controller

jlsherrill:


jlsherrill:
[test]

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      api :POST, "/job_invocations/", N_("Create a job template")
+      param_group :job_invocation, :as => :create
+      def create
+        @job_invocation = JobInvocation.new(params[:job_template])
+        @composer = JobInvocationApiComposer.new(@job_invocation, params[:job_invocation])
+        process_response @composer.save!
+      end
+
+      def resource_name
```
ares: is this needed? the controller name seems standard

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+          templates = JobTemplate.where(:job_name => params[:job_invocation][:job_name])
+          if templates.pluck(:provider_type).uniq.length != templates.length
+            raise Foreman::Exception, _("Duplicate remote execution providers found for specified Job, please specify a single template_id.")
+          end
+        end
+
+        raise Foreman::Exception, _("No templates associated with specified Job Name") if templates.empty?
+      end
+
+      def resource_class
```
ares: is this needed? `resource_name` should be ok and resource_class_for(resource) should work with it

app/models/job_invocation_api_composer.rb
```diff
@@ -0,0 +1,62 @@
+class JobInvocationApiComposer
+  attr_accessor :params, :job_invocation, :host_ids, :search_query
+
+  delegate :job_name, to: :job_invocation
```
ares: not sure if this is important or not but we use hash rocket style in whole Foreman AFAIK, we don't have to follow it but I just want to make sure that everyone is OK with this

app/models/job_invocation_api_composer.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+    input_values.compact
+  end
+
+  def template_input(template, name)
+    template.template_inputs.find_by_name(name)
+  end
+
+  def valid?
+    targeting.valid? & job_invocation.valid? & !template_invocations.map(&:valid?).include?(false)
```
ares: `!template_invocations.map(&:valid?).include?(false)` is equal to `template_invocations.all?(&:valid?)`

app/models/job_invocation_api_composer.rb
```diff
@@ -0,0 +1,62 @@
...
+    self.job_invocation.targeting = build_targeting
+    self.job_invocation.template_invocations = [build_template_invocation]
+  end
+
+  delegate :save!, to: :job_invocation
+
+  def build_targeting
+    targeting ||= job_invocation.targeting || job_invocation.build_targeting
+    targeting.targeting_type = params[:targeting_type] if params[:targeting_type]
+    targeting.bookmark = Bookmark.find(params[:bookmark_id]) if params[:bookmark_id]
```
ares: should use `available_bookmarks`?

app/models/job_invocation_api_composer.rb
```diff
@@ -0,0 +1,62 @@
...
+  def find_job_template
+    templates = JobTemplate.authorized(:view_job_templates).where(:job_name => job_name)
+    template = templates.find(params[:template_id]) if params[:template_id]
+    template = templates.first if templates.count == 1
+    template
+  end
+
+  def build_input_values(invocation)
+    input_values = params.fetch(:inputs, []).map do |input_hash|
+      if input = template_input(invocation.template, input_hash[:name])
```
ares: wrap condition in brackets if you want to assign the value instead of ==

app/models/template_input.rb
```diff
@@ -91,11 +91,12 @@ def resolved_value
 
   class UserInputResolver < InputResolver
     def ready?
-      !input_value.nil?
+      @renderer.invocation
     end
 
     def resolved_value
-      input_value.value
+      raise(ValueNotReady, _("Required input not specified")) if @input.required? && input_value.try(:value).blank?
```
ares: this check should be part of `ready?` I think, also we should probably accept blank values, only input_value or inpute_value.value being nil should result in "not ready"

app/models/template_invocation.rb
```diff
@@ -1,13 +1,24 @@
...
 
   def to_action_input
     { :id => id, :name => template.name }
   end
+
+  private
+
+  def provides_required_input_values
+    required_inputs = self.template.template_inputs.where(:required => true)
+    unless (required_inputs - self.input_values.map(&:template_input)).empty?
```
ares: N+1 query, probably best to just compare ids
```ruby
required_inputs = self.template.template_inputs.where(:required => true).pluck(:id)
unless (required_inputs & self.input_values.pluck(:template_input_id)).empty?
```

app/views/api/v2/job_invocations/main.json.rabl
```diff
@@ -0,0 +1,3 @@
+object @job_invocation
+
+extends "api/v2/job_invocations/base"
```
ares: missing EOL

app/views/api/v2/job_invocations/show.json.rabl
```diff
@@ -0,0 +1,14 @@
...
+node :targeting do
+  attributes :bookmark_id, :search_query, :targeting_type, :user_id
+end
+
+node :template_invocations do
+  attributes :template_id
+  child :input_values do
+    attributes :input_id, :value
+  end
+end
```
ares: missing EOL

app/views/api/v2/job_templates/base.json.rabl
```diff
@@ -1,3 +1,3 @@
 object @job_template
 
-attributes :id, :name, :job_name, :provider_type
+attributes :id, :name, :job_name, :provider_type
```
ares: missing EOL

config/routes.rb
```diff
@@ -22,6 +22,9 @@
 
   namespace :api, :defaults => {:format => 'json'} do
     scope "(:apiv)", :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2, :default => true) do
+      resources :job_invocations, :except => [:new, :edit, :update, :destroy] do
+      end
```
ares: empty block can be omitted

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    class JobInvocationsControllerTest < ActionController::TestCase
+      setup do
+        @invocation = FactoryGirl.create(:job_invocation, :with_template)
+        @template = FactoryGirl.create(:job_template, :with_input)
+      end
+
+      test "should get index" do
+        get :index
+        invocations = ActiveSupport::JSON.decode(@response.body)
+        assert !invocations.empty?, "Should response with invocation"
```
ares: `refute_empty invocations, 'Should response with invocation'` should work

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -0,0 +1,45 @@
...
+        assert_response :success
+      end
+
+      test "should get invocation detail" do
+        get :show,  :id => @invocation.id
+        assert_response :success
+        template = ActiveSupport::JSON.decode(@response.body)
+        assert !template.empty?
```
ares: refute_empty

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -0,0 +1,45 @@
+require 'test_plugin_helper'
```
ares: we use rspec syntax in unit tests and this one in functional so far (job_templates_controller_test.rb), I think we should start using just one

app/models/job_invocation_api_composer.rb
```diff
@@ -0,0 +1,62 @@
...
+    self.job_invocation.template_invocations = [build_template_invocation]
+  end
+
+  delegate :save!, to: :job_invocation
+
+  def build_targeting
+    targeting ||= job_invocation.targeting || job_invocation.build_targeting
+    targeting.targeting_type = params[:targeting_type] if params[:targeting_type]
+    targeting.bookmark = Bookmark.find(params[:bookmark_id]) if params[:bookmark_id]
+    targeting.search_query = params[:search_query]
```
ares: what happens if search_query and bookmark_id were passed at the same time?

app/models/job_invocation_api_composer.rb
```diff
@@ -0,0 +1,62 @@
...
+    targeting
```
ares: where does the user gets saved? it's later required for resolving hosts

app/models/job_invocation_api_composer.rb
jlsherrill: I think rubocop auto-corrected this when it created them.  Will update

test/functional/api/v2/job_invocations_controller_test.rb
jlsherrill: I'm not sure i fully understand what you mean by this comment?  This was modeled after the existing job_templates_controller api controller

ares: I meant that syntax of this test file is not in rspec style but that's not a big deal. Let's ignore it for now.

app/models/template_input.rb
jlsherrill: @ares - i can move that check to ready.  

I would think an exception should be thrown if .value is called and a required value is not present.  But i do agree that .ready? should return false.

When you say 'accept blank values', do you mean for required inputs (treating those as provided?)

ares: well, raise can remain in here as well but resolved_value should never be called if ready is false (I guess this is defensive check)

by accept blank value I mean that `input_value.try(:value).blank?` is also true for empty string which can be valid value I think, so `input_value.nil?` should be used instead?

jlsherrill: I don't think an empty string is valid for a required value?

jlsherrill: Also, i'm curious how a user would know what the issue is with a specific input if an exception is never raised.  They'd just see some message about an 'input not being ready'.

ares: So imagine "yum custom options" that could contains things like -y or --nogpgcheck but does not have to contain anything hence empty string. If I understand your change correctly, if I want to rely on input to exist, the input value must exist and valid value is also empty string. Otherwise I'd not be able to write template like `yum install sshd <%= input('custom_args') %>`, if it's not required, the input would never be ready

ares: The exception would be raised when I try to use resolved value for input which is not ready https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/template_input.rb#L74

jlsherrill: I don't think those options would be marked as required then?

jlsherrill: I was under the impression required meant "it requires some value"

ares: ok, I have just probably a different view on what's considered some value, e.g. when we have default value set to '-y' and you'd want to disable it I still see emptry string as valid value...

nevermind :-) we'll see if it causes any issue, definitely not a blocker

jlsherrill: So i think in that instance empty string is completely valid, and you just wouldn't mark the input as required.

ares: so when I don't mark it as required and it has default value, what will be result of that input macro?

app/views/api/v2/job_invocations/index.json.rabl
```diff
@@ -0,0 +1,3 @@
+collection @job_templates
```
ares: `@job_invocations`, unfortunately functional tests don't cover this

app/views/api/v2/job_invocations/show.json.rabl
```diff
@@ -0,0 +1,14 @@
...
+extends "api/v2/job_invocations/main"
+
+node :targeting do
+  attributes :bookmark_id, :search_query, :targeting_type, :user_id
+end
+
+node :template_invocations do
+  attributes :template_id
+  child :input_values do
+    attributes :input_id, :value
```
ares: s/input_id/template_input_id/

app/views/api/v2/job_invocations/show.json.rabl
```diff
@@ -0,0 +1,14 @@
+object @job_invocation
+
+extends "api/v2/job_invocations/main"
+
+node :targeting do
+  attributes :bookmark_id, :search_query, :targeting_type, :user_id
+end
+
+node :template_invocations do
```
ares: child, otherwise I'm getting some weird output

app/views/api/v2/job_invocations/show.json.rabl
```diff
@@ -0,0 +1,14 @@
+object @job_invocation
+
+extends "api/v2/job_invocations/main"
+
+node :targeting do
```
ares: child

app/models/template_invocation.rb
```diff
@@ -1,13 +1,24 @@
...
 
   def to_action_input
     { :id => id, :name => template.name }
   end
+
+  private
+
+  def provides_required_input_values
+    required_input_ids = self.template.template_inputs.where(:required => true).pluck(:id)
+    unless (required_input_ids - self.input_values.pluck(:template_input_id)).empty?
```
ares: this does not work with in-memory objects that are not persisted yet, when I do a new job_invocation and specify values, template invocations are invalid since no input value is found

if you do `.map(&:template_input_id)` it works

app/views/api/v2/job_invocations/base.json.rabl
```diff
@@ -0,0 +1,3 @@
+object @job_invocation
+
+attributes :id, :job_name, :targeting_id
```
ares: we should display some overall status

app/models/template_invocation.rb
jlsherrill: heh, trading testing performance for real time performance?

In this case it really won't matter, so i don't mind changing it.

app/views/api/v2/job_invocations/base.json.rabl
jlsherrill: added below

app/models/template_invocation.rb
ares: it's not only about testing, it didn't work for new records that were not saved yet since

```ruby
self.input_values # => return all including in-memory objects
self.input_values.pluck(...) # => returns only persisted objects
```

which means in case of new record `required_input_ids - always_empty_array` which is not empty if there's any required input regardless of provided input values

jlsherrill: Ahhh right, good call!

iNecas:
rebase wanted

jlsherrill:
rebased!

jlsherrill:
rebased again

ares:
[test]

ares:
Tested again, works nicely, ACK => merge at will

ares:
We agreed with @iNecas that there's no need to wait for author to merge his own PR after it's ACKed, so I'm merging this one. Thanks @jlsherrill!

PR #18, opened 2015-08-19
Refs #10753 - Support for setting future time of execution

adamruzicka:
requires theforeman/foreman-tasks#131

### DIFF
--- a/app/assets/javascripts/template_invocation.js
+++ b/app/assets/javascripts/template_invocation.js
@@ -31,6 +31,11 @@ function job_invocation_form_binds() {
     $('#job_template_' + $(this).val()).show();
   });
 
+  $('input.trigger_mode_selector').on('click', function () {
+     $("#trigger_mode_future").hide();
+     $('#trigger_mode_' + $(this).val()).show();
+   });
+
   $('select#job_invocation_job_name').on('change', refresh_execution_form);
 
   $('button#refresh_execution_form').on('click', refresh_execution_form);
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -11,14 +11,18 @@ module RemoteExecutionHelper
     options = { :class => 'statistics-pie small', :expandable => true, :border => 0, :show_title => true }
 
     if (bulk_task = invocation.last_task)
+      failed_tasks = bulk_task.sub_tasks.select { |sub_task| %w(warning error).include? sub_task.result }
+      cancelled_tasks, failed_tasks = failed_tasks.partition { |task| task_cancelled? task }
       success = bulk_task.output['success_count'] || 0
-      failed = bulk_task.output['failed_count'] || 0
-      pending = bulk_task.output['pending_count'] || bulk_task.sub_tasks.count
+      cancelled = cancelled_tasks.length
+      failed = failed_tasks.length
+      pending = (bulk_task.output['pending_count'] || bulk_task.sub_tasks.count)
 
       flot_pie_chart('status', job_invocation_status(invocation),
                      [{:label => _('Success'), :data => success, :color => '#5CB85C'},
                       {:label => _('Failed'), :data => failed, :color => '#D9534F'},
-                      {:label => _('Pending'), :data => pending, :color => '#DEDEDE'}],
+                      {:label => _('Pending'), :data => pending, :color => '#DEDEDE'},
+                      {:label => _('Cancelled'), :data => cancelled, :color => '#B7312D'}],
                      options)
     else
       content_tag(:h4, job_invocation_status(invocation))
@@ -28,11 +32,19 @@ module RemoteExecutionHelper
   def job_invocation_status(invocation)
     if invocation.last_task.blank?
       _('Job not started yet 0%')
+    elsif invocation.last_task.state == 'scheduled'
+      _('Job set to execute at %s') % invocation.last_task.start_at
+    elsif invocation.last_task.state == 'stopped' && invocation.last_task.result == 'error'
+      invocation.last_task.execution_plan.errors.map(&:message).join("\n")
     else
       label = invocation.last_task.pending ? _('Running') : _('Finished')
       label + ' ' + (invocation.last_task.progress * 100).to_i.to_s + '%'
     end
 
+  end
+
+  def task_cancelled?(task)
+    task.execution_plan.errors.map(&:exception).any? { |exception| exception.class == ::ForemanTasks::Task::TaskCancelledException }
   end
 
   def host_counter(label, count)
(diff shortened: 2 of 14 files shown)


app/assets/javascripts/template_invocation.js
```diff
@@ -26,6 +26,12 @@ function job_invocation_form_binds() {
     $('#job_template_' + $(this).val()).show();
   });
 
+  $('input.execution_type_selector').on('click', function () {
+     parent = $(this).closest('div');
```
ohadlevy: consider using an id / class as it would not break on html structure changes.

app/models/job_invocation.rb
```diff
@@ -15,7 +15,40 @@ class JobInvocation < ActiveRecord::Base
...
     { :id => id, :name => job_name }
   end
+
+  def execution_type
+    @execution_type || :immediate
+  end
+
+  def execution_type=(value)
+    return @execution_type if @execution_type
+    if [:immediate, :future].map(&:to_s).include?(value)
```
ohadlevy: why not %w(immediate future) instead of symbol to string conversion

app/models/job_invocation.rb
```diff
@@ -15,7 +15,40 @@ class JobInvocation < ActiveRecord::Base
...
+
+  def start_at_parsed
+    @start_at.present? && Time.strptime(@start_at, time_format)
+  end
+
+  def start_at
+    @start_at ||= Time.now.strftime(time_format)
+  end
+
+  def start_at=(value)
```
ohadlevy: attr_writer :start_at ?

ares:
Tests are green but Rubocop is unhappy

```
222 tests, 228 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Functional Tests, Unit Tests to /var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.1/foreman/coverage. 14345 / 18290 LOC (78.43%) covered.
Running RuboCop...
Inspecting 42 files
...C..............C.......C...............

Offenses:

/var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.1/plugin/app/lib/actions/middleware/bind_job_invocation.rb:29:4: C: Final newline missing.
end
   
/var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.1/plugin/app/models/job_invocation.rb:31:5: C: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
    if %w(immediate future).include?(value)
    ^^
/var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.1/plugin/app/helpers/remote_execution_helper.rb:32:36: C: Favor format over String#%.
      _('Job set to execute at %s' % invocation.last_task.start_at)
                                   ^
```

iNecas:
I did not get though the whole PR yet, but some notes from testing:

* it will need to rebase and fix against the latest features (there are some conflicts and something new to count on)
* the basic functionality worked well, one thing that could be improved is better handling of job cancellation: I would expect the task would be displayed as failed, with details on the reason.

app/models/job_invocation.rb
```diff
@@ -15,7 +15,35 @@ class JobInvocation < ActiveRecord::Base
...
   scoped_search :on => [:job_name], :complete_value => true
 
+  attr_accessor :start_before
+  attr_writer :start_at
+
   def to_action_input
     { :id => id, :name => job_name }
   end
+
+  def execution_type
```
iNecas: I know I've came initially with this naming, but I'm still not sure that's the best name for the value we are collecting, Anyone suggestions on better name?

ares: `%w(execution_mode invocation_time timing trigger_type trigger_mode).sample`

app/views/job_invocations/_tab_hosts.html.erb
```diff
@@ -10,24 +10,24 @@
...
-          <% if (host = task.locks.where(:resource_type => 'Host::Managed').first.try(:resource)) %>
-            <td><%= link_to_if_authorized host.name, hash_for_host_path(host).merge(:auth_object => host, :permission => :view_hosts) %></td>
-          <% else %>
-            <td><%= _('host not linked') %></td>
-          <% end %>
+          <% task = job_invocation.last_task.sub_tasks.select { |task| task.locks.where(:resource_type => 'Host::Managed', :resource_id => host.id).any? }.first %>
+          <% template_invocation = task.locks.where(:resource_type => 'TemplateInvocation').first.try(:resource) unless task.nil? %>
+          <td><%= link_to_if_authorized host.name, hash_for_host_path(host).merge(:auth_object => host, :permission => :view_hosts) %></td>
 
           <td><%= template_invocation.nil? ? _('N/A') : _(RemoteExecutionProvider.provider_for(template_invocation.template.provider_type)) %></td>
```
iNecas: I guess we could extract this into `remote_execution_provider_for_task(task)` helper, the `template_invocation` itself is not used anywhere else

app/helpers/remote_execution_helper.rb
```diff
@@ -28,6 +28,8 @@ def job_invocation_chart(invocation)
   def job_invocation_status(invocation)
     if invocation.last_task.blank?
       _('Job not started yet 0%')
+    elsif invocation.last_task.state == 'scheduled'
+      _('Job set to execute at %s') + invocation.last_task.start_at
```
iNecas: I'm getting `can't convert ActiveSupport::TimeWithZone into String` when invoking the job scheduled to future

iNecas: Probably `s/+/%`

iNecas:
Still getting 'Finished 100%`, when cancelling the task: I know it's logically correct, but still, something like 'task cancelled', would be more appropriate

adamruzicka:
For now the tasks are divided to failed and cancelled by matching their error messages against `[Cc]ancel` since I can't figure out any other way how to distinguish them apart from adding another step state to Dynflow

iNecas:
What about raising specific exception on the cancallation, and chedking that type of error?

app/models/job_invocation.rb
```diff
@@ -42,4 +45,37 @@ def failed_hosts
...
+
+  def total_hosts_count
+    if targeting.resolved_at.nil?
+      _('N/A')
+    else
+      targeting.hosts.count
+    end
+  end
+
+  def execution_type
```
ares: this wasn't resolved

@inecas:
> I know I've came initially with this naming, but I'm still not sure that's the best name for the value we are collecting, Anyone suggestions on better name?

`%w(execution_mode invocation_time timing trigger_type trigger_mode).sample`

adamruzicka: I'd go with `trigger_mode`

iNecas: +1 for `trigger_mode`

iNecas:
I've tested the code and worked as expected, all the concerns fixed: still it would be nice to add more unit tests for the new code.

app/models/job_invocation.rb
```diff
@@ -42,4 +45,48 @@ def failed_hosts
...
+  def trigger_mode
+    @trigger_mode || :immediate
+  end
+
+  def trigger_mode=(value)
+    return @trigger_mode if @trigger_mode
+    if %w(immediate future).include?(value)
+      @trigger_mode = value.to_sym
+    else
+      raise ::Foreman::Exception, _("Job Invocation trigger mode must be one of 'immediate', 'future'")
```
iNecas: Would be better to have the possible values listed in some constant and use the values both in condition and exception

iNecas: Also, now it raises the execption when opening new job invocation, as it tries to set the value to `nil`: we need to ensure default value still

iNecas:
Last two comments: @adamruzicka you can squash together with addressing those

PR #19, opened 2015-08-21
Extend ForemanTasks after its loaded

stbenjam:
The plugin doesn't load in production, because ForemanTasks gets loaded after RemoteExecution.

Not sure if this is the right place, but it does solve them problem for me.  @ehelms also suggested we could try the finisher hook.  Any thoughts?

```
undefined method `has_many' for Rake::Task:Class (NoMethodError)
  /opt/rh/ruby193/root/usr/share/gems/gems/foreman_remote_execution-0.0.2/app/models/concerns/foreman_remote_execution/foreman_tasks_task_extensions.rb:6:in `block in <module:ForemanTasksTaskExtensions>'
  /opt/rh/ruby193/root/usr/share/gems/gems/activesupport-3.2.8/lib/active_support/concern.rb:119:in `class_eval'
```

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -21,6 +21,7 @@ module ForemanRemoteExecution
     end
 
     initializer "foreman_remote_execution.require_dynflow", :before => "foreman_tasks.initialize_dynflow" do |app|
+      ForemanTasks::Task.send(:include, ForemanRemoteExecution::ForemanTasksTaskExtensions)
       ForemanTasks.dynflow.require!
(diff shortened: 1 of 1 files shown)


iNecas:
This is strange, the to_prepare block should be called after the initialization process (but I will rather check that), the reason why the mixing should happen in the to_prepare phase is it gets called even in development while life-reloading the code, so this variant would probably break once the code gets reloaded in development, as the reloaded classes would not get the modules included

iNecas:
The #20 should fix the issue: I was able to reproduce the problem, but only when the katello was presented, the #20 seemed to work both in Katello and non-Katello environments

stbenjam:
Closed in favor of #20

PR #20, opened 2015-08-24
Fix random loading issues when extending ForemanTasks::Task model

iNecas:
We need to explicitly force to load the Task model due to Rails loader having
issues with resolving it to Rake::Task otherwise

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -107,6 +107,10 @@ module ForemanRemoteExecution
       end
       Bookmark.send(:include, ForemanRemoteExecution::BookmarkExtensions)
       HostsHelper.send(:include, ForemanRemoteExecution::HostsHelperExtensions)
+
+      # We need to explicitly force to load the Task model due to Rails loader
+      # having issues with resolving it to Rake::Task otherwise
+      require_dependency 'foreman_tasks/task'
       ForemanTasks::Task.send(:include, ForemanRemoteExecution::ForemanTasksTaskExtensions)
     end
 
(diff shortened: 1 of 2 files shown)


lib/foreman_remote_execution/engine.rb
```diff
@@ -107,6 +107,10 @@ class Engine < ::Rails::Engine
       end
       Bookmark.send(:include, ForemanRemoteExecution::BookmarkExtensions)
       HostsHelper.send(:include, ForemanRemoteExecution::HostsHelperExtensions)
+
+      # We need to explicitly force to load the Task model due to Rails loader
+      # having issues with resolving it to Rake::Task otherwise
+      require_dependency 'foreman_tasks/task'
```
ohadlevy: this feels like a hack? shouldn't rails initialization phase allow you to specify it post rails bootup?

iNecas: The problem is we need this to run code exactly in the to_prepare section, for the Rails reloading to work both in development and production http://apidock.com/rails/Rails/Configuration/to_prepare

iNecas: Other initializers run only once, so we would need to sacrifice the ForemanTasks::Task reloading, which is something I would like to avoid

stbenjam:
Seems like its the proper use case of the method http://apidock.com/rails/ActiveSupport/Dependencies/Loadable/require_dependency.

What is Katello doing that's interfering? I almost only work with Katello loaded.

@iNecas Can you also bump the gem version?

iNecas:
@stbenjam it might be just a particular change of load order: probably might happen with other plugins as well

iNecas:
the version updated

stbenjam:
APJ

iNecas:
v0.0.3 just released

stbenjam:
RPM - https://github.com/theforeman/foreman-packaging/pull/781

Will do DEB later, have to address the assets issue later today.

PR #22, opened 2015-08-26
Remove .scss from stylesheets so they precompile

stbenjam:


iNecas:
Tested by applying the patch, precompiling, removing ruby193-rubygem-sass, and loading the UI: worked well (din't work without the path).ACK

PR #25, opened 2015-08-27
fixes #12125 - use git ls-files in gemspec

### LINKED ISSUE #24: v0.0.3 is a large .gem
Note that one of your releases, v0.0.3 is very large (7MB): https://rubygems.org/gems/foreman_remote_execution/versions

    7.3M    doc/tmp/images/plantuml/

Perhaps use `git ls-files` in your gemspec so that only version tracked files are included in the release instead of whatever's in the working directory, as this dir is gitignored.

stbenjam:
fixes #24

`git ls-files` seems to be somewhat controversial based on Googling, but it works fine and ensures only committed files get released.

### DIFF
--- a/foreman_remote_execution.gemspec
+++ b/foreman_remote_execution.gemspec
@@ -14,9 +14,9 @@ Gem::Specification.new do |s|
   s.description = 'A plugin bringing remote execution to the Foreman, completing the config ' +
                   'management functionality with remote management functionality.'
 
-  s.files =            Dir['{app,config,doc,db,lib,locale}/**/*', 'LICENSE', 'Rakefile', 'README.md']
-  s.test_files =       Dir['test/**/*']
-  s.extra_rdoc_files = Dir['doc/**/*', 'README*', 'LICENSE']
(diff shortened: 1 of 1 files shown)


foreman_remote_execution.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.description = 'A plugin bringing remote execution to the Foreman, completing the config ' +
                   'management functionality with remote management functionality.'
 
-  s.files =            Dir['{app,config,doc,db,lib,locale}/**/*', 'LICENSE', 'Rakefile', 'README.md']
+  s.files =            `git ls-files`.split("\n")
   s.test_files =       Dir['test/**/*']
```
iNecas: You need to do the same for `test_files` and `extra_rdoc_files`, I've tested the code and it still doesn't fix the issue with extra files in `doc`

iNecas: @stbenjam ping

iNecas:
Thanks @stbenjam !

PR #26, opened 2015-08-27
fixes #11469 - Ship some default templates

stbenjam:
Provides:
 - Package Action
 - Puppet Run Once
 - Run Command
 - Service Action

Templates are seeded by YAML, essentially  a cleaned up `template.to_json(:includes => :template_inputs)`.  We could also go the Foreman route and put the metadata in an ERB comment at the top.

I add a new `select_options` input type, I wanted this for the package action where a user could select from install/update/etc. 

Here's a gif if you just want to see:

rex

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -37,6 +37,25 @@ class JobTemplate < ::Template
   end
   self.table_name = 'templates'
 
+  # Import a template ERB, with metadata in the first YAML comment
+  def self.import(template, options = {})
+    metadata = parse_metadata(template)
+    return if metadata.blank? || metadata.delete('kind') != 'job_template' || self.find_by_name(metadata['name'])
+
+    inputs = metadata.delete('template_inputs')
+
+    # This awkward dance is because you can't instantiate a new template with :locked => true
+    template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')))
+    template.update_attributes(options)
+    template.assign_taxonomies
+
+    inputs.each do |input|
+      template.template_inputs << TemplateInput.create(input)
+    end
+
+    template
+  end
+
   # Override method in Taxonomix as Template is not used attached to a Host,
   # and matching a Host does not prevent removing a template from its taxonomy.
   def used_taxonomy_ids(type)
@@ -51,7 +70,19 @@ class JobTemplate < ::Template
     dup
   end
 
+  def assign_taxonomies
+    if default
+      organizations << Organization.all if SETTINGS[:organizations_enabled]
+      locations << Location.all if SETTINGS[:locations_enabled]
+    end
+  end
+
   private
+
+  def self.parse_metadata(template)
+    match = template.match(/<%\#(.+?).-?%>/m)
+    match.nil? ? {} : YAML.load(match[1])
+  end
 
   # we can't use standard validator, .provider_names output can change but the validator does not reflect it
   def provider_type_whitelist
--- a/app/models/template_input.rb
+++ b/app/models/template_input.rb
@@ -6,7 +6,8 @@ class TemplateInput < ActiveRecord::Base
             :puppet_parameter => N_('Puppet parameter') }.with_indifferent_access
 
   attr_accessible :name, :required, :input_type, :fact_name, :variable_name,
-                  :puppet_class_name, :puppet_parameter_name, :description, :job_template_id
+                  :puppet_class_name, :puppet_parameter_name, :description, :job_template_id,
+                  :options
 
   belongs_to :template
   has_many :template_invocation_input_values, :dependent => :destroy
@@ -40,6 +41,10 @@ class TemplateInput < ActiveRecord::Base
 
   def value(renderer)
     get_resolver(renderer).value
+  end
+
+  def options_array
+    self.options.blank? ? [] : self.options.split(/\r?\n/).map(&:strip)
   end
 
   private
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -49,7 +49,11 @@
                   <%= job_template_fields.fields_for :input_values do |input_values_fields| %>
                     <% job_template.template_inputs.where(:input_type => 'user').each do |input| %>
                       <%= input_values_fields.fields_for input.id.to_s, @composer.template_invocation_input_value_for(input) do |input_fields| %>
-                        <%= text_f input_fields, :value, :label => input.name, :help_inline => input.description, :required => input.required %>
+                        <% unless input.options.blank? %>
+                          <%= selectable_f input_fields, :value, input.options_array, {:include_blank => !input.required }, :require => input.required, :label => input.name, :help_inline => input.description %>
+                        <% else %>
+                          <%= textarea_f input_fields, :value, :label => input.name, :help_inline => input.description, :required => input.required, :rows => 2 %>
+                        <% end %>
                       <% end %>
                     <% end %>
                   <% end %>
--- a/app/views/templates/package_action.erb
+++ b/app/views/templates/package_action.erb
@@ -1,0 +1,52 @@
+<%#
+kind: job_template
+name: Package Action - SSH Default
+job_name: Package Action
+provider_type: Ssh
+template_inputs:
+- name: pre_script
+  description: A script to run prior to the package action
+  input_type: user
+  required: false
+- name: action
+  description: 'The package action: install, update, or remove'
+  input_type: user
+  required: true
+  options: "install\nupdate\nremove"
+- name: package
+  description: The name of the package, if any
+  input_type: user
+  required: false
+- name: post_script
+  description: A script to run after the package action
+  input_type: user
+  required: false
+%>
+
+die() {
+  echo "${1}, exiting..."
+  exit $2
+}
+
+<% unless input("pre_script").blank? -%>
+  # Pre Script
+  <%= input("pre_script") %>
+  RETVAL=$?
+  [ $RETVAL -eq 0 ] || die "Pre script failed" $RETVAL
+<% end -%>
+
+# Action
+<% if @host.operatingsystem.family == 'Redhat' -%>
+  yum -y <%= input("action") %> <%= input("package") %>
+<% elsif @host.operatingsystem.family == 'Debian' -%>
+  apt-get -y <%= input("action") %> <%= input("package") %>
+<% end -%>
+RETVAL=$?
+[ $RETVAL -eq 0 ] || die "Package action failed" $RETVAL
+
+<% unless input("post_script").blank? -%>
+  # Post Script
+  <%= input("post_script") %>
+  RETVAL=$?
+  [ $RETVAL -eq 0 ] || die "Post script failed" $RETVAL
+<% end -%>
--- a/app/views/templates/puppet_run_once.erb
+++ b/app/views/templates/puppet_run_once.erb
@@ -1,0 +1,12 @@
+<%#
+kind: job_template
+name: Puppet Run Once - SSH Default
+job_name: Puppet Run Once
+provider_type: Ssh
+template_inputs:
+- name: puppet_options
+  description: Additional options to pass to puppet
+  input_type: user
+  required: false
+%>
+puppet agent --onetime --no-usecacheonfailure <%= input("puppet_options") %>
--- a/app/views/templates/run_command.erb
+++ b/app/views/templates/run_command.erb
@@ -1,0 +1,12 @@
+<%#
+kind: job_template
+name: Run Command - SSH Default
+job_name: Run Command
+provider_type: Ssh
+template_inputs:
+- name: command
+  description: Command to run on the host
+  input_type: user
+  required: true
+%>
+<%= input("command") %>
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.sw?
 .idea
 .bundle/
 log/*.log
--- a/app/models/template_invocation_input_value.rb
+++ b/app/models/template_invocation_input_value.rb
@@ -5,4 +5,6 @@ class TemplateInvocationInputValue < ActiveRecord::Base
 
   validates :value, :presence => true, :if => proc { |v| v.template_input.required? }
 
+  validates :value, :inclusion => { :in => proc { |v| v.template_input.options_array } },
+                    :if => proc { |v| v.template_input.input_type == 'user' && v.template_input.options_array.present? }
 end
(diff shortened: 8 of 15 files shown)


app/models/template_input.rb
```diff
@@ -2,11 +2,16 @@ class TemplateInput < ActiveRecord::Base
   class ValueNotReady < ::Foreman::Exception
   end
 
-  TYPES = { :user => N_('User input'), :fact => N_('Fact value'), :variable => N_('Variable'),
+  class ValueNotPermitted < ::Foreman::Exception
+  end
+
+  TYPES = { :user => N_('User input'), :select_options => N_('User selectable options'),
```
ares: While I really like the select box feature, I planned to wait with advanced forms for outcome of parameters refactoring so we can potentially reuse validators, types etc. If you think this is a must for our first version and should be implemented as a part of our plugin I think that it shouldn't be a new input type since it is user input. I think we just need to extend user input to support select options, the evaluation logic should be the same (except for validation).

ares:
> Templates are seeded by YAML, essentially a cleaned up template.to_json(:includes => :template_inputs). We could also go the Foreman route and put the metadata in an ERB comment at the top.

For consistency I'd prefer ERB with metadata. I hope to improve foreman_templates plugin later to work with any kind of template so the same format would help. But I'm definitely open for discussion about this.

app/models/template_input.rb
stbenjam: It's not a "must" but I think it's very worthwhile. I made it just part of User Input now.

stbenjam:
I think the YAML is tidier, but I moved them to ERB to be consistent with Foreman.

app/models/job_template.rb
```diff
@@ -37,6 +37,28 @@ def self.base_class
...
   self.table_name = 'templates'
 
+  # Import a template ERB, with metadata in the first YAML comment
+  def self.import(template, options = {})
+    return unless (match = template.match(/<%\#(.+?).-?%>/m)) && (metadata = YAML.load(match[1])) && metadata.delete('kind') == 'job_template'
+
+    if self.find_by_name(metadata['name']).blank?
+      inputs = metadata.delete('template_inputs')
+      template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>/m, '')))
+      template.update_attributes(options)
```
stbenjam: This awkward dance is because you can't instantiate a new record with `:locked => true` in Foreman

app/views/templates/package_action.erb
```diff
@@ -0,0 +1,46 @@
...
+# Pre Script
+<%= input("pre_script") %>
+PRE_RETVAL=$?
+
+# Action
+<% if @host.operatingsystem.family == 'Redhat' %>
+yum -y <%= input("action") %> <%= input("package") %>
+<% elsif @host.operatingsystem.family == 'Debian' %>
+apt-get -y <%= input("action") %> <%= input("package") %>
+<% end %>
```
jlsherrill: For the katello case (without provisioning) this would require the OS to be created and set for each system).  I do think this makes sense but isn't something i had considered before.  Will make a note for host unification.

app/views/templates/package_action.erb
```diff
@@ -0,0 +1,46 @@
+<%#
+kind: job_template
+name: Package Action - SSH Default
+job_name: PackageAction
```
iNecas: Any reason fro not usnig space in the name? 'Package Action' would look more nicely in the UI

stbenjam: Not really, that does looks better.  In my head I was thinking of it more like an identifier.

app/views/templates/package_action.erb
iNecas: I've hit an issue with this as well, when running a job against unmanged host (and even with managed host, one is able to get into state where the operating system is not set). 

The question might be if, in this case, we need to use the server parameters, or rather writing bash conditions for checking, which command is available? Opinions?

app/views/templates/puppet_run_once.erb
```diff
@@ -0,0 +1,14 @@
...
+provider_type: Ssh
+template_inputs:
+- name: puppet_options
+  description: Additional options to pass to puppet
+  input_type: user
+  required: false
+%>
+
+puppet agent --onetime --no-usecacheonfailure <%= input("puppet_options") %>
+
```
iNecas: extra line

app/views/templates/run_command.erb
```diff
@@ -0,0 +1,13 @@
...
+name: Run Command - SSH Default
+job_name: RunCommand
+provider_type: Ssh
+template_inputs:
+- name: command
+  description: Command to run on the host
+  input_type: user
+  required: true
+%>
+
```
iNecas: Every seeded template has two blank lines at the beginning:

```


<%= input("command") %>
```

would be nice to just start with the code, but maybe this is something that would need to be handled in Foreman core?

iNecas:
I've tested the code and works as expected, some comments inline

app/views/templates/package_action.erb
iNecas: Identifier doesn't mean it can't have spaces (although it usually does:) Thanks

app/views/templates/package_action.erb
jlsherrill: yeah, a fallback to checking what is available on the system would be a good method if we want to support 'unmanaged' hosts.  It does complicate the script and you'd need to do this in every script but it would make them quite a bit more robust.

jlsherrill: When using something like ansible this may be a bit harder or impossible though.

app/views/templates/run_command.erb
stbenjam: It's in my code, I fixed the regex.  I remove the metadata at the top. Foreman leaves it which imho many users find confusing, and our metadata is 3-4x longer.

app/views/templates/package_action.erb
stbenjam: This is also only the first iteration of the package service script.  Once community users get their hands on this - we're going to get PR's to add SuSE, FreeBSD, Arch, etc.  That's really a lot messier to do in a shell script than checking the host object.

app/views/templates/package_action.erb
```diff
@@ -0,0 +1,45 @@
...
+  required: false
+- name: post_script
+  description: A script to run after the package action
+  input_type: user
+  required: false
+%>
+# Pre Script
+<%= input("pre_script") %>
+PRE_RETVAL=$?
+
```
jlsherrill: Any thoughts on failing out if the pre script fails?

stbenjam: Oh great point, updated

app/views/templates/package_action.erb
stbenjam: I guess it could be emoji too, let's call it :package: :dart:  :-)

app/views/templates/package_action.erb
jlsherrill: I wonder if we should solidify this requirement in some way? hitting the error at template rendering might be frustrating

app/models/job_template.rb
```diff
@@ -37,14 +37,45 @@ def self.base_class
...
   self.table_name = 'templates'
 
+  # Import a template ERB, with metadata in the first YAML comment
+  def self.import(template, options = {})
+    metadata = parse_metadata(template)
+    return if metadata.blank? || metadata.delete('kind') != 'job_template' || self.find_by_name(metadata['name'])
+
+    inputs = metadata.delete('template_inputs')
+
+    # This awkward dance is because you can't instantiate a new template with :locked => true
```
ares: would it make sense to fix it in core so locked attribute is checked only for persisted records? (nothing that would block this PR)

app/models/template_input.rb
```diff
@@ -42,6 +46,10 @@ def value(renderer)
     get_resolver(renderer).value
   end
 
+  def options_array
+    self.options.blank? ? [] : self.options.split(/\r?\n/).map(&:squish)
```
ares: just a nit, would a pure ruby `strip` be sufficient?

app/models/template_input.rb
```diff
@@ -95,7 +103,13 @@ def ready?
     end
 
     def resolved_value
-      input_value.value
+      value = input_value.value
+
+      if @input.options_array.blank? || @input.options_array.include?(value)
+        value
+      else
+        raise(ValueNotPermitted, "Value '#{value}' is not a valid selection for #{@input.name}")
```
ares: I think this should be handled in `ready?` so it returns false for invalid value, otherwise when you preview a template with invalid input it will fail instead of displaying $INPUT placeholder

app/views/job_invocations/_form.html.erb
```diff
@@ -47,9 +47,13 @@
               <fieldset id="job_template_<%= job_template.id %>" class="job_template_form <%= 'hidden' if !@composer.job_template_ids.include?(job_template.id) && !@composer.only_one_template_available? %>">
                 <%= job_templates_fields.fields_for job_template.id.to_s do |job_template_fields| %>
                   <%= job_template_fields.fields_for :input_values do |input_values_fields| %>
-                    <% job_template.template_inputs.where(:input_type => 'user').each do |input| %>
+                    <% job_template.template_inputs.where(:input_type => ['user']).each do |input| %>
```
ares: just a nit, array is not needed here anymore (difference is in SQL query using `=` vs `IN`)

app/models/job_template.rb
stbenjam: I don't mind fixing it in Foreman, but it takes forever to get a review lately...

https://github.com/theforeman/foreman/pull/2652

app/views/job_invocations/_form.html.erb
stbenjam: Fixed, just a leftover from when I created the select type.

app/models/job_template.rb
ares: I just wanted to make sure we have PR open, thank you!

app/models/template_input.rb
stbenjam: I should probably just validate at creation of the TemplateInput, which I can do with a normal rails callback.

I put it here to validate at run time, but maybe that's not a good idea.  Do we do anything if the admin changes the provided values? Maybe the admin gives a user limited access to run job X with values A, B, and C, and later decides to not let them do C anymore.  Saved jobs would still let the user do C.  Maybe that's OK.

I guess we have that problem with everything, maybe they have saved jobs running on hosts they no longer have access to.

app/views/templates/package_action.erb
stbenjam: Maybe show a :warning: with a tooltip next to the Resolved Hosts count if a host is lacking key attributes?

app/models/template_input.rb
```diff
@@ -91,7 +99,7 @@ def resolved_value
 
   class UserInputResolver < InputResolver
     def ready?
-      !input_value.nil?
+      !input_value.value.nil?
```
ares: input_value can be also nil here which should result in not ready input, it would raise exception here

app/models/template_input.rb
ares: Permission are evaluated when the run is scheduled (respecting view_hosts and view_job_templates permissions), until we add execute permission there are some gaps that will open more widely with delayed execution. We don't have designed permission granularity on input values, not sure if it's necessary. We can always hardcode parameters to templates if you want to limit user capabilities.

stbenjam: I know, maybe I didn't explain correctly.  The select box lets an admin limit a users choices. That's what A, B, and C are in my example above.  But maybe it doesn't matter.

stbenjam: And if permissions are only evaluated at scheduling then the behavior is fine, I'll check check the values in a normal rails callback.

app/models/template_input.rb
stbenjam: Yes I didn't mean to change it

app/models/template_input.rb
ares: ah I see what you mean, I think that's fine for now

ares:
:+1:, please go ahead and merge, :clap:

iNecas:
:+1:

ohadlevy:
would you add those to foreman templates repo? sounds like we should extend the template plugin to sync those too?

ares:
@ohadlevy that's the goal, ideally we'd like to improve foreman_template codebase a bit during this

app/models/job_template.rb
eLobato: Merged, should be in for 1.10

stbenjam: Thanks!

app/views/templates/package_action.erb
```diff
@@ -0,0 +1,52 @@
+<%#
```
ohadlevy: should we not defined os family or something other way of limiting?

app/models/job_template.rb
ares: btw it's set to be in 1.9.1, tracking issue at http://projects.theforeman.org/issues/11620

app/views/templates/package_action.erb
stbenjam: I think that's way out of scope for this PR and should be a separate issue if we do it at all. There's no connection between OS and JobTemplates exposed to the user today.

app/views/templates/package_action.erb
jlsherrill: Yeah something like that would work.  I'll think about it some more

app/views/job_invocations/_form.html.erb
```diff
@@ -49,7 +49,11 @@
                   <%= job_template_fields.fields_for :input_values do |input_values_fields| %>
                     <% job_template.template_inputs.where(:input_type => 'user').each do |input| %>
                       <%= input_values_fields.fields_for input.id.to_s, @composer.template_invocation_input_value_for(input) do |input_fields| %>
-                        <%= text_f input_fields, :value, :label => input.name, :help_inline => input.description, :required => input.required %>
+                        <% unless input.options.blank? %>
+                          <%= selectable_f input_fields, :value, input.options_array, {:include_blank => !input.required }, :require => input.required, :label => input.name, :help_inline => input.description %>
+                        <% else %>
+                          <%= textarea_f input_fields, :value, :label => input.name, :help_inline => input.description, :required => input.required, :rows => 2 %>
```
ares: this renders big fields even for things like numbers, how about row => 1, user can enlarge in browser if it's needed

ares:
tested by deep-dive, works nicely, one small comment inline, otherwise :+1:

iNecas:
I've tested the latest version on clear setup, everything worked as expected, good to merge from my side. Awesome work @stbenjam!

app/views/job_invocations/_form.html.erb
ares: OK, no strong opinion on this, we'll see with more template examples

stbenjam: Ok thanks!

ares:
:fireworks: good job

PR #27, opened 2015-08-31
Fixes #11601 - do the preview for specific host if possible

ares:


app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -27,7 +27,12 @@
     <h5>
       <b><%= template_invocation.template.name %></b> <%= 'through' %> <%= _(RemoteExecutionProvider.provider_for(template_invocation.template.provider_type)) %>
     </h5>
-    <pre><%= InputTemplateRenderer.new(template_invocation.template, nil, template_invocation).preview %></pre>
+    <% target = template_invocation.targeting.hosts.first %>
```
stbenjam: Based on the discussion here, https://github.com/theforeman/foreman_remote_execution/pull/26#discussion_r38207845, maybe we pick the first host with an OS?

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -27,7 +27,12 @@
...
+    <% if target.present? %>
+      <%= _('Preview for target %s') % target.name %>
+    <% end %>
+    <pre><%= InputTemplateRenderer.new(template_invocation.template, target, template_invocation).preview %></pre>
```
stbenjam: If, for some reason, the template doesn't render the user just sees a "false" - could we show something more informative?

app/views/job_invocations/_tab_overview.html.erb
ares: I incline to solve this through inputs somehow. This bug was more about completely missing host, so even if host had operating system, @host.operatingsystem resulted on calling operatingsystem on nil... Anyway I changed it to prefer host with OS for now and fallback to any host from targeting.

app/views/job_invocations/_tab_overview.html.erb
ares: Right, improved to print out renderer's error.

stbenjam:
Looks good, ACK from me

ares:
Thanks, merging

PR #28, opened 2015-08-31
Fixes #11584 - rerun feature

ares:
not sure how to test foreman tasks queries correctly but I think it's not critical

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -2,16 +2,22 @@ class JobInvocationsController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
 
   def new
-    @composer = JobInvocationComposer.new(JobInvocation.new,
-                                          :host_ids => params[:host_ids],
-                                          :targeting => {
-                                            :targeting_type => Targeting::STATIC_TYPE,
-                                            :bookmark_id => params[:bookmark_id]
-                                          })
+    @composer = JobInvocationComposer.new.compose_from_params(
+      :host_ids => params[:host_ids],
+      :targeting => {
+        :targeting_type => Targeting::STATIC_TYPE,
+        :bookmark_id => params[:bookmark_id]
+      })
+  end
+
+  def rerun
+    job_invocation = resource_base.find(params[:id])
+    @composer = JobInvocationComposer.new.compose_from_invocation(job_invocation)
+    render :action => 'new'
   end
 
   def create
-    @composer = JobInvocationComposer.new(JobInvocation.new, params)
+    @composer = JobInvocationComposer.new.compose_from_params(params)
     if @composer.save
       @task = ForemanTasks.async_task(::Actions::RemoteExecution::RunHostsJob, @composer.job_invocation)
       redirect_to job_invocation_path(@composer.job_invocation)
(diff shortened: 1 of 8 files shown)


app/models/job_invocation.rb
```diff
@@ -18,4 +18,28 @@ class JobInvocation < ActiveRecord::Base
...
+    else
+      ForemanTasks::Task.for_action_types('Actions::RemoteExecution::RunHostJob').where('1=0')
+    end
+  end
+
+  def failed_template_invocation_tasks
+    template_invocations_tasks.where(:result => 'warning')
+  end
+
+  def failed_host_ids
```
iNecas: where is this used, I could imaging seeing the usage in https://github.com/theforeman/foreman_remote_execution/pull/28/files#diff-5becfffb6454c87c9b39dd84ae3676d2R18

ares: exactly there it was planned to be used :-) note that `failed_hosts` is not used anywhere on purpose

iNecas:
`Resolve hosts` button doesn't work, getting

```
2015-09-01T10:26:28 [app] [F]
 | NoMethodError (undefined method `composer_from_params' for #<JobInvocationComposer:0x0000000a114b80>):
 |   /home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/controllers/job_invocations_controller.rb:45:in `refresh'
 |   app/controllers/concerns/application_shared.rb:13:in `set_timezone'
 |   app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 |   lib/middleware/catch_json_parse_errors.rb:9:in `call'
 |

```

iNecas:
After fixing the typo (`composer_from_params` vs `compose_from_params`) the code worked as expected

app/helpers/remote_execution_helper.rb
```diff
@@ -57,6 +57,16 @@ def template_invocation_status(task)
   def job_invocation_task_buttons(task)
     buttons = []
     buttons << link_to(_('Refresh'), {}, :class => 'btn btn-default', :title => _('Refresh this page'))
+    if authorized_for(:permission => :create_job_invocations)
+      buttons << link_to(_("Rerun"), rerun_job_invocation_path(:id => task.locks.where(:resource_type => 'JobInvocation').first.resource),
+                         :class => "btn btn-default",
+                         :title => _('Rerun the job'))
+    end
+    if authorized_for(:permission => :create_job_invocations)
+      buttons << link_to(_("Rerun in failed"), rerun_job_invocation_path(:id => task.locks.where(:resource_type => 'JobInvocation').first.resource, :failed_only => 1),
```
iNecas: I'm not sure about the `in` word. What about `Rerun failed`?

ares: or maybe `on failed?`

app/controllers/job_invocations_controller.rb
```diff
@@ -2,16 +2,28 @@ class JobInvocationsController < ApplicationController
...
+        :targeting_type => Targeting::STATIC_TYPE,
+        :bookmark_id => params[:bookmark_id]
+      })
+  end
+
+  def rerun
+    job_invocation = resource_base.find(params[:id])
+    @composer = JobInvocationComposer.new.compose_from_invocation(job_invocation)
+
+    if params[:failed_only]
```
iNecas: I wonder, if this logic shouldn't be in the composer?

iNecas:
One question inline, but :+1: otherwise: this is really helpful feature both for development and production

app/controllers/job_invocations_controller.rb
ares: I think this parameter is specific to the controller action, maybe we could change the `params[:host_ids]` so we reuse the composer logic but this seems cleaner to me

iNecas: Ok, let it here for now: if the need for sharing this logic appears, we can move the code on more sharable place

iNecas:
APJ

ares:
I'll close this PR and reopen another one, github does not pick the new commit from my branch for some reason.

PR #32, opened 2015-09-01
Fixes #11145 - template invocations are destroyed with template

ares:


iNecas:
Tests? We do so for other deletions https://github.com/theforeman/foreman_remote_execution/blob/master/test/unit/targeting_test.rb#L57

ares:
Here you are, rebase is for free.

ares:
yet another rebase

iNecas:
```doge
very green
            much like
   such ack
         wow
```

ares:
Thanks, merging

PR #33, opened 2015-09-02
fixes #11554, #11827 - execution proxy through subnet, with load balancing

stbenjam:
## Background

Currently, when you schedule an action, we determine the proxy to handle the request by some magic, in the following order:

1. We look at all the proxies attached to the host (Puppet, TFTP, DNS) and see if any have remote execution features we're looking for.  We take the first we find.

1. If you enable the setting `remote_execution_global_proxy`, and we don't find a proxy on the Host, we'll expand our search to all proxies.

1. If nothing is found, fail for that host.

This works, but we need some more granular control over this in order to be deterministic in the proxies we choose in multi-proxy environments, and account for complex networking setups. We want to do it without making the user's life too difficult. 

## Execution Interface

This PR adds the concept of an "Execution Interface."  New hosts will automatically get one, and you can change it to be a different interface if desired:

newhost

The proxy to use is determined through the subnet assigned to the execution interface. You can actually assign *multiple* remote execution providers.  This buys us a nice way to have load balancing / HA..

subnet

### New Work Flow

The new process for determining the remote execution proxy:

1. If the host has an execution interface with a subnet, then we use the proxies associated to that subnet, if any.

1. Same as 1 above, except now gated by a new setting: `remote_execution_fallback_proxy`

1. Same as 2 above

1. Same as 3 above

## Load Balancing / HA

This is not yet implemented, bu we'll round robin load balance requests across the smart proxies for that subnet.  

## Some questions

1. How does this setup feel from a usability perspective? Any suggestions?

1. Should you *also* be able to assign remote execution proxies directly to hosts? The default action would always be inherit from subnet, but a user could want in some cases some one-off overrides.

1. Are there environments where this setup will not work well?

1. Any concerns about integration with Katello?

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -8,12 +8,17 @@ module Actions
 
       include ::Dynflow::Action::Cancellable
 
-      def plan(job_invocation, host, connection_options = {})
+      def plan(job_invocation, host, template_invocation, proxy, connection_options = {})
         action_subject(host, :job_name => job_invocation.job_name)
-
-        template_invocation = find_template_invocation(job_invocation, host)
         hostname = find_ip_or_hostname(host)
-        proxy = find_proxy(template_invocation, host)
+
+        raise _("Could not use any template used in the job invocation") if template_invocation.blank?
+
+        settings =  { :global_proxy   => 'remote_execution_global_proxy',
+                      :fallback_proxy => 'remote_execution_fallback_proxy' }
+
+        raise _("Could not use any proxy. Consider configuring %{global_proxy} " +
+                "or %{fallback_proxy} in settings") % settings if proxy.blank?
 
         renderer = InputTemplateRenderer.new(template_invocation.template, host, template_invocation)
         script = renderer.render
@@ -45,52 +50,18 @@ module Actions
         _('Run %{job_name} on %{host}') % { :job_name => input[:job_name], :host => input[:host][:name] }
       end
 
-      def find_template_invocation(job_invocation, host)
-        providers = available_providers(job_invocation, host)
-        providers.each do |provider|
-          job_invocation.template_invocations.each do |template_invocation|
-            if template_invocation.template.provider_type == provider
-              return template_invocation
-            end
+      def find_ip_or_hostname(host)
+        %w(execution primary provision).each do |flag|
+          if host.send("#{flag}_interface") && host.send("#{flag}_interface").ip.present?
+            return host.execution_interface.ip
           end
         end
 
-        raise _("Could not use any template used in the job invocation")
-      end
-
-      def find_ip_or_hostname(host)
         host.interfaces.each do |interface|
           return interface.ip unless interface.ip.blank?
         end
-        return host.name
-      end
 
-      def available_providers(job_invocation, host)
-        # TODO: determine from the host and job_invocation details
-        return ['Ssh']
-      end
-
-      def find_proxy(template_invocation, host)
-        provider = template_invocation.template.provider_type.to_s
-        all_host_proxies(host).each do |proxies|
-          if proxy = proxies.joins(:features).where("features.name = ?", provider).first
-            return proxy
-          end
-        end
-        raise _("Could not use any proxy: assign a proxy with provider '%{provider}' to the host or set '%{global_proxy_setting}' in settings") %\
-            { :provider => provider, :global_proxy_setting => 'remote_execution_global_proxy' }
-      end
-
-      def all_host_proxies(host)
-        Enumerator.new do |e|
-          host.interfaces.each do |interface|
-            if interface.subnet
-              e << ::SmartProxy.where(:id => interface.subnet.proxies.map(&:id))
-            end
-          end
-          e << host.smart_proxies
-          e << ::SmartProxy.authorized if Setting[:remote_execution_global_proxy]
-        end
+        return host.fqdn
       end
     end
   end
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -3,17 +3,55 @@ module ForemanRemoteExecution
     extend ActiveSupport::Concern
 
     included do
+      alias_method_chain :build_required_interfaces, :remote_execution
+      alias_method_chain :reload, :remote_execution
+      alias_method_chain :becomes, :remote_execution
+
       has_many :targeting_hosts, :dependent => :destroy, :foreign_key => 'host_id'
     end
 
-    # create or overwrite instance methods...
-    def instance_method_name
+    def execution_interface
+      get_interface_by_flag(:execution)
     end
 
-    module ClassMethods
-      # create or overwrite class methods...
-      def class_method_name
+    def remote_execution_proxies(provider)
+      proxies = {}
+      proxies[:subnet]   = execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface && execution_interface.subnet
+      proxies[:fallback] = smart_proxies.with_features(provider) if Setting[:remote_execution_fallback_proxy]
+
+      if Setting[:remote_execution_global_proxy]
+        proxy_scope = if Taxonomy.enabled_taxonomies.any?
+                        ::SmartProxy.with_taxonomy_scope_override(location, organization)
+                      else
+                        proxy_scope = ::SmartProxy
+                      end
+
+        proxies[:global] = proxy_scope.authorized.with_features(provider)
       end
+
+      proxies
+    end
+
+    def drop_execution_interface_cache
+      @execution_interface = nil
+    end
+
+    def becomes_with_remote_execution(*args)
+      became = becomes_without_remote_execution(*args)
+      became.drop_execution_interface_cache
+      became
+    end
+
+    def reload_with_remote_execution(*args)
+      drop_execution_interface_cache
+      reload_without_remote_execution(*args)
+    end
+
+    private
+
+    def build_required_interfaces_with_remote_execution(attrs = {})
+      build_required_interfaces_without_remote_execution(attrs)
+      self.primary_interface.execution = true if self.execution_interface.nil?
     end
   end
 end
--- a/app/models/target_remote_execution_proxy.rb
+++ b/app/models/target_remote_execution_proxy.rb
@@ -1,0 +1,4 @@
+class TargetRemoteExecutionProxy < ActiveRecord::Base
+  belongs_to :remote_execution_proxy, :class_name => 'SmartProxy'
+  belongs_to :target, :polymorphic => true
+end
--- a/app/overrides/foreman/nics/_execution_interface.html.erb
+++ b/app/overrides/foreman/nics/_execution_interface.html.erb
@@ -1,0 +1,1 @@
+<%= checkbox_f f, :execution, :help_inline => _("The execution interface is used for remote execution"), :class => :interface_execution %>
--- a/app/overrides/foreman/subnets/_rex_tab_pane.html.erb
+++ b/app/overrides/foreman/subnets/_rex_tab_pane.html.erb
@@ -1,0 +1,5 @@
+<div class="tab-pane" id="rex_proxies">
+  <%= fields_for :subnet do |f| %>
+    <%= multiple_selects f, :remote_execution_proxies, SmartProxy.authorized.with_features(RemoteExecutionProvider.provider_names), @subnet.remote_execution_proxy_ids, {:label => _("Proxies"), :help_inline => _("Select as many remote execution proxies as applicable for this subnet.  When multiple proxies with the same provider are added, actions will be load balanced among them.")} %>
+  <% end %>
+</div>
--- a/app/services/proxy_load_balancer.rb
+++ b/app/services/proxy_load_balancer.rb
@@ -1,0 +1,31 @@
+class ProxyLoadBalancer
+  def initialize
+    @tasks   = {}
+    @offline = []
+  end
+
+  # Get the least loaded proxy from the given list of proxies
+  def next(proxies)
+    exclude = @tasks.keys + @offline
+    @tasks.merge!(get_counts(proxies - exclude))
+    next_proxy = @tasks.select { |proxy, _| proxies.include?(proxy) }.min_by { |_, job_count| job_count }.first
+    @tasks[next_proxy] += 1
+    next_proxy
+  end
+
+  private
+
+  def get_counts(proxies)
+    proxies.inject({}) do |result, proxy|
+      begin
+        proxy_api = ProxyAPI::ForemanDynflow::DynflowProxy.new(:url => proxy.url)
+        result[proxy] = proxy_api.tasks_count('running')
+        result
+      rescue
+        @offline << proxy
+        Rails.logger.error("Could not fetch task counts from #{proxy}, skipped.")
+        result
+      end
+    end
+  end
+end
--- a/db/migrate/20150826191632_create_target_remote_execution_proxies.rb
+++ b/db/migrate/20150826191632_create_target_remote_execution_proxies.rb
@@ -1,0 +1,14 @@
+class CreateTargetRemoteExecutionProxies < ActiveRecord::Migration
+  def change
+    create_table :target_remote_execution_proxies do |t|
+      t.integer :remote_execution_proxy_id
+      t.integer :target_id
+      t.string :target_type
+
+      t.timestamps
+    end
+
+    add_index :target_remote_execution_proxies, :remote_execution_proxy_id, :name => 'target_remote_execution_proxies_proxy_id'
+    add_index :target_remote_execution_proxies, [:target_id, :target_type], :name => 'target_remote_execution_proxies_target_id_target_type'
+  end
+end
--- a/test/test_plugin_helper.rb
+++ b/test/test_plugin_helper.rb
@@ -6,3 +6,21 @@ require 'dynflow/testing'
 # Add plugin to FactoryGirl's paths
 FactoryGirl.definition_file_paths << File.join(File.dirname(__FILE__), 'factories')
 FactoryGirl.reload
+
+# Foreman's setup doesn't handle cleaning up for Minitest::Spec
+DatabaseCleaner.strategy = :transaction
+
+class Minitest::Spec
+  class << self
+    alias_method :context, :describe
+  end
+
+  before :each do
+    DatabaseCleaner.start
+    Setting::RemoteExecution.load_defaults
+  end
+
+  after :each do
+    DatabaseCleaner.clean
+  end
+end
--- a/test/unit/concerns/host_extensions_test.rb
+++ b/test/unit/concerns/host_extensions_test.rb
@@ -1,0 +1,90 @@
+require 'test_plugin_helper'
+
+describe ForemanRemoteExecution::HostExtensions do
+  let(:provider) { 'Ssh' }
+
+  before do
+    User.current = FactoryGirl.build(:user, :admin)
+  end
+
+  after { User.current = nil }
+
+  context 'host has multiple nics' do
+    let(:host) { FactoryGirl.build(:host, :with_execution) }
+
+    it 'should only have one execution interface' do
+      host.interfaces << FactoryGirl.build(:nic_managed)
+      host.interfaces.each { |interface| interface.execution = true }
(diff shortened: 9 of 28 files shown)


stbenjam:
[test] 

weird rubocop failure `private method color called for #<String:0x0000000c814988>` on 8/9 tests

stbenjam:
Rubocop failure is because deface pulls in `colorize`, which interferes with `rainbow`'s monkey patching.

https://github.com/bbatsov/rubocop/pull/2205

ares:
> Should you also be able to assign remote execution proxies directly to hosts? The default action would always be inherit from subnet, but a user could want in some cases some one-off overrides.

I think the answer is yes. 
* some hosts might not have interfaces correctly configured (if importing interfaces is disabled) and there exist just one dummy interface with IP (with no subnet associations)
* for hosts we can simply add bulk action to assign remote execution proxy

Ideally I'd like to have some generic module that can be included in any object allowing you to specify rex proxies and some class responsible for determining the right proxy. I can imagine proxy being associated to Location and even Organization. It could make sense to have different proxy per Environment. The class that would determine the proxy for an invocation would find all objects, assign some weigh to them and find the best option. Later we could also add other attributes affecting the selection to the algorithm, like load balancing. Obviously, what I've described above does not have to be there from day 1, but I'd like to hear others opinions about this and agree upon the goal. The only thing from this paragraph that I'd like to see in this PR is extracting the logic of finding a proxy to a separate class.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -55,6 +55,10 @@ def find_template_invocation(job_invocation, host)
       end
 
       def find_ip_or_hostname(host)
+        if host.execution_interface && host.execution_interface.ip.present?
+          return host.execution_interface.ip
+        end
+
```
ares: would it make sense to prefer `provision` and then `primary` interface before we go through all? sorry, I know it's not strictly related to your change

db/migrate/20150826192731_add_execution_to_interface.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def type
+    Nic::Managed
+  end
+end
+
+class AddExecutionToInterface < ActiveRecord::Migration
+  def up
+    add_column :nics, :execution, :boolean, :default => false
+    add_column :hosts, :execution_interface, :integer
```
ares: missing in down, but I don't think this column is used, or at least is covered by method of same name

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -3,16 +3,45 @@ module HostExtensions
...
+
+    private
+
+    def build_required_interfaces_with_remote_execution(attrs = {})
+      build_required_interfaces_without_remote_execution(attrs)
+      self.primary_interface.execution = true if self.execution_interface.nil?
+    end
+
+    def candidate_proxies_for(provider)
+      [
```
ares: I find the following a bit more readable than if conditions nested in array

```ruby
candidates = []
candidates << execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface && execution_interface.subnet
candidates << smart_proxies.with_features(provider) if Setting[:remote_execution_fallback_proxy]
candidates << ::SmartProxy.authorized.with_features(provider) if Setting[:remote_execution_global_proxy]
candidates.flatten.compact
```

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -3,16 +3,45 @@ module HostExtensions
...
+          smart_proxies.with_features(provider)
+        end,
+
+        if Setting[:remote_execution_global_proxy]
+          ::SmartProxy.authorized.with_features(provider)
+        end
+      ].flatten.compact
+    end
+
+    def execution_interface_count
```
ares: this won't catch adding second execution interface through interfaces API, otoh validation on interface level can be tricky, it would have to check host's in memory interfaces as well

test/test_plugin_helper.rb
```diff
@@ -6,3 +6,21 @@
...
+    alias_method :context, :describe
+  end
+
+  before :each do
+    DatabaseCleaner.start
+    Setting::RemoteExecution.load_defaults
+  end
+
+  after :each do
+    DatabaseCleaner.clean
```
ares: why do we need to clean the DB after each test? this will probably slow down the test suite noticeable

test/unit/concerns/host_extensions_test.rb
```diff
@@ -0,0 +1,84 @@
...
+  before { User.current = FactoryGirl.build(:user, :admin) }
+  after { User.current = nil }
+
+  context 'host has multiple nics' do
+    let(:host) { FactoryGirl.build(:host, :with_execution) }
+
+    it 'should only have one execution interface' do
+      host.interfaces << FactoryGirl.build(:nic_managed)
+      host.interfaces.each { |interface| interface.execution = true }
+      refute host.valid?
```
ares: I'd add `assert host.errors.has_key?(:nics)` since the host can be invalid for many reasons...

test/unit/concerns/host_extensions_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    it 'returns the execution interface' do
+      assert host.execution_interface
```
ares: `host.execution_interface.must_be_kind_of Nic::Managed` would be probably better

test/unit/concerns/host_extensions_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+  end
+
+  context 'host with subnet' do
+    let(:host) { FactoryGirl.build(:host, :with_execution) }
+
+    it 'gets remote execution proxy from subnet' do
+      assert_equal host.remote_execution_proxy(provider), host.subnet.remote_execution_proxies.first
```
ares: could you please convert assert to must/wont syntax? this seems unnatural for rspec-like test syntax

test/unit/concerns/nic_extensions_test.rb
```diff
@@ -0,0 +1,9 @@
+require 'test_plugin_helper'
+
+describe ForemanRemoteExecution::NicExtensions do
+  let(:host) { FactoryGirl.create(:host) }
+
+  it 'sets the first interface as the execution interface' do
+    assert host.interfaces.first.execution
```
ares: it seems that it sets this flag on primary interface rather than first

test/test_plugin_helper.rb
stbenjam: Tests fail without it for me locally. Things ended up getting stored persistently in the database, like hosts and proxies I was creating with FactoryGirl.

stbenjam: Also if you notice above I have ` alias_method :context, :describe`, otherwise context isn't defined in the extensions tests...and I get an exception.

I've never used `Minitest::Spec` before, is there some rails magic that's maybe not working with testing the concerns?

app/models/concerns/foreman_remote_execution/host_extensions.rb
stbenjam: How does Foreman do it? I can't see where Foreman actually checks that you only have one primary/provision interface.

ares: see https://github.com/theforeman/foreman/blob/develop/app/models/nic/base.rb#L181

test/test_plugin_helper.rb
jlsherrill: hrmmmm, this seems kinda odd.  I wouldn't think using a transaction strategy would be too slow as it would be doing what the test suites are doing for other projects.  I don't just quite understand why its needed.

stbenjam: It is odd, and you won't see it on jenkins because you start with a fresh DB.  Feel free to run them locally twice, and you'll see objects created during tests aren't cleaned out of the database without this.  At least that's what it does for me on my katello-dev box which uses postgresql.

stbenjam: And transaction strategy is fast, there's no noticeable performance difference.

ares: Usually, unit tests do not touch database, but it's tricky with ActiveRecord so we need things like db cleaner. At least from my experience with mysql, this made a big difference. Anyway I haven't got to testing yet so I have no idea why it is necessary.

stbenjam:
Added PoC for the load balancing so the idea is clearer.

Needs https://github.com/theforeman/foreman-tasks/pull/138 and https://github.com/theforeman/smart_proxy_dynflow/pull/7

iNecas:
@stbenjam I like the approach, let's make the tests green and I'm ok with getting this in

ares:
the test issue is unrelated to the PR and is already fixed in master, could you please rebase the branch?

app/services/proxy_load_balancer.rb
```diff
@@ -0,0 +1,28 @@
+class ProxyLoadBalancer
+  def initialize
+    all_proxies = SmartProxy.unscoped.with_features(RemoteExecutionProvider.providers.keys)
+    @jobs = get_counts(all_proxies)
+  end
+
+  # Get the least loaded proxy from the given list of proxies
+  def next(proxies)
+    ids = proxies.map(&:id)
```
iNecas: You can use the proxy objects directly in `@jobs` keys, which allows us to avoid loading the proxy object again for every host iteration

app/services/proxy_load_balancer.rb
```diff
@@ -0,0 +1,28 @@
+class ProxyLoadBalancer
+  def initialize
+    all_proxies = SmartProxy.unscoped.with_features(RemoteExecutionProvider.providers.keys)
+    @jobs = get_counts(all_proxies)
```
iNecas: Lazily loading the counts would be a bit better

iNecas: I mean calculating them the first time the `next` method is called with the given proxy

app/services/proxy_load_balancer.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  # Get the least loaded proxy from the given list of proxies
+  def next(proxies)
+    ids = proxies.map(&:id)
+    proxy = @jobs.reject { |id, _| !ids.include?(id) }.min_by { |_, job_count| job_count }.first
```
iNecas: `select { |proxy, _| proxies.include?(proxy) }` would be bit more natural (both using the proxy as key and using select instead of reject here)

app/services/proxy_load_balancer.rb
iNecas: `s/@jobs/@count_by_proxy/` or something else more descriptive?

app/services/proxy_load_balancer.rb
```diff
@@ -0,0 +1,28 @@
...
+  private
+
+  def get_counts(proxies)
+    proxies.inject({}) do |result, proxy|
+      begin
+        proxy_api = ProxyAPI::ForemanDynflow::DynflowProxy.new(:url => proxy.url)
+        result[proxy.id] = proxy_api.tasks_count('running')
+        result
+      rescue
+        result # skip proxy with error
```
iNecas: We should at least log the exception

app/overrides/foreman/nics/_execution_interface.html.erb
```diff
@@ -0,0 +1 @@
+<%= checkbox_f f, :execution, :help_inline => _("The Execution interface is used for Remote Execution"), :class => :interface_execution %>
```
iNecas: Probably no need for capital leters in `Execution` and `Remote Execution`, or?

app/overrides/foreman/subnets/_rex_tab_pane.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="tab-pane" id="rex_proxies">
+  <%= fields_for :subnet do |f| %>
+    <%= multiple_selects f, :remote_execution_proxies, SmartProxy.authorized.with_features(RemoteExecutionProvider.provider_names), @subnet.remote_execution_proxy_ids, {:label => _("Proxies"), :help_inline => _("Select as many Remote Execution proxies as applicable for this subnet.  When multiple proxies with the same provider are added, actions will be load balanced among them.")} %>
```
iNecas: Same here about the capitals `Remote Execution`

app/models/target_remote_execution_proxy.rb
```diff
@@ -0,0 +1,4 @@
+class TargetRemoteExecutionProxy < ActiveRecord::Base
+  belongs_to :remote_execution_proxy, :class_name => 'SmartProxy'
+  belongs_to :target, :polymorphic => true
```
iNecas: It's polymorphic to enable association directly to host? Just checking, if I understand the intention properlỵ...

db/migrate/20150826191632_create_target_remote_execution_proxies.rb
```diff
@@ -0,0 +1,11 @@
+class CreateTargetRemoteExecutionProxies < ActiveRecord::Migration
+  def change
+    create_table :target_remote_execution_proxies do |t|
+      t.integer :remote_execution_proxy_id
```
iNecas: Indexes on foreign keys can help us later (on `:remote_execution_proxy_id` and `[:target_id, :target_type]`, especially if somebody used the per-host target association

app/services/proxy_load_balancer.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  private
+
+  def get_counts(proxies)
+    proxies.inject({}) do |result, proxy|
+      begin
+        proxy_api = ProxyAPI::ForemanDynflow::DynflowProxy.new(:url => proxy.url)
+        result[proxy.id] = proxy_api.tasks_count('running')
+        result
+      rescue
```
ares: I think that you can deduplicate with `ensure` block returning `result`

app/models/target_remote_execution_proxy.rb
stbenjam: Yeah, polymorphic to add host later, was going to add it in a separate PR, unless you really want it here

iNecas: Ok for another PR

app/services/proxy_load_balancer.rb
iNecas: Also: it should handle a case, where there is an empty set of `@jobs`

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -3,17 +3,45 @@ module HostExtensions
...
-    # create or overwrite instance methods...
-    def instance_method_name
+    def execution_interface
+      get_interface_by_flag(:execution)
+    end
+
+    def remote_execution_proxies(provider)
+      proxies = {}
+      proxies[:subnet]   = execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface && execution_interface.subnet
+      proxies[:fallback] = smart_proxies.with_features(provider) if Setting[:remote_execution_fallback_proxy]
```
iNecas: What bout taxonomy strategy: using all proxies in the enabled taxonomies for the host? It would allow to model the proxies around the org/locations, without too much effort on our side…

stbenjam: For fallback? All the host proxies should already be confined to the host's taxonomy. If not, that's Foreman's problem IMHO to handle.

Or do you mean global? To me that means truly global and unconfined to an org.

stbenjam: Oh do you mean an additional strategy between fallback and global? I think we already have a lot, because we're going to add a host strategy as well.

app/services/proxy_load_balancer.rb
stbenjam: Ensure's return result is discarded

app/models/concerns/foreman_remote_execution/host_extensions.rb
iNecas: Personally, I think the taxonomy strategy would be more useful than the host one. Yes something between fallback and global. I believe, with pretty small number of code, we would get something, that the locations were trying to get.

stbenjam: That's almost a simple implementation of @ohadlevy's idea, which is fine, but it's adding too many different ways to determine the proxy, it's going to confuse users.

If you can use taxonomy-based (i.e. put a bunch of proxies in one location), then why bother using subnet strategy at all?

stbenjam: Not that I'm suggesting scrapping the subnet-based, because I believe through the interface is the best way in a large environment.

iNecas: In case you don't use taxonomies (which is a valid thing to do) :) I agree too many options might be confusing. How do others see it? @jlsherrill @ares @adamruzicka ?

adamruzicka: I'm not entirely sure what you mean by the host strategy.
Right now we have 3 ways to determine the proxy (subnet, fallback, global) and want to add topology (and/or host?) one?
Too many options might be confusing for some users but I think 5 options might be the middle ground between forcing the user to do things "our" way and being overly complicated.

iNecas: What about extending the global strategy to respect the taxonomies… If somebody is using taxonomies, the proxies would be selected from given taxonomies?

stbenjam: That might be OK, one might think of global being scoped to taxonomies. A 'universal' proxy search would be unscoped ;-)

iNecas: :+1: :)

db/migrate/20150826191632_create_target_remote_execution_proxies.rb
stbenjam: Done

app/models/concerns/foreman_remote_execution/host_extensions.rb
stbenjam: https://github.com/theforeman/foreman_remote_execution/pull/33/files#diff-e3b1be04f41d99a66e199b0060b84e63R22

app/services/proxy_load_balancer.rb
```diff
@@ -0,0 +1,28 @@
...
+  private
+
+  def get_counts(proxies)
+    proxies.inject({}) do |result, proxy|
+      begin
+        proxy_api = ProxyAPI::ForemanDynflow::DynflowProxy.new(:url => proxy.url)
+        result[proxy] = proxy_api.tasks_count('running')
+        result
+      rescue
+        Rails.logger.error("Could not fetch task counts from #{proxy}, skipped.")
```
iNecas: if count request fails, the caching would not work fully: the proxy that is down would get requested every time it's part of some proxy pool: we might want to store something into `result[proxy]` also, when the request fails (`nil` might be ok, if we count on that on other places

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -3,17 +3,55 @@ module HostExtensions
...
-      # create or overwrite class methods...
-      def class_method_name
+    def remote_execution_proxies(provider)
+      proxies = {}
+      proxies[:subnet]   = execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface && execution_interface.subnet
+      proxies[:fallback] = smart_proxies.with_features(provider) if Setting[:remote_execution_fallback_proxy]
+
+      if Setting[:remote_execution_global_proxy]
+        proxy_scope = if Taxonomy.enabled_taxonomies.any?
+                        ::SmartProxy.with_taxonomy_scope_override(organization, location)
```
iNecas: https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L38  - reversed order of taxonomies is expected: `location, organization`

app/services/proxy_load_balancer.rb
stbenjam: Ok, updated

stbenjam:
Ready for another look

iNecas:
Looks ok, going through the final testing before acking

iNecas:
ACK, tested both via subets and the global strategy (including taxonomies)

stbenjam:
Thanks @iNecas !

PR #34, opened 2015-09-07
Refs #11361 - Allows to set retry count a timeout interval on a job

adamruzicka:
timeouts require theforeman/foreman-tasks#132

### DIFF
--- a/app/lib/actions/remote_execution/run_proxy_command.rb
+++ b/app/lib/actions/remote_execution/run_proxy_command.rb
@@ -14,16 +14,24 @@ module Actions
       end
 
       def on_data(data)
-        super(data)
-        error! _("Script execution failed") if failed_run?
+        if data[:result] == 'initialization_error'
+          handle_connection_exception(data[:metadata][:exception_class]
+                                       .constantize
+                                       .new(data[:metadata][:exception_message]))
+        else
+          super(data)
+          error! _("Script execution failed") if failed_run?
+        end
       end
+
 
       def rescue_strategy
         ::Dynflow::Action::Rescue::Skip
       end
 
       def failed_run?
-        exit_status && proxy_output[:exit_status] != 0
+        output[:result] == 'initialization_error' ||
+          (exit_status && proxy_output[:exit_status] != 0)
       end
 
       def exit_status
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -8,7 +8,7 @@ module Actions
 
       include ::Dynflow::Action::Cancellable
 
-      def plan(job_invocation, host)
+      def plan(job_invocation, host, connection_options = {})
         action_subject(host, :job_name => job_invocation.job_name)
 
         template_invocation = find_template_invocation(job_invocation, host)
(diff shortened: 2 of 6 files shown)


app/lib/actions/remote_execution/run_proxy_command.rb
```diff
@@ -23,12 +66,62 @@ def rescue_strategy
...
+        end
+      end
+
+      def format_exception(exception)
+        { output[:proxy_task_id] =>
+          { :exception_class => exception.class.name,
+            :execption_message => exception.message } }
+      end
+
+      def with_connection_error_handling(event = nil)
```
iNecas: Shouldn't this be part of foreman-tasks, as it's more about the connection between forman-tasks and smart-proxy-dynflow, than about the remote execution itself?

app/lib/actions/remote_execution/run_proxy_command.rb
```diff
@@ -5,17 +5,60 @@ class RunProxyCommand < Actions::ProxyAction
...
+                              time,
+                              Timeout.new
+            metadata[:timeout] = time.to_s
+          end
+          super
+        end
+      end
+
+      def check_task_status
+        if output[:proxy_task_id]
```
iNecas: This logic could also go into https://github.com/theforeman/foreman-tasks/pull/132/files#diff-03dde90ec7e9caa322015dee15f4e956R51

app/lib/actions/remote_execution/run_proxy_command.rb
```diff
@@ -5,17 +5,60 @@ class RunProxyCommand < Actions::ProxyAction
...
+
+      def trigger_proxy_task
+        suspend do |suspended_action|
+          output[:metadata] ||= {}
+          unless metadata[:timeout]
+            time = Time.now + input[:connection_options][:timeout]
+            @world.clock.ping suspended_action,
+                              time,
+                              Timeout.new
+            metadata[:timeout] = time.to_s
```
iNecas: Another candidate for getting into foreman-tasks, and just providing the timeout data via some method

iNecas:
Would it be possible to have at least configuration option to set the default parameters?

iNecas:
How it should behave, when the host is not available at the moment (but will be in a minute?)

iNecas:
What should happen, when the task starts on the proxy and I restart the proxy then?

adamruzicka:
Having that possibility would be nice, I'll look into it.

With the default settings it would try running the command of the host 4 times in 15 second intervals and fail afterwards. The user can then see that it failed with 'host unreachable', bring the host up and rerun the job on the failed hosts. The user could also tweak the connection options for the job (reminds me there's no UI for that) to be tolerant to minor downtimes.

The way it is now it would wait for the timeout period for the proxy task to finish, after which it would ask the proxy about the tasks' status, get 404 as response and fail. This could be fixed by polling the proxy about the status more frequently or making the proxy report to foreman that it is back online and restarting all the tasks that should be running on that proxy.

iNecas:
+1 for making the intervals configurable (global configuration so far).

I'm ok with not retrying on host not being available.

I don't think having the same interval for script startup and script finishing is a good idea, I would suggest for now just applying the timeout on just start of it.

ares:
Please rebase the branch.

adamruzicka:
The issues have been addressed in theforeman/foreman-tasks#139

iNecas:
Tested and works as expected. Please squash and update dependency on foreman-tasks to 0.7.5, as I released new version with the https://github.com/theforeman/foreman-tasks/pull/139

iNecas:
Thanks @adamruzicka

PR #35, opened 2015-09-08
Fixes #11722 - disable select2 on input_type

iNecas:
The input type has a limited number of options and the select2 was causing
issues with dynamically generated fields.


ares:
ACK

iNecas:
The mysql tests failling for unrealted reasons, merging

PR #36, opened 2015-09-09
Fixes #11364 - Live output for template invocation and job invocation detail

ares:
I plan to add some optimizations but it does not prevent others from reviewing current work. Note it depends on 
* https://github.com/theforeman/foreman-tasks/pull/136
* https://github.com/theforeman/foreman-tasks/pull/137

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -24,12 +24,6 @@ module Actions
       def run(event = nil)
         super unless event == Dynflow::Action::Skip
       end
-
-      def output
-        result = super
-        result[:pending_count] = (result[:total_count] || 0) - (result[:failed_count] || 0) - (result[:success_count] || 0)
-        result
-      end
     end
   end
 end
--- a/app/views/template_invocations/_output_line.html.erb
+++ b/app/views/template_invocations/_output_line.html.erb
@@ -1,0 +1,8 @@
+<% @_line_counter ||= local_assigns[:line_counter] || 0 %>
+
+<%= content_tag :div, :class => 'line ' + output_line['output_type'], :data => { :timestamp => output_line['timestamp'] } do %>
+  <% output_line['output'].strip.split("\n").each do |output| %>
+    <%= content_tag(:span, (@_line_counter += 1).to_s.rjust(4).gsub(' ', '&nbsp;').html_safe + ':', :class => 'counter', :title => Time.at(output_line['timestamp'])) %>
+    <%= content_tag(:div, (output.empty? ? '&nbsp;' : output).html_safe, :class => 'content') %>
+  <% end %>
+<% end %>
--- a/app/views/template_invocations/show.html.erb
+++ b/app/views/template_invocations/show.html.erb
@@ -1,0 +1,39 @@
+<% title _('Detail of %s run') % @template_invocation.job_invocation.job_name %>
+<% stylesheet 'template_invocation' %>
+<% javascript 'template_invocation' %>
+
+<div id="title_action">
+  <div class="btn-toolbar pull-right">
+    <%= link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation)) %>
+    <%= link_to_function(_('Toggle command'), '$("div.preview").toggle()') %>
+    <%= link_to_function(_('Toggle STDERR'), '$("div.line.stderr").toggle()') %>
+    <%= link_to_function(_('Toggle STDOUT'), '$("div.line.stdout").toggle()') %>
+  </div>
+</div>
+
+<h3><%= _('Target: ') %><%= link_to(@host.name, host_path(@host)) %></h3>
+
+<div class="preview hidden">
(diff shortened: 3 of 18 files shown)


app/views/template_invocations/show.html.erb
```diff
@@ -0,0 +1,35 @@
+<% title (_('Detail of %{name} run on %{host}') % {:name => @template_invocation.job_invocation.job_name, :host => link_to(@host.name, host_path(@host)) }).html_safe %>
```
iNecas: is ` @template_invocation.job_invocation.job_name` html_safe?

ares: I thought we were working on remote execution :-) ...good catch

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -27,7 +27,6 @@ def run(event = nil)
 
       def output
         result = super
-        result[:pending_count] = (result[:total_count] || 0) - (result[:failed_count] || 0) - (result[:success_count] || 0)
         result
```
jlsherrill: was this intentional?  If so can you just call super?

ares: oops :-) removed the whole method, it's no longer necessary (after this was merged https://github.com/Dynflow/dynflow/commit/4c918383f1234c22decdc3e75d4e0b75af0c73c1)

app/views/template_invocations/show.html.erb
```diff
@@ -0,0 +1,39 @@
...
+<div id="title_action">
+  <div class="btn-toolbar pull-right">
+    <%= link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation)) %>
+    <%= link_to_function(_('Toggle command'), '$("div.preview").toggle()') %>
+    <%= link_to_function(_('Toggle STDERR'), '$("div.line.stderr").toggle()') %>
+    <%= link_to_function(_('Toggle STDOUT'), '$("div.line.stdout").toggle()') %>
+  </div>
+</div>
+
+<h3><%= _('Target: ') %><%= link_to(@host.name, hosts_path(@host)) %></h3>
```
iNecas: `s/hosts_path/host_path/`

iNecas:
with template `sleep 15; yum update -y`, I'm getting

```
undefined method `[]' for nil:NilClass

33:   <% if @template_invocation_task.pending? %>
34:     $(function() {
35:       delayed_refresh($("div.terminal").data('refresh-url'),
36:               {"since": "<%= @template_invocation_task.main_action.output.last['timestamp'] %>", "line_counter": $("div.terminal div.line").size()});
37:     });
38:   <% end %>
39: </script>

/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/views/template_invocations/show.html.erb:36:in `__home_inecas__rojects_ws_foreman_rex_foreman_remote_execution_app_views_template_invocations_show_html_erb___3304946041196362430_70149937491960'
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/controllers/template_invocations_controller.rb:11:in `show'
app/controllers/concerns/application_shared.rb:13:in `set_timezone'
app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
lib/middleware/catch_json_parse_errors.rb:9:in `call'
```

As there was no output generated yet

app/views/template_invocations/show.js.erb
```diff
@@ -0,0 +1,21 @@
...
+<% if params[:since] %>
+  target.append(lines);
+<% else %>
+  target.html(lines);
+<% end %>
+
+<% if @template_invocation_task.pending? %>
+  $(function() {
+    delayed_refresh($("div.terminal").data('refresh-url'),
+            {"since": "<%= @lines.last['timestamp'] %>", "line_counter": $("div.terminal div.line").size()});
```
iNecas: `@lines` can be empty, when there were no new lines since last time: it should fallback to previous `since` value

iNecas:
When hitting the issues, I've noticed that the errors from ajax calls are not being shown to the user and the reloading just stopped.

app/views/template_invocations/_output_line.html.erb
```diff
@@ -0,0 +1,8 @@
+<% @_line_counter ||= local_assigns[:line_counter] || 0 %>
```
iNecas: why using instance variable vs. local one?

ares: Because I need to keep the value between `_output_line` renders, local one would always start from 0. Unfortunately we can't reuse built-in counter because some lines contain multiple rows...

iNecas: Thanks for explanation

iNecas:
I've released foreman-tasks 0.7.4, please update the depednency in the spec, squash and rebase. ACK after doing so and passing tests

ares:
The test issue is unrelated to this PR, keeping as separate commit. I bumped the foreman-tasks version and using the ACK to merge.

PR #39, opened 2015-09-15
refs #11805 - use before_create hook instead of after_create

stbenjam:
If you want it for ~ 1.9


iNecas:
I've simplified the workaround a bit  #40, so that we don't duplicate the code: it should work with foreman with https://github.com/theforeman/foreman/pull/2723

PR #40, opened 2015-09-17
Refs #11805 - use before_create instead of after_create for default templates

iNecas:
Otherwise we hit issues described in rails/rails#12180 (comment), will not be
needed once https://github.com/theforeman/foreman/pull/2723 is in and we don't
support older versions

### DIFF
--- a/app/models/concerns/foreman_remote_execution/taxonomy_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/taxonomy_extensions.rb
@@ -4,6 +4,13 @@ module ForemanRemoteExecution
 
     included do
       has_many :job_templates, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'JobTemplate'
+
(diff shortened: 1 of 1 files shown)


iNecas:
Since there is a demand for the plugin to work with stable versions of Foreman and Katello, let's fix the texonomies issues here as well until we run only the foreman with fixed taxonomies (in terms of templates assignment :)

stbenjam:
ACK I guess, although I don't learn anything from you if you take over the PR, could've just suggested skip_callback on mine.

iNecas:
@stbenjam sry about that, this way just looked faster (given I had the patch in my code-base and it was working), I will try to play more nicely next time :v:

stbenjam:
Thanks :-) I know this was just a little thing but you (and @ares) are quite a bit more talented than me, so I appreciate the opportunity to get good feedback from you, when there's time.

ares:
[test] the issue is unrelated to the PR, should be fixed in master now

PR #42, opened 2015-09-24
Fixes #11864 - introduce remote execution status

ares:


stbenjam:
Tests run against 1.9, so this would need to be compatible with both.

ares:
I thought it was (https://github.com/theforeman/foreman_remote_execution/pull/42/files#diff-24c9f52e5c8d0dc8ba4b5d2b7ec5de44R68) but I added host extensions later, will wrap that in condition as well.

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -4,16 +4,27 @@ module HostExtensions
...
-      end
+    def execution_status_label(options = {})
+      @execution_status_label ||= get_status(HostStatus::ExecutionStatus).to_label(options)
     end
+
+    # module ClassMethods
+    #   # create or overwrite class methods...
+    #   def class_method_name
+    #   end
+    # end
```
stbenjam: Also please remove this boilerplate

ares:
OK, I think this is a candidate for 1.10 bump, I'll remove not working 1.9 support. Let's keep it open until iteration 2 is done. The issue is that the preloader (in production env) tries to load the execution status which can't find its superclass.

stbenjam:
I think you could wrap the entire class in a check, something like this: https://github.com/theforeman/foreman_salt/blob/4d3af2f4d4b4f4cf68a267a4b8fc389f47e9b9c4/app/lib/actions/foreman_salt/report_import.rb

Up to you though

ares:
yeah, but I think this falls into unnecessary hack category :-)

ares:
@stbenjam could you please change the tests to run against 1.10? This PR is no longer blocked.

app/models/host_status/execution_status.rb
```diff
@@ -0,0 +1,38 @@
+module HostStatus
```
ares: this should be also under `ForemanOpenscap` module and hence in app/models/foreman_openscap/host_status (host status might not be needed, but leaving this up to you)

stbenjam:
Develop or 1.10? https://github.com/theforeman/foreman-infra/pull/191

ares:
I think 1.10 would be better, but please wait until new release, we realized we'll have to do 0.0.8 that will be still 1.9 compatible (a lot of issues with 0.0.7)

stbenjam:
[test]

stbenjam:
Maybe try what I have here: https://github.com/theforeman/foreman_remote_execution/pull/55/files#diff-24c9f52e5c8d0dc8ba4b5d2b7ec5de44R107

I don't think we need to extend anything but Host::Managed, even unmanaged hosts are still Host::Managed with managed=false.

ares:
I removed Host::Base, but I think we should extend all other descendants (I know only about Host::Discovery but there might be others in future).

stbenjam:
Ok, we might need a separate managed extensions then if we extend managed methods only (like params in my other PR).  Or if aspects is happening anytime soon it can provide a way for aspect parameters.

Anyway, looks like just a typo now.

ares:
[test]

stbenjam:
I bet if you were to change engine.rb to only extend Host::Managed, tests would pass

ares:
should be green after rebasing on top of https://github.com/theforeman/foreman_remote_execution/pull/55 or when https://github.com/theforeman/foreman/pull/2868 gets in

stbenjam:
@ares Needs rebase due to #55

app/models/host_status/execution_status.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def self.status_name
+    N_('Execution')
+  end
+
+  def to_label(options = {})
+    case to_status(options)
+      when OK
+        N_('Last execution succeeded')
```
stbenjam: Sorry one last thought, a host with no executions will have a status that says `Last execution succeeded`, would that be confusing?

Does it make sense to have a different status for that? Like "No executions" that's rated as a Global::OK?

ares: You're right but two different values would result in different numbers so when you search for hosts that does not have any failed execution you'd have to search for `OK or NONE`. So how about changing just to wording to "Last execution succeeded or nothing executed yet"?

ares: or maybe the status should be relevant only if there was some execution?

stbenjam: That description is really long :open_mouth:  Unless you can think of a way to phrase it in a shorter way, I think I prefer making it relevant or not.

ares:
So I changed it to be irrelevant until there's at least one task. In case another status calls refresh and this task becomes relevant (there's at least one task but none is stopped) we print different message while still remaining OK.

stbenjam:
Works great ! ACK

ares:
Thanks, merging

PR #44, opened 2015-09-24
Refs #10755 - Recurring actions, job invocation task groups

adamruzicka:
requires theforeman/foreman-tasks#141

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -1,9 +1,13 @@
 class JobInvocationsController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
 
+  before_filter :find_or_create_triggering, :only => [:create, :refresh]
+
   def new
     @composer = JobInvocationComposer.new.compose_from_params(
       :host_ids => params[:host_ids],
+      :job_invocation => {
+      },
       :targeting => {
         :targeting_type => Targeting::STATIC_TYPE,
         :bookmark_id => params[:bookmark_id]
@@ -24,16 +28,9 @@ class JobInvocationsController < ApplicationController
 
   def create
     @composer = JobInvocationComposer.new.compose_from_params(params)
-    action = ::Actions::RemoteExecution::RunHostsJob
     if @composer.save
       job_invocation = @composer.job_invocation
-      if job_invocation.trigger_mode == :future
-        ForemanTasks.delay action,
-                           job_invocation.delay_options,
-                           job_invocation
-      else
-        ForemanTasks.async_task(action, job_invocation)
-      end
+      @composer.triggering.trigger(::Actions::RemoteExecution::RunHostsJob, job_invocation)
       redirect_to job_invocation_path(job_invocation)
     else
       render :action => 'new'
@@ -42,13 +39,13 @@ class JobInvocationsController < ApplicationController
 
   def show
     @job_invocation = resource_base.find(params[:id])
-    @auto_refresh = @job_invocation.last_task.try(:pending?)
+    @auto_refresh = @job_invocation.task.try(:pending?)
     hosts_base = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
     @hosts = hosts_base.search_for(params[:search], :order => params[:order] || 'name ASC').paginate(:page => params[:page])
   end
 
   def index
-    @job_invocations = resource_base.search_for(params[:search]).paginate(:page => params[:page]).with_last_task.order(params[:order] || 'id DESC')
+    @job_invocations = resource_base.search_for(params[:search]).paginate(:page => params[:page]).with_last_task.order(params[:order] || '"job_invocations"."id" DESC')
   end
 
   # refreshes the form
@@ -68,6 +65,10 @@ class JobInvocationsController < ApplicationController
   end
 
   private
+
+  def find_or_create_triggering
+    @triggering ||= ::ForemanTasks::Triggering.new_from_params(params[:triggering])
+  end
 
   def action_permission
     case params[:action]
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -3,14 +3,17 @@ module Actions
     class RunHostsJob < Actions::ActionWithSubPlans
 
       middleware.use Actions::Middleware::BindJobInvocation
+      middleware.use Actions::Middleware::RecurringLogic
 
-      def delay(delay_options, job_invocation)
+      def delay(delay_options, job_invocation, locked = false)
+        task.add_missing_task_groups(job_invocation.task_group)
         job_invocation.targeting.resolve_hosts! if job_invocation.targeting.static?
-        action_subject(job_invocation)
-        super(delay_options, job_invocation, true)
+        super delay_options, job_invocation, locked
       end
 
       def plan(job_invocation, locked = false, connection_options = {})
+        job_invocation.task_group.save! if job_invocation.task_group.try(:new_record?)
+        task.add_missing_task_groups(job_invocation.task_group) if job_invocation.task_group
         action_subject(job_invocation) unless locked
         job_invocation.targeting.resolve_hosts! if job_invocation.targeting.dynamic? || !locked
         input.update(:job_name => job_invocation.job_name)
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -74,15 +74,7 @@
     </fieldset>
   <% end %>
 
-  <div class="form-group trigger_mode">
-    <legend><%= _('Schedule Job') %></legend>
-    <%= radio_button_f f, :trigger_mode, :value => 'immediate', :text => _("Run Job Now"), :class => 'trigger_mode_selector' %>
-    <%= radio_button_f f, :trigger_mode, :value => 'future', :text => _("Schedule Future Job"), :class => 'trigger_mode_selector' %>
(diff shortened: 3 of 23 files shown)


ares:
request for rebase :-)

app/controllers/job_invocations_controller.rb
```diff
@@ -31,6 +31,11 @@ def create
         ForemanTasks.delay action,
                            job_invocation.delay_options,
                            job_invocation
+      elsif job_invocation.trigger_mode == :recurring
```
iNecas: `case job_invocation.trigger_mode` might work better than if/elsif here

app/views/job_invocations/_form.html.erb
```diff
@@ -75,10 +75,59 @@
...
     <%= radio_button_f f, :trigger_mode, :value => 'immediate', :text => _("Run Job Now"), :class => 'trigger_mode_selector' %>
     <%= radio_button_f f, :trigger_mode, :value => 'future', :text => _("Schedule Future Job"), :class => 'trigger_mode_selector' %>
+    <%= radio_button_f f, :trigger_mode, :value => 'recurring', :text => _("Set up recurring Job"), :class => 'trigger_mode_selector' %>
     <fieldset id="trigger_mode_future" class="trigger_mode_form <%= 'hidden' if @composer.job_invocation.trigger_mode != :future %>">
       <%= text_f f, :start_at, :label => _('Start at') %>
       <%= text_f f, :start_before, :label => _('Start before') %>
     </fieldset>
+    <fieldset id="trigger_mode_recurring" class="trigger_mode_form <%= 'hidden' if @composer.job_invocation.trigger_mode != :recurring %>">
+      <%= selectable_f f, :input_type, %w(cronline monthly weekly daily hourly), {}, :label => _("Repeats") %>
+      <%= f.fields_for 'recurring_options' do |recurring_options_fields| %>
```
iNecas: A potential helper that  might go info foreman-tasks for the field-set for recurring logic?

db/migrate/20150923125825_add_job_invocation_task_group.rb
```diff
@@ -0,0 +1,7 @@
+class AddJobInvocationTaskGroup < ActiveRecord::Migration
+  def change
+    alter_table :foreman_tasks_task_groups do |t|
+      t.integer :job_invocation_id, index: true
```
iNecas: This association would be better on the `job_invocations` part to prevent the sparse tables

iNecas:
Did another set of testing, the new version, with middleware works much intuitively.

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -3,14 +3,17 @@ module RemoteExecution
     class RunHostsJob < Actions::ActionWithSubPlans
 
       middleware.use Actions::Middleware::BindJobInvocation
+      middleware.use Actions::Middleware::RecurringLogic
 
-      def delay(delay_options, job_invocation)
+      def delay(delay_options, job_invocation, locked = false)
+        task.add_missing_task_groups(job_invocation.task_group)
```
iNecas: Currently, the iterations after the first one don't have their own job invocation instance, so that the events are not shown in job invocations index, so we then can't track status from the job perspective then. We should check here, and in case the job has already a task, we should clone the invocation, rather than use the current one.

Also, since we're mapping the job invocation to 0-1 tasks, we should rename 'last_task_id', to just 'task_id'.

iNecas:
It should be nice, to be able to get to the recurring logic details, where one would be able to get to the next jobs of belonging the the recurring logic group etc.

ares:
This needs rebase :-(

adamruzicka:
@ares @iNecas @stbenjam : rebased and ready for another round of reviews

iNecas:
Putting invalid date raise an exception instead of correct validation error at the right field (try inserting blank cronline)

iNecas:
Refreshing the form (updating the targeting search query, changing the job type or template), makes the form for schedule job disappear

iNecas:
When using non-cronline form, I'm getting

```
undefined method `cronline_hash' for #<Class:0x00000008ac0848>
Rails.root: /home/inecas/Projects/ws/foreman-rex/foreman

/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/models/foreman_tasks/recurring_logic.rb:107:in `new_from_triggering'
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/controllers/job_invocations_controller.rb:39:in `create'
app/controllers/concerns/application_shared.rb:13:in `set_timezone'
app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
lib/middleware/catch_json_parse_errors.rb:9:in `call'
```

adamruzicka:
undefined_method error fixed [here](https://github.com/adamruzicka/foreman-tasks/commit/d90616ac63bb7d71d799e649ee09dc8a7aaae697)

adamruzicka:
@iNecas all the comments should be addressed

iNecas:
When using `hourly` recurring option, I'm getting:

```
invalid strptime format - `%Y-%m-%d %H:%M'
Rails.root: /home/inecas/Projects/ws/foreman-rex/foreman

/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/models/foreman_tasks/triggering.rb:78:in `parse_start_at!'
/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/models/foreman_tasks/triggering.rb:9:in `block in <class:Triggering>'
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/models/job_invocation_composer.rb:45:in `save'
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/controllers/job_invocations_controller.rb:31:in `create'
app/controllers/concerns/application_shared.rb:13:in `set_timezone'
app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
lib/middleware/catch_json_parse_errors.rb:9:in `call'
```

Debugging showed that the `start_at_raw` was empty.

db/migrate/20150923125825_add_job_invocation_task_group.rb
```diff
@@ -0,0 +1,9 @@
+class AddJobInvocationTaskGroup < ActiveRecord::Migration
+  def up
+    add_column :job_invocations, :task_group_id, :integer, :index => true
```
iNecas: Foreign key?

adamruzicka:
Should be fixed in https://github.com/adamruzicka/foreman-tasks/commit/20cf8ea22031d3b25afe5f756f6cf2316b8555e7

db/migrate/20151116105412_add_triggering_to_job_invocation.rb
```diff
@@ -0,0 +1,9 @@
+class AddTriggeringToJobInvocation < ActiveRecord::Migration
+  def up
+    add_column :job_invocations, :triggering_id, :integer, :index => true
```
iNecas: Foreign key again?

db/migrate/20150923125825_add_job_invocation_task_group.rb
adamruzicka: Yes. Should I add the database constraint or can we allow a job invocation to have no task group?

db/migrate/20151116105412_add_triggering_to_job_invocation.rb
adamruzicka: Again, yes. A job invocation could be created without a triggering so it made sense to me not to enforce the association

db/migrate/20150923125825_add_job_invocation_task_group.rb
iNecas: AFAIK you can have foreign key on nullable fields -  it should check the contraint just on the case, when the field is entered.

adamruzicka: You're right, I'll add the constraints

iNecas:
Looks better every iteration: from the last testing (some comments might apply to the foreman-tasks PR):

* future execution doesn't have a default values set - this was quite useful to have set to current time, instead of having to write the time on my own, I would welcome having that back
* related resources in the job invocation seems strange (in the recurring logics it might have some sense, although even there, it would be nice just to be able to link to the job invocation detail), having 'Recurring logic' tab there would be better IMO
* canceling the recurring logic leads to 'Template missing' error message
* `Finished: running` seems strange, `status` might be better, also, I would perhaps choose `Active` over `Running`, as the `Running` might feel the task actually running at the moment and is a bit misleading

adamruzicka:
> - future execution doesn't have a default values set - this was quite useful to have set to current time, instead of having to write the time on my own, I would welcome having that back
> - canceling the recurring logic leads to 'Template missing' error message
> - Finished: running seems strange, status might be better, also, I would perhaps choose Active over Running, as the Running might feel the task actually running at the moment and is a bit misleading

addressed in https://github.com/adamruzicka/foreman-tasks/commit/8ee423e9e16441778996d3314b33831a5b6b95cb

> - related resources in the job invocation seems strange (in the recurring logics it might have some sense, although even there, it would be nice just to be able to link to the job invocation detail), having 'Recurring logic' tab there would be better IMO

#### In recurring logic details
Shows type of the resource and count per type, links to job invocations index filtered by the recurring logic id
in recurring logic

#### In job invocation details
Should this show the resources related to the recurring logic too?
in job invocation

iNecas:
The default time for future execution works for new invocations, but still missing when using the re-run. Minor, but still nice to have enhancement.

iNecas:
Waiting for https://github.com/theforeman/foreman-tasks/pull/141 to get released

iNecas:
Waiting for rubocop and squash

iNecas:
No other reasons for not getting this into master. Merging now. Thanks @adamruzicka

PR #46, opened 2015-09-25
Fixes #11895 - Better error handling for the live output

iNecas:
Also include debug data in the output (for example when the host, or the proxy
is not approachable)

### DIFF
--- a/app/controllers/template_invocations_controller.rb
+++ b/app/controllers/template_invocations_controller.rb
@@ -7,15 +7,10 @@ class TemplateInvocationsController < ApplicationController
     @template_invocation_task = ForemanTasks::Task.find(params[:id])
     @template_invocation = @template_invocation_task.locks.where(:resource_type => 'TemplateInvocation').first.try(:resource)
     @host = @template_invocation_task.locks.where(:resource_type => 'Host::Managed').first.try(:resource)
-
-    respond_to do |format|
-      format.html
-      format.js do
-        @lines = @template_invocation_task.main_action.output
-        if params[:since]
-          @lines = @lines.select { |o| o['timestamp'].to_f > params[:since].to_f }
-        end
-      end
-    end
+    @auto_refresh = @template_invocation_task.pending?
+    @since = params[:since].to_f if params[:since].present?
+    @line_sets = @template_invocation_task.main_action.live_output
+    @line_sets = @line_sets.drop_while { |o| o['timestamp'].to_f <= @since } if @since
+    @line_counter = params[:line_counter].to_i
   end
 end
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -111,7 +111,24 @@ module RemoteExecutionHelper
                          :disabled => !task.cancellable?,
                          :method => :post)
     end
-    return button_group(*buttons)
+    return buttons
+  end
+
+  def template_invocation_task_buttons(task)
+    buttons = []
+    if authorized_for(:permission => :view_foreman_tasks, :auth_object => task)
+      buttons << link_to(_("Task Details"), foreman_tasks_task_path(task),
+                         :class => "btn btn-default",
+                         :title => _('See the task details'))
+    end
+    if authorized_for(:permission => :edit_foreman_tasks, :auth_object => task)
+      buttons << link_to(_("Cancel Job"), cancel_foreman_tasks_task_path(task),
+                         :class => "btn btn-danger",
+                         :title => _('Try to cancel the job on a host'),
+                         :disabled => !task.cancellable?,
+                         :method => :post)
+    end
+    return buttons
   end
 
   def link_to_invocation_task_if_authorized(invocation)
--- a/app/lib/actions/remote_execution/run_proxy_command.rb
+++ b/app/lib/actions/remote_execution/run_proxy_command.rb
@@ -39,18 +39,56 @@ module Actions
       end
 
       def live_output
-        output = self.output[:proxy_output].present? ? self.output[:proxy_output] : fetch_output['output']
-        output_records(output)
+        records = connection_messages
+        if !task.pending?
+          records.concat(finalized_output)
+        else
+          records.concat(current_proxy_output)
+        end
+        records.sort_by { |record| record['timestamp'].to_f }
       end
 
       private
 
-      def output_records(base_output)
-        base_output['result']
+      def connection_messages
+        metadata.fetch(:failed_proxy_tasks, []).map do |failure_data|
+          format_output(_('Initialization error: %s') % "#{failure_data[:exception_class]} - #{failure_data[:exception_message]}", 'debug', failure_data[:timestamp])
+        end
       end
 
-      def fetch_output
-        proxy.status_of_task(output[:proxy_task_id])['actions'].detect { |action| action['class'] == proxy_action_name }
+      def current_proxy_output
+        return [] unless output[:proxy_task_id]
+        proxy_data = proxy.status_of_task(output[:proxy_task_id])['actions'].detect { |action| action['class'] == proxy_action_name }
+        proxy_data.fetch('output', {}).fetch('result', [])
+      rescue => e
+        ::Foreman::Logging.exception("Failed to load data for task #{task.id} from proxy #{ input[:proxy_url] }", e)
+        [exception_to_output(_("Error loading data from proxy"), e)]
+      end
+
+      def finalized_output
+        records = []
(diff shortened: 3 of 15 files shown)


iNecas:
@ares I'm going to add some tests as well, to cover the logic, but a review in the meantime would be nice

app/helpers/remote_execution_helper.rb
```diff
@@ -82,6 +82,23 @@ def job_invocation_task_buttons(task)
...
+                         :title => _('See the task details'))
+    end
+    if authorized_for(:permission => :edit_foreman_tasks, :auth_object => task)
+      buttons << link_to(_("Cancel Job"), cancel_foreman_tasks_task_path(task),
+                         :class => "btn btn-danger",
+                         :title => _('Try to cancel the job on a host'),
+                         :disabled => !task.cancellable?,
+                         :method => :post)
+    end
+    return button_group(*buttons)
```
ares: I'd prefer not wrapping buttons in group in this helper, it's hard to reuse/customize later (especially from plugins)

app/views/template_invocations/_output_line.html.erb
```diff
@@ -1,8 +1,6 @@
-<% @_line_counter ||= local_assigns[:line_counter] || 0 %>
-
 <%= content_tag :div, :class => 'line ' + output_line['output_type'], :data => { :timestamp => output_line['timestamp'] } do %>
   <% output_line['output'].strip.split("\n").each do |output| %>
-    <%= content_tag(:span, (@_line_counter += 1).to_s.rjust(4).gsub(' ', '&nbsp;').html_safe + ':', :class => 'counter', :title => Time.at(output_line['timestamp'])) %>
+    <%= content_tag(:span, (@line_counter += 1).to_s.rjust(4).gsub(' ', '&nbsp;').html_safe + ':', :class => 'counter', :title => (output_line['timestamp'] && Time.at(output_line['timestamp']))) %>
```
ares: so this is changing variables defined out of partial, I'm not sure if it's better or worse than the previous counter hack :-)

app/controllers/template_invocations_controller.rb
```diff
@@ -7,15 +7,14 @@ def show
...
-        @lines = @template_invocation_task.main_action.output
-        if params[:since]
-          @lines = @lines.select { |o| o['timestamp'].to_f > params[:since].to_f }
-        end
-      end
+    @auto_refresh = @template_invocation_task.pending?
+    @lines = @template_invocation_task.main_action.live_output
+    @since = params[:since].to_f if params[:since].present?
+    if @since
+      @line_counter = params[:line_counter].to_i
```
ares: maybe take this one out of condition like this

`@line_counter = params[:line_counter].to_i`

result is the same and line_counter might not be dependent only `since` parameter later

app/lib/actions/remote_execution/run_proxy_command.rb
```diff
@@ -39,18 +39,56 @@ def exit_status
       end
 
       def live_output
-        output = self.output[:proxy_output].present? ? self.output[:proxy_output] : fetch_output['output']
-        output_records(output)
+        records = connection_messages
```
ares: does this mean that failing messages will be always first regardless of timestamp?

ares:
In general looks good, just few nits or questions. I'd like to test it as well after it's complete.

iNecas:
When testing, make sure you have also https://github.com/theforeman/foreman-tasks/pull/142 (especially when testing the case, when the proxy is down)

app/views/template_invocations/_output_line.html.erb
iNecas: Since the logic around the `@line_counter` is a bit more conplex now (counting on a case where timestamp might be nil), it was just logical step to move that logic into the controller. It's better for sure :)

app/lib/actions/remote_execution/run_proxy_command.rb
iNecas: The assumption is the connection error will always happen sooner, than other execution output: in other words, once the connection is established, no other connection errors are expected to occur. But I can make a final sort, just in case

app/controllers/template_invocations_controller.rb
iNecas: it is: in case the timestamp at some line was nil, you don't filter anything, therefore you always start from 0. I added this at the beginning of the PR, where not all lines had timestamps. I can revisit that: and make sure the timestamps are always present, what about that?

app/lib/actions/remote_execution/run_proxy_command.rb
iNecas: Added the sort

iNecas:
@ares I've addressed all the issues I accepted as issues :) + fixed an issue with line numbers (see "fix line counting" commit), working on tests now, but should be ready for rereview/testing

iNecas:
Tests added

ares:
[Tests broken](http://ci.theforeman.org/job/test_plugin_pull_request/1472/database=mysql,label=fast,ruby=1.9.3/testReport/junit/(root)/ForemanRemoteExecution__RunProxyCommandTest___live_output__when%20there%20was%20some%20connection%20error%20while%20running%20the%20command/test_0001_reports_the_failure_as_part_of_the_live_output/)

```
Error Message
<"Initialization error: RuntimeError - Connection error"> expected but was
<"Initialization error:  - ">. (test_0001_reports the failure as part of the live output)
/var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/1.9.3/plugin/test/unit/actions/run_proxy_command_test.rb:110    
```

iNecas:
The tests are failing due to https://github.com/theforeman/foreman-tasks/pull/142 not being released yet, please review regardless, so that I could iterate on this (I don't want to release foreman-tasks yet just in case some comments on this PR would mean changes on foreman-tasks side)

iNecas:
@ares I've fixed the issue we run into yestraday (with the exception from RestClient), as part of the work, I've also refactored foreman-tasks to optimize the humanized data calculation https://github.com/theforeman/foreman-tasks/pull/143

ares:
Running tests locally reveals some issues, might be just my setup though... does following failures ring any bell?

```
1) Failure:
Api::V2::JobInvocationsControllerTest#test_0003_should create valid without template_id [/home/ares/Projekty/Zdrojaky/foreman_remote_execution/test/function
al/api/v2/job_invocations_controller_test.rb:31]:
<"job name 6"> expected but was
<nil>.

  2) Failure:
Api::V2::JobInvocationsControllerTest#test_0004_should create valid with template_id [/home/ares/Projekty/Zdrojaky/foreman_remote_execution/test/functional/
api/v2/job_invocations_controller_test.rb:40]:
<"job name 8"> expected but was
<nil>.

  3) Error:
InputTemplateRenderer::renderer for template with variable input used::with matching input defined::rendering::with host specified::with existing variable i
mplemented as host parameter::preview#test_0001_should render preview:
SystemStackError: stack level too deep
    /home/ares/Projekty/Zdrojaky/foreman_remote_execution/app/models/concerns/foreman_remote_execution/host_extensions.rb:36

  4) Error:
InputTemplateRenderer::renderer for template with variable input used::with matching input defined::rendering::with host specified::with existing variable i
mplemented as host parameter::rendering#test_0001_renders the value from host parameter:
SystemStackError: stack level too deep
    /home/ares/Projekty/Zdrojaky/foreman_remote_execution/app/models/concerns/foreman_remote_execution/host_extensions.rb:36
```

this one is likely caused by my local plugins
```
  5) Error:
Targeting::can delete a host#test_0001_anonymous:
RuntimeError: Executing dynflow action inside a transaction is not a good idea
    /home/ares/Projekty/Zdrojaky/foreman-tasks/app/models/foreman_tasks/concerns/action_triggering.rb:122:in `ensure_not_in_transaction!'
    /home/ares/Projekty/Zdrojaky/foreman-tasks/app/models/foreman_tasks/concerns/action_triggering.rb:109:in `dynflow_task_wrap'
    /home/ares/Projekty/Zdrojaky/foreman-tasks/app/models/foreman_tasks/concerns/action_triggering.rb:34:in `destroy'
    /home/ares/Projekty/Zdrojaky/foreman_remote_execution/test/unit/targeting_test.rb:61:in `block (3 levels) in <top (required)>'
```

iNecas:
I don't see any of them locally (running against foreman-1.9 stable now)

ares:
Ok, we have ACK also from Mr. Jenkins, merging then. Thanks @iNecas!

PR #47, opened 2015-09-29
fixes #11620 - create template in one step

stbenjam:
Small refactor, we can create templates locked initially now that https://github.com/theforeman/foreman/pull/2652 is merged and released.

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -38,16 +38,14 @@ class JobTemplate < ::Template
   end
   self.table_name = 'templates'
 
-  # Import a template ERB, with metadata in the first YAML comment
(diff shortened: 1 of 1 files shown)


ares:
Cool! I suppose this implies foreman >= 1.10 so either add this to engine.rb or wait with merge after another 1.10 PR (e.g. https://github.com/theforeman/foreman_remote_execution/pull/42). Otherwise :+1:

stbenjam:
@ares Oh it should be OK for 1.9.1+: https://github.com/theforeman/foreman/blob/1.9-stable/app/models/template.rb#L9

ares:
Ah I wasn't aware that this was backported. +1 for merging then :-)

PR #50, opened 2015-10-01
Fixes #12024 - balancer recovers if no proxy is usable

ares:


ares:
@stbenjam would you mind reviewing?

stbenjam:
Oh, didn't realize min_by could return nil there.  Looks good - can you add a test?

ares:
sure, added

stbenjam:
APJ

ares:
Result is read because of rubocop findings in already merged code. I disable the cop since I don't see anything wrong with using inject (it suggests to use each_with_object instead)

```
Offenses:

/var/lib/workspace/workspace/test_plugin_pull_request/database/postgresql/label/fast/ruby/1.9.3/plugin/app/services/proxy_load_balancer.rb:19:13: C: Use each_with_object instead of inject.
    proxies.inject({}) do |result, proxy|
            ^^^^^^
```

stbenjam:
Rubocop weirdness, it was green on the PR that added the inject.

Anyway,  APJ.

ares:
It's green, thanks for review!

PR #51, opened 2015-10-07
Fixes #12084 - adds default order for hosts in job invocation

ares:


ares:
closing, the scoped search support was easier than I thought

PR #52, opened 2015-10-08
Fixes #12065 - refresh hosts in invocation detail

ares:


app/assets/javascripts/template_invocation.js
```diff
@@ -46,6 +46,7 @@ function job_invocation_form_binds() {
 }
 
 function delayed_refresh(url, data){
+
```
ohadlevy: nitpick: whitepsace

app/assets/javascripts/template_invocation.js
```diff
@@ -57,3 +58,16 @@ function delayed_refresh(url, data){
...
+  return {
+    hosts_needs_refresh: $("div#hosts").data('needs-refresh'),
+    host_ids_needing_status_update: fetch_ids_of_hosts('status'),
+    host_ids_needing_provider_update: fetch_ids_of_hosts('provider'),
+    host_ids_needing_actions_update: fetch_ids_of_hosts('actions')
+  }
+}
+
+function fetch_ids_of_hosts(attribute){
+  return _.map($('div#hosts td.host_' + attribute + '[data-needs-refresh="true"]'), function(elem) { return $(elem).data('id') });
```
ohadlevy: nitpick: maybe refresh_require=true?

app/views/job_invocations/show.js.erb
```diff
@@ -1,6 +1,22 @@
 $('div#status_chart').html('<%=j job_invocation_chart(@job_invocation) %>');
 $('div#status').flot_pie();
 
+<% if params[:hosts_needs_refresh] == 'true' && @job_invocation.resolved? %>
+  var hosts_tab = $('div#hosts')
+  hosts_tab.html('<%=j render('tab_hosts', :job_invocation => @job_invocation, :hosts => @hosts) %>');
+  hosts_tab.data('needs-refresh', '');
```
ohadlevy: I dont know why, but i really dislike show.js.erb :)

ares: do you know any better way that does not duplicate the view logic? the only alternative that I can imagine is to render some template html and send just data from server to client, then use the template to clone html and fill in the data but 
  1. we'd duplicate at least translation data or logic like "status 1 means Failed with cross icon"
  2. I saw too many issues with this approach in existing Foreman forms (compute profiles, interface, parameters)

Probably best would be to use some templating language that can be used in both ruby na js (e.g. [mustache](https://mustache.github.io/)) but I don't want to introduce another templating solution :-)

ares:
tests failed because of some dependency installation issue on 1.9.3

iNecas:
It replaces #51, right?

ares:
correct

iNecas:
The build issue should go away now https://github.com/theforeman/foreman/pull/2786

iNecas:
Tested and works nicely, ACK

ares:
Thanks, merging

PR #53, opened 2015-10-08
Fixes #12064 - allow empty values for non-required inputs

iNecas:
Also, make the API error message more descriptive, when some issue is there on
provided inputs.

### DIFF
--- a/app/models/concerns/foreman_remote_execution/errors_flattener.rb
+++ b/app/models/concerns/foreman_remote_execution/errors_flattener.rb
@@ -1,0 +1,62 @@
+module ForemanRemoteExecution
+  module ErrorsFlattener
+    def flattened_validation_exception
+      ActiveRecord::RecordNotSaved.new(I18n.t("activerecord.errors.messages.record_invalid", :errors => flattened_errors.join(', ')))
+    end
+
+    def flattened_errors
+      errors = Hash.new { |h, k| h[k] = [] }
+      self.errors.keys.each do |key|
+        messages = self.errors[key]
+        invalid_objects = invalid_objects_for_attribute(key)
+        if invalid_objects.blank?
+          errors[key] = messages
+        else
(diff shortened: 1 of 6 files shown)


app/models/concerns/foreman_remote_execution/errors_flattener.rb
```diff
@@ -0,0 +1,62 @@
...
+        end
+      end
+      errors.map { |key, messages| self.errors.full_message(key, messages.join(', ')) }
+    end
+
+    private
+
+    def invalid_objects_for_attribute(attribute)
+      if self.respond_to?(attribute)
+        invalid_object = self.send(attribute)
```
ares: I think `public_send` should be enough

app/models/template_invocation.rb
```diff
@@ -16,9 +22,10 @@ def to_action_input
   private
 
   def provides_required_input_values
-    required_input_ids = self.template.template_inputs.where(:required => true).pluck(:id)
-    unless (required_input_ids - self.input_values.map(&:template_input_id)).empty?
-      errors.add(:base, _("Not all required inputs have values."))
+    required_inputs = self.template.template_inputs.where(:required => true)
```
ares: `.includes(:input_values)` might be handy because of line below

ares: dismiss, I misread that

ares:
Tested, works nice, just one nit (`public_send`) and good to go.

iNecas:
[test]

app/models/concerns/foreman_remote_execution/errors_flattener.rb
iNecas: Good point, new rubocop would have found that as well https://github.com/bbatsov/rubocop/commit/242a761e0cafc2e4634bd62a3362bed2b70fcec2 (but we need to rely on the one in foreman right now)

PR #54, opened 2015-10-09
Refs #12065 - fix foreman 1.9 compatibility with refresh js

iNecas:
Foreman 1.9 didn't include underscore.js, but we are ok with jQuery in this
case.


iNecas:
This is the commit that introduces underscore.js, and it's post 1.9 https://github.com/theforeman/foreman/commit/09ce8a634bd5924de5f9c1d865a9a765a5b0174f

ares:
oops, good catch, ACK

PR #57, opened 2015-10-20
Fixes #12317 - Missing title on new job invocation

dLobatog:
**Problem**
When you click on 'Run Job' on a host, there is no title on that page.

**Solution**
Add a title to the new page for Job invocations.


iNecas:
Looks good, waiting for theforeman-bot to be happy about the PR

dLobatog:
Updated!

ares:
I think this should replace h1 few lines below since we'd have two headers
title

stbenjam:
Thanks @dLobatog !

PR #58, opened 2015-10-20
Fixes #12246 - Validate job_name on template creation

dLobatog:
**Problem**
You can create job templates without a job_name currently. This leads to a situation where you cannot use the template in a job invocation because it lacks a job_name. 

**Solution**
Validate that the job_name is present on save on every job template creation.

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -28,6 +28,7 @@ class JobTemplate < ::Template
                   end
                 }
 
+  validates :job_name, :presence => true, :unless => ->(job_template) { job_template.snippet }
   validates :provider_type, :presence => true
   validate :provider_type_whitelist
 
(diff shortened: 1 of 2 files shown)


iNecas:
Looks good. Would you mind pairing it with a foreman_remote_execution redmine ticket?

app/models/job_template.rb
```diff
@@ -28,6 +28,7 @@ class JobTemplate < ::Template
                   end
                 }
 
+  validates :job_name, :presence => true
```
stbenjam: Should check to make sure its not a snippet

dLobatog:
Yup, I'll write a test and check that's not an snippet.

stbenjam:
Thanks @dLobatog !!

ACK

PR #59, opened 2015-10-21
Fixes #12233 - clean the interface cache only when becoming Managed::Host

iNecas:
Without the patch, discovery reboot fails on

``` NoMethodError: undefined method `drop_execution_interface_cache' for
```

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -38,7 +38,7 @@ module ForemanRemoteExecution
 
     def becomes_with_remote_execution(*args)
       became = becomes_without_remote_execution(*args)
-      became.drop_execution_interface_cache
+      became.drop_execution_interface_cache if became.respond_to? :drop_execution_interface_cache
       became
     end
 


ares:
maybe use `became.respond_to?`, I'm not sure what host types we'll ever see :-) Otherwise :+1:

stbenjam:
Isn't this more a problem of us extending the descendants of Host::Base? I still don't think we need to do that, the UI doesn't allow executing jobs on discovery hosts.  That *might* be worthwhile eventually in the Discovery case, but like @ares said, we don't know what host types we'll ever see, yet we assume now we should extend them all with our remote execution methods.

iNecas:
No, actually, this is hapending with your PR https://github.com/theforeman/foreman_remote_execution/pull/55/files#diff-24c9f52e5c8d0dc8ba4b5d2b7ec5de44R107, I noticed that, when I was testing the #55, So in master, it actually might work even now

iNecas:
@stbenjam I'm ok with closing this PR if you're going to address that in your PR (until now, I haven't realized that issue might be related to the #55)

stbenjam:
Yea @ares ran into a similar problem on another PR and we had a discussion about it, the outcome was to keep the descendants but remove Host::Base itself, but this is a good example of why not to do the descendants as well. 

@ares is there a reason to keep the descendants you think? If so we can take this and I can change #55, I don't really have a strong opinion about it

stbenjam:
Per discussion on hangouts, APJ from me

stbenjam:
Filed http://projects.theforeman.org/issues/12240

PR #60, opened 2015-10-21
Update README to include features, link to manual

dLobatog:
**Problem**
People come to the repo and don't know how to install the project, what does it exactly do, etc.. People ask in #theforeman IRC channel about basic help on how to install the plugin because they didn't see any documentation in the README.

**Solution**
Add a few basic information screenshots, points of contact, and links to the manual to ensure new users don't get lost.

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -28,6 +28,7 @@ class JobTemplate < ::Template
                   end
                 }
 
+  validates :job_name, :presence => true, :unless => ->(job_template) { job_template.snippet }
   validates :provider_type, :presence => true
   validate :provider_type_whitelist
 
--- a/test/unit/job_template_test.rb
+++ b/test/unit/job_template_test.rb
@@ -1,6 +1,19 @@
 require 'test_plugin_helper'
 
 describe JobTemplate do
+  context 'when creating a template' do
+    let(:job_template) { FactoryGirl.build(:job_template, :job_name => '') }
+
+    it 'needs a job_name' do
+      refute job_template.valid?
+    end
+
+    it 'does not need a job_name if it is a snippet' do
+      job_template.snippet = true
+      assert job_template.valid?
+    end
+  end
+
   context 'cloning' do
     let(:job_template) { FactoryGirl.build(:job_template, :with_input) }
 


README.md
```diff
@@ -8,33 +8,65 @@
...
 management functionality with remote management functionality.
 
+* Website: [theforeman.org](http://theforeman.org)
+* ServerFault tag: [Foreman](http://serverfault.com/questions/tagged/foreman)
+* Issues: [foreman_remote_execution Redmine](http://projects.theforeman.org/projects/foreman_remote_execution/issues)
+* Wiki: [Foreman wiki](http://projects.theforeman.org/projects/foreman/wiki/About)
+* Community and support: [#theforeman](https://kiwiirc.com/client/irc.freenode.net/?#theforeman) for general support, [#theforeman-dev](https://kiwiirc.com/client/irc.freenode.net/?#theforeman-dev) for development chat in [Freenode](irc.freenode.net)
+* Mailing lists:
+    * [foreman-users](https://groups.google.com/forum/?fromgroups#!forum/foreman-users)
+    * [foreman-dev](https://groups.google.com/forum/?fromgroups#!forum/foreman-dev)
```
stbenjam: I'm not so sure I want to include all this here, I'd rather link to a support URL on theforeman.org. Some of these things could get outdated quickly.

I'd say just link to foreman_remote_execution plugin docs, and theforeman.org support manual

README.md
```diff
@@ -8,33 +8,65 @@
...
+![job templates](http://theforeman.org/plugins/foreman_remote_execution/0.0/job_template_form.png)
+
 ## Installation
 
-See [How_to_Install_a_Plugin](http://projects.theforeman.org/projects/foreman/wiki/How_to_Install_a_Plugin)
-for how to install Foreman plugins
+Please see the Foreman manual for appropriate instructions:
 
-## Usage
+* [Foreman: How to Install a Plugin](http://theforeman.org/plugins/)
```
stbenjam: Should link to http://theforeman.org/plugins/foreman_remote_execution

stbenjam: Which doesn't exist, we need a landing page I guess, and I think the readme should just link to that.

dLobatog: No, that's only about installing the repo - actual usage was moved under installation below to just a link to the REX manual.

README.md
dLobatog: Sounds good. Should I link to just http://theforeman.org/support.html then? 

I basically copied the boilerplate from - https://github.com/theforeman/foreman-docker and https://github.com/theforeman/foreman_host_rundeck . What exactly could get outdated though, so I can change it in the other READMEs?

stbenjam: I guess, I feel like there's a lot of duplicate data here and we're better off redirecting users to the web site.  Maybe some other support avenue will open up, and it wouldn't be in this README, it's not really on our radar to keep this in sync.

README.md
stbenjam: Right, I meant the install instructions don't belong here at all. As given, they're not really even enough to use the plugin, the manual has the full instructions and because things are moving quickly I think we should just have them in one place.

dLobatog:
@stbenjam Updated!

stbenjam:
Weird, https://github.com/theforeman/foreman_remote_execution/pull/58 and have the same commit

dLobatog:
I opened a new one in https://github.com/theforeman/foreman_remote_execution/pull/66 not sure what happened

PR #63, opened 2015-10-23
Fixes #12290 - fixes detail link permission

ares:


stbenjam:
Works for me, created a limit user and it can see the details link

ACK

ares:
Thanks for review, merging

PR #64, opened 2015-10-26
fixes #12308 - preview hosts in invocation creation

stbenjam:
I have a really hard time working with javascript so I yoinked this one since it seemed simple enough to try to learn how to do it.

Does it make sense?

One thing I think I'm lacking is I'd probably like to replace the button icon with the spinny waiting thing while it's making the ajax request.  Do you know of any examples in foreman of it doing that?

### DIFF
--- a/app/assets/javascripts/template_invocation.js
+++ b/app/assets/javascripts/template_invocation.js
@@ -24,6 +24,33 @@ function refresh_search_query(value){
   $('textarea#targeting_search_query').val($('span#bookmark_query_map span#bookmark-' + id).data('query'));
 }
 
+function show_preview_hosts_modal() {
+  var modal_window = $('#previewHostsModal');
+
+  var form = $('form#job_invocation_form');
+  var data = form.serializeArray();
+
+  request = $.ajax({
+    data: data,
+    type: 'GET',
+    url: modal_window.attr('data-url'),
+    success: function(request) {
+      modal_window.find('.modal-body').html(request);
+    },
+    complete: function() {
+      modal_window.modal({'show': true});
+      modal_window.find('a[rel="popover-modal"]').popover();
+    }
+  });
+}
+
+function close_preview_hosts_modal() {
+  var modal_window = $('#previewHostsModal');
+  modal_window.modal('hide');
+  modal_window.removeData();
+  modal_window.find('.modal-body').html('');
+}
+
 function job_invocation_form_binds() {
   $('input.job_template_selector').on('click', function () {
     parent_fieldset = $(this).closest('fieldset');
@@ -39,6 +66,8 @@ function job_invocation_form_binds() {
   $('select#job_invocation_job_name').on('change', refresh_execution_form);
 
   $('button#refresh_execution_form').on('click', refresh_execution_form);
+
+  $('button#preview_hosts').on('click', show_preview_hosts_modal);
 
   $('textarea#targeting_search_query').on('change', refresh_execution_form);
 
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -17,6 +17,9 @@
         <%= button_tag(:type => 'button', :class => 'btn btn-default btn-sm', :title => _("Refresh"), :id => 'refresh_execution_form') do %>
           <%= icon_text('refresh') %>
         <% end %>
+        <%= button_tag(:type => 'button', :class => 'btn btn-default btn-sm', :title => _("Preview"), :id => 'preview_hosts') do %>
+          <%= icon_text('eye-open') %>
+        <% end %>
       </div>
     </div>
 
@@ -80,6 +83,6 @@
       <%= text_f f, :start_before, :label => _('Start before') %>
     </fieldset>
   </div>
-
+  <%= render :partial => 'preview_hosts_modal' %>
   <%= submit_or_cancel f %>
 <% end %>
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -56,11 +56,24 @@ class JobInvocationsController < ApplicationController
     @composer = JobInvocationComposer.new.compose_from_params(params)
   end
 
+  def preview_hosts
+    composer = JobInvocationComposer.new.compose_from_params(params)
+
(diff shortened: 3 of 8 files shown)


app/assets/javascripts/template_invocation.js
```diff
@@ -24,6 +24,33 @@ function refresh_search_query(value){
...
 
+function show_preview_hosts_modal() {
+  var modal_window = $('#previewHostsModal');
+
+  var form = $('form#job_invocation_form');
+  var data = form.serializeArray();
+
+  request = $.ajax({
+    data: data,
+    type: 'POST',
```
ohadlevy: why post?

stbenjam: No reason, is put better?

ohadlevy: since you are pulling host lists, shouldn't you use GET ? (POST == CREATE Object, PUT == update object, GET == read only)

stbenjam: I have to send the form back to be able to render the hosts list based on the current value of your query

app/views/job_invocations/_form.html.erb
```diff
@@ -81,5 +84,20 @@
     </fieldset>
   </div>
 
+  <!-- modal window -->
```
ares: maybe extract this to partial or maybe even better, extract it to helper that would receive arguments like path, title, content, so it can be reused somewhere else?

app/assets/javascripts/template_invocation.js
```diff
@@ -24,6 +24,33 @@ function refresh_search_query(value){
...
+    url: modal_window.attr('data-url'),
+    success: function(request) {
+      modal_window.find('.modal-body').html(request);
```
ares: would it make sense to use `#previewHostsModal .modal-body` just in case there are more modal windows on the page?

ares:
looks good to me, I'm looking forward to test tomorrow

app/views/job_invocations/_form.html.erb
stbenjam: I like the idea of making it nicer, but there's a lot of stuff in here like onclick's.  This is a very simple modal, but they get a lot more complicated when you look at some of the others in Foreman, including additional onclicks, additional sections like footers, default data inside the body, etc.  I think that's maybe why they haven't helper-ized it yet.

ares: I don't think we'd have to migrate all modal windows in core but I though this could be a good start which could be extended when needed as we might need more modal windows in rex... but I don't insist, I just wanted to have the form partial clean, so at least partial would be good

ares:
Works nicely, one nit from testing - for some reason, cross button doesn't close the modal window. One idea, would it be beneficial to add information about dynamic and static search on top of the list? Something like "Job will be executed on this list of hosts" vs "This is a preview of target hosts which and it may change before the job will start because dynamic targeting is selected". Last idea, not sure if it's worth it, hosts names could link to host details (with target _blank)?

stbenjam:
>  for some reason, cross button doesn't close the modal window.

What browser/version? Works for me on Firefox + Chrome

ares:
Weird, after I recheckout the same branch I used for testing, it works. It was Chrome 46. Sorry for false alert, seems like some local issue.

stbenjam:
Thanks @ares! Updated - nice suggestions, looks better

preview

ohadlevy:
@stbenjam what if the list is very long? I assume you might want to a. ensure a max hight (and scroll) and consider pagination?

app/assets/javascripts/template_invocation.js
ohadlevy: I don't think that conflicts? also, do you need to send the entire form or just the search term /bookmark used?

stbenjam:
It does scroll, that's the overflow-y setting:

scroll

Although maybe the modal is too tall, how big should I make it? My CSS isn't very good.  Pagination also adds a lot of complication.

I want to finish up other things this week for iteration 3, so if that's going to hold this PR, I'll have to revisit it later.

ohadlevy:
I could see us skipping paginiation (e.g. limit to first xx hosts) and
potentially linking to the host list with the same bookmark/search.
i dont think you should hold the pr for that.

ares:
needs a trivial rebase

ares:
tested with manual rebase, works as a charm, ACK (pending green tests after rebase)

app/assets/javascripts/template_invocation.js
stbenjam: The form, because I also need to know if it's a dynamic query or not.  Refresh uses a post, too.

stbenjam:
Can you take a quick look again @ares?

Based on @ohadlevy's feedback, I only show the first `Setting[:entries_per_page]` hosts (default would be 20, here its 2).

and3more

app/assets/javascripts/template_invocation.js
stbenjam: It works with GET, so I changed it

ares:
I like the count limiting, maybe 20 (default) is to small number but we'll see. It's easy to add `*5` later if needed...

Last thing I just spot, `#preview_hosts` requires `:view_job_invocations` permission but it's actually only used when you create new invocation, should it be `:create_job_invocations`? (similar to refresh)

app/assets/javascripts/template_invocation.js
ares: Sorry to jump on this too late, if we send the search field it can be very long (bulk action on 500 hosts, FQDN is ~20 chars), GET search query size is limited by some browsers (IE ~ 2k), Apache refuse after 8k (according to doc) so I think POST would be justified here.

stbenjam:
> I like the count limiting, maybe 20 (default) is to small number but we'll see. It's easy to add *5 later if needed...

Yea, I just figured using `Setting[:entries_per_page]` made sense, but we could add pagination to the modal later, or just use our own value if we think its not enough.

> Last thing I just spot, #preview_hosts requires :view_job_invocations permission but it's actually only used when you create new invocation, should it be :create_job_invocations? (similar to refresh)

Yea that makes sense, updated.

ares:
APJ

stbenjam:
Thanks!

PR #65, opened 2015-10-26
fixes #12316 - make global proxy turned on by default

stbenjam:
With this as a default, then the plugin just works out of the box.  More
advanced users will likely want to turn it off and use subnet-based
proxy determination for segmented networks while getting proper HA, load
balancing, etc.


ares:
Thanks @stbenjam, merging! Could you please change this in [manual](http://theforeman.org/plugins/foreman_remote_execution/0.0/index.html#4.1Determiningthesmartproxyforhost) of 0.1?

PR #67, opened 2015-10-29
fixes #12269 - only show rerun failed when there's failed tasks

stbenjam:


stbenjam:
@iNecas How about this one for Iteration 3 too?

iNecas:
Tested and works great. I know @ares  preference is rather to disable the buttons than hiding them (as even if everything works, you know that you could rerun if something went wrong).

stbenjam:
That makes sense, its less confusing than magically disappearing buttons.

Updated

iNecas:
ACK! Tested and works for me

stbenjam:
Thanks!

PR #68, opened 2015-10-30
Fixes #10758 - allow execution limiting by permissions

ares:
There are couple of things that has to be solved
* we have to link host of template_invocation somehow
* autocompletion in filters does not work, didn't have time to find why the path is wrong
* if the execution fails we don't display any nice message, in RunHostsJob we just know that sub task failed

### DIFF
--- a/db/migrate/20151215114631_add_host_id_to_template_invocation.rb
+++ b/db/migrate/20151215114631_add_host_id_to_template_invocation.rb
@@ -1,0 +1,29 @@
+class AddHostIdToTemplateInvocation < ActiveRecord::Migration
+  class FakeTemplateInvocation < ActiveRecord::Base
+    set_table_name 'template_invocations'
+  end
+
+  def up
+    add_column :template_invocations, :host_id, :integer
+    add_foreign_key "template_invocations", "hosts", :name => "template_invocations_hosts_id_fk", :column => 'host_id'
+    FakeTemplateInvocation.reset_column_information
+
+    say 'Migrating existing execution locks to explicit relations, this may take a while'
+    FakeTemplateInvocation.all.each do |template_invocation|
+      task = ForemanTasks::Task.for_action_types('Actions::RemoteExecution::RunHostJob').joins(:locks).where(
+        :'foreman_tasks_locks.resource_type' => 'TemplateInvocation',
+        :'foreman_tasks_locks.resource_id' => template_invocation.id).first
+      next if task.nil? # skip invocations from very early versions of remote executions
+      host_id = task.locks.where(:'foreman_tasks_locks.resource_type' => 'Host::Managed').first.resource_id
+      next unless Host.find_by_id(host_id)
+
+      template_invocation.host_id = host_id
+      template_invocation.save!
+    end
+  end
+
+  def down
+    remove_foreign_key "template_invocations", :name => "template_invocations_hosts_id_fk"
+    remove_column :template_invocations, :host_id
+  end
+end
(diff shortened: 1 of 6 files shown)


stbenjam:
> autocompletion in filters does not work, didn't have time to find why the path is wrong

Maybe you'll to override it like Katello does? https://github.com/Katello/katello/blob/master/lib/katello/plugin.rb#L145

ares:
I changed the relation so we can use scoped search for host/hostgroup on TemplateInvocation. Also auto completion now works, index route has to be defined as well, we don't need overrides in this plugin.

iNecas:
[test]

iNecas:
When running migration, I got:

```
undefined method `host_id=' for #<AddHostIdToTemplateInvocation::FakeTemplateInvocation:0x00000009302f60>/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/db/migrate/20151215114631_add_host_id_to_template_invocation.rb:20:in `block in up'
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/db/migrate/20151215114631_add_host_id_to_template_invocation.rb:12:in `each'
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/db/migrate/20151215114631_add_host_id_to_template_invocation.rb:12:in `u
```

db/migrate/20151215114631_add_host_id_to_template_invocation.rb
```diff
@@ -0,0 +1,29 @@
+class AddHostIdToTemplateInvocation < ActiveRecord::Migration
+  class FakeTemplateInvocation < ActiveRecord::Base
+    set_table_name 'template_invocations'
+  end
+
+  def up
+    # add_column :template_invocations, :host_id, :integer
```
iNecas: Forgot to uncomment?

iNecas:
What about seeding some pre-defined remote execution role, mainly for documenation purposes - it's not trivial to find out what are all the permissions one needs for making the remote execution for some non-admin user to work

iNecas:
I have this filter on execute_job_template: `host.name = host-1.foreman.test OR host_group.name = allow`, I got the execution allowed for the `host-1`, but not for the host in `allow` host_group

iNecas:
Nm - the last issue was on my side. To wrap the things up, after uncommenting the lines in migration, it works nicely. I would suggest to seed remote execution role with this permissions (this is a minimal set I needed to get it working):

* Job template	view_job_templates	
* Job invocation	view_job_invocations, create_job_invocations	
* Template invocation	execute_template_invocation		
* Host/managed	view_hosts, edit_hosts	
* Smart proxy	view_smart_proxies

iNecas:
Tested and works nicely (after some fixes described above), ACK after the comments are addressed

ares:
I fixed the migration :bug: :hammer: and added two roles, the edit_hosts permissions would be only useful to make checkboxes on host page active, while user can still assign mupltiple hosts using search in invocation form. I think we shouldn't add this permission to the role. The reason is editing hosts would be too powerful (changing IP to get to different host etc). Otherwise I got the same permissions set.

Please give it another look and let's see what Mr. Jenkins say.

iNecas:
ACK, APJ

ares:
Rubocop should be happy now

iNecas:
[test] - hitting the "HostIntegrationTest::create new host page.test_0003_enables override of hostgroup defaults" starts to be annoying

ares:
re[test]

iNecas:
The jenkins is pretty false positive right now with the "HostIntegrationTest::create new host page.test_0003_enables override of hostgroup defaults", I'm ok with merging and waiting for it to pass in master

ares:
Merging rebased version, thanks for the review.

PR #69, opened 2015-11-02
fixes #11634 - allow customization of ssh user

stbenjam:
Greatly simplified after discussing with @iNecas  this morning:

- Global setting `remote_execution_ssh_user` is used by default
- Parameter of same name can be set to allow customization - let user set user names as granularly as they want by using parameter hierarchy: per host, per host group, environment, etc.

Accompanies https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/10

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -25,7 +25,8 @@ module Actions
         link!(job_invocation)
         link!(template_invocation)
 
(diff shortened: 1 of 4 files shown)


iNecas:
ACK, works well - I've tested both upading the remote user on per-host basis, as well as chaning the global one. The only issue I've hit was not related directly to this, but this feature enabled to get into this condition, I've opened that as separate issue http://projects.theforeman.org/issues/12422

PR #70, opened 2015-11-05
fixes #12218 - check if query and bookmark are blank not nil

stbenjam:


iNecas:
One thing I've inoticed while testing: although the validation prevents from continuing, the information about number of resolved hosts is still not valid : it says 73 hosts, while it should be either 0 or info about search query missing

stbenjam:
@iNecas  I think you're seeing what https://github.com/theforeman/foreman_remote_execution/pull/71 fixes

iNecas:
@stbenjam LOL, ACK then, tested and works great!

PR #71, opened 2015-11-05
fixes #12407 - show 0 hosts in count and preview when there's no query

stbenjam:
Otherwise when there's no query, it shows you all hosts


iNecas:
ACK

PR #73, opened 2015-11-11
Fixes #12437, refs #12288 - better handling of unexpected states

iNecas:
Don't raise cryptic errors when there is no reasonable output comming from the
proxy.

Also, improve the error message a proxy in missing (and proxy command fails in planning
phase). There is still work to be done to get to the erors from the job hosts list,
but this requires some changes on dynflow side that I don't want to rush to get
into stable relese. But at least, when going to the `Monitor -> Tasks` and showing
the failed `RuhHostJob` task, one doesn't see:

    undefined method `live_output' for nil:NilClass (NoMethodError)

### DIFF
--- a/app/lib/actions/remote_execution/helpers/live_output.rb
+++ b/app/lib/actions/remote_execution/helpers/live_output.rb
@@ -1,0 +1,17 @@
+module Actions
+  module RemoteExecution
+    module Helpers
+      module LiveOutput
+        def exception_to_output(context, exception, timestamp = Time.now)
+          format_output(context + ": #{exception.class} - #{exception.message}", 'debug', timestamp)
+        end
+
+        def format_output(message, type = 'debug', timestamp = Time.now)
+          { 'output_type' => type,
+            'output' => message,
+            'timestamp' => timestamp.to_f }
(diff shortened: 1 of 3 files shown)


iNecas:
I put the two commits into one PR, as there would be conflicts when merging otherwise

stbenjam:
@iNecas There seems to be some regression here, if I try to run a command where /tmp/foreman-proxy is unreadable by the ssh user, the task returns successful instead of failing

iNecas:
This is not a regression introduced by this changes IMO, but I see issues with handling encoding on proxy side, that is causing that… I will open a PR against smart_proxy_remote_execution_ssh

stbenjam:
Ah ok, this PR does work and handles the case with no output so ACK

iNecas:
Here is a fix for the issue you've discovered https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/15 (description at http://projects.theforeman.org/issues/12456), I think, It can also be related to http://projects.theforeman.org/issues/12437

iNecas:
Thanks @stbenjam

PR #74, opened 2015-11-12
Fixes #12288 - disable transactions for plan phase in run host job action

iNecas:
Otherwise we lose the link from the task to the job invocation. Since
in this case, the execution plan is not driving a transaction between
local database and external systems, we can turn this logic off for
this action.

With this patch (that depends on https://github.com/Dynflow/dynflow/pull/174),
when there is an issue with running the job on proxy (as in case when the proxy
is just down), we show the tasks on per-host part as failed, with posibility to
show details, as with any other error.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -2,6 +2,7 @@ module Actions
   module RemoteExecution
     class RunHostJob < Actions::EntryAction
 
+      middleware.do_not_use Dynflow::Middleware::Common::Transaction
       include Actions::RemoteExecution::Helpers::LiveOutput
 
       def resource_locks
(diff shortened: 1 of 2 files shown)


ares:
I suppose tests failed because it needs newer dynflow?

iNecas:
Yes

iNecas:
The tests are green now

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -10,6 +11,9 @@ def resource_locks
 
       def plan(job_invocation, host, template_invocation, proxy, connection_options = {})
         action_subject(host, :job_name => job_invocation.job_name)
+        link!(job_invocation)
+        link!(template_invocation)
+
```
adamruzicka: The reason for moving those three lines is to create the links before any error can be raised?

iNecas: We need to do that as soon as possible, before any other error occurs, so that one can get to the errors form the linked resource

adamruzicka:
Works as expected, ACK

PR #75, opened 2015-11-19
Refs #12145 - Adds time format and explanation

adamruzicka:


iNecas:
Rebase please. No comments from code point of view

stbenjam:
This is fine, but if we're going to touch it, I'd rather fix it better via http://projects.theforeman.org/issues/12686, and use a date/time picker.

adamruzicka:
Closing since it has to be fixed in foreman-tasks

PR #76, opened 2015-11-19
Refs #11994 - Adds invocation times to index

adamruzicka:


domcleal:
Check your ticket number, this probably isn't the right one.

adamruzicka:
@domcleal Oh, wrong number, sorry

stbenjam:
A few thoughts, I realize the second one makes this a bit more complicated but it'd be far more useful:

* I'd rather we use the `time_ago_in_words` helper to stay consistent with Foreman's use (see Reports) page.  It's easier to parse than a date stamp by humans
* I'd rather we make start and end time a property of the `JobInvocation` itself, so that we can leverage column sorting and scoped search to query (e.g. search for all jobs from Friday)

stbenjam:
One more:

Do we need to show both times on the index? It makes the table pretty big, perhaps only show End time in the index and both on the details page?

adamruzicka:
@stbenjam I wanted to stay consistent with tasks index, where we use plain timestamps for both times. However I'm not against `time_ago_in_words` or not including the second time at all.

Job invocation is bound one-to-one with a task (in the recurring logic PR there's a commit to rename job invocation's last_task to just task), so there isn't a reason to have the times saved on both ends of the association. The scoped search and sorting can be done through this association too.

stbenjam:
IIRC, the column sorting through associations with the scoped_search helpers didn't work, but maybe you can get it working.  If it does, we should give the user the ability to sort all the columns on that page.

adamruzicka:
I'll look into it tomorrow

adamruzicka:
I managed to make the table sortable by start/end time through the association, searching should work too.

I'm not sure how sorting/searching by task counts could be done. What about opening a new issue for it since it's a bit out of scope of [#1194](http://projects.theforeman.org/issues/11994)?

app/controllers/job_invocations_controller.rb
```diff
@@ -48,7 +48,7 @@ def show
   end
 
   def index
-    @job_invocations = resource_base.search_for(params[:search]).paginate(:page => params[:page]).order('id DESC')
+    @job_invocations = resource_base.search_for(params[:search]).paginate(:page => params[:page]).with_task.order(params[:order] || 'id DESC')
```
stbenjam: Should this be `.with_last_task`?

app/models/job_invocation.rb
```diff
@@ -24,6 +24,13 @@ class JobInvocation < ActiveRecord::Base
 
   scoped_search :on => [:job_name], :complete_value => true
 
+  scoped_search :in => :last_task, :on => :started_at, :rename => 'last_task.started_at'
+  scoped_search :in => :last_task, :on => :ended_at, :rename => 'last_task.ended_at'
```
stbenjam: Can we just rename these to started_at/ended_at? It looks more user-friendly in the search bar

stbenjam: Also add `:complete_value => true`, it adds some nice autocompletion for dates:

complete

stbenjam:
The sorting works well.  We can deal with the other columns, I agree it's out of scope.

I do still think I'd prefer `time_ago_in_words` to stay with Foreman style, the time stamps are hard to parse for humans.  Maybe @iNecas or @ares has an opinion about it

app/controllers/job_invocations_controller.rb
adamruzicka: Oh, right, forgot to change it here

app/models/job_invocation.rb
adamruzicka: Renamed and enabled the autocompletion

adamruzicka:
Removed the `ended_at` column, changed the format to `time_ago_in_words` with exact time in popup on hover (similar to foreman audits)
img

stbenjam:
I like it, ACK from me

iNecas:
ACK

PR #77, opened 2015-11-19
[Foreman 1.11] Fixes #12538 - Support rubocop 0.35.1

daviddavis:

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -14,8 +14,8 @@ module Actions
 
         raise _("Could not use any template used in the job invocation") if template_invocation.blank?
 
-        settings =  { :global_proxy   => 'remote_execution_global_proxy',
-                      :fallback_proxy => 'remote_execution_fallback_proxy' }
+        settings = { :global_proxy   => 'remote_execution_global_proxy',
+                     :fallback_proxy => 'remote_execution_fallback_proxy' }
 
         raise _("Could not use any proxy. Consider configuring %{global_proxy} " +
                 "or %{fallback_proxy} in settings") % settings if proxy.blank?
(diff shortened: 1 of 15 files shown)


stbenjam:
@daviddavis  What was the reason for closing this? Is it still needed?

daviddavis:
@stbenjam I ran rubocop against master with 0.28 (or whatever the previous version was) and it looked like there were errors so I just assumed that you guys weren't checking your code with rubocop. Would be happy to reopen this though if you guys want to use rubocop in jenkins or locally. Let me know.

daviddavis:
I take that back. It does look like your tests are running rubocop but they're using 0.28.0. Not sure how though...

iNecas:
Given you changed the title, you probably figured out we run the tests against the foreman 1.10, as our release cycle is not synchonized with foreman (and we do releases that are usually compatible both with latest stable and nightly). What we could do is trying to get the changes that work both on old and new version now, and keeping the things that depend on the new rubocop version for later?

daviddavis:
@inecas cool I updated the PR and commented out the 0.35.1 cops. When this project updates to foreman 1.11, we can just uncomment them and fix any breakages and then we should be good to go.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -14,7 +14,7 @@ def plan(job_invocation, host, template_invocation, proxy, connection_options =
 
         raise _("Could not use any template used in the job invocation") if template_invocation.blank?
 
-        settings =  { :global_proxy   => 'remote_execution_global_proxy',
+        settings = { :global_proxy   => 'remote_execution_global_proxy',
                       :fallback_proxy => 'remote_execution_fallback_proxy' }
```
iNecas: One space to the left would look slightly nicer :)

iNecas:
Reopending new PR after rebase as #123

PR #79, opened 2015-12-03
Fixes unknown column errors on last_task

adamruzicka:
#44 broke jobs index and maybe something more because of renaming `last_task_id` column on `job_invocations` table to `task_id`.

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -45,7 +45,7 @@ class JobInvocationsController < ApplicationController
   end
 
   def index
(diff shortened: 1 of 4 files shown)


stbenjam:
Needs a redmine issue in the commit message, but it looks OK otherwise

stbenjam:
Looks good, and works for me - ACK

iNecas:
Since it's breaking master, I'm merging before @adamruzicka

adamruzicka:
@iNecas thank you and sorry again for causing this

PR #80, opened 2015-12-03
Fixes #12489 - Effective user support

iNecas:
Allow to specify effective user settings on template and job invocation level.

On job template, on chooses:

  * value - what effective user value should be used (defaults to Settings[:remote_executioneffective_user]
  * current_user - the current user login will be used as the effective user
  * overridable - the user can change the effective user value at invocation time - when false, the current user value is enforced

There are two methods of effective user supported now - su and sudo. It's determined by host param `remote_execution_effective_uesr_method` and the default is configured in settings.

Requires https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/16
on smart proxy side.

Also, I refactored and unified the UI and API job invocation composer - I really disliked
the duplicity of code, when adding the effective user on both sides. It is also preparation
for the developer API feature, where we will need also to compose the job invocation.


stbenjam:
> Also, I refactored and unified the UI and API job invocation composer - I really disliked
the duplicity of code, when adding the effective user on both sides.

+1

Could we also move this to `app/services`? IMHO it doesn't belong in models

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -29,6 +29,7 @@ def show
             param :name, String, :required => true
             param :value, String, :required => true
           end
+          param :effective_user, String, :required => false
```
stbenjam: The downside to this is, it's only useful for SSH (or some other provider that needs this setting).  What about taking a hash of provider options? 

Just wondering what happens if you need to specify some playbook-specific setting for Ansible, the hammer --help and apidocs will be a mess and not clear what options are appropriate when.

iNecas: I went for this path just because it should just work in hammer then - with nested hash, I guess there will be some additional work needed on the hammer side. But I agree, since it's provider specific , it would be worth the effort. What about nesting this into under 'ssh' key?

iNecas:
@stbenjam ad. moving the compose into `app/lib/services` I know @ares has more preferences on keeping that in model. It really depends, if you look at model just as a pure database wrapper, or also a layer on top of that. I can imagine this being in `models` or `services`

stbenjam:
> It really depends, if you look at model just as a pure database wrapper, or also a layer on top of that. I can imagine this being in models or services

Ok, it's not a big deal, and it is creating ActiveRecords so I guess it's relevant for `app/models`.

app/controllers/api/v2/job_invocations_controller.rb
stbenjam: >  What about nesting this into under 'ssh' key?

I like that. It keeps the provider-specific settings grouped together and makes it clear what goes with what.

iNecas: Updated

iNecas:
Ready for re-review

app/models/job_invocation_composer.rb
```diff
@@ -1,49 +1,186 @@
...
+    def template_invocations_params
+      template = template_with_default
+      template_invocation_params = { :template_id => template.id, :effective_user => api_params[:effective_user] }
+      template_invocation_params[:input_values] = api_params.fetch(:inputs, []).map do |input_params|
+        input = template.template_inputs.find_by_name(input_params[:name])
+        unless input
+          raise ::Foreman::Exception, _('Unknown input %{input_name} for template %{template_name}') %
+              { :input_name => input_params[:name], :template_name => template.name }
+        end
+        { :template_input_id => input.id, :value => input_params[:value] }
```
stbenjam: This section is going to need to be updated to take into account the changes I've made to the API to take a hash

stbenjam: https://github.com/theforeman/foreman_remote_execution/commit/da840d90cda1bf9f703e1a8dacc40f32417ba58e#diff-7bedc2583949107cd5e4a5407a080a6bR33

iNecas: I've rebased with refelcting the changes in API (both `s/template_id/job_template_id/` and `inputs` format)

iNecas:
Rebased and updated to the latest changes

app/views/job_invocations/_form.html.erb
```diff
@@ -65,6 +65,9 @@
                       <% end %>
                     <% end %>
                   <% end %>
+                  <% if job_template.effective_user.overridable? %>
+                    <%= text_f job_template_fields, :effective_user, :label => _('Effective user') %>
```
stbenjam: Can we have a description next to this, similar to the Setting?

stbenjam:
This works great! :+1:  Just one comment above and it's good to go.

app/views/job_invocations/_form.html.erb
iNecas: Something like this? https://github.com/theforeman/foreman_remote_execution/commit/28a786d44f92ffaf80bce7092654b5ba7f8b2bb5

stbenjam: Yup, thanks

stbenjam:
APJ

iNecas:
I've squashed the commits. Will merge after jenkins passes.

iNecas:
[test]

PR #81, opened 2015-12-08
Refs #11730 - Adds description for job invocation

adamruzicka:
- [x] tests
- [x] edit description format from job template UI
- [x] more user-friendly descriptions

Obligatory gif

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -40,13 +40,15 @@ class JobInvocation < ActiveRecord::Base
 
   default_scope -> { order('job_invocations.id DESC') }
 
-  attr_accessor :start_before
+  attr_accessor :start_before, :description_format
   attr_writer :start_at
 
   def deep_clone
     JobInvocationComposer.from_job_invocation(self).job_invocation.tap do |invocation|
       invocation.task_group = JobInvocationTaskGroup.new.tap(&:save!)
       invocation.triggering = self.triggering
+      invocation.description_format = self.description_format
+      invocation.description = self.description
       invocation.template_invocations = self.template_invocations.map(&:deep_clone)
       invocation.save!
     end
@@ -107,5 +109,20 @@ class JobInvocation < ActiveRecord::Base
   def output(host)
     return unless (task = sub_task_for_host(host)) && task.main_action && task.main_action.live_output.any?
     task.main_action.live_output.first['output']
+  end
+
+  def generate_description!
+    key_re = /%\{([^\}]+)\}/
+    template_invocation = template_invocations.first
+    input_names = template_invocation.template.template_input_names
+    hash_base = Hash.new { |hash, key| hash[key] = "%{#{key}}" }
+    input_hash = hash_base.merge Hash[input_names.zip(template_invocation.input_values.pluck(:value))]
+    input_hash.update(:job_name => job_name)
+    description_format.scan(key_re) { |key| input_hash[key.first] }
+    self.description = description_format
+    input_hash.each do |k, v|
+      self.description.gsub!(Regexp.new("%\{#{k}\}"), v)
+    end
+    save!
   end
 end
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -1,5 +1,6 @@
 class JobTemplate < ::Template
-  attr_accessible :job_name, :provider_type, :effective_user_attributes
+
+  attr_accessible :job_name, :provider_type, :description_format, :effective_user_attributes
 
   include Authorizable
   extend FriendlyId
@@ -85,6 +86,19 @@ class JobTemplate < ::Template
 
   def effective_user
     super || (build_effective_user.tap(&:set_defaults))
+  end
+
+  def generate_description_format
+    if description_format.blank?
+      generated_description = '%{job_name}'
+      unless template_inputs.empty?
+        inputs = template_inputs.map(&:name}.map { |name| %Q(#{name}="%{#{name}}") }.join(' ')
+        generated_description << " with inputs #{inputs}"
+      end
+      generated_description
+    else
+      description_format
+    end
   end
 
   private
--- a/app/views/templates/puppet_run_once.erb
+++ b/app/views/templates/puppet_run_once.erb
@@ -2,6 +2,7 @@
 kind: job_template
 name: Puppet Run Once - SSH Default
 job_name: Puppet Run Once
+description_format: '%{job_name} with "%{puppet_options}"'
 provider_type: Ssh
 template_inputs:
 - name: puppet_options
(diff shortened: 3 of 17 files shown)


app/views/templates/puppet_run_once.erb
```diff
@@ -8,5 +8,6 @@ template_inputs:
   description: Additional options to pass to puppet
   input_type: user
   required: false
+description_format: puppet_options="%{puppet_options}"
```
iNecas: nit-pick - would look nicer if it was under the job name

iNecas:
I'm missing the ability to edit the format form the template form

iNecas:
Missing tests

iNecas:
Two fields in the job invocation just for the description seem too much for me, in ideal case, the user should not even care about the description (but should be able to override it, when needed). What about showing just description field, and when overriding, letting the user to edit the description format: one would loose the ability to preview, but still, it would be one field less than it is right now. My assumption is folks will not care about the description format, unless they want to customize that, and I guess they would much likely want to do that in template form than in the invocation form

iNecas:
Also, what about getting also the part "Run Run command…" into the description format, the "Run Run command" seems pretty strange anyway, this would allow us to format something more user-friendly such as "install package(s) zsh"

adamruzicka:
> What about showing just description field, and when overriding, letting the user to edit the description format

So there would be one field used for two different things (displaying description preview/taking description format)? Wouldn't it be confusing?

iNecas:
----- Original Message -----
> > What about showing just description field, and when overriding, letting the
> > user to edit the description format
> 
> So there would be one field used for two different things (displaying
> description preview/taking description format)? Wouldn't it be confusing?

Or the user can edit the lastly-generated desription, instead of the format
(which might be also confusing for somebody not familiar with it). So the
only place where one would actually see the description template would be
the template form.

Also, I wonder if that should not be the last field in the form,
so that the user fills in the other fields, so that the generated message has enough
data to show. I can imagine somebody wanting to edit the message just because the inputs
are not filled in just yet.

stbenjam:
I didn't try it yet, but the gif looks great :clap:

ares:
Very nice! Could you please just rename labels so a Description would be named "Description preview" and Description format would be "Description template" or just "Description"? Also would it make sense to hide it under some "advanced" link? We have a lot of fields in this form already? Opinions on above are welcome from anybody...

stbenjam:
> Also would it make sense to hide it under some "advanced" link? We have a lot of fields in this form already?

Good point, it's going to get crowded on that page.  Can we do an Advanced tab maybe?

adamruzicka:
Moving the descriptions to another tab would be troublesome. How about putting the scheduling options into separate tab and making the description template hide/show with the checkbox?
[Images here](http://imgur.com/a/S4c9Q)

stbenjam:
I actually like breaking the scheduling out, makes the form less cluttered. :+1: 

What do you think about this also?

1) Rename "Description preview" to just "Description"
2) Move it to the top of the form (above job name)

Either that, or add a divider between the inputs and the description.

adamruzicka:
@stbenjam 
1) I had "Description" instead of "Description preview" in the first version (the gif in the OP), @ares suggested to rename it this way. Maybe just "Description" would be better if the template is hidden.
2) It doesn't look nice at the top when you don't have anything filled in

A divider makes the whole thing look odd, like it's splitting the form in a wrong place - [image](http://i.imgur.com/Bzuu4ls.png)

stbenjam:
Alright, that's fine then, we'll have time with Kyle next week as the whole form probably needs some UX improvements.

I think just Description makes sense when the template is hidden, otherwise it's not clear what's being previewed

app/models/job_invocation.rb
```diff
@@ -108,4 +110,16 @@ def output(host)
...
   end
+
+  def generate_description!
+    key_re = /%\{([^\}]+)\}/
+    template_invocation = template_invocations.first
+    input_names = template_invocation.template.template_input_names
+    hash_base = Hash.new { |hash, key| hash[key] = "%{#{key}}" }
+    input_hash = hash_base.merge Hash[input_names.zip(template_invocation.input_values.pluck(:value))].symbolize_keys
+    input_hash.update(:job_name => job_name)
+    description_format.scan(key_re) { |key| input_hash[key.first.to_sym] }
```
iNecas: Calling `to_sym` on user input has security consequences (DOS due to symbols not getting garbage collected). What about replacing missnig keys in description_format with "%%{missing_key}", the symbolize keys has a similar issues (the set of inputs is an user input as well). It seems the easies thing would be to implement the '%{}' substitution using `gsub` or similar.

app/models/job_template.rb
```diff
@@ -87,6 +88,11 @@ def effective_user
     super || (build_effective_user.tap(&:set_defaults))
   end
 
+  def generate_description_format
+    format = description_format || template_inputs.map(&:name).map { |name| %Q(#{name}="%{#{name}}") }.join(' ')
+    format == '' ? '%{job_name}' : format
```
iNecas: The `%{job_name}` should be part of the generate format even when the format is not empty (when there is some input). And maybe in format of `Run %{job_name} with inputs…`

app/models/job_template.rb
```diff
@@ -87,6 +88,20 @@ def effective_user
...
   end
 
+  def generate_description_format
+    if description_format.blank?
+      names = template_inputs.map(&:name)
+      if names.empty?
+        '%{job_name}'
+      else
+        inputs = names.map { |name| %Q(#{name}="%{#{name}}") }.join(' ')
+        "%{job_name} with inputs #{inputs}"
```
iNecas: `input_names` name would be more descriptive than just `names`. Also, to reduce the duplicity, it could look like this:

```ruby
generated_description = '%{job_name}'
unless template_inputs.empty?
  inputs = template_inputs.map(&:name}.map { |name| %Q(#{name}="%{#{name}}") }.join(' ')
  generated_description << " with inputs #{inputs}"
end
generated_description
```

iNecas:
One minor comment, ACK otherwise, tested and it works nicely

iNecas:
[test]

app/models/job_template.rb
```diff
@@ -87,6 +88,19 @@ def effective_user
     super || (build_effective_user.tap(&:set_defaults))
   end
 
+  def generate_description_format
+    if description_format.blank?
+      generated_description = '%{job_name}'
+      unless template_inputs.empty?
+        inputs = template_inputs.map(&:name}.map { |name| %Q(#{name}="%{#{name}}") }.join(' ')
```
ares: syntax error `map(...}`

ares:
I'm not sure whether scheduling as a separate tab is a good option but let's see what users say, we can always put it back... I put few fixes into https://github.com/adamruzicka/foreman_remote_execution/pull/1

iNecas:
Closing in favor of #88

PR #82, opened 2015-12-10
fixes #12753 - exit status should be last item

stbenjam:
The problem here is `task.ended_at` is according to the Foreman server's time, whereas other output lines happen in respect to the proxy's.  In my case, the two servers were several seconds off, causing "Exit status" to render in the middle of the output:

exit status

The simplicity of sorting by time though is nice, so just put exit status at  Sat, 20 Nov 2286 17:46:39 GMT, a.k.a 9999999999 seconds from the epoch.

Any other suggestions to handle this?

### DIFF
--- a/app/lib/actions/remote_execution/run_proxy_command.rb
+++ b/app/lib/actions/remote_execution/run_proxy_command.rb
@@ -76,11 +76,16 @@ module Actions
         end
 
         if exit_status
-          records << format_output(_("Exit status: %s") % exit_status, 'stdout', task.ended_at)
+          records << format_output(_("Exit status: %s") % exit_status, 'stdout', final_timestamp(records))
         elsif run_step && run_step.error
-          records << format_output(_("Job finished with error") + ": #{run_step.error.exception_class} - #{run_step.error.message}", 'debug', task.ended_at)
+          records << format_output(_("Job finished with error") + ": #{run_step.error.exception_class} - #{run_step.error.message}", 'debug', final_timestamp(records))
         end
         return records
+      end
+
+      def final_timestamp(records)
+        return task.ended_at if records.blank?
+        records.last.fetch('timestamp', task.ended_at) + 1
       end
 
       def proxy_result


adamruzicka:
How about sorting the "regular" output first and appending the exit status record afterwards? Something along the lines of https://github.com/theforeman/foreman_remote_execution/compare/master...adamruzicka:fix/exit-status

Not that I dislike magic constants, but...

stbenjam:
The time stamp is used in order to make live output work, we only fetch newer records, so the position in the array itself doesn't matter much, if I understand that example correctly.

For example, the API with hammer:

https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb#L48

https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/1/files#diff-e5e95ce732d7d825d38bbd6e81023cfeR51


The javascript UI code does the same thing, the exit status really does need the latest time stamp, or we need to make exit_status not part of the output but a separate property.

adamruzicka:
I know the timestamp is used, but it should work anyway. I assume a task cannot generate more output once it's finished.

For example in [hammer](https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/1/files#diff-e5e95ce732d7d825d38bbd6e81023cfeR51) you receive an `output`, sort already sorted `output[:lines]` by the timestamp and use the timestamp of the last line to request newer output if the `task` is still `pending?`. This means that if you receive an `output`, where `output[:refresh]` is set, then the `output[:lines]` is sorted properly and does not contain the exit record and we can request newer batch of output lines using the timestamp of the last line.

If you receive an `output` where the `output[:refresh]` is not set, then it means the task is finished, `output[:lines]` contains the exit record as the last element and it doesn't matter if the exit record is in the "right" position by the timestamp since the timestamp is not used anymore.

I really hope I'm not missing something big there.

stbenjam:
Ah, I see, this fixes the problem when you're viewing the output after the fact, but not when viewing live output.  Both the API and UI controllers will drop records with timestamps before `:since`, so the order in the array doesn't matter:

https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb#L55

When the Foreman and Proxy timestamps have diverged, the exit result would have a time stamp in the middle of the execution and never get sent at all. Does that make sense?

Not easy to explain :grin:

iNecas:
What about using for this timestamp the timestamp of last output line? Alternatively, we can add the exit status info at the proxy part, instead of the server part, so that we have the ordering correct

stbenjam:
> What about using for this timestamp the timestamp of last output line? 

Ok, that makes sense - so how about this?

iNecas:
[test]

stbenjam:
@iNecas updated

iNecas:
ACK, tested and works great

stbenjam:
Thanks!

PR #84, opened 2015-12-15
Fixes #12826 - tasks relation explicit class name

ares:


adamruzicka:
ACK

ares:
Merging, thanks for quick look.

PR #85, opened 2015-12-15
Fixes #12769 - fix ajax refresh of invocation form

iNecas:
Use `replaceWith` instead of `html` for updating the form data
and re-initiate the bootstrap `popup` to fix the inline help.


ares:
APJ(:+1:)
works nice!

iNecas:
[test]

ares:
re [test] seems like intermittent issue

PR #88, opened 2015-12-16
Fixes #11730 - Adds description for job invocation

iNecas:


iNecas:
I took https://github.com/theforeman/foreman_remote_execution/pull/81 and https://github.com/adamruzicka/foreman_remote_execution/pull/1 and opened a new PR, as @adamruzicka is not here today and we would like to finalize the iteration 4

ares:
conditional ACK :-), I opened "redmine issue 12847":http://projects.theforeman.org/issues/12847 to add API in next iteration, also I'd like to see some refactorings so taht job_invocation does not have to be saved twice and `generate_description!` could be split into smaller methods as it's hard to read now

iNecas:
[test]

PR #89, opened 2015-12-16
Refs #10753 - update foreman-tasks to 0.7.8

iNecas:
There were some changes since 0.7.6 that are required by the recurring logic
so we should make sure we have the latest version available.


stbenjam:
ACK

stbenjam:
I think @iNecas's is already gone for the year, merging

PR #90, opened 2015-12-17
Fixes #12850 - Binds job_invocations with tasks using task groups

adamruzicka:


db/migrate/20151217092555_migrate_to_task_groups.rb
```diff
@@ -0,0 +1,17 @@
+class MigrateToTaskGroups < ActiveRecord::Migration
+  class FakeJobInvocation < ActiveRecord::Base
+    set_table_name 'job_invocations'
+  end
+
+  def up
+    say 'Migrating from locks to task groups'
+    FakeJobInvocation.where('task_group_id IS NULL').each do |job_invocation|
+      next if job_invocation.task_id.nil?
```
ohadlevy: include it in the original query?

db/migrate/20151217092555_migrate_to_task_groups.rb
```diff
@@ -0,0 +1,17 @@
...
+      task_group = JobInvocationTaskGroup.new
+      task_group.tasks << ForemanTasks::Task.find(job_invocation.task_id)
```
ohadlevy: why not
```ruby
tasg_groups.task_ids << job_invocation.task_id
```

db/migrate/20151217092555_migrate_to_task_groups.rb
adamruzicka: Good point

stbenjam:
@adamruzicka  Thanks much! :+1:  I can view my old invocations now.  Can you squash the commits into 1?

stbenjam:
Adam's gone and want to get the gem's done today.  Squashed in https://github.com/theforeman/foreman_remote_execution/pull/91

PR #91, opened 2015-12-17
Fixes #12850 - Binds job_invocations with tasks using task groups

stbenjam:


ares:
ACK since it's 1:1 to #90 :-)

stbenjam:
Thanks @ares and @adamruzicka !

PR #92, opened 2015-12-22
Fixes #12817 - swap links

ares:

PR #93, opened 2015-12-23
Fixes #12926 - add rerun to host row in job detail

ares:
based on https://github.com/theforeman/foreman_remote_execution/pull/92 to avoid conflicts, please review just the second commit or ideally review the second PR first...

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -77,9 +77,10 @@ module RemoteExecutionHelper
     end
   end
 
-  def template_invocation_actions(task, host)
+  def template_invocation_actions(task, host, job_invocation)
     [
-      display_link_if_authorized(_('View host'),  hash_for_host_path(host).merge(:auth_object => host, :permission => :view_hosts))
+      display_link_if_authorized(_('View host'), hash_for_host_path(host).merge(:auth_object => host, :permission => :view_hosts)),
+      display_link_if_authorized(_('Rerun on %s') % host.name, hash_for_rerun_job_invocation_path(:id => job_invocation, :host_ids => [ host.id ])),
     ]
   end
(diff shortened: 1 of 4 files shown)


app/helpers/remote_execution_helper.rb
```diff
@@ -77,12 +77,11 @@ def template_invocation_status(task)
...
-  def template_invocation_actions(task, host)
-    if task.nil?
-      []
-    else
-      [display_link_if_authorized(_("Details"), hash_for_template_invocation_path(:id => task).merge(:auth_object => host, :permission => :view_hosts))]
-    end
+  def template_invocation_actions(task, host, job_invocation)
+    [
+      display_link_if_authorized(_('View host'), hash_for_host_path(host).merge(:auth_object => host, :permission => :view_hosts)),
+      display_link_if_authorized(_('Rerun on %s') % host.name, hash_for_rerun_job_invocation_path(:id => job_invocation, :host_ids => [ host.id ])),
```
stbenjam: Is there a reason to show the hostname? Don't think we do that anywhere else in Foreman, do we?  I think it looks OK, but it'd be weird if there were multiple action items because you already know the host based on the row the button's in.

ares: I wanted to be clear about on which host it's going to be run, without following the whole line to the left. But I can easily change it to 'Rerun' only if you prefer. Or maybe 'Rerun on this host'

ares:
rebased, still ongoing discussion in outdated diff

app/helpers/remote_execution_helper.rb
stbenjam: No strong preference either way, the way it is now is OK

stbenjam:
APJ

PR #95, opened 2015-12-23
Fixes #12808 - add extra run job buttons

ares:


ares:
Images worth of thousand words :-)

rex1

_green button top right corner_

rex2

_first template action, only for job templates and ignoring snippets_

stbenjam:
ACK

ares:
Thanks, merging.

PR #96, opened 2016-01-04
Fixes #12976 - Rails 4 compatibility

dLobatog:


foreman_remote_execution.gemspec
```diff
@@ -19,7 +19,6 @@ Gem::Specification.new do |s|
   s.extra_rdoc_files = `git ls-files doc`.split("\n") + Dir['README*', 'LICENSE']
 
   s.add_dependency "deface"
-  s.add_dependency "rails", "~> 3.2.8"
```
dLobatog: There's already a dependency on Rails on core, no need to have it on plugins?

test/unit/job_invocation_composer_test.rb
```diff
@@ -21,16 +21,16 @@
     User.current.save
   end
 
-  let(:testing_job_template_1) { FactoryGirl.create(:job_template, :job_name => 'testing_job_template_1', :name => 'testing1', :provider_type => 'Ssh') }
```
dLobatog: Eh, this one is kind of dumb but Minitest 5 doesn't allow to write `let(:test**` and fails with ``let': name cannot begin with "test" (ArgumentError)`

app/models/targeting.rb
```diff
@@ -38,8 +38,8 @@ def resolve_hosts!
     raise ::Foreman::Exception, _('Cannot resolve hosts without a bookmark or search query') if bookmark.nil? && search_query.blank?
 
     self.search_query = bookmark.query if dynamic? && bookmark.present?
-    self.touch(:resolved_at)
     self.save!
+    self.touch(:resolved_at)
```
dLobatog: On Rails 4 you cannot touch objects that have not been saved yet.

iNecas:
Is this supposed to be a full compatibilty update for the rails 4? I did some work on this before leaving for christmas and I had more updates there (maily about find_by…) Should I open a PR with those changes and incorporate your as well?

lib/foreman_remote_execution/engine.rb
```diff
@@ -27,7 +29,7 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_remote_execution.register_plugin', after: :finisher_hook do |_app|
       Foreman::Plugin.register :foreman_remote_execution do
-        requires_foreman '>= 1.10'
+        requires_foreman '>= 1.11'
```
iNecas: Is this necessary?

test/unit/input_template_renderer_test.rb
```diff
@@ -288,7 +288,11 @@
           end
 
           context 'with existing variable implemented as smart variable' do
-            let(:puppet_class) { FactoryGirl.create(:puppetclass, :environments => [environment], :hosts => [renderer.host]) }
+            let(:puppet_class) do
```
dLobatog: I had to do this because the host_classes created by the old factory would not validate, because hosts on Puppetclass is not mass_assignable on Foreman anymore.

lib/foreman_remote_execution/engine.rb
dLobatog: Notice the change in line 17 about how migrate paths are loaded, the Railties API changed in a non-backwards compatible way. Other than that, I tried to keep changes backwards-compatible.

iNecas: I fail to see the backward incompatibility on the line 17: I see the `+` operator is not defined on the new structure (as it used to be an array before), but `<<` should work on both

stbenjam:
@dLobatog Thanks for this.  I tested it, and so far everything seems to work on Rails 4.  If you take the suggestions from @iNecas regarding using the `<<` operator and keeping the 1.10 requirement, we can see if goes green.

I'd like to get this in ASAP

test/unit/job_invocation_composer_test.rb
```diff
@@ -45,29 +45,29 @@
     describe '#available_templates' do
       it 'obeys authorization' do
         composer # lazy load composer before stubbing
-        JobTemplate.expects(:authorized).with(:view_job_templates).returns(JobTemplate.scoped)
```
iNecas: In rails 3, `all` returns array, while `scoped` returns the relation. Since it seems it's the only thing that beaks the Foreman 1.10 cpmopatibilty. I suggest replacing `scoped` with `where({})` instead, not the nicest (we can leave a TODO comment to replace it once there are more reasons to deprecate 1.10

lib/foreman_remote_execution/engine.rb
iNecas: Confirmed by testing: this doesn't break Foreman 1.10 compatibilty

test/unit/job_invocation_composer_test.rb
iNecas: hah :)

iNecas:
When running the tests on Rails 4, one test if failing due to this line: https://github.com/theforeman/foreman_remote_execution/blob/13fe40457f2dfa08139550339807b16ae3d1c967/test/unit/actions/run_proxy_command_test.rb#L53

When changing it to

```
task.ended_at = Time.at(timestamp + 1)
```

Seems like Rails 4 are more strict about what can be set to an attribute based on type. @dLobatog could you please fix this line as part of the PR?

iNecas:
I've tested the PR, and with the proposed changes, it worked both with Foreman 1.10/rails 3 and nightly/rails 4.

iNecas:
Here are the changes I'm talking about http://fpaste.org/307276/19991291/raw/

dLobatog:
Updated, thanks!

iNecas:
ACK, pending rubocop fix

dLobatog:
Rubocop should pass now :smile:

iNecas:
There are some Foreman integration tests failing, but they don't seem to be related to anything in here

ares:
I've seen this already, seems intermittent.

stbenjam:
[test]

iNecas:
It turns out some fix for foreman 1.11 in foremant-tasks broke running tests with 1.10, https://github.com/theforeman/foreman-tasks/pull/164 to fix that

iNecas:
[test] hopefully last try :)

iNecas:
This PR has some spell on it :) in the mean-time net-ldap caused incomapatibility with ruby 1.9, the fix is not in foreman, waiting for master build to be successful http://ci.theforeman.org/job/test_plugin_foreman_remote_execution_master/

iNecas:
It's all green now :fireworks: :birthday:

iNecas:
Thanks @dLobatog !

PR #98, opened 2016-01-04
Fixes #12928 - Keeps description fields if job invocation is invalid

adamruzicka:


app/views/job_invocations/_form.html.erb
```diff
@@ -88,11 +88,18 @@
                       <% end %>
 
                     <% end %>
-                    <%= textarea_f f, :description, :label => _("Description"), :readonly => "true", :id => "description", :rows => 2, :help_inline => content_tag(:span, check_box_tag('description_format_override', job_template.generate_description_format, true, :onchange => "description_override(this);") + ' ' + _("Use default description template")) %>
-                    <div id="description_format_container" class="hidden">
+
+                    <% description_checkbox = content_tag(:span, check_box_tag('description_format_override',
```
ohadlevy: consider using an helper instead?

app/helpers/remote_execution_helper.rb
```diff
@@ -194,4 +194,28 @@ def time_ago(time)
     end
   end
 
+  # Requires JS from template_invocation.js
```
ares: sorry I know you were told to extract it to helper but to me it sounds more natural to become a partial as this helper does not have other than rendering logic or do I miss something? maybe a helper could be only `description_format_override_checkbox_f`  and `description_format_textarea_f` that would make more sense to me

ares:
except one small thing it looks good, feel free to squash and please rebase so I can test it

ares:
It seems to fix the issue only for templates with some description format, with blank description format, I'm getting %{job_name} as new description template no matter what I type in. Note that this template has one input. Also when the template does not have any input and description format, description preview is blank when I enter the form, it gets filled after I click the checkbox for the first time. Interestingly when the template does not have any input, description template works just fine.

iNecas:
rebase needed

ares:
:cop: 
```
Offenses:

/var/lib/workspace/workspace/test_plugin_pull_request@2/database/mysql/label/fast/ruby/2.1/plugin/app/helpers/remote_execution_helper.rb:214:1: C: Trailing whitespace detected.
```

ares:
It persist it on failed invocation, that's good but unfortunately it gets cleared when I click resolv hosts button or I change Search query.

ares:
[test]

ares:
Works fine now, thanks @adamruzicka, merging!

adamruzicka:
Hooray :tada:, thanks for bearing with me

PR #99, opened 2016-01-05
Fixes #12815 - unify start time information

ares:
While it's not blocked by https://github.com/theforeman/foreman-tasks/pull/161 it would be better to get that one in first, I'd remove the TODO and use delegate instead. We'll see how fast would that PR get in, but it shouldn't blocking the review.


ares:
I'll rebase one new foreman-tasks so we can use the delegate

ares:
rebased using new foreman-tasks, ready for reviewing

stbenjam:
Looks good, ACK

ares:
Thanks, merging.

PR #100, opened 2016-01-06
Fixes #13011 - use standard way to join with task

iNecas:
Otherwise, the scoped search has issues producing valid sql

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -36,7 +36,11 @@ class JobInvocation < ActiveRecord::Base
   belongs_to :triggering, :class_name => 'ForemanTasks::Triggering'
   has_one :recurring_logic, :through => :triggering, :class_name => 'ForemanTasks::RecurringLogic'
 
-  scope :with_task, -> { joins('LEFT JOIN foreman_tasks_tasks ON foreman_tasks_tasks.id = job_invocations.task_id') }
+  if Rails::VERSION::MAJOR >= 4
+    scope :with_task, -> { references(:task) }
+  else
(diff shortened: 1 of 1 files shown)


app/models/job_invocation.rb
```diff
@@ -36,7 +36,7 @@ class JobInvocation < ActiveRecord::Base
   belongs_to :triggering, :class_name => 'ForemanTasks::Triggering'
   has_one :recurring_logic, :through => :triggering, :class_name => 'ForemanTasks::RecurringLogic'
 
-  scope :with_task, -> { joins('LEFT JOIN foreman_tasks_tasks ON foreman_tasks_tasks.id = job_invocations.task_id') }
+  scope :with_task, -> { joins(:task) }
```
ohadlevy: :+1: what was the original reasoning before?

ares: it was originally added in https://github.com/theforeman/foreman_remote_execution/commit/a9c05c8c7dfe8d6dfcf91c782d9b9985f80a19fb#diff-d3ff599b762e4ff95c9b2776d2be1079R32 in https://github.com/theforeman/foreman_remote_execution/pull/76, maybe @adamruzicka would remember... could be some issue in rails 3 AR or it just slipped through

iNecas: Ok, we needed the outer join, which turned out to be present in Rails 4 via `references`

iNecas:
In case we are dropping Foreman 1.10 starting iteration 5, we can remove the 1.10 compatibility condition

ares:
ACK

PR #101, opened 2016-01-06
Fixes #13016 - introduce job invocation status

ares:


ares:
please see redmine issue for more details and motivation

iNecas:
[test]

app/models/host_status/execution_status.rb
```diff
@@ -37,6 +41,47 @@ def to_label(options = {})
...
 
+  class ExecutionTaskStatusMapper
+    attr_accessor :task
+
+    def self.sql_conditions_for(status)
+      status = STATUS_NAMES.key(status) if status.is_a?(String)
+
+      case status
+        when OK
+          [ "state = 'stopped' AND (result = 'success' OR result = 'warning')" ]
```
iNecas: result `warning` probably deserves separate state: It's a bit strange for a job with some failed hosts to be marked `success`, possible variants `warning` or `partial sucess` or something else?

app/models/host_status/execution_status.rb
```diff
@@ -1,17 +1,21 @@
 class HostStatus::ExecutionStatus < HostStatus::Status
+  # execution finished successfully
   OK = 0
```
iNecas: what about using `:symbols` instead nothing-telling integers?

ares: We should never use that integer directly, it's just storage value so why we'd store them as symbol/text? All host statuses are stored as numbers in Foreman (and this state already used OK = 0 and ERROR = 1) I just added two new.

app/models/host_status/execution_status.rb
ares: I don't think it's a good idea on this level, if the job succeeded (no matter if someone had to help it) you want to find at among successful finished ones, if you really care about details, you can go to foreman task. Maybe we could add a warning icon next to status linking to the task.

app/models/host_status/execution_status.rb
iNecas: What's the advantage of numbers over symbols. Just minor readability enhancement. If nothing else, debugging the variable with the status would be nicer.

app/models/host_status/execution_status.rb
iNecas: I thing marking a job with some failed hosts as `success` is a bug

ares: But warning does not mean there's some host that failed? That would end up as error I thought. I think there's not even many situation in rex in which a task can end up with warning, only I can think of is some dynflow issue after which you have to resume the task manually? If I get it wrong and warning always means that something failed and was not fixed, then I'm happy to change it to ERROR.

app/models/host_status/execution_status.rb
ares: I think these were reasons why host status was implemented with numbers
* compatibility with bitmask field that's used in configuration status (same is used by openscap status)
* column size

Another benefit I see is that if some state changes later because we find better word for it, we don't have to change existing records nor the value we display in API.

If we introduced symbol representation we'd have to provide some mapper, since our ExecutionStatus inherits from core HostStatus::Status object. I don't thing it's worth it.

app/models/host_status/execution_status.rb
iNecas: in rex, warning in dynflow RunHostsJob means mainly that some sub-task has failed - so it's this case (try it). I still thing it would deserve separate state than error (where error is that nothing even started), but it would be still better, than success.

app/models/host_status/execution_status.rb
iNecas: Ah, I haven't realized it's coming from the Foreman host status… nm then. Thanks for explanation

app/models/host_status/execution_status.rb
ares: oh, I'd consider this as a bug in foreman-tasks then, anyway moving warning to ERROR state for now because I'd expect "partial failed" being failed as well

iNecas:
ACK

ares:
thanks, merging

PR #102, opened 2016-01-06
Fixes #12814 - ship default bookmarks for invocations

ares:
Based on https://github.com/theforeman/foreman_remote_execution/pull/101 which is prerequisity.

### DIFF
--- a/db/seeds.d/90-bookmarks.rb
+++ b/db/seeds.d/90-bookmarks.rb
@@ -1,0 +1,19 @@
+Bookmark.without_auditing do
+  [
+    { :name => 'failed', :query => "status = failed", :controller => 'job_invocations' },
+    { :name => 'succeeded', :query => 'status = succeeded', :controller => 'job_invocations'},
+    { :name => 'running', :query => 'status = running', :controller => 'job_invocations'},
+    { :name => 'pending', :query => 'status = queued or status = running', :controller => 'job_invocations'},
+    { :name => 'recurring', :query => 'recurring = true', :controller => 'job_invocations'},
+    { :name => 'recent', :query => 'started_at > "24 hours ago"', :controller => 'job_invocations'},
+  ].each do |input|
+    next if Bookmark.where(:controller => 'job_invocations').find_by_name(input[:name])
+    # TODO audit should be fixed once core #13109 gets merged
+    next if audit_modified? Bookmark, input[:name]
+    attributes = { :public => true }.merge(input)
+    b = Bookmark.where(:name => input[:name], :controller => input[:controller]).first || Bookmark.new
+    b.attributes = attributes
+    b.save
+    raise "Unable to create bookmark: #{format_errors b}" if b.errors.any?
+  end
+end
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -17,8 +17,6 @@ class JobInvocation < ActiveRecord::Base
 
   scoped_search :on => :job_name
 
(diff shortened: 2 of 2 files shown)


db/seeds.d/90-bookmarks.rb
```diff
@@ -0,0 +1,18 @@
+Bookmark.without_auditing do
+  [
+    { :name => 'failed', :query => "status = failed", :controller => 'job_invocations' },
+    { :name => 'succeeded', :query => 'status = succeeded', :controller => 'job_invocations'},
+    { :name => 'running', :query => 'status = running', :controller => 'job_invocations'},
+    { :name => 'pending', :query => 'status = queued or status = running', :controller => 'job_invocations'},
+    { :name => 'recurring', :query => 'recurring = true', :controller=> 'job_invocations'},
```
stbenjam: Can we have another bookmark - maybe "recent"? Like `started_at > "24 hours ago"` - only because I don't think it's super intuitive for how that works, so it gives the user something to look at as an example

app/helpers/remote_execution_helper.rb
```diff
@@ -30,15 +30,17 @@ def job_invocation_chart(invocation)
...
+      when HostStatus::ExecutionStatus::QUEUED
+        _('queued')
+      when HostStatus::ExecutionStatus::RUNNING
+        _('running %s%') % invocation.progress
+      when HostStatus::ExecutionStatus::OK
+        _('succeeded')
+      when HostStatus::ExecutionStatus::ERROR
+        _('failed')
+      else
+        _('unknown status')
```
stbenjam: Is it possible to show these as labels in the table columns? I kind of like how it looks now, e.g. `Running 50%` with a capital letter.  Maybe that's really nitpicky I guess

db/seeds.d/90-bookmarks.rb
ares: sure, but let's wait until https://github.com/theforeman/foreman_remote_execution/pull/99 as currently there's no single start_at column

app/helpers/remote_execution_helper.rb
ares: hm, I wanted to be consistent with time info we display in start column but if you want I can change it back to capital letters, not a big deal :-) also the percentage will hopefully be later replaced by some small chart (as it was in wireframes)

db/seeds.d/90-bookmarks.rb
stbenjam: started_at would cover jobs that already started, right? That's there now, and that's all "recent" would be interested in I think.

app/helpers/remote_execution_helper.rb
stbenjam: It's fine, it can stay like this.  I can't really tell you any justification for capitalizing it other than it looks better to me.

ares: I also wanted to display the value in a same way as is used in search syntax, we could support both there but I'd prefer downcase

db/seeds.d/90-bookmarks.rb
ares: ok, added

iNecas:
[test]

iNecas:
One comments, good otherwise

ares:
@iNecas should be fixed now, also sent the fix against Foreman which has the same issue

iNecas:
ACK

ares:
just rebased, merging after Mr Jenkins approval

ares:
thanks @iNecas and Mr. Jenkins for reviewing, merging

PR #104, opened 2016-01-07
Fixes #11724 - load the job template

ares:


iNecas:
It it supposed to work with foreman 1.10? 
new-template-preview

ares:
Interesting, I didn't tested with 1.10 but I don't see anything incompatible. Does this happen for new templates or when you edit existing one (I guess the first case)?

iNecas:
First case

ares:
@iNecas this should work with 1.10 and develop, would you mind retesting?

iNecas:
Works nicely. One idea: nicer error message when template inputs are defined, explaining that the template needs to be saved in order to work (or fixing the inputs case as well:)

ares:
Like this?

iNecas:
works for me, ACK

ares:
Thanks, merging

PR #103, opened 2016-01-06
fixes #12847 - api for job descriptions

stbenjam:
~~Built on top of Daniels' rails4 PR since I've migrated my dev environment.  Will rebase once https://github.com/theforeman/foreman_remote_execution/pull/96 is merged~~

PR #105, opened 2016-01-08
Fixes #13071 - add inline help for job_name placeholder

ares:


app/views/job_templates/_custom_tabs.html.erb
```diff
@@ -10,7 +10,8 @@
 
   <%= text_f f, :description_format,
              :help_inline => popover(_('Explanation'), _('Description template determines the job name once it is submitted. Input values can become part of the name
-                         if they are specified using interpolation syntax, e.g. %{fqdn} where fqdn is the name of interpolated input.')) %>
+                         if they are specified using interpolation syntax, e.g. %{fqdn} where fqdn is the name of interpolated input.
+                         An exception is %{job_name} which is always replaced by template name.')) %>
```
stbenjam: %job_name renders as the *template* name? That sounds more like a bug.

ares: oops, it's a invocation job name, I took this from another PR that was replacing job name with template name, will fix the help

ares:
@stbenjam I reworded it, does it make more sense now?

stbenjam:
Not really :grin: The text itself is a little confusing to me. E.g. "Description template determines the job name." It doesn't, it determines the description.  

How about this?

*This template is used to generate the description. Input values can be used using the syntax %{fqdn}. You may also include the job name using %{job_name}.*

adamruzicka:
+1 for @stbenjam's suggestion

ares:
ok, I've changed it to your suggestion, just note that we display this description under job name column in job invocations table...

ares:
[test]
```Fatal server error:
(EE) Server is already active for display 3
	If this server is no longer running, remove /tmp/.X3-lock
	and start again.
```

stbenjam:
> ok, I've changed it to your suggestion, just note that we display this description under job name column in job invocations table...

Which also needs fixing at some point, it's not the job name

ares:
agreed, got it captured from today meeting, I guess it's not blocking this though?

stbenjam:
Nope, I didn't notice the problem was deeper than just that popover so we can fix the rest as part of a separate issue.  

ACK

stbenjam:
Merging since I want the same description text in the API for https://github.com/theforeman/foreman_remote_execution/pull/103/

stbenjam:
Thanks @ares

PR #107, opened 2016-01-08
fixes #13088 - use !blank? instead of present? in ssh key snippet

stbenjam:
Reported on IRC by a user.  present? not allowed in safemode on Array. I thought I tested this with safemode enabled, but guess not.  stbenjam--


stbenjam:
Also proposed https://github.com/svenfuchs/safemode/pull/14, because I tend to do this every now and then with these methods, but we don't need to wait for that to change the snippet.

ares:
Thanks @stbenjam, merging.

PR #108, opened 2016-01-11
Fixes #11406 - Rename Ssh to SSH

ares:


stbenjam:
Do you think it's worth renaming the provider's key too for consistency's sake? https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L152

stbenjam:
It also looks like `with_features` is case-sensitive, so this thingy doesn't return any proxies for the tests which still use Ssh in a few places. https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/concerns/foreman_remote_execution/host_extensions.rb#L43

ares:
right, I renamed it everywhere except proxy action name, could you please give it another look? (note that you have to rerun the migration since the provider type of job template has been changed as well)

stbenjam:
Awesome! Tested and works great, ACK

ares:
Thanks for review, merging!

PR #109, opened 2016-01-12
Fixes #12812 - replace job name by job category

ares:


stbenjam:
A couple of initial thoughts/questions:

1. What would be involved in renaming the column to job_category? I think the main concern was the amount of logic tied up in the concept of job name that we're moving to the category, but just a rename should be possible? Otherwise it shows up too many places for the user and I think we're going to make it more confusing for them - in the API, scoped search fields, description template, etc.

2. Can we put all existing templates to be in the "Miscellaneous" category? I think it looks nicer if all our defaults are just in one category.  The user can re-categorize them later if they want.

stbenjam:
3. Would it be possible to make the API able to only take a template name? Not sure I like the usability of requiring a category and a template name like it is now

ares:
1. plain renaming the attribute should be ok, it will change API and parameters structure but I guess in this stage of development, it's not a big deal, I'll look at that tomorrow

2. do we allow to change category of locked template? and if we do, is that expected? :smiley: I actually like some existing categories that illustrate why they are useful, I'd maybe restructure them a bit, so we'd have "Service", "Package", "Commands". Run puppet and arbitrary commands would be in Commands. In package we could have Package action (or 3 templates for install/update/remove) and in Service we could have restart/enable/disable. Just a brainstorming, what do you think?

3. re API I'll take a look how much it would change, I think it should be easily doable in API param mapper, it would load the job category automatically.

stbenjam:
> do we allow to change category of locked template? and if we do, is that expected? :smiley: 

You can in seeds or a migration but not in the UI. The only broken thing with locks right now is inputs - those can be changed regardless, there's an open Redmine for that.

Should users be able to change the category? We can make that one of the things that's allowed to be changed.

> I actually like some existing categories that illustrate why they are useful, I'd maybe restructure them a bit, so we'd have "Service", "Package", "Commands". Run puppet and arbitrary commands would be in Commands. In package we could have Package action (or 3 templates for install/update/remove) and in Service we could have restart/enable/disable. Just a brainstorming, what do you think?

Yea that sounds better, just the existing job names didn't sound like categories, which is why I wanted to rename them.  I do like just one template 'package action' supporting multiple actions, but we could discuss that more to see if it makes sense to break them into one-template-per-action.

ares:
I reported the template name in invocation api as http://projects.theforeman.org/issues/13239, I'd prefer to keep it separate as it implies unique template names. User can already use id so it shouldn't block this PR.

I kept commits separate for easier review, could you please take a look?

app/views/job_invocations/_form.html.erb
```diff
@@ -45,31 +61,21 @@
...
+          <span class="help-inline"><%= popover(_('Explanation'),
+                                        _("Type has impact on when is the query evaulated to hosts.<br><ul><li><b>Static</b> - evaluates just after you submit this form</li><li><b>Dynamic</b> - evaluates just before the execution is started, so if it's planed in future, targeted hosts set may change before it</li></ul>")) %>
+          </span>
         </div>
       <% end %>
 
       <% @composer.displayed_provider_types.each do |provider_type| %>
         <fieldset id="provider_<%= provider_type %>" class="provider_form">
-          <legend><%= _('Provider') + ': ' + provider_type %></legend>
+          <!--<legend><%= _('Provider') + ': ' + provider_type %></legend>-->
```
iNecas: History became legend. Legend became myth. And for two and a half thousand years somebody removed this commented code from the repository (I've been watching Lord of the rings recently:)

app/views/job_templates/_custom_tabs.html.erb
```diff
@@ -1,17 +1,18 @@
...
+    <%= auto_complete_search(:job_category,
+                             f.object.job_category,
+                             :placeholder => _("Job category") + ' ...',
+                             :name => 'job_template[job_category]',
                              :id => 'search') %>
   <% end %>
 
   <%= text_f f, :description_format,
              :help_inline => popover(_('Explanation'), _('This template is used to generate the description.
                                                           Input values can be used using the syntax %{fqdn}.
```
iNecas: I wonder, if the `%{fqdn}` is a good example here, as the description is generated for a group of hosts. I know it's not a direct subject of this PR, but while touching inline help, I'm ok with improving it.

app/controllers/api/v2/job_templates_controller.rb
```diff
@@ -30,10 +30,11 @@ def show
...
           param :name, String, :required => true, :desc => N_('Template name')
-          param :job_name, String, :required => true, :desc => N_('Job name')
-          param :description_format, String, :required => false, :desc => N_('This template is used to generate the description. Input values ' +
-                                                                             'can be used using the syntax %{fqdn}. You may also include the ' +
-                                                                             'job name using %{job_name}.')
+          param :job_category, String, :required => true, :desc => N_('Job category')
+          param :description_format, String, :required => false, :desc => N_('This template is used to generate the description.' +
+                                                                             'Input values can be used using the syntax %{fqdn}.' +
+                                                                             'You may also include the job category and template' +
+                                                                             'name using %{job_category} and %{template_name}.')
```
stbenjam: Nitpick, please add spaces after the periods, the first and second sentences would render like `'description.Input'`

app/views/job_templates/auto_complete_job_category.json.erb
```diff
@@ -0,0 +1,3 @@
+[
+  <%= @job_categories.map { |name| '{"completed": "","part": "' + name + '","label": "' + name + '","category": ""}' }.join(',').html_safe %>
```
iNecas: Shouldn't we use a hash serialized to json via some JSON library: it's not that hard to create a name that would break this code as we're missing escaping of the values.

app/models/job_invocation.rb
```diff
@@ -29,7 +29,7 @@ class JobInvocation < ActiveRecord::Base
 
   has_many :tasks, :through => :task_group, :class_name => 'ForemanTasks::Task'
 
-  scoped_search :on => [:job_name], :complete_value => true
+  scoped_search :on => [:job_category], :complete_value => true
```
stbenjam: How's this different than Line #18?

db/migrate/20160114120200_rename_job_categories.rb
```diff
@@ -0,0 +1,20 @@
+class RenameJobCategories < ActiveRecord::Migration
+  MAP = {
+    'Package Action - SSH Default' => ['Package Action', 'Packages'],
```
iNecas: The conclusion was not to rename the template name and keep the `SSH Default` there, right?

app/views/job_templates/_custom_tabs.html.erb
stbenjam: Ah, good point, maybe `%{package}` or `%{input_name}`

stbenjam:
I think we should have a migration to update the already seeded templates, or make the db:seeds update them on every update, but we would need to disable Foreman's strict requirement on seeds being idempotent or fix the logic.  Otherwise the old description template is rendered and I see this:


title

app/views/templates/run_command.erb
```diff
@@ -1,8 +1,8 @@
 <%#
 kind: job_template
 name: Run Command - SSH Default
-job_name: Run Command
-description_format: "%{job_name} %{command}"
+job_category: Commands
+description_format: "%{job_category} %{command}"
```
stbenjam: I think this needs a better description, because now it renders `Commands hostname -f`

Maybe something simple like `Run %{command}`

stbenjam:
Tested, and it works well other than notes above.  Very, very nice :+1:

db/migrate/20160114120200_rename_job_categories.rb
ares: yes since we'll introduce uniqueness validation soon

ares:
I think I've addressed all comments, please re-review so we can get this in.

db/migrate/20160114120200_rename_job_categories.rb
iNecas: Ah sry, I've misread the code

app/controllers/job_templates_controller.rb
```diff
@@ -6,8 +6,9 @@ def load_vars_from_template
     @organizations    = @template.organizations
   end
 
-  def auto_complete_job_name
-    @job_names = resource_base.where(['job_name LIKE ?', "%#{params[:search]}%"]).pluck(:job_name).uniq
+  def auto_complete_job_category
+    @job_categories = resource_base.where(['job_category LIKE ?', "%#{params[:search]}%"]).pluck(:job_category).uniq
+    render :json => @job_categories.map { |name| { 'completed' => '', 'part' => name, 'label' => name, 'category' => '' } }.to_json
```
iNecas: Much better! :+1:

iNecas:
Haven't tested but :+1: codewise from me

stbenjam:
Tested with changes, works very nicely, APJ from me! :tada:

ares:
Thanks guys, merging! I added changes in manual to "missing documentation" list.

PR #112, opened 2016-01-13
Fixes #13173 - Documentation button links for job invocations/templates

dLobatog:
As the rest of Foreman, this plugin can benefit from links to the
documentation straight from the app. Two easy places to start are the
job invocations/ job templates indexes.

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -90,11 +90,10 @@ module RemoteExecutionHelper
   end
 
   def job_invocations_buttons
-    if authorized_for(hash_for_new_job_invocation_path)
-      link_to(_("Run Job"), new_job_invocation_path,
-              :class => "btn btn-default",
-              :title => _('Run Job'))
-    end
+    [
+      documentation_button_rex('3.2ExecutingaJob'),
+      display_link_if_authorized(_("Run Job"), hash_for_new_job_invocation_path)
+    ]
   end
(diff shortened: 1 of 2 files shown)


app/helpers/remote_execution_helper.rb
```diff
@@ -194,4 +193,13 @@ def time_in_words_span(time)
     end
   end
 
+  def documentation_button_rex(section = '')
+    url = "http://theforeman.org/plugins/foreman_remote_execution/\
+#{ForemanRemoteExecution::VERSION}/index.html#" + section
```
ares: could you please change it to
```ruby
url = 'http://theforeman.org/plugins/foreman_remote_execution/' +
  "#{ForemanRemoteExecution::VERSION}/index.html#" +
  section
```

ares:
otherwise looks good

app/helpers/remote_execution_helper.rb
stbenjam: Wouldn't we need something like `ForemanRemoteExecution::VERSION.split('.').take(2).join('.')` since this resolves to 0.1.2?

ares: right, I didn't test the PR yet, but makes sense

dLobatog:
@ares Updated to disregard the patch version on the URL

ares:
Thanks @dLobatog, merging!

PR #113, opened 2016-01-13
fixes #12684 - introduce api for scheduling / recurring invocations

stbenjam:
Needs the views from https://github.com/theforeman/foreman-tasks/pull/165, still a WIP, having everything not be first class citizens on JobInvocation causes so much need for hokiness.

Anyway, first shot at the API - thoughts?

### DIFF
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -61,7 +61,7 @@ class JobInvocationComposer
     def params
       { :job_category => job_category,
         :targeting => targeting_params,
-        :triggering => {},
+        :triggering => triggering_params,
         :description_format => api_params[:description_format] || template_with_default.generate_description_format,
         :template_invocations => template_invocations_params }.with_indifferent_access
     end
@@ -73,6 +73,29 @@ class JobInvocationComposer
     def targeting_params
       raise ::Foreman::Exception, _('Cannot specify both bookmark_id and search_query') if api_params[:bookmark_id] && api_params[:search_query]
       api_params.slice(:targeting_type, :bookmark_id, :search_query).merge(:user_id => User.current.id)
+    end
+
+    def triggering_params
+      raise ::Foreman::Exception, _('Cannot specify both recurrence and scheduling') if api_params[:recurrence].present? && api_params[:scheduling].present?
+
+      if api_params[:recurrence].present?
+        {
+          :mode => :recurring,
+          :cronline => api_params[:recurrence][:cron_line],
+          :end_time => format_datetime(api_params[:recurrence][:end_time]),
+          :input_type => :cronline,
+          :max_iteration => api_params[:recurrence][:max_iteration]
+        }
+      elsif api_params[:scheduling].present?
+        {
+          :mode => :future,
+          :start_at_raw => format_datetime(api_params[:scheduling][:start_at]),
+          :start_before_raw => format_datetime(api_params[:scheduling][:start_before]),
+          :end_time_limited => api_params[:scheduling][:start_before] ? true : false
+        }
+      else
+        {}
+      end
     end
 
     def template_invocations_params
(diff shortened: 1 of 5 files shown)


stbenjam:
Some examples:

### Recurring

POST /api/job_invocations

```json
{
  "job_invocation": {
    "job_name": "Run Command",
    "inputs": {
      "command": "ping -c 1 google.com"
    },
    "recurrence": {
      "cron_line": "*/5 * * * *"
    },
    "search_query": "name ~ rex02",
    "targeting_type": "static_query"
  }
}
```

### Scheduled

POST /api/job_invocations

```json
{
  "job_invocation": {
    "job_name": "Run Command",
    "inputs": {
      "command": "ping -c 1 google.com"
    },
    "scheduling": {
      "start_at": "2016-01-14T13:00:00+00:00"
    },
    "search_query": "name ~ rex02",
    "targeting_type": "static_query"
  }
}
```

adamruzicka:
I like it so far. Do we want to allow only cronline input through api or will the other modes (daily, weekly, etc) be added later?

stbenjam:
I'm not sure, TBH.  Another topic to discuss for tomorrow maybe.  I think we could show all the various options, like daily, monthly, etc, if its needed.  Not sure if that makes the API or CLI too confusing with so many options, the UI is easier since you can show/hide things based on context.

ares:
please rebase so I can test (sorry, there's happening too much at the same time these days)

app/models/job_invocation_composer.rb
```diff
@@ -75,6 +75,29 @@ def targeting_params
       api_params.slice(:targeting_type, :bookmark_id, :search_query).merge(:user_id => User.current.id)
     end
 
+    def triggering_params
+      raise ::Foreman::Exception, _('Cannot specify both recurrence and scheduling') if api_params[:recurrence].any? && api_params[:scheduling].any?
+
+      if api_params[:recurrence].any?
+        {
+          :mode          => :recurring,
+          :cronline      => api_params[:recurrence][:cron_line],
```
ares: Could you please avoid aligning hash :rocket:s? When we later change one line we have to realign all lines which makes harder to find the real change in git diff.

app/models/job_invocation_composer.rb
```diff
@@ -98,6 +121,12 @@ def template_with_default
       return template
     end
 
+    private
+
+    def format_datetime(datetime)
+      return datetime if datetime.blank?
+      DateTime.parse(datetime).strftime('%Y-%m-%d %H:%M')
```
ares: I think this could raise ArgumentError if user pass invalid date format, which would result in 500 (not tested, can do after rebase)

ares:
please rebase, except one style comment and one small thing it looks good, looking forward to test

app/models/job_invocation_composer.rb
stbenjam: But aligned hash :rocket: are so pretty :heart_eyes: I get what you're saying though, fixed

app/models/job_invocation_composer.rb
stbenjam: Yes probably but I think the API controller should bubble the message up to the user no? Will test it and see

ares:
there seems to be many test failures

stbenjam:
Most of them are expected, as the foreman-tasks PR linked in the description is needed.

iNecas:
rebase needed

iNecas:
ACK, tested and works great

stbenjam:
Right, this needs all of the foreman-tasks stuff first, it makes use of a recurring logic partial.  It doesn't have to, I guess.  But that PR is https://github.com/theforeman/foreman-tasks/pull/165.

stbenjam:
But the advantage of using the partial for recurrence, is in case we add some things later to recurrence, then it shows up in the invocation API too.

iNecas:
Right, let me test recurring logic PR and get that in first

iNecas:
[test]

iNecas:
foreman-tasks 0.7.11 with required changes was just pushed to rubygems

stbenjam:
@iNecas Should we think about running tests with foreman-tasks from git?

stbenjam:
[test]

stbenjam:
Green now, thanks

PR #78, opened 2015-11-20
refs #12490 - changes needed for hammer cli

stbenjam:
Some API tweaks for the hammer cli to work well.  Goes with https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/1.

 :heavy_check_mark: TODO
- [x] Template Input API
- [x] Invocation output API
- [x] template_id => job_template_id in params for hammer id resolution
- [x] RABL clean-up
- [x] Additional API test coverage
- [x] Permissions

PR #114, opened 2016-01-13
fixes #13147 - rename job name to job description on invocations index

stbenjam:


app/models/job_invocation.rb
```diff
@@ -15,7 +15,8 @@ class JobInvocation < ActiveRecord::Base
   validates :job_name, :presence => true
   validates_associated :targeting, :template_invocations
 
-  scoped_search :on => :job_name
+  scoped_search :on => :job_name, :complete_value => true
+  scoped_search :on => :description, :complete_value => true
```
adamruzicka: I'm not sure whether allowing autocomplete on description is a good idea since it does not escape quotes and searching fails because of it. See [here](https://imgur.com/a/fOwjW).

ares: I'd say it's generic scoped search issue and searching can still still useful, 3/4 templates we ship do not contain quotes. @stbenjam would you mind opening scoped_search and a tracking issue and linking it to #13147 for this? I think it does not block this PR to be merged right?

ares:
Since https://github.com/theforeman/foreman_remote_execution/pull/78/files#diff-d3ff599b762e4ff95c9b2776d2be1079R41 the sorting by job_name (and now description) does not work since the order is always set by default scope. Could you please remove default scope and set default ordering explicitly? Default scopes are bad, they cause too many surprises.

stbenjam:
Up to you, it's a little related since we're touching the field here. I have no problem doing it unless you want to take care of it

ares:
I'll fix it, I love deleting `default_scope`s :-) Reported as http://projects.theforeman.org/issues/13226

app/models/job_invocation.rb
stbenjam: If you look at @adamruzicka's example, default descriptions render quotes around the input values.  I think it makes sense to remove the auto complete for now.

Filed https://github.com/wvanbergen/scoped_search/issues/138

stbenjam:
Ok, no problem, don't forget about the API controller, it needs the order too so hammer displays the newest first.

ares:
Thanks @stbenjam, merging!

PR #115, opened 2016-01-13
fixes #13134 - Use patternfly icons

stbenjam:


dLobatog:
@stbenjam  `job_invocations/_tab_hosts` and `_form` also use glyphicons

stbenjam:
Thanks, updated

app/helpers/remote_execution_helper.rb
```diff
@@ -60,19 +60,19 @@ def host_counter(label, count)
...
           else
-            content_tag(:i, '&nbsp'.html_safe, :class => 'glyphicon glyphicon-exclamation-sign') + content_tag(:span, _('failed'), :class => 'status-error')
+            icon_text('error-circle-o', _('failed'), :kind => 'pficon')
           end
         when 'success'
-          content_tag(:i, '&nbsp'.html_safe, :class => 'glyphicon glyphicon-ok-sign') + content_tag(:span, _('success'), :class => 'status-ok')
+          icon_text('ok', _('success'), :kind => 'pficon')
         when 'pending'
-          content_tag(:i, '&nbsp'.html_safe, :class => 'glyphicon glyphicon-question-sign') + content_tag(:span, _('pending'))
+          icon_text('question', _('pending'), :kind => 'fa')
```
iNecas: The question mark on pending might be misleading for somebody (or the whole pending word), running would be better probably (but not sure I like the patterfly running icon), anyway, not necessary to be done as part of this PR

iNecas:
one comment, but necesary needed to do on this PR, tested, ACK

app/helpers/remote_execution_helper.rb
stbenjam: I agree, do you want to file a separate issue? The question mark itself looks like a possible error as well, certainly could be confusing.

iNecas: +1 for an issue

PR #117, opened 2016-01-14
Fixes #13130 - save template invocations for each host

ares:
This commit introduces so called pattern template invocation which is
created by job composer as a pattern for real template invocation that
being created during dynflow action run. Pattern template invocation
holds all information from composer (e.g. input values) except target
host which is evaluated just during dynflow action run phase.

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -3,17 +3,24 @@ class JobInvocation < ActiveRecord::Base
 
   include ForemanRemoteExecution::ErrorsFlattener
   FLATTENED_ERRORS_MAPPING = {
-    :template_invocations => lambda do |template_invocation|
+    :pattern_template_invocations => lambda do |template_invocation|
       _('template') + " #{template_invocation.template.name}"
     end
   }
 
   belongs_to :targeting, :dependent => :destroy
-  has_many :template_invocations, :inverse_of => :job_invocation, :dependent => :destroy
+  has_many :all_template_invocations, :inverse_of => :job_invocation, :dependent => :destroy, :class_name => 'TemplateInvocation'
+  if Rails::VERSION::MAJOR >= 4
+    has_many :template_invocations, -> { where('host_id IS NOT NULL') }, :inverse_of => :job_invocation
+    has_many :pattern_template_invocations, -> { where('host_id IS NULL') }, :inverse_of => :job_invocation, :class_name => 'TemplateInvocation'
+  else
+    has_many :template_invocations, :conditions => 'host_id IS NOT NULL', :inverse_of => :job_invocation
+    has_many :pattern_template_invocations, :conditions => 'host_id IS NULL', :inverse_of => :job_invocation, :class_name => 'TemplateInvocation'
+  end
 
   validates :targeting, :presence => true
   validates :job_category, :presence => true
-  validates_associated :targeting, :template_invocations
+  validates_associated :targeting, :all_template_invocations
 
   scoped_search :on => :job_category, :complete_value => true
   scoped_search :on => :description # FIXME No auto complete because of https://github.com/wvanbergen/scoped_search/issues/138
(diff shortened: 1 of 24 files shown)


app/models/job_invocation.rb
```diff
@@ -94,24 +98,16 @@ def to_action_input
     { :id => id, :name => job_name }
   end
 
-  def template_invocation_tasks
```
ares: replaced by association, see above

app/models/job_invocation_composer.rb
```diff
@@ -282,20 +282,11 @@ def template_invocation_input_value_for(input)
   end
 
   def job_template_ids
-    job_invocation.template_invocations.map(&:template_id)
+    job_invocation.pattern_template_invocations.map(&:template_id)
   end
 
   private
 
-  def dup_template_invocations(job_invocation)
```
ares: replaced by existing deep_clone (that no longer save records), effective_user was useless anyway since dup copied that value as well

test/unit/actions/run_hosts_job_test.rb
```diff
@@ -39,8 +39,7 @@ class RunHostsJobTest < ActiveSupport::TestCase
     end
 
     it 'triggers the RunHostJob actions on the resolved hosts in run phase' do
-      template_invocation = job_invocation.template_invocation_for_host(host)
-      action.expects(:trigger).with(Actions::RemoteExecution::RunHostJob, job_invocation, host, template_invocation, proxy)
+      action.expects(:trigger).with() { |*args| args[0] == Actions::RemoteExecution::RunHostJob }
```
ares: hard-coding of API makes it fragile, also the template_invocation is now built (not saved) as part of `RunHostsJob` run phase so can't be found here

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -13,8 +13,9 @@
   <pre><%= job_invocation.targeting.search_query %></pre>
 
   <%= _('Evaluated at:') %> <%= job_invocation.targeting.resolved_at %><br>
-  <% if job_invocation.template_invocations.size > 1 %>
-    <% job_invocation.template_invocations.each do |template_invocation| %>
+  <% if job_invocation.pattern_template_invocations.size > 1 %>
+    <% job_invocation.pattern_template_invocations.each do |template_invocation| %>
+      <%= template_invocation.inspect %>
```
iNecas: Intentional?

iNecas:
went over the code and tested. Works nicely, just one question inline.

app/views/job_invocations/_tab_overview.html.erb
ares: seems I need :eyeglasses:

ares:
fixed, @iNecas would you mind give it another round?

app/models/job_template.rb
```diff
@@ -9,7 +9,9 @@ class JobTemplate < ::Template
 
   audited :allow_mass_assignment => true
   has_many :audits, :as => :auditable, :class_name => Audited.audit_class.name
-  has_many :template_invocations, :dependent => :destroy, :foreign_key => 'template_id'
+  has_many :all_template_invocations, :dependent => :destroy, :foreign_key => 'template_id', :class_name => 'TemplateInvocation'
+  has_many :template_invocations, -> { where('host_id IS NOT NULL') }, :foreign_key => 'template_id'
```
iNecas: Time to drop `1.10` or adding a temporary condition on rails version?

ares: I will add condition, 1.10 users deserves to get this fix

ares:
rebased and added rails 3 compatibility

ares:
[test] Foreman 1.10 was fixed (https://github.com/theforeman/foreman/commit/7a7e215f18876a2cd7f8db34de968842bd701d7f)

ares:
got :green_apple: do I hear ACK @iNecas?

iNecas:
ACK

ares:
Thanks, merging.

PR #118, opened 2016-01-14
fixes #13210, #13211, #13204 - json and apipie

thomasmckay:


stbenjam:
You should be able to do `fixes #13210, #13211, #13204` instead, that should make Mr. @theforeman-bot happy.

app/views/api/v2/template_inputs/main.json.rabl
```diff
@@ -2,7 +2,8 @@ object @template_input
 
 extends "api/v2/template_inputs/base"
 
-attributes :template_id, :fact_name, :variable_name, :puppet_parameter_name
+attributes :template_id, :fact_name, :variable_name, :puppet_parameter_name, :puppet_class_name
+attributes :description, :required
```
stbenjam: It can span multiple lines here with only 1 attributes call

```ruby
attributes :template_id, :fact_name, :variable_name, :puppet_parameter_name,
  :puppet_class_name, :description, :require
```

thomasmckay: TBH, my preference would be to further divide them just for readability (putting the puppet ones on a line, etc.). I'm fine with combining, though.

iNecas:
Tested and works well

ares:
Thanks @thomasmckay, merging!

PR #119, opened 2016-01-15
Fixes #13240 - ability to render one template from another

iNecas:
Usage:

In template, on can reference another template by using:

```
render_template("Run Command - SSH Default", 'command' => 'hello world')
```

It's also possible to import inputs defined on one template into another. It's
done by adding 'Foreign Input Set' to the template (in the job tab under 
inputs definition). One can either load all inputs from the target template, 
define some exceptions (values that will be overridden by the template and 
makes no sense to ask user for them) or explicitly include some.

By default, when rendering a sub-template, it looks at the foreign input sets 
and uses the values when a match is found. To explicitly provide the values
regardless of the foreign input sets one can use:

```
render_template("Run Command - SSH Default", {'command' => 'hello world'},
                                             {:with_foreign_input_set => false})
```

In the current behaviour, we check that two inputs with the same name are not
included twice in the template (we need to do it both on template preparation
validation, as well as at job invocation time to cover some edge cases).

TODO:
- [x] API/hammer cli updates
- [ ] Document in manual


app/models/foreign_input_set.rb
```diff
@@ -0,0 +1,45 @@
+class ForeignInputSet < ActiveRecord::Base
+  class CircularDependencyError < Foreman::Exception
```
ares: `::Foreman::Exception` is needed I think

app/models/foreign_input_set.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  TYPES = { :user => N_('User input'), :fact => N_('Fact value'), :variable => N_('Variable'),
```
ares: how is this different from `TemplateInput::TYPES`?

app/models/foreign_input_set.rb
```diff
@@ -0,0 +1,45 @@
...
+    inputs = inputs.reject { |input| excluded_names.include?(input.name) }
+    return inputs
+  end
+
+  def included_names
+    self.include.to_s.split(',').map(&:strip)
+  end
+
+  def excluded_names
+    self.exclude.to_s.split(',').map(&:strip)
```
ares: could you please dry this up with included_names? some private method would do...

app/models/foreign_input_set.rb
iNecas: I think it's the case more when in dynflow actions, (because there is `Actions::Foreman` module), I haven't seen issues with this here

app/models/foreign_input_set.rb
iNecas: probably just copied over one line more than I wanted

iNecas:
[test]

app/models/foreign_input_set.rb
ares: :white_check_mark:

app/models/input_template_renderer.rb
```diff
@@ -4,19 +4,26 @@ class UndefinedInput < ::Foreman::Exception
...
   include UnattendedHelper
 
-  attr_accessor :template, :host, :invocation, :error_message
+  attr_accessor :template, :host, :invocation, :input_values, :error_message
 
   # takes template object that should be rendered
   # host and template invocation arguments are optional
   # so we can render values based on parameters, facts or user inputs
-  def initialize(template, host = nil, invocation = nil)
+  # rubocop:disable Metrics/ParameterLists
```
ares: :cop: :gun: :question:

iNecas: What do you suggest? Replacing the args with options hash?

app/models/input_template_renderer.rb
```diff
@@ -40,7 +49,49 @@ def input(name)
     end
   end
 
+  def render_template(template_name, input_values = {}, options = {})
+    options.assert_valid_keys(:with_foreign_input_set)
```
ares: I hope someone won't need to extend this from plugin

iNecas: It's not a Rails, but still would be doable, just would need to not pass the additional options to overridden method.

app/models/input_template_renderer.rb
```diff
@@ -40,7 +49,49 @@ def input(name)
...
+      raise error_message
+    else
+      out
+    end
+  end
+
+  def foreign_input_set_values(target_template, overrides = {})
+    input_set = @template.foreign_input_sets.where(:target_template_id => target_template).first
+    return {} if input_set.nil?
+    if input_set.include_all?
```
ares: could we use `ForeignInputSet#inputs`? If the issue is that it would get inputs recursively then maybe it could take `recursive` bool argument?

app/models/input_template_renderer.rb
```diff
@@ -40,7 +49,49 @@ def input(name)
...
+      inputs_to_generate = input_set.target_template.template_inputs.pluck(:name)
+    else
+      inputs_to_generate = input_set.include.to_s.split(',').map(&:strip)
+    end
+    inputs_to_generate -= overrides.keys + input_set.exclude.to_s.split(',').map(&:strip)
+
+    included_renderer = self.class.new(input_set.target_template, host, invocation, nil, @preview, @templates_stack)
+    input_values = inputs_to_generate.inject({}) do |hash, input_name|
```
ares: this might be worth of extracting to a input_values method of the renderer itself

app/views/job_templates/_custom_tabs.html.erb
```diff
@@ -23,6 +23,13 @@
       </div>
     <% end %>
     <%= add_child_link '+ ' + _("Add Input"), :template_inputs, { :title => _('add a input for this template')} %>
+    <%= new_child_fields_template(f, :foreign_input_sets, { :partial => "template_inputs/foreign_input_set_form" }) %>
+    <%= f.fields_for :foreign_input_sets do |ff| %>
+      <div class="template_input_form fields">
+        <%= render 'template_inputs/foreign_input_set_form', :f => ff %>
+      </div>
+    <% end %>
+    <%= add_child_link '+ ' + _("Add Foreign Input Set"), :foreign_input_sets, { :title => _('add a input set for this template (to reference a different template)')} %>
```
ares: `add a input set for this template (to reference a different template`
maybe
`add an input set for this template to reference a different template inputs`

app/views/template_inputs/_foreign_input_set_form.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="fields">
+  <div class="form-group">
+    <%= field_set_tag _("Foreign input set").html_safe +
+                          ' ' +
+                          remove_child_link('x', f, {:rel => 'twipsy', "data-title" => _('remove template input set'), :'data-placement' => 'left', :class => 'fr badge badge-danger'}).html_safe do %>
+      <%= selectable_f f, :target_template_id, JobTemplate.all.map { |t| [ t.name, t.id ] }, {}, :class => 'input_type_selector without_select2' %>
+      <%= checkbox_f f, :include_all %>
+      <%= text_f f, :include %>
```
ares: inline help with "comma separated values" would be nice I think, same for exclude

test/unit/input_template_renderer_test.rb
```diff
@@ -95,6 +95,147 @@
...
+        User.current = users :admin
+        command_template.save!
+        package_template.save!
+        job_invocation.template_invocations << template_invocation
+      end
+
+      describe 'foreign input set' do
+        describe 'with include_all' do
+          let(:template) do
+            FactoryGirl.create(:job_template, :template => '<%= render_template("package action", "action" => "install) %>').tap do |template|
```
ares: would fail on missing `"` after `install` but it seems you only test inputs that were found :-)

test/unit/input_template_renderer_test.rb
```diff
@@ -95,6 +95,147 @@
...
+
+        describe 'with include_all and some excludes' do
+          let(:template) do
+            FactoryGirl.create(:job_template, :template => '<%= render_template("package action", "action" => "install") %>').tap do |template|
+              template.foreign_input_sets << FactoryGirl.build(:foreign_input_set, :target_template => package_template, :include_all => true, :exclude => 'action,debug')
+            end
+          end
+
+          it 'includes all inputs from the imported template except the listed once' do
+            template.template_inputs_with_foreign.map(&:name).sort.must_equal ['package']
```
ares: could you please add the same test but with both include and exclude set to empty string (UI vs API)

ares:
Would you mind adding a JS that would disable `include` field if `include_all` is checked?

app/models/input_template_renderer.rb
ares: to shoot the this cop entirely

app/models/input_template_renderer.rb
ares: :white_check_mark:

ares:
+ needs rebase

iNecas:
@ares thanks for the comments, I've updated the PR and all the issues should be addressed (except the missing API and manual update)

ares:
sorry, need another rebase

ares:
visually looks OK so you can squash at the same time, I'll do the testing then

iNecas:
Rebased + added api

iNecas:
Hammer CLI command opened https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/7

stbenjam:
Master is failing again with transient failures :sob:

stbenjam:
Oh did one of you already do it? It shows 147 pending: http://ci.theforeman.org/view/Plugins/job/test_plugin_foreman_remote_execution_master/

ares:
[test] failed only one postgres 1.9.3, interesting error though -
```undefined method `minutes' for nil:NilClass (NoMethodError)
/var/lib/workspace/workspace/test_plugin_pull_request/database/postgresql/label/fast/ruby/1.9.3/foreman/app/controllers/concerns/foreman/controller/session.rb:25:in `update_activity_time'
```

stbenjam:
That's one of the transient errors I've seen.  Jenkins was quite overloaded yesterday, if every PR needs testing 2-3 times. Maybe we just let the merge happen if it's clearly such a case.

iNecas:
@stbenjam if it's clearly an intermittent issue, I'm ok with not rerunning the tests again

test/unit/input_template_renderer_test.rb
ares: :white_check_mark:

ares:
@iNecas it needs one more rebase, ideally if you could add foreign set locking that was introduced by #120, I resolved the conflict manually and I'll start testing current version meanwhile

iNecas:
Rebased and updated to include the checks on `locked?` should be ready for final review

ares:
Awesome, works perfectly. ACK when we have confirmation from Mr. Jenkins

iNecas:
[test]

iNecas:
Intermittent issue, ignoring and merging

PR #120, opened 2016-01-19
fixes #12497 - ensure template inputs uneditable and forms disabled when locked

stbenjam:


stbenjam:
[test]

ares:
the same will be required for foreign input sets from #119

app/helpers/remote_execution_helper.rb
```diff
@@ -205,4 +205,11 @@ def documentation_button_rex(section = '')
       url,
       :rel => 'external', :class => 'btn btn-info', :target => '_blank')
   end
+
+  def template_input_header(template)
+    header = _('Template input')
+    header += ' ' + remove_child_link('x', f, {:rel => 'twipsy', 'data-title' => _('remove template input'), :'data-placement' => 'left',
```
ares: just wondering why mixing 'data-title' and :'data-placement' (string and symbol) types as keys

stbenjam: I moved this from the view to the helper since it needed some logic, it came with mixed keys. I can fix it - which do you prefer as keys?

ares: I don't care :-) symbols would be slightly better (even the horrible :'data-something' syntax)

stbenjam:
[test] *sigh*

ares:
works nicely, thanks @stbenjam, merging!

PR #124, opened 2016-01-21
fixes #13308 - call correct method on job invocation

stbenjam:
There's a debate about calling these method names correctly, and splitting deep_clone (no save) with deep_clone! which saves, but @ares will file a separate issue to make this consistent.


ares:
APJ, I filed http://theforeman.org/issues/13310

ares:
[test]

stbenjam:
Integration test failure, thought we were going to ignore these?

ares:
I wanted to give it a try when there's smaller load on CI. Will do the merging in few moments.

ares:
Merging, thanks @stbenjam!

PR #126, opened 2016-01-21
fixes #13317, #12538 - fix rubocop and mass assignment for foreman 1.11

stbenjam:
Based on the failures of http://ci.theforeman.org/job/test_plugin_pull_request/2133/, which tested the rubocop PR

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,13 @@ inherit_from:
 
 AllCops:
   RunRailsCops: true # always run the rails cops
+  Exclude:
+    - 'doc/plugins/**'
+
+# TODO: uncomment when on rubocop 0.35.1
+#Rails/Date:
+  #Exclude:
+    #- foreman_remote_execution.gemspec
 
 # Don't enforce documentation
 Style/Documentation:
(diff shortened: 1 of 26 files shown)


iNecas:
ACK, I'm just wondering why it was not failing on sqlite or mysql…

stbenjam:
@iNecas Integration tests only run on postgresql

iNecas:
@stbenjam that explains a lot :)

iNecas:
Unfotrunately, it seems there is some issue with whitelisting the attribute this way (seems the original whitelisting is lost or something:/

stbenjam:
Huh, in other cases attr_accessible adds to the existing set, other plugins do it too - I wonder what's gone wrong here :-\

stbenjam:
It will pass against 1.11

```
09:07 | Dominic > you're running that on 1.10-stable which doesn't have any attr_accessible calls, so you're enabling protected attributes on the whole model 
```

iNecas:
ah, ok

stbenjam:
Against develop: http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/2151/

iNecas:
nice and green :)

stbenjam:
Is that an ACK? :)

stbenjam:
The job is also switched to develop a few minutes ago, as well

iNecas:
ACKy ACK

PR #72, opened 2015-11-05
[Foreman 1.11] fixes #12312 - bulk actions opens job invocation in new page directly

stbenjam:
Requires https://github.com/theforeman/foreman/pull/2892

PR #129, opened 2016-01-22
fixes #13331 - drop 1.10 support and bump version to 0.3.0

stbenjam:


stbenjam:
The right commit message would help, I really shouldn't release things on a Friday :grin:

I think this is all we need to do to drop 1.10, right?

iNecas:
Yes, while dropping the version, we can remove some conditions, see `grep -r Rails::VERSION .` for the parts that can be updated

dLobatog:
Two questions arise:

1) Why? Are there any non-backwards compatible changes other than the patternfly icons thing?
2) I don't know what kind of versioning schema does this project follow. Minor versions are supposed to be fixes & small features compatible with older versions, so this ought to be 1.0 instead? That's assuming this follows http://semver.org/ as other plugins/core/proxy/etc

stbenjam:
> 1) Why? Are there any non-backwards compatible changes other than the patternfly icons thing?

https://github.com/theforeman/foreman_remote_execution/pull/72 which is a nice usability improvement.   Do note we have 0.2.0 which is compatible with 1.10, which was just released with all the new features that were demo'd.

iNecas:
@dLobatog yes, we try to follow the semver, but note http://semver.org/#spec-item-4 - there is a special case for 0-starting releases. We might switch to first API-stable release, but I would still give it some time until we gain more experience with the API and how it fits users.

stbenjam:
Updated to remove the Rails 3 bits

ares:
also please drop https://github.com/theforeman/foreman_remote_execution/blob/master/app/assets/stylesheets/job_invocations.css.scss#L24 since https://github.com/theforeman/foreman/blob/develop/app/assets/stylesheets/status-colors.scss

otherwise looks good to me and works fine

stbenjam:
Done

ares:
Thanks @stbenjam, merging!

PR #130, opened 2016-01-22

dLobatog:
It's a bit difficult to notice some sections in the job invocations
page. Similarly, the piechart is strangely aligned to the right with a
big space in between.
To try to overcome these issues and homogenize the interface, we can use
Patternfly cards
https://rawgit.com/patternfly/patternfly/master/tests/cards.html

This is how it used to look like: 
After this change, it looks like: screenshot from 2018-02-14 13-25-08
screenshot from 2018-02-14 13-25-19


stbenjam:
BTW, I'd delete your job templates and reseed them so you get the new description templates for the templates we ship.  Hopefully eventually we'll get the community templates plugin to work for jobs so you can keep them in sync more easily.

dLobatog:
@stbenjam I don't follow what does that have to do with this PR, you want me to show the template description somewhere?

ohadlevy:
@dLobatog why not include the chart in card? IMHO it does provide value, even if the data repeat itself.
further, I would convert more to cards, the fact that half the screen is cards vs not imho feels incomplete.

stbenjam:
>  @stbenjam I don't follow what does that have to do with this PR, you want me to show the template description somewhere?

Nothing, except you keep showing screenshots with these really long default descriptions.  The two you show here I guess are your own templates, but I thought I saw ones using our templates that hadn't been updated to include the description template.

stbenjam:
As far as the actual design, I'l say what I said on Friday here, I think the tabs should go away, people really do want to see which hosts succeeded or failed very quickly.  I'd prefer if this looked like a normal index page (e.g. All Hosts) but just had some cards at the top as a kind of header, and removing the template preview entirely.

If people want to see the rendered template, they could be able to do so for a specific host using action buttons.

ohadlevy:
:+1:

dLobatog:
@ohadlevy , I had a short UI design meeting with @stbenjam @ares @iNecas and @kybaker and we agreed in a few points to improve this PR:

* Show sub-tasks in the host actions table 
* Move template previews to a separate tab
* Use piechart to display the list of results
* Use the following layout
screenshot from 2016-01-25 15-48-10

Also we want to display live output right here, when you click on a host, it should open a modal with the live output. That's an important and frequent action when the job fails. But that'd be a separate PR :smile:

angystardust:
``echo trololo > /etc/motd`` :smile:
You simply rock, guys!

ares:
hey, someone read our manual ;-)

dLobatog:
I did all of the things you can see on my previous comment, now it looks like this:

screenshot from 2016-01-26 20-43-02

I don't think the chart looks quite right, but it's because it's obviously not matching with Patternfly. It should use https://www.patternfly.org/patterns/donut-chart/ instead. Then it'd be perfect :sparkles: 

IMO it's already an improvement over what we have, I'd argue the whole charts conversion to Patternfly donuts :doughnut: should happen in core, it should provide helpers for this kind of stuff.

I also tried two alternative layouts which fill the blank space, perhaps you prefer one of these:

screenshot from 2016-01-26 20-48-36
screenshot from 2016-01-26 20-49-06

Let me know what you think!

kybaker:
@dLobatog - this layout has a much nicer information hierarchy. Much more readable. I like the second image the best purely from a readability standpoint but either the second or third image will work.

ohadlevy:
:+1: for the second image as well, i think that if the target host card had
a blue top it will align nicely as well.
can you adjust the height of both cards to fit? (they are nearly the same).

stbenjam:
:+1: to either the second or third images.  Awesome.

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -1,47 +1,127 @@
...
+        </h2>
+      </div>
+      <div class="card-pf-body">
+        <p>
+        <% if job_invocation.bookmark.present? %>
+          <%= _('Bookmark') %> <%= job_invocation.bookmark.name %><br>
+        <% else %>
+          <%= _('Manual selection') %>
+        <% end %>
+        <%= 'using ' %>
```
ares: I know it's not your code, but could you please add `_()`? also could you fix the indentation of this `<p>` (strong and pre tags)?

iNecas:
:+1: for second one

ares:
I also vote for 2nd now, just note that with further changes we might have to redesign the top part again since the donut will be smaller and the minicards are too wide already. But it's much better that what we have now so +1.

During testing I found that after clicking search button, no tab is active so I got the blank page (should be small fix). Also I'd revert the previewing for each host to just first, since for 1000 hosts it would generate too big page. I'd like to look if we can update select2 so it can use ajax to load data (IIRC it required rails 4+ which is no longer a blocker). Then I'd add previewing per host using it.

adamruzicka:
The second one is really nice. I wonder if the block with the cards have the same width as the table of hosts and the divider line below tabs?

dLobatog:
Updated, now 'search' works fine, and I added the missing translation to 'using'. It's using the layout on the last picture of my  previous comment

stbenjam:
Looks good, is there a way to get the two columns in the header to be the same height?

Also, when starting a new invocation I get an error in the view, although the invocation runs successfully.

invocation-error

stbenjam:
And the pie chart doesn't seem to replace the old one when updating live, it just adds another one:

chart

app/helpers/job_invocations_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+    template_name = content_tag(:strong, template.name)
+    provider = template.provider.humanized_name
+    link_content = template_name + ' - ' + provider + ' ' +
+      icon_text('edit', '', :kind => 'pficon')
+    link_to_if_authorized(link_content,
+                          hash_for_edit_job_template_path(:id => template.id))
+  end
+
+  def preview_hosts(template_invocation)
+    hosts = template_invocation.targeting.hosts
```
ares: this is iterating over all hosts of invocation, it can be thousands of hosts, either add some limit here or revert it to display just one as a temporary solution until we have the selection

iNecas: +1 for just one host for now, as it was before (but in separate tab as it's done here)

app/views/job_invocations/show.html.erb
```diff
@@ -7,18 +7,25 @@
...
-  <div class="tab-pane <%= job_invocation_active_tab(:overview, params) %>" id="primary">
-    <%= render 'tab_overview', :job_invocation => @job_invocation %>
+  <div class="tab-pane active" id="primary">
+    <%= render 'tab_overview', :job_invocation => @job_invocation, :hosts => @hosts %>
   </div>
 
-  <div class="tab-pane <%= job_invocation_active_tab(:hosts, params) %>" id="hosts" data-refresh_required="<%= @job_invocation.resolved? ? '' : 'true' %>">
-    <%= render 'tab_hosts', :job_invocation => @job_invocation, :hosts => @hosts %>
+  <div class="tab-pane" id="preview_templates"
+    data-refresh_required="<%= @job_invocation.resolved? ? '' : 'true' %>">
```
dLobatog: Self note, this is what's blocking the auto-refresh

dLobatog:
Should be ready now, with a Patternfly donut. I limited the preview templates hosts to 20.

Please test with https://github.com/theforeman/foreman/pull/3128 :smile: 

screenshot from 2016-01-28 12-41-42

ohadlevy:
looks good!
some comments on the card content:
target hosts card - imho you should use a footer (for the evaluated at) or
just change it to a tool tip.
maybe even the header should not be target hosts, rather just Hosts or Host
Selection?

I would even consider merging the host counter card with it.

further, when running, does it make sense to show a progress bar instead of
a pie? i mean I'm both interested in how many hosts are running and in
their status, and we can aggregate the number of pending, running, error
etc to a total that is meaningful in addition to a percentage ?

dLobatog:
@ohadlevy I added the footer for 'Evaluated at'. I haven't changed the title to 'Host selection' or anything yet, I'd want to hear others' opinions too. 
About the progress bar, I don't mind - I thought about having 4 horizontal bars, for 'success', 'failed' etc.. but ended up doing this as it looked similar to how it was before but better. Not sure what you mean about aggregate these figures, isn't that what we have below the donut chart already?

stbenjam:
I get `undefined method `donut_chart' for #<#<Class:0x00000008a3e410>:0x0000000934b3c8>` on the most recent commit

ohadlevy:
@dLobatog I meaned stacked - see http://getbootstrap.com/components/#progress-stacked

dLobatog:
@stbenjam Are you testing this with https://github.com/theforeman/foreman/pull/3128 ? I moved the donut_chart method there so that other plugins could use it.

stbenjam:
Ah sorry, I missed that comment above.

It looks great! :+1: 

Now it doesn't align with the cards anymore but I'm not sure it's a big deal.  Anything we can do to make the two columns more balanced? 

Also it shoes NaN% for a short period of time at the beginning, anyway to avoid that? 

GIF:
recorded

ares:
This is waiting for core PR and last comment does not seem to be addressed. Moving to WoC

ares:
the core chart PR was merged in core, @dLobatog would you mind to rebase?

dLobatog:
screenshot from 2017-11-23 11-51-42
screenshot from 2017-11-23 11-51-53

This is how it looks already - no changes needed in Foreman core at all.

adamruzicka:
Thank you for picking this up again. I like it, however the scheduling info ("scheduled to start at" and friends) probably should get its own card

ares:
That's pretty huge improvement already. I agree with @adamruzicka , if we could make scheduling info a card next to hosts count, it might be slightly better

Rohoover:
I agree with @ares @adamruzicka 

Seems like the information is important enough to warrant it's own card.

webpack/index.js
```diff
@@ -0,0 +1,17 @@
...
+import componentRegistry from 'foremanReact/components/componentRegistry';
+import JobInvocationContainer from './react_app/components/job_invocations';
+import './react_app/redux/reducers'
+
+componentRegistry.register(
+  { name: 'JobInvocationContainer', type: JobInvocationContainer}
+);
+
+if(window.location.href.match(/job_invocations/)) {
+  const jobInvocationId = parseInt(window.location.href.split('/').slice(-1)[0])
```
ohadlevy: might consider using URLSearchParams or urijs instead

webpack/react_app/components/job_invocations/AggregateStatus/index.js
```diff
@@ -0,0 +1,42 @@
...
+  render() {
+    const {
+      statuses,
+    } = this.props;
+
+    return (
+      <div id="aggregate_statuses">
+        <p className="card-pf-aggregate-status-notifications">
+          <span className="card-pf-aggregate-status-notification">
+            <a id='success_count' href="#">
```
ohadlevy: is href required?

webpack/react_app/redux/actions/job_invocations/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+import {
+  JOB_INVOCATIONS_GET_JOB_INVOCATIONS,
+  JOB_INVOCATIONS_POLLING_STARTED,
+} from '../../consts';
+
+import API from 'foremanReact/API';
+
+const defaultJobInvocationsPollingInterval = 10000;
+const jobInvocationsInterval = isNil(process.env.JOB_INVOCATIONS_POLLING) ?
```
ohadlevy: this seems to be repeating it self as a pattern, maybe it makes more sense to extract it to a common functions in core and reuse here and in notifications?

app/views/job_invocations/_card_target_hosts.html.erb
```diff
@@ -0,0 +1,33 @@
+<% template_invocations = @job_invocation.pattern_template_invocations %>
+<div class="card-pf card-pf-accented">
```
ares: We'll need cards on other places soon too (auth sources page) and we also have some cards on smart proxy page. I wonder if it would be a good time to add some helpers to core. patternfly-react does not have cards component yet, Dominik might look into contributing this there but it will take some time. But if you think it's too much work, I'm fine with having card tags here :-)

ohadlevy: I would suggest trying to contribute cards regardless, it should not be a huge effort imho just to add the html structure to patternfly-react.

adamruzicka:
Maybe the hosts table should get its own tab or something. Either that or this https://imgur.com/a/27OGI probably needs to be fixed somehow.

webpack/react_app/redux/actions/job_invocations/index.js
dLobatog: Indeed - that's the kind of abstraction ActionCable provides for ws except we can't use it (yet).

app/helpers/job_invocations_helper.rb
```diff
@@ -0,0 +1,80 @@
+module JobInvocationsHelper
+   def minicard(icon, number, text)
+    content_tag(:div, :class => 'card-pf card-pf-accented
+                card-pf-aggregate-status card-pf-aggregate-status-mini') do
+      content_tag(:h2, :class => 'card-pf-title', :style => 'line-height: 1.1') do
+        icon_text(icon, '', :kind => 'pficon') +
+          content_tag(:a) do
```
ares: this makes the title link that leads nowhere, it's decorated on hover but clicking has no effect, could you please remove it? I tested it and works just fine

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -1,48 +1,22 @@
...
+  </div>
 
-  <%= _('Evaluated at:') %> <%= job_invocation.targeting.resolved_at %><br>
-  <% if job_invocation.pattern_template_invocations.size > 1 %>
-    <% job_invocation.pattern_template_invocations.each do |template_invocation| %>
-      <%= host_counter template_invocation.template.provider.humanized_name, ForemanTasks::Task::DynflowTask.for_action(Actions::RemoteExecution::RunHostJob).for_resource(template_invocation).uniq.size %>
+  <div class='col-xs-12 col-sm-6 col-md-6'>
+  <%= render :partial => 'card_target_hosts', :job_invocation => job_invocation %>
+  <% template_invocations.each do |template_invocation| %>
+    <% if template_invocation.effective_user %>
```
ares: this and total host cards seems to take too much space, they could live next to each other

dLobatog: I agree, however it's easier to manage when there's no effective user set. If I put them next to each other, when there's no effective user set it looks weird to have a card occupy half the size of the one above

ares:
@dLobatog would you mind bumping foreman requirement at https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L34 and version at https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/version.rb to 1.5.0 so we don't forget to do it after this is merged? The code changed a lot so I guess the review might take a bit longer.

dLobatog:
@ares I don't think the code changed that much - I'll comment the changes I did, sure I'll update the min Foreman version

webpack/react_app/redux/actions/job_invocations/index.js
```diff
@@ -0,0 +1,60 @@
...
+  process.env.JOB_INVOCATIONS_POLLING;
+
+const getJobInvocations = url => (dispatch) => {
+  const isDocumentVisible =
+    document.visibilityState === 'visible' ||
+    document.visibilityState === 'prerender';
+
+  if (isDocumentVisible) {
+    API.get(url)
+      .then(onGetJobInvocationsSuccess)
```
dLobatog: This is the key part that changed because of https://github.com/theforeman/foreman/pull/5080

app/helpers/job_invocation_output_helper.rb
```diff
@@ -29,7 +28,7 @@ def colorize_line(line)
     parts.each do |console_line|
       if console_line.include?('{{{format')
         if (color_index = console_line[/color:(\d+)/, 1]).present?
-          current_color = CONSOLE_COLOR[color_index]
+          current_color = CONSOLE_COLOR[color_index] || 'default'
```
dLobatog: Setting a default value on a frozen hash isn't possible - so this is the best next option.

app/helpers/remote_execution_helper.rb
```diff
@@ -99,7 +107,6 @@ def job_invocation_task_buttons(task)
     job_invocation = task.task_groups.find { |group| group.class == JobInvocationTaskGroup }.job_invocation
     task_authorizer = Authorizer.new(User.current, :collection => [task])
     buttons = []
-    buttons << link_to(_('Refresh'), {}, :class => 'btn btn-default', :title => _('Refresh this page'))
```
dLobatog: I removed this as I did not see a reason for it to exist anymore - page elements are updated.

app/helpers/remote_execution_helper.rb
```diff
@@ -200,17 +208,6 @@ def preview_box(template_invocation, target)
     end
   end
 
-  def job_invocation_active_tab(tab, params)
```
dLobatog: Irrelevant anymore as the `hosts` tab doesn't exist

app/views/job_invocations/_tab_hosts.html.erb
```diff
@@ -38,7 +38,7 @@
     </tbody>
   </table>
 
-  <%= will_paginate_with_info @hosts %>
+  <%= will_paginate_with_info @hosts, :'data-no-turbolink' => 'true', :container => true %>
```
dLobatog: This is what allows the React component to mount when you click on next/prev page on the hosts table

dLobatog:
http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4571 tests running here

lib/foreman_remote_execution/engine.rb
```diff
@@ -54,7 +54,7 @@ class Engine < ::Rails::Engine
           permission :lock_job_templates, { :job_templates => [:lock, :unlock] }, :resource_type => 'JobTemplate'
           permission :create_job_invocations, { :job_invocations => [:new, :create, :refresh, :rerun, :preview_hosts],
                                                 'api/v2/job_invocations' => [:create] }, :resource_type => 'JobInvocation'
-          permission :view_job_invocations, { :job_invocations => [:index, :show, :auto_complete_search], :template_invocations => [:show],
+          permission :view_job_invocations, { :job_invocations => [:index, :chart, :show, :auto_complete_search], :template_invocations => [:show],
```
dLobatog: This little thing was missing!

dLobatog:
http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4574 tests running here (with charts permission)

dLobatog:
[Core tests are failing with](http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4574/#showFailuresLink), looks related to https://github.com/theforeman/foreman_remote_execution/commit/5e59c8f93f5c70cbafc208a365a17d62d10bf4a8  @ares any ideas?

`The asset "execution_interface.js" is not present in the asset pipeline.Falling back to an asset that may be in the public folder`

mmoll:
I guess https://github.com/theforeman/foreman_remote_execution/blob/1d23156f5646f412aeb424e1fac6987d4b1e2998/lib/foreman_remote_execution/engine.rb#L123 needs to get adjusted.

ares:
this has been resolved by #312 thanks to @mmoll, rebasing should help

dLobatog:
Awesome, rebased @ares

dLobatog:
Tests running at http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4577

dLobatog:
@ares The failed tests fail in the same way REX master is failing currently, still not sure yet. In any case as you can see postgres/sqlite passed

webpack/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
```
ohadlevy: I don't think you need ReactDom?

webpack/index.js
```diff
@@ -0,0 +1,18 @@
...
+import URI from 'urijs';
+import { mount } from 'foremanReact/common/MountingService';
+import componentRegistry from 'foremanReact/components/componentRegistry';
+import JobInvocationContainer from './react_app/components/job_invocations';
+import './react_app/redux/reducers'
```
ohadlevy: that shouldn't be required here?

webpack/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+componentRegistry.register(
+  { name: 'JobInvocationContainer', type: JobInvocationContainer}
+);
+
+if(window.location.href.match(/job_invocations/)) {
```
ohadlevy: why do this way? maybe you should use react-router like katello does? I'm worried this always require a full page reload for this code to be executed.

webpack/react_app/components/job_invocations/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+  render() {
+    const {
+      job_invocations,
+      statuses,
+    } = this.props;
+
+    return (
+      <div id="job_invocations_chart_container">
+        <PieChart
```
ohadlevy: you should consider using patterfly-react chart components directly? (see https://github.com/theforeman/foreman/pull/5137 for background)

webpack/react_app/components/job_invocations/index.js
```diff
@@ -0,0 +1,47 @@
+import { groupBy, isUndefined } from 'lodash';
```
ohadlevy: please setup linting on your project, these are unused.

webpack/react_app/redux/actions/job_invocations/index.js
```diff
@@ -0,0 +1,60 @@
+import { isNil } from 'lodash';
```
ohadlevy: this is no longer used even in core, see https://github.com/theforeman/foreman/commit/dfbb65cf13f06408770205dd91acbe274d8f68e4#diff-2abb6d1defb26fc866851a507eef315c

webpack/react_app/redux/actions/job_invocations/index.js
ohadlevy: IMHO not just AC/WS - but code wise this is falling into the category of DRY

webpack/react_app/components/job_invocations/index.js
dLobatog: Makes sense, but after that PR is merged we can update REX here. My main concern is that this PR is already only 1.17 compatible because of Axios, and using Patternfly donut chart from that PR would make it 1.18

webpack/react_app/redux/actions/job_invocations/index.js
sharvit: We are trying to avoid using `lodash` (it's heavy)

In your case, i think you can just do:
```jsx
const jobInvocationsInterval = process.env.JOB_INVOCATIONS_POLLING || defaultJobInvocationsPollingInterval;  
```

webpack/react_app/components/job_invocations/index.js
```diff
@@ -0,0 +1,47 @@
...
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = state => {
+  const {
+    job_invocations,
+    statuses,
+  } = state.foreman_remote_execution_reducers.job_invocations;
```
sharvit: While it is 100% fine (especially because it's a plugin), consider using camelCase only when writing javaScript.

webpack/react_app/redux/actions/job_invocations/index.js
```diff
@@ -0,0 +1,61 @@
...
+      .then(onGetJobInvocationsSuccess)
+      .catch(onGetJobInvocationsFailed)
+      .then(triggerPolling);
+  } else {
+    // document is not visible, keep polling without api call
+    triggerPolling();
+  }
+};
+
+export const startJobInvocationsPolling = url => (dispatch, getState) => {
```
dLobatog: `eslint` wants this to be 'export default` but I couldn't make it work...

dLobatog:
[test]

dLobatog:
[test default]

lib/foreman_remote_execution/version.rb
```diff
@@ -1,3 +1,3 @@
 module ForemanRemoteExecution
-  VERSION = '1.4.5'.freeze
+  VERSION = '1.5.0'.freeze
```
adamruzicka: We do version bumps in separate commits, so you don't have to worry about it here.

dLobatog: I was asked in https://github.com/theforeman/foreman_remote_execution/pull/130#issuecomment-362210308 😕

webpack/react_app/redux/actions/job_invocations/index.js
sharvit:
```jsx
const startJobInvocationsPolling = url => (dispatch, getState) => { ... };
export default startJobInvocationsPolling;
```

or
```jsx
export default (url) => (dispatch, getState) => { ... };
```

sharvit:
@dLobatog, you should be able to mock the `foremanReact` module.
https://facebook.github.io/jest/docs/en/es6-class-mocks.html

dLobatog:
@sharvit I realize I can do that, my issue was actually testing components that render charts (which come from foreman) - it's a little clunky to copy the HTML into fixtures and test that, it actually can change in foreman but it would pass tests here

sharvit:
@dLobatog - you can mock the `Chart` component and do **snapshot testing** with a **shallow renderer**.

Instead of test that the `Chart` component getting **rendered properly**,
you will test the the the `Chart` component getting **called properly**.

An example in foreman for shallow snapshot:
https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/Alert/Alert.test.js

dLobatog:
@sharvit I see, thank you. Any changes in the code you'd like to see?

iNecas:
@dLobatog hopefully, I've enabled the hound hooks now, you would probably need to re-open a PR for it to start with the comments

.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": [
+    "env",
+    "react"
+  ],
+  "plugins": [
+    "transform-class-properties"
+  ]
+}
```
sharvit: I think it makes sense to use the same `.babelrc` configuration from the foreman.
At production build time, it will use them anyway.

.eslintrc
```diff
@@ -0,0 +1,30 @@
+{
```
sharvit: Same here, why not using the same configurations as in foreman?

webpack/index.js
```diff
@@ -0,0 +1,22 @@
+import URI from 'urijs';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { mount } from 'foremanReact/common/MountingService';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import componentRegistry from 'foremanReact/components/componentRegistry';
+import JobInvocationContainer from './react_app/components/job_invocations';
+import './react_app/redux/reducers';
```
sharvit: Prefer to import variables and use them instead importing code that supposes to run and work.
Can migrate to:

`webpack/react_app/redux/reducers/index.js`
```jsx
export default combineReducers({
  jobInvocations,
});
```

`webpack/index.js`
```jsx
import { registerReducer } from 'foremanReact/common/MountingService';
import rootReducer from './react_app/redux/reducers';

registerReducer('foreman_remote_execution_reducers', rootReducer);
```

webpack/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+componentRegistry.register({
+  name: 'JobInvocationContainer',
+  type: JobInvocationContainer,
+});
+
+if (window.location.href.match(/job_invocations/)) {
```
sharvit: What if turbolinks will navigate to the URL? will it still work?

In Foreman, we usually mount the components from inside the view file:
https://github.com/theforeman/foreman/blob/45d1359a8bb5d7ab75f1c0b83d28e8fff8eb8193/app/views/dashboard/_status_chart_widget.html.erb#L4-L6

I guess it will even be easier to get the `jobInvocationId` because you can get it from rails.

dLobatog: Unfortunately I don't think that's possible to do in plugins. At the point when Rails loads the page, plugin components are not available yet, so we have to 'hack it'.

```
JobInvocationContainer among BookmarkContainer, PieChart, StatisticsChartsList, PowerStatus, NotificationContainer, ToastNotifications, StorageContainer, PasswordStrength
componentRegistry.js:45:12
 action TOASTS_CLEAR @ 11:19:40.599
```

Katello also needs to do this for the experimental UI component - https://github.com/Katello/katello/blob/master/webpack/index.js#L18

cc @waldenraines @danseethaler

dLobatog: Turbolinks works with this

dLobatog:
@iNecas It appears that linting works, although not with this repo rules https://github.com/theforeman/foreman_remote_execution/pull/315#discussion_r167825431 is picked up, but there is another error that is not. Also please if you can enable Travis, it would run tests on JS.

adamruzicka:
@dLobatog Travis should be enabled now

dLobatog:
@adamruzicka Any idea why the testing gets aborted sometimes? It seems that mysql/sqlite went OK

adamruzicka:
@dLobatog Jenkins is configured to kill jobs if they don't finish within two hours. Maybe the slaves were under heavy load so they didn't finish in time.

> Build timed out (after 120 minutes). Marking the build as aborted.
Build was aborted

dLobatog:
[test]
[test default]

dLobatog:
screenshot from 2018-02-14 13-25-08
screenshot from 2018-02-14 13-25-19

Some more recent screenshots

app/views/job_invocations/_user_input.html.erb
```diff
@@ -0,0 +1,21 @@
+<div id='preview_<%= target.id %>' class='collapse out'>
+  <%= preview_box(template_invocation, target) %>
+  <% if template_invocation.input_values.present? %>
+    <table class='datatable table table-striped table-bordered dataTable no-footer'>
```
ares: we just finished converting to table_css_classes helper provided by core (see https://github.com/theforeman/foreman_remote_execution/pull/313/files ) could you please update it here too?

lib/foreman_remote_execution/version.rb
ares: @adamruzicka I know this is how it was done in past but I also remember people forgot to bump it later, so I'd prefer if we start bumping the version right away, unless there's some concern

ares:
oh and also I'm now getting this when I install npm modules

```
npm WARN foreman_remote_execution@1.0.0 No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
```

lib/foreman_remote_execution/version.rb
adamruzicka: I don't feel all that strongly about this, as long as we don't double bump the version

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -1,48 +1,22 @@
...
+  </div>
 
-  <%= _('Evaluated at:') %> <%= job_invocation.targeting.resolved_at %><br>
-  <% if job_invocation.pattern_template_invocations.size > 1 %>
-    <% job_invocation.pattern_template_invocations.each do |template_invocation| %>
-      <%= host_counter template_invocation.template.provider.humanized_name, ForemanTasks::Task::DynflowTask.for_action(Actions::RemoteExecution::RunHostJob).for_resource(template_invocation).uniq.size %>
+  <div class='col-xs-12 col-sm-6 col-md-6'>
+  <%= render :partial => 'card_target_hosts', :locals => { :job_invocation => job_invocation } %>
+  <% template_invocations.each do |template_invocation| %>
+    <% if template_invocation.effective_user %>
```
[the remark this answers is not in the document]
dLobatog: It looks like the effective user is always displayed, however I can't show the 'total hosts' next to that. "Total hosts" displays the total number of hosts in the job invocation, but "effective user" is showed on a per-template invocation basis. I can put the "Total hosts" on top, and then the list of template invocation cards (effective user/user inputs). Usually there's only one.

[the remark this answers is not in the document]
ares: Are you sure? The template invocation IIRC is only one, since this is in fact pattern template invocation. I tried running on multiple hosts and I always get just one card, which is IMHO correct.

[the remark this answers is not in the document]
dLobatog: I didn't know about `pattern template invocation`, I've just noticed the assignment on top, IMO it was confusing. That should only be one, unless ... you run a `JobInvocation` for 2 providers, in that case it's 2. We should really remove this asap.

.eslintrc
dLobatog: It's using Foreman's config now

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -1,59 +1,29 @@
...
-  <% job_invocation.pattern_template_invocations.each do |template_invocation| %>
-    <h5>
-      <b><%= template_invocation.template.name %></b> <%= 'through' %> <%= template_invocation.template.provider.humanized_name %>
-    </h5>
-    <% target = template_invocation.targeting.hosts.with_os.first || template_invocation.targeting.hosts.first %>
-    <%= _('Preview for target %s') % target.try(:name) || 'N/A' %>
-
-    <%= preview_box(template_invocation, target) %>
-
+<% template_invocations = job_invocation.pattern_template_invocations %>
```
dLobatog: This should go away alongside support for a JobInvocation for multiple providers at the same time IMO

webpack/index.js
```diff
@@ -0,0 +1,29 @@
...
+    10,
+  );
+
+  const mountJobInvocationContainer = () => {
+    mount('JobInvocationContainer', '#status_chart', {
+      url: `/job_invocations/chart?id=${jobInvocationId}`,
+    });
+  };
+
+  document.addEventListener('page:change', mountJobInvocationContainer);
```
dLobatog: This catches the turbolink event, so it mounts the component every time the page is changed

dLobatog:
I've updated the `JobInvocation#finished?` test to depend on `task.pending?` as @adamruzicka showed me it's a reliable way of finding whether the JobInvocation had finished or not. Prior to that I was checking if `task.state` was `cancelled` or `finished`

dLobatog:
Problems with jenkins & webpack.. [test default] [test]

dLobatog:
[test default] [test]

dLobatog:
[test default] [test]

dLobatog:
Anything I can do to move this forward @ares @iNecas @adamruzicka ?

adamruzicka:
I'm still running into some issues with the table. Now switching pages doesn't make the chart disappear, however after a while (probably on next chart refresh) the table switches back to page 1. The same happens when you try to use the search bar above the table, again on next refresh it resets.

dLobatog:
@adamruzicka Actually I see it now - it's a bug that existed before this change, not related with React. The table is updated using `js.erb` hence React polling doesn't affect it. 

I already started reactifying the table so if that could happen in another PR I'd appreciate it, as fixing the js.erb should go away soonish

dLobatog:
I had to rebase to fix the conflict on `.hound.yml`

adamruzicka:
Leaving it to another PR sounds good to me, this one is already pretty large. Can't think of anything I'd like to see added or changed here anymore, big +1 from me

dLobatog:
@adamruzicka ~800 net LOC & tomorrow marks the 2 year anniversary of this PR :birthday:

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -1,59 +1,29 @@
...
-    <% unless job_invocation.execution_timeout_interval.nil? %>
-      <b><%= _("Timeout to kill after") %></b>: <%= job_invocation.execution_timeout_interval %> <%= _('seconds') %><br>
+      <%= render :partial => 'card_user_input', :locals => { :template_invocation => template_invocation } %>
     <% end %>
   <% end %>
+  </div>
 </div>
+<hr/>
+<% if job_invocation.resolved? %>
+  <div id="hosts" data-refresh_required="<%= job_invocation.resolved? ? '' : 'true' %>">
```
dLobatog: @iNecas @adamruzicka The table went back to the 1st page after every reload because on a previous update, this was set to 'true' all the time. I've changed it to this conditional and now it's fine.

webpack/react_app/redux/actions/jobInvocations/index.js
```diff
@@ -0,0 +1,74 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import API from 'foremanReact/API';
+
+import {
+  JOB_INVOCATIONS_GET_JOB_INVOCATIONS,
+  JOB_INVOCATIONS_POLLING_STARTED,
+  JOB_INVOCATIONS_JOB_FINISHED,
+} from '../../consts';
+
+const defaultJobInvocationsPollingInterval = 10000;
```
iNecas: Imo this should be 1000, otherwise it feels like a regression to what we had before

webpack/react_app/components/jobInvocations/AggregateStatus/index.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+const AggregateStatus = statuses => (
```
sharvit: If you will migrate it to
```jsx
const AggregateStatus = ({ statuses }) => (...);
```

You would be able to do `{statuses.success}` instead `{statuses.statuses.success}`

dLobatog: Fixed, that was goofy...

app/helpers/job_invocations_chart_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+  def job_invocation_chart(invocation)
+    donut_chart('#status_chart',
+                job_invocation_data(invocation)[:columns],
+                job_invocation_data(invocation)[:groups])
+  end
+
+  def job_invocation_data(invocation)
+    success = job_invocation_success_count(invocation)
+    cancelled = job_invocation_cancelled_count(invocation)
+    failed = job_invocation_failed_count(invocation)
```
iNecas: I believe we already have more accurate implementation of getting the counts https://github.com/theforeman/foreman_remote_execution/blob/f0cd7abce067b119d06d1e2bcddbcca7a95bd04c/app/helpers/remote_execution_helper.rb#L16

Perhaps some missed rebase?

dLobatog: Definitely, I will implement this. The PR right now can't be merged anyway until https://github.com/theforeman/foreman/pull/5137 is merged too, it's been difficult to keep this PR updated & still making sense ( > 2 years of rebases)

iNecas: Sure, I'm quite confident we're getting to the successful merge soon :)

dLobatog:
https://github.com/theforeman/foreman-infra/issues/490

dLobatog:
[test]
[test default]

iNecas:
The chart still doesn't represent the numbers correctly. Notice that the numbers are correct (both below the chart as well as in the chart api call response), but the chart still shows some part as gray ("Pending").

rex-new-chart

app/helpers/job_invocations_chart_helper.rb
```diff
@@ -0,0 +1,70 @@
...
+
+module JobInvocationsChartHelper
+  def job_invocation_chart(invocation)
+    donut_chart('#status_chart',
+                job_invocation_data(invocation)[:columns],
+                job_invocation_data(invocation)[:groups])
+  end
+
+  def job_invocation_data(invocation)
+    success = @job_invocation.progress_report[:success]
```
iNecas: I guess we should use argument `invocation`, not instance variable `@job_invocation` here. Although it doesn't seem to have an effect at the final state.

dLobatog:
@iNecas I noticed that about the chart, it also happens on the dashboard if I'm not wrong. I tend to think it's a pf-react bug, but I will investigate. Thanks for the review!

iNecas:
[test]

iNecas:
[test default] dunno which thing re-triggers the tests on plugin. Anyway jenkins is the only thing we're waiting for before merging

dLobatog:
[test]
[test default] problem with integration tests

iNecas:
[test]
[test default]

adamruzicka:
@dLobatog One more thing, could you please add information about "Start before" to the "Schedule" card? It's missing completely

dLobatog:
@iNecas BTW the problem you saw with the chart not updating the last time was investigated by @amirfefer and it's a problem in patternfly-react, it does not allow to pass some options to c3. Here's the workflow:

 https://github.com/bcbcarl/react-c3js/blob/master/src/index.js#L85 

1. First, new data is added for all columns, failed, pending, etc..
2. On each reload, we get data about failed, pending, success, cancelled
3. When the job is finishing, we will STOP getting **pending** data, so the component starts getting less data, 'failed', '

This is being fixed on https://github.com/theforeman/foreman/pull/5279

iNecas:
Thanks, I'm ready to merge this as soon as the master gets stable (which is broken currently by the uglifierjs thing in foreman).

amirfefer:
@dLobatog, 
when the job is finished, how does the fetched data look?
like this?
```javascript
[['sucess', 3], ['failed', 2],['pending', 0], ['canceled', 0]]
```
or this, without `pending` category?
```javascript
[['sucess', 3], ['failed', 2], ['canceled', 0]]
```

dLobatog:
@amirfefer `[['sucess', 3], ['failed', 2]]`, with only the categories that have at least 1 job.

ares:
I don't want to block this any longer, two things I found during testing but I'm fine with addressing in separate PRs:

1) when I paginate during run on 50 hosts, the chart disappears, only the percent label gets updated
2) preview templates seems to be limited to only 20 templates, still improvement :-)

dLobatog:
[test]
[test default]

iNecas:
[test]
[test default]

iNecas:
It seems like the jenkins failures are actually related? Perhaps some issue with npm install I don't understand. I see 'npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none was installed.' in the jenkins output, but not sure it's relevant.

iNecas:
The difference I see is that the failures on the master are usually around 1,2-10 failed tests, while with this PR, it's 600+ failed tests. I re-triggered the tests to see if this is consistent behavior. http://ci.theforeman.org/job/test_plugin_pull_request/4788/ One thing this PR adds is additional npm dependencies, so I wonder if that could be source of the issues.

iNecas:
There is still significant difference in number of failed tests in this PR and in master, where it seems more like intermittent issues on few foreman tests, while here it seems like more serious.

iNecas:
This would be one approach:

```patch
diff --git a/definitions/procedures/service/base.rb b/definitions/procedures/service/base.rb
index a78a3a1..15005b3 100644
--- a/definitions/procedures/service/base.rb
+++ b/definitions/procedures/service/base.rb
@@ -1,6 +1,12 @@
 module Procedures
   module Service
     class Base < ForemanMaintain::Procedure
+      def self.common_params(context)
+        context.instance_eval do
+          param :only, 'A comma-separated list of services to include', :array => true
+          param :exclude, 'A comma-separated list of services to skip', :array => true
+        end
+      end
       def run_service_action(action, options)
         action_noun = feature(:service).action_noun(action).capitalize
         puts "#{action_noun} the following service(s):\n"
diff --git a/definitions/procedures/service/enable.rb b/definitions/procedures/service/enable.rb
index 8ccf4a6..5ceccd3 100644
--- a/definitions/procedures/service/enable.rb
+++ b/definitions/procedures/service/enable.rb
@@ -4,8 +4,7 @@ module Procedures::Service
   class Enable < Base
     metadata do
       description 'Enable applicable services'
-      param :only, 'A comma-separated list of services to include', :array => true
-      param :exclude, 'A comma-separated list of services to skip', :array => true
+      Base.common_params(self)
     end
 
     def run
```

iNecas:
Ups, wrong PR for the last patch :)

mmoll:
[test default]

mmoll:
[test]

iNecas:
Webpack still fails to compile

dLobatog:
:green_apple:  cc @adamruzicka

iNecas:
Btw. it's this the plugins pr test? https://github.com/theforeman/foreman-infra/blob/a018d4c03ff8f40552b97285a5ac15dab03301a4/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh

iNecas:
@dLobatog last question before merging: will there be any changes needed to be done on packaging side?

dLobatog:
@iNecas That looks like the same code, but I can't find the link between it and test_plugin_pull_request. Also the job has not been updated by Puppet. I'll update it anyway.

package.json
```diff
@@ -0,0 +1,62 @@
...
+    "eslint": "^4.10.0",
+    "eslint-config-airbnb": "^16.0.0",
+    "eslint-plugin-import": "^2.8.0",
+    "eslint-plugin-jest": "^21.2.0",
+    "eslint-plugin-jsx-a11y": "^6.0.2",
+    "eslint-plugin-react": "^7.4.0",
+    "jest": "^21.2.1"
+  },
+  "dependencies": {
+    "babel-polyfill": "^6.26.0",
```
iNecas: Looking at the dependencies listed in foreman_ansible, I wonder why there is so much difference between this two https://github.com/theforeman/foreman_ansible/blob/master/package.json#L10?

dLobatog: The dependencies here are all in Foreman, however it's necessary to keep them in this package.json in order to be able to run `npm run test`. 

Tests for JS in the webpack run without Foreman, so the deps need to be local in order to be able to run them. Since Ansible has no tests for the webpack stuff (it just makes 1 call to 1 library https://github.com/theforeman/foreman_ansible/blob/master/webpack/components/ReportJsonViewer.js), it doesn't need the deps in the package.json

iNecas:
Thanks @dLobatog, it took just 2 years, 75 days and 10 minutes, but it got it eventually :)

PR #131, opened 2016-01-22
Fixes #13334 - Job live output is very hard to read

dLobatog:
The CSS on that page makes the SSH output very difficult to read.

See a couple of examples:

The CSS from the foreman-docker console output seems a lot more
readable, see how it would look like in this plugin:

### DIFF
--- a/app/assets/stylesheets/template_invocation.css.scss
+++ b/app/assets/stylesheets/template_invocation.css.scss
@@ -19,27 +19,21 @@ div.terminal {
   display: block;
   padding: 9.5px;
   margin: 0 0 10px;
-  font-size: 13px;
-  line-height: 1.428571429;
+  font-size: 12px;
   word-break: break-all;
   word-wrap: break-word;
-  color: #B2B2B2;
-  background-color: #000000;
+  color: rgba(255, 255, 255, 1);
+  background-color: rgba(47, 47, 47, 1);
   border: 1px solid #000000;
-  border-radius: 4px;
-  font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
+  border-radius: 0px;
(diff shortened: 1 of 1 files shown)


stbenjam:
I like this better, :+1: from me - anybody else have an opinion on this?

ares:
Interesting, I found your colors harder to read on my monitor :-) But I don't want to implement color profiles so if more people prefer bright colors I'm OK with it. Btw the original colors are default terminal colors. I guess what matters is how long the terminal output is, we wanted to make it fixed height in future so there would be more of dark background.

ohadlevy:
also, one comment for reviewer, can you make sure that when selecting text
(e.g. via the mouse) for coping, it changes the background of the entire
line?
thanks

dLobatog:
@ares I have a scrollable terminal in foreman-docker, we should probably reuse that if we want to make it fixed height.

ares:
@dLobatog what would be the best way to share it? could we get it to core somehow? I can imagine others rejecting the PR because there's no use of it in core, I guess for VNC console canvas is used...

dLobatog:
@ares Those concerns would be valid in my book, via a shared gem with assets that only the plugins that need it use it. I'll speak with the patternfly folks to see if they'd be interesting on offering recommendations on how to show terminal output too.

ares:
This discussion probably does not belong here but I'd say Foreman should be also a platform for other plugins we support, I could argue that plugin API is also not used by core... Maybe we could build something like foreman_extras with similar components. Anyway I'm interested in your findings about patternfly integration.

ares:
OK, no one else seems to be interested and it's 2:1. I'm not sure what exactly @ohadlevy meant by entire line but the PR also fixes the height of selected background. There are still gaps between lines but that's a separate issues. Thanks @dLobatog merging!

PR #132, opened 2016-01-23
Fixes #13342 - Align Remote execution checkbox in NICs modal

dLobatog:
Notice the label change to 'Remote execution', which I think it's more
clear. I also changed the help inline to a popover as per Patternfly
guidelines https://www.patternfly.org/patterns/field-level-help/

before: 
after:

### DIFF
--- a/app/overrides/foreman/nics/_execution_interface.html.erb
+++ b/app/overrides/foreman/nics/_execution_interface.html.erb
@@ -1,1 +1,9 @@
-<%= checkbox_f f, :execution, :help_inline => _("The execution interface is used for remote execution"), :class => :interface_execution %>
+<%= checkbox_f(
+  f,
+  :execution,
+  :help_inline => popover(
+    '', _("The execution interface is used for remote execution"),
+    :rel => 'popover-modal'),
+  :label_size => 'col-md-3',
+  :label => _('Remote execution'),
+  :class => :interface_execution) %>


ohadlevy:
Any idea why the icon is not centralized?

dLobatog:
@ohadlevy The popover icon? No idea, it wasn't centred with bootstrap either. We're supposed to move it next to the label with patternfly - https://www.patternfly.org/patterns/field-level-help/

stbenjam:
Tested and looks good now - thanks!

We'd probably have to update this again once you match the patternfly style over in Foreman.  What do you think about extracting the checkboxes in Foreman to a nic helper?  So instead of this huge chunk of checkbox code: https://github.com/theforeman/foreman/blob/develop/app/views/nic/_base_form.html.erb#L50

It could look something like this instead:


```erb
<%= interface_flag :managed, _('Should this interface be managed via DHCP and DNS smart proxy and should it be configured during provisioning?') if SETTINGS[:unattended] %>
<%= interface_flag :primary, _('The Primary interface is used for constructing the FQDN of the host') %>
<%= interface_flag :provision, _('The Provisioning interface is used for TFTP of PXELinux (or SSH for image-based hosts)') %>
```

dLobatog:
I think that makes sense, but unless I can land that on 1.11, this checkbox would be misaligned until 1.12 if we wait for that change in Foreman

stbenjam:
Sure, no problem, this is good to go, I filed http://projects.theforeman.org/issues/13360

Thanks @dLobatog !

PR #134, opened 2016-01-25
Refactor multiple parts to make outside provider integration easier

dLobatog:
I normally don't do this but here's a provider, jobs and renderer for
Ansible and REX. It needs https://github.com/iNecas/smart_proxy_ansible
to work.

I tried to refactor stuff out from SSH and reuse it as much as possible.
At this stage I'm in a bit of a rush to present it at
http://cfgmgmtcamp.eu/schedule/speakers/DanielLobatoGarcia.html , so I
can't really make huge changes to it until then.

This PR is just a way for me to get feedback in case I'm doing something
terribly wrong and some kind REX maintainer has the time to take a quick
look. :)

Props to @inecas for giving me a place to start with this (his ansible
branch of REX contains basically this)


angystardust:
Awsome! Absolutely love this! :heart:

dLobatog:
Template renderer now doesn't rely on a special Job template. Now it can use normal playbooks. The host variables and group variables are resolved via  host parameters and host group parameters.

iNecas:
The only drawback right now is, that part of the template invocation is now directly mapped inside the host, but probably not a big deal (depends on how would the users want to use that). On the other hand, we could add the mapping to the template later (maybe via inputs…)?

dLobatog:
@inecas That aside, a problem I'm facing is that it's not sending the remote user properly and it's trying to call SSH with the foreman-proxy user. Obviously my VMs normally don't have a foreman-proxy user hence my jobs fail. I changed `/etc/ansible.cfg` so that the remote_user is always root by default but we should be able to send it from here to the proxy

iNecas:
@dLobatog ack

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,33 @@
+module JobInvocationOutputHelper
+
+  def colorize_line(line)
```
dLobatog: This method may look a little odd - I explain. Ansible output is colored some times and these colors are printed as regular bash color codes in live output without this method.

I haven't tried, but I'm almost certain this is also relevant to SSH. I can remove this from this PR and submit it separately.

dLobatog:
@stbenjam @iNecas @ares This is lacking tests and everything - however it's fully working. Can you recommend a path for this to get in? I understand it's a bit tough to review, so if you want, I can submit this in 4 parts:

1. Support for coloring lines (also relevant for SSH I think)? 
2. Refactoring of RunHostJob and RunHostsJob - not sure this needs tests
3. Refactoring of RunProxyCommand - not sure this needs tests
4. Ansible provider registration, RunHostAnsible, RunProxyAnsibleCommand, and AnsibleTemplateRenderer. this needs tests

Sounds good? Any other idea?

iNecas:
Let me first to see it on one place here before the splitting, I will allocate time for review later this week. Does it work for you?

dLobatog:
@iNecas ping?

iNecas:
Sorry, other priorities did not allowed me to look into this at the moment. Anyway, I was thinking on how hard it would be to have the ansible provider in a separate plugin, probably in the foreman_ansible one?

dLobatog:
@iNecas I'll try to do it. The only issue though, is that there are a number of things I made here I wouldn't have available, such as a generic RunProxyCommand, refactorings of RunHost and RunHostsJob, etc... - I can do without them but then foreman_ansible will duplicate a lot of the SSH provider code.

iNecas:
I'm ok with keeping the refactorings here, assuming that will make it easier for other providers as well.

app/helpers/job_invocation_output_helper.rb
ohadlevy: afair there is a client side (javascript) lib that convert terminal to html colors.

ohadlevy: oh look i already created an issue for it :) http://projects.theforeman.org/issues/12105

dLobatog:
Closing for https://github.com/theforeman/foreman_ansible/pull/18 and https://github.com/theforeman/foreman_remote_execution/pull/160

PR #136, opened 2016-01-27
Fixes #13137 - Improve table styles

ares:
before
tables1
after
tables2

and similarly for job invocations


stbenjam:
Job templates page looks better, for invocations I feel like the Start column isn't big enough, 
column

stbenjam:
md-2 for start column looks better to me, what do you think?

md2

ares:
right, I missed it since all of my texts were short, good catch!

stbenjam:
APJ

ares:
Thanks, merging!

PR #137, opened 2016-01-27
Fixes #13148 - style statistics as on report page

ares:
before
tables3
after
tables4

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -165,6 +165,13 @@ module RemoteExecutionHelper
     end
   end
 
+  def invocation_result(invocation, key)
+    unknown = '&mdash;'
+    result = invocation_count(invocation, :output_key => key, :unknown_string => unknown.html_safe )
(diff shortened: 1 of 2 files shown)


stbenjam:
I'm not sure I really like the Foreman reports page index, but at least it looks consistent.  The only thing I'm not sure about is these N/A's.

stbenjam:
How about dashes?

foo

ares:
Dashes seems a bit better, maybe we could use one column with the statistics that @dLobatog adds in the job invocation detail under the chart. But yeah, it wouldn't be consistent with reports so maybe we should change it there first.

ohadlevy:
:+1: to fix on both places, I wasnt too happy with the reports page in
foreman after patternfly either.

stbenjam:
Looks good, although I think I prefer `&mdash;` to the double `--`

ares:
right, changed to `&mdash;`

stbenjam:
:+1: ACK

ares:
Thanks, merging

PR #139, opened 2016-01-27
Fixes #13142 - introduce advanced flag

ares:
RFC, I'm not sure if some icon next to link would help or maybe move schedule selection above inputs?

### DIFF
--- a/app/assets/javascripts/template_invocation.js
+++ b/app/assets/javascripts/template_invocation.js
@@ -74,6 +74,19 @@ function job_invocation_form_binds() {
   $('textarea#targeting_search_query').on('change', refresh_execution_form);
 
   $('select#targeting_bookmark_id').on('change', refresh_search_query);
+
+  $('a.advanced_fields_switch').on('click', toggle_advanced_fields )
+}
+
+function toggle_advanced_fields() {
+  switcher = $('a.advanced_fields_switch');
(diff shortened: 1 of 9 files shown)


ares:
invo1
invo2

ohadlevy:
:+1: IMHO Schedule could also be advanced?

regarding icon - probably fa-angle-down or similar?

   -

ares:
How about hiding `Type of query`, that's useful only for future and recurring scheduling. I'd prefer to keep scheduling visible since it's important feature and it's good for users to know about it. Also it doesn't take much space. And yeah, it looks better with icon 
advanced

ares:
Latest version
full
there are too many radio buttons but advanced will be hidden most of the time I think

stbenjam:
This is such a huge improvement!! :+1: :+1: 

I think the arrows should be different though - `fa-angle-right` when closed and `fa-angle-down` when open.  I think that's how most of these work? At least that's how @dLobatog's over in https://github.com/theforeman/foreman_remote_execution/pull/130 work for the template preview.

I don't much care just as long as we're consistent about it

ares:
I changed the icons to be consistent, please take another look.

stbenjam:
ackity ACK, works great!

ares:
Thanks you, merging

PR #141, opened 2016-01-28
Fixes #13439 - authorized_files is created at the user directory but owner is root

dLobatog:
As noticed by ewoud on
https://github.com/theforeman/community-templates/pull/233 , during
provision the authorized_keys file is created under the 'ssh_user'
directory but the user may not be able to read or write it after that,
since it was created by root.

Fix is simple, just chown the directory and file as 'ssh_user'.

### DIFF
--- a/app/views/unattended/snippets/_remote_execution_ssh_keys.erb
+++ b/app/views/unattended/snippets/_remote_execution_ssh_keys.erb
@@ -15,4 +15,5 @@ EOF
 
 chmod 700 <%= ssh_path %>
 chmod 600 <%= ssh_path %>/authorized_keys
+chown -R <%= "#{ssh_user}:#{ssh_user}" %> <%= ssh_path %>
 <% end %>


stbenjam:
As theforeman/community-templates#233 was merged, I think we should just drop this from the plugin entirely.  We'll need to make sure foreman develop is synced with community-templates

ares:
@stbenjam I think we need this fix if we release 0.3.0 before Foreman 1.11 branching, since templates are synced during branching AFAIK. After sync, we can merge https://github.com/theforeman/foreman_remote_execution/pull/151

stbenjam:
We can sync templates at our leisure, see https://github.com/theforeman/community-templates/pull/233#issuecomment-176227677.

But I don't really care if you want to merge this, but I would prefer if we get the PR at least open that removes the templates first, I don't want us to forget.

stbenjam:
I see https://github.com/theforeman/foreman_remote_execution/pull/151 was opened, thanks.

But sorry one last thing - per the discussion on https://github.com/theforeman/community-templates/pull/233#discussion_r51133047, it shouldn't have been merged to community-templates like it was.

The chown shouldn't include the ssh_user as the group, as it is not guaranteed to be the same, This PR and community-templates needs to be updated as well to fix that.

stbenjam:
I filed a PR https://github.com/theforeman/community-templates/pull/236

stbenjam:
Disregard, we don't need this, Foreman templates will be synced before we release 0.3.0 (https://github.com/theforeman/foreman/pull/3167).

Thanks @dLobatog for getting this moved over to community-templates :+1:

PR #142, opened 2016-01-29
Fixes #13461 - add preview information to template context

ares:


stbenjam:
APJ

iNecas:
I'm curious more about the use case for this? An example?

stbenjam:
The way I read @ares' issue was a very advanced user who knows ruby may want to do something crazy.

What if you had a template that called one of the proxy API's, say `::ProxyAPI::Realm` to create a record in IPA w/ the one-time password, and then rendered a template to call `ipa-client-install` via remote execution with the retrieved OTP?

stbenjam:
So you could have a template like this - I actually think it's a quite cool idea:

```erb
<%
  otp = if @preview
           '$[OTP]'
         else
           api = ::ProxyAPI::Realm.new(:url => SmartProxy.find_by(name: 'realm-proxy.example.com'))
           otp = api.create(:hostname => @host.fqdn)
         end
%>

ipa-client-install -w <%= otp %>
```

iNecas:
I see, and yes, that is interesting :)

ares:
hey, no spoilers! :-) but yeah, @stbenjam :dart: 

another usecases that came to my mind - you may want to hide sensitive info, e.g. private key that you distribute (I know, not a good idea anyway) and we could use it to hide `die()` helpers, public key content etc in preview until we have folding for snippets

maybe a better approach would be to add macro like
```
preview_stub('stub text') do
  $crazy_and_dangerous_stuff
end
```

maybe I'll wrap it into helper at least?

ares:
I've added a `preview?` wrapper/macro to avoid using variables, please take another look.

adamruzicka:
Looks good to me and works well.

I sort of like the idea of `preview_stub(placeholder = '', &block)` macro, considering probably all usages of this will be in the form of
```ruby
if preview?
  something # Or possibly nothing at all
else
  dragons_n_stuff
end
```

ares:
when I added the macro with block I quickly realized that string will be usually too long, keeping it as simple `preview?` method gives the biggest flexibility a better readability

adamruzicka:
Ok, no arguing against that

ares:
@adamruzicka so should I consider it as ACK? :-)

adamruzicka:
@ares Definitely :)

ares:
Thanks, merging.

PR #144, opened 2016-02-01
fixes #10756, #13500 - developer api and PXE-less reprovisioning

iNecas:
This PR adds ability to run the script provision template via remote
execution and reboot the host, which effectively brings PXEless
provisioning into Foreman.

pxe-less-reprovision-with-rex
This feature uses a new developer API of the remote execution plugin.

Any plugin (or core) and define a feature at initialization:

```ruby 
RemoteExecutionFeature.register(:reprovision, N_("Reprovision"),
                               :description => "Reprovision the host via script",
                               :provided_inputs => ["script"])
```

The mapping between the feature and templates can be made via
'Administer -> Remote Execution Features'.

```ruby
composer = JobInvocationComposer.for_feature(:reprovision, host, :script => "grubby...") composer.save! composer.trigger
```

When seeding a template, there is a new meta `feature`, which allows to
assign some template as default to the specified feature.

The template mapping is now limited to just one template, but in future, we
can extend that to allow different mappings through organizations, but lets
keep it simple first.

TODO:

- [ ] moar tests
- [ ] API and Hammer CLI for reprovision
- [ ] permissions
- [ ] documenting in manual

### DIFF
--- a/app/controllers/concerns/foreman_remote_execution/hosts_controller_extensions.rb
+++ b/app/controllers/concerns/foreman_remote_execution/hosts_controller_extensions.rb
@@ -1,0 +1,36 @@
+module ForemanRemoteExecution
+  module HostsControllerExtensions
+    extend ActiveSupport::Concern
+    include Foreman::Renderer
+
+    included do
+      alias_method_chain(:action_permission, :remote_execution)
+    end
+
+    def reprovision
+      find_resource
+      script_template = @host.provisioning_template(:kind => 'script')
+      if script_template.nil?
+        process_error :redirect => :back, :error_msg => _("No script provisioning template available")
+        return
+      end
+      @host.setBuild
+      script = unattended_render(script_template.template, @template_name)
+      composer = JobInvocationComposer.for_feature(:reprovision, @host, :script => script)
+      composer.save!
+      composer.trigger
+      process_success :success_msg => _("Reprovision job started. The host should reboot soon."), :success_redirect => :back
+    end
+
+    private
+
+    def action_permission_with_remote_execution
+      case params[:action]
+      when 'reprovision'
+        :edit_host
+      else
+        action_permission_without_remote_execution
+      end
+    end
+  end
+end
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -11,9 +11,13 @@ module ForemanRemoteExecution
       multiple_actions_without_remote_execution + [[_('Run Job'), new_job_invocation_path, false]]
     end
 
-    def host_title_actions_with_run_button(*args)
-      title_actions(button_group(link_to(_('Run Job'), new_job_invocation_path(:host_ids => [args.first.id]), :id => :run_button)))
-      host_title_actions_without_run_button(*args)
+    def host_title_actions_with_run_button(host)
+      links = [link_to(_('Run Job'), new_job_invocation_path(:host_ids => [host.id]), :id => :run_button)]
+      if RemoteExecutionFeature.feature(:reprovision).template && @host.provisioning_template(:kind => 'script')
+        links << link_to(_('Reprovision'), reprovision_host_path(host.id), { :method => :post, :disabled =>  @host.build })
+      end
+      title_actions(button_group(*links))
+      host_title_actions_without_run_button(host)
     end
   end
 end
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -50,7 +50,7 @@ module ForemanRemoteExecution
                         proxy_scope = ::SmartProxy
                       end
 
-        proxies[:global] = proxy_scope.authorized.with_features(provider)
+        proxies[:global] = proxy_scope.with_features(provider)
       end
 
       proxies
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -57,10 +57,8 @@ module Api
       def create
         composer = JobInvocationComposer.from_api_params(job_invocation_params)
         composer.save!
-        @job_invocation = composer.job_invocation
-        @job_invocation.generate_description! if @job_invocation.description.blank?
-        composer.triggering.trigger(::Actions::RemoteExecution::RunHostsJob, @job_invocation)
-        process_response @job_invocation
+        composer.trigger
+        process_response composer.job_invocation
       end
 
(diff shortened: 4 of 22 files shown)


iNecas:
The reprovisioning in action:
pxe-less-reprovision-with-rex

I had to update the script template slightly to work on Centos7 (ommiting the --copy-default):

```
grubby --add-kernel=$KERNEL --initrd=$INITRD --title "<%= @host.operatingsystem %>" --make-default  --args="ks=<%= foreman_url('provision')%>"
```

iNecas:
It gives the best results in combination with https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/20

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -50,7 +50,7 @@ def remote_execution_proxies(provider)
                         proxy_scope = ::SmartProxy
                       end
 
-        proxies[:global] = proxy_scope.authorized.with_features(provider)
```
iNecas: I had to do this in order to ssh keys snippet to work with global proxy, I will probably open a separate PR and issue for this.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -11,9 +11,13 @@ def multiple_actions_with_remote_execution
       multiple_actions_without_remote_execution + [[_('Run Job'), new_job_invocation_path, false]]
     end
 
-    def host_title_actions_with_run_button(*args)
-      title_actions(button_group(link_to(_('Run Job'), new_job_invocation_path(:host_ids => [args.first.id]), :id => :run_button)))
-      host_title_actions_without_run_button(*args)
+    def host_title_actions_with_run_button(host)
+      links = [link_to(_('Run Job'), new_job_invocation_path(:host_ids => [host.id]), :id => :run_button)]
+      if RemoteExecutionFeature.feature(:reprovision).template && @host.provisioning_template(:kind => 'script')
+        links << link_to(_('Reprovision'), reprovision_host_path(host.id), { :method => :post, :disabled =>  @host.build })
```
stbenjam: Could we somehow add this to the build button as a dropdown? I would guess maybe not easy to extend from plugin.

I only wonder because  the host actions can get quite big if you install a lot of plugins, and this is really related to the build.

app/controllers/concerns/foreman_remote_execution/hosts_controller_extensions.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    included do
+      alias_method_chain(:action_permission, :remote_execution)
+    end
+
+    def reprovision
+      find_resource
+      script_template = @host.provisioning_template(:kind => 'script')
+      if script_template.nil?
+        process_error :redirect => :back, :error_msg => _("No script provisioning template available")
```
stbenjam: Single quote string literals (and in a bunch of other places) :smile:  rubocop can fix all of these.

app/controllers/concerns/foreman_remote_execution/hosts_controller_extensions.rb
```diff
@@ -0,0 +1,36 @@
...
+        return
+      end
+      @host.setBuild
+      script = unattended_render(script_template.template, @template_name)
+      composer = JobInvocationComposer.for_feature(:reprovision, @host, :script => script)
+      composer.save!
+      composer.trigger
```
stbenjam: We do this 3-step dance in a lot of places, we don't have to change it now, but any reason not to just save and trigger in one step for this and other controllers? Something like 

```ruby
JobInvocationComposer.trigger_for_feature!(:reprovision, @host, :script => script)
```

stbenjam:
I like the approach in general, do we have a way to get the results back to the caller? This is also a separate problem by being able to show some structured results (e.g. package install case -> return the version installed), but we should at least support the ability to send a success/fail message back, perhaps by letting the call register itself as an observer?

app/controllers/concerns/foreman_remote_execution/hosts_controller_extensions.rb
iNecas: so there would be both `JobInvocationComposer.trigger_for_feature(:reprovision, @host, :script => script)` and `JobInvocationComposer.trigger_for_feature!(:reprovision, @host, :script => script)` (the difference is in `save` and `save!`). Maybe `JobInvocationComposer.for_feature(:reprovision, @host, :script => script).trigger` and `JobInvocationComposer.for_feature(:reprovision, @host, :script => script).trigger!`

iNecas:
Even in katello today, we're using the task object to get the output. So we might just get the task data simpler to get after triggering, so that the plugin can consume that.

app/controllers/concerns/foreman_remote_execution/hosts_controller_extensions.rb
stbenjam: Yea, that would work.  It's really not a big deal, I just noticed we keep making minor changes to how this gets called and figured maybe it could be DRYer.

stbenjam:
If a Katello user kicks off a package install, the UI will probably stay the same as it is today, so that `composer.trigger` should return the task ID and Katello can show the same progress bar it does now, right?

I was thinking of a case where you have a plugin that kicks off a remote execution task, and wants to get the results of that sent somewhere else (eventually).

For example, foreman_salt has a 'Run Salt' button on the Hosts page. I'd like to notify a ReportImporter service when it completes, because then the service could look at the remote execution task output, and parse that output as a Report.

Maybe there's other ways to accomplish this already - like call Remote Execution from within a task itself and just block on the REX task to complete, instead of doing something like the observer pattern.

Hopefully I'm explaining what I mean :confused: :question:

iNecas:
I see. In your case, I would probably rather go into triggering the task from within the orchestration action in salt: so you would get the composer from the developer API and the calling `plan_action(RunHostsJob, composer)` from your action: thin this case, the action would get part of your orchestration and you could chain the actions as dynflow allows to do. We might provide some helper on doing so (maybe we will get to it even with Katello… we will see)

iNecas:
FYI: I've opened separate PR with just developer API https://github.com/theforeman/foreman_remote_execution/pull/148, in order to move other features forward (such as Katello integration), and getting some time to get the pxe-less reprovision finished (API/CLI and UI updates)

iNecas:
I've opened the power action templates in separate PR https://github.com/theforeman/foreman_remote_execution/pull/155

app/models/concerns/foreman_remote_execution/host_extensions.rb
iNecas: https://github.com/theforeman/foreman_remote_execution/pull/156

iNecas:
closing in favor of #159, I will try also to extend the build button with the toggle there

PR #145, opened 2016-02-03
fixes #11948 - setup remote execution for localization

stbenjam:


locale/action_names.rb
```diff
@@ -0,0 +1,7 @@
+# Autogenerated!
+_("Action with sub plans")
+_("Create architecture")
+_("Delete architecture")
+_("Update architecture")
+_("Import facts")
+_("")
```
domcleal: Remove?

stbenjam: This is probably caused by https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L63, I had to comment it out to get the rake task to run, otherwise I get this error:

```
NoMethodError: undefined method `[]' for nil:NilClass
/home/vagrant/foreman_remote_execution/app/lib/actions/remote_execution/run_host_job.rb:63:in `humanized_name'
```

locale/en/foreman_plugin_template.po
```diff
@@ -0,0 +1,19 @@
+# foreman_plugin_template
```
domcleal: This file's misnamed.

domcleal: Oh, it shouldn't even exist - you have one named after the plugin.

stbenjam: Ah, sorry, I refetched the en directory from the plugin template.  I'll remove it.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+"\n"
+
+msgid ""
+"Another interface is already set as execution. Are you sure you want to use th"
+"is one instead?"
+msgstr ""
+
+msgid ""
+"<div class=\"line error\">There was an error while updating the status, try <a h"
+"ref=\"javascript:window.location.href=window.location.href\">refreshing</a> the "
```
domcleal: You should extract the link out of here, which is particularly tricky without fragmenting the sentence, but the translation shouldn't contain JS.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+"page</div>"
+msgstr ""
+
+msgid "List foreign input sets"
+msgstr ""
+
+msgid "Show foreign input set details"
+msgstr ""
+
+msgid "Target template id"
```
domcleal: usually ID is in capitals

locale/action_names.rb
stbenjam: @iNecas Any idea how to get it to not try the interpolation when it extracts the action names?

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+msgid "Create a template input"
+msgstr ""
+
+msgid "Delete a template input"
+msgstr ""
+
+msgid "Update a template input"
+msgstr ""
+
+msgid ", note that you must save template input changes before you try to preview it."
```
domcleal: Appears to be a sentence fragment, starting with a comma.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+msgid "Pending"
+msgstr ""
+
+msgid "Cancelled"
+msgstr ""
+
+msgid "queued"
+msgstr ""
+
+msgid "running %s%"
```
domcleal: Trailing %?

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+msgstr ""
+
+msgid "Description template"
+msgstr ""
+
+msgid "Explanation"
+msgstr ""
+
+msgid ""
+"This template is used to generate the description.\n"
```
domcleal: Whitespace is a bit of a mess here, you might need to use heredoc or change how it's represented in the source.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+"ates just before the execution is started, so if it's planed in future, target"
+"ed hosts set may change before it</li></ul>"
+msgstr ""
+
+msgid ""
+"The final host list may change because the selected query is dynamic.  It will"
+" be rerun during execution."
+msgstr ""
+
+msgid "...and %{count} more"
```
domcleal: This should probably use `n_()` for pluralisation, even if it's the same in English.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+msgstr ""
+
+msgid "Status"
+msgstr ""
+
+msgid "Actions"
+msgstr ""
+
+msgid ""
+"\"The dynamic query '#{job_invocation.targeting.search_query}' was not resolved"
```
domcleal: This variable must be interpolated with %s.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+" yet. The list of hosts to which it would resolve now can be seen\""
```
domcleal: There seems to be an extra leading and trailing quotation mark, and a missing full stop.

locale/en/foreman_remote_execution.po
```diff
@@ -1,19 +1,760 @@
...
+msgstr ""
+
+msgid "Job"
+msgstr ""
+
+msgid "Type"
+msgstr ""
+
+msgid ""
+"This template is used to generate the description.\n"
```
domcleal: Whitespace is a mess here, may need reformatting in the source.

locale/en/foreman_remote_execution.po
stbenjam: It shows a percentage

domcleal:
I'm commenting on the POT file to indicate bad string extractions - they need to be fixed in the code and re-extracted before they're suitable for translation.

locale/action_names.rb
domcleal: Not sure I see the connection between that code and this file.  I'm not sure what's autogenerated in this file, or which rake task runs that?

stbenjam: ForemanTasks extends the gettext task, to localize humanized_names.

stbenjam: And it explodes on the given code, so I commented it out to generate this PR, which generated an empty string, I guess.

locale/en/foreman_remote_execution.po
domcleal: I'm not sure what to suggest here, but will say that it's highly likely to get mistranslated on occasion.  While you can't just append the percent sign and to localise it properly means it should be included in the string, you might want to rewrite it as %{percent}% and/or add a TRANSLATORS command in the source code immediately before to say it's a percentage and should contain the % sign.  Hopefully then it won't be deleted.

locale/action_names.rb
domcleal: Perhaps you need to handle a nil `input[:host]` at https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L64, e.g. the next line has a `.try` to do the same.

locale/en/foreman_remote_execution.po
stbenjam: Yea, it's a bit hokey in the code, it's concatenated with a "here" link.  Could I do this instead?

```ruby
        <%= _("The dynamic query '%{query}' was not resolved yet. The list of hosts to which it would resolve now can be seen %{here}." % {:query => job_invocation.targeting.search_query,
                                                                                                                                           :here => link_to(_('here.'), hosts_url(:search => job_invocation.targeting.search_query))}) %>
```

domcleal: There's not any good way to do this, and we also have instances of it in Foreman.  The problem with the way you suggest is that the two different strings in that line will not be translated together, and the translator will have no idea or context to say that "here." is a continuation of the first sentence.

IMHO it's probably best, but much clunkier to either do:

    <%= _("Foo %{link_start}refresh%{link_end}.") % {:link_start => "<a href=\"#{foo_path}\">", :link_end => "</a>"}
    <%= _("Foo: %s") % link_to(_("describe refresh explicitly"), ..) %>

depending on how important the inline link is, and preferring to have a more standalone fragment in the latter example than a single word/partial sentence "here".

locale/action_names.rb
```diff
@@ -0,0 +1,7 @@
+# Autogenerated!
+_("Action with sub plans")
+_("Create architecture")
+_("Delete architecture")
+_("Update architecture")
+_("Import facts")
+_(" on ")
```
stbenjam: Not much better here, shouldn't the `%{description} on %{host}` be shown here to give translator's context?

app/helpers/remote_execution_helper.rb
```diff
@@ -35,8 +35,8 @@ def job_invocation_status(invocation)
       when HostStatus::ExecutionStatus::QUEUED
         _('queued')
       when HostStatus::ExecutionStatus::RUNNING
-        _('running %s%') % invocation.progress
-      when HostStatus::ExecutionStatus::OK
+        _('running %{percent}%') % {:percent => invocation.progress}
+      when HostStatus::ExecutionStatus::K
```
adamruzicka: I'm afraid we can't make `K` an official host status

stbenjam: Thanks nice catch :laughing:

locale/action_names.rb
adamruzicka: @stbenjam [foreman_remote_execution/pull/153](https://github.com/theforeman/foreman_remote_execution/pull/153) should fix this issue by renaming the action to `Remote action` and using `humanized_input` for displaying the rest

stbenjam:
This should be good to go if someone could review it again, I'd like to get the Transifex projects online

app/assets/javascripts/template_invocation.js
```diff
@@ -96,7 +96,7 @@ function delayed_refresh(url, data){
         url: url,
         data: data,
         dataType: "script",
-        error: function() { $("div.terminal div.printable").append(__('<div class="line error">There was an error while updating the status, try <a href="javascript:window.location.href=window.location.href">refreshing</a> the page</div>')) }
+        error: function() { $("div.terminal div.printable").append(__('<div class="line error">There was an error while updating the status, try refreshing the page.</div>')) }
```
domcleal: HTML should be outside of the translated string, concatenate them together.

I'm not actually sure that Javascript translation is in place for plugins anyway, I think you'll find that it's always English as there's no mechanism to generate and include app/assets/javascripts/locale/ files like Foreman core does.

stbenjam: Updated, but FWIW, there's certainly a bunch of strings in foreman core with HTML not extracted out of the string.

domcleal: Indeed, it's far from perfect.  I just try to catch these as early as possible to prevent unrelated HTML updates causing translations to be discarded and having to be re-done, make translator's lives easier and to reduce bugs occurring in translations themselves.

domcleal:
POT looks good for translation.

stbenjam:
Thanks, and it's green.

iNecas:
ACK

stbenjam:
Thanks!

PR #148, opened 2016-02-04
fixes #10756 - developer api for registering features and templates

iNecas:
Any plugin (or core) and define a feature at initialization:

```
RemoteExecutionFeature.register(:reprovision, N_("Reprovision"),
                                :description => "Reprovision the host via script",
                                :provided_inputs => ["script"])
```

The mapping between the feature and templates can be made via
'Administer -> Remote Execution Features'.

```
composer = JobInvocationComposer.for_feature(:reprovision, host, :script => "grubby...")
composer.save!
composer.trigger
```

When seeding a template, there is a new meta `feature`, which allows
to assign some template as default to the specified feature.

The template mapping is now limited to just one template, but in future,
we can extend that to allow different mappings through organizations,
but lets keep it simple first.

- [x] - API/cli for feature mapping setting
- [x] -tests for API/CLI

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -16,7 +16,7 @@ class JobTemplate < ::Template
   has_many :template_invocations, -> { where('host_id IS NOT NULL') }, :foreign_key => 'template_id'
   has_many :pattern_template_invocations, -> { where('host_id IS NULL') }, :foreign_key => 'template_id', :class_name => 'TemplateInvocation'
 
-  has_many :remote_execution_features, :dependent => :nullify, :foreign_key => 'template_id'
+  has_many :remote_execution_features, :dependent => :nullify
 
   # these can't be shared in parent class, scoped search can't handle STI properly
   # tested with scoped_search 3.2.0
@@ -76,7 +76,7 @@ class JobTemplate < ::Template
     end
 
     if metadata['feature'] && (feature = RemoteExecutionFeature.feature(metadata['feature']))
-      feature.template_id ||= template.id
+      feature.job_template_id ||= template.id
       feature.save!
     end
 
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -16,6 +16,8 @@ class JobTemplate < ::Template
   has_many :template_invocations, -> { where('host_id IS NOT NULL') }, :foreign_key => 'template_id'
   has_many :pattern_template_invocations, -> { where('host_id IS NULL') }, :foreign_key => 'template_id', :class_name => 'TemplateInvocation'
 
+  has_many :remote_execution_features, :dependent => :nullify, :foreign_key => 'template_id'
+
   # these can't be shared in parent class, scoped search can't handle STI properly
   # tested with scoped_search 3.2.0
   include Taxonomix
@@ -66,11 +68,16 @@ class JobTemplate < ::Template
 
     inputs = metadata.delete('template_inputs')
 
-    template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options))
+    template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')).except('feature').merge(options))
     template.assign_taxonomies
 
     inputs.each do |input|
       template.template_inputs << TemplateInput.create(input)
+    end
+
+    if metadata['feature'] && (feature = RemoteExecutionFeature.feature(metadata['feature']))
+      feature.template_id ||= template.id
+      feature.save!
     end
 
     template
--- a/app/models/remote_execution_feature.rb
+++ b/app/models/remote_execution_feature.rb
@@ -1,0 +1,36 @@
+class RemoteExecutionFeature < ActiveRecord::Base
+  attr_accessible :label, :name, :provided_input_names, :description, :template_id
+
+  validate :label, :name, :presence => true, :unique => true
+
+  belongs_to :template
+
+  extend FriendlyId
+  friendly_id :label
+
+  def provided_input_names
+    self.provided_inputs.to_s.split(',').map(&:chomp)
+  end
+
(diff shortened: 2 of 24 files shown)


db/migrate/20160201154900_change_template_invocation_input_values_value.rb
```diff
@@ -0,0 +1,14 @@
+class CreateRemoteExecutionFeatures < ActiveRecord::Migration
```
ehelms: Looks like you have the same migration twice here.

ehelms:
I want to see if I understand the idea correctly. A plugin or core can define a feature, which is a specific, hard coded remote execution action with an associated template? 

 1) How does a plugin declare what template the feature should use?
 2) Admin's can change the default template for a given feature?
   2a) What is an example of where an Admin would need to change a feature template for say Katello package install for example?
 3) Is a given feature tied to a provider type?

app/models/job_template.rb
```diff
@@ -70,13 +72,18 @@ def self.import(template, options = {})
...
 
-    template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options))
+    template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')).except('feature').merge(options))
     template.assign_taxonomies
 
     inputs.each do |input|
       template.template_inputs << TemplateInput.create(input)
     end
 
+    if metadata['feature'] && feature = RemoteExecutionFeature.feature(metadata['feature'])
```
stbenjam: Nitpick, put assignments like this in parens

lib/foreman_remote_execution/engine.rb
```diff
@@ -95,6 +103,9 @@ class Engine < ::Rails::Engine
         register_custom_status HostStatus::ExecutionStatus
         # add dashboard widget
         # widget 'foreman_remote_execution_widget', name: N_('Foreman plugin template widget'), sizex: 4, sizey: 1
+
+        RemoteExecutionFeature.register(:reprovision, N_("Reprovision"),
+                                        :description => "Reprovision the host via a script")
```
stbenjam: Reprovision not part of this pr right?

iNecas:
@ehelms conceptually:

1. the plugin (or core) declares what features are going to use remote execution for something, such as here: https://github.com/Katello/katello/pull/5756/files#diff-84f0b46367dabf5e0710a0fe3aca7627R176
2. the plugin might also seed a default template for the features (such as here) https://github.com/Katello/katello/pull/5756/files#diff-7ed9ec7ba0d7a85f8b68f8e5b3e509dfR6, that ends up as defalut mapping for the feature
3. the admin can change the mapping if they have a use-case for that. The use-case is, when there is some customization needed for the feature, that is not there by default. Also, this is preparation for multi-provider case, where the Katello will not probably ship templates for all possible providers
4. the feature itself is not tied to provider - this is given by the mapping for a template. In the future, there should be possibility to map more templates to feature, and have more complex decision algorithm on which one to use for a given host, but we don't want to overcomplicate that right now.

Did this help with understanding of the concepts?

ehelms:
That helped a lot, thanks @iNecas. I wasn't picking up when reviewing the Katello PR that the block you pointed to (https://github.com/Katello/katello/pull/5756/files#diff-7ed9ec7ba0d7a85f8b68f8e5b3e509dfR6) was declarative. The github styling made me think it was simply a comment.

stbenjam:
> The github styling made me think it was simply a comment. 

The ERB comment holds the template's metadata, and we have an importer that parses it (https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_template.rb#L63).  foreman_templates plugin uses the same convention to store template metadata in the community-templates.

db/migrate/20160201154900_change_template_invocation_input_values_value.rb
iNecas: I was not too careful on the rebase, should be gone now

app/models/job_template.rb
iNecas: updated

lib/foreman_remote_execution/engine.rb
iNecas: right, thanks for spotting

iNecas:
Hammer CLI PR opened here https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/11

iNecas:
API and Hammer tests added

iNecas:
I've renamed the `remote_execution_features#template_id` to `remote_execution_features#job_template_id`. I've edited the original migration, if you already run it, the fix should be:

```
# running this sql command
alter table remote_execution_features rename template_id to job_template_id

# then regenerate the schema file:
rake db:migrate
```

iNecas:
@stbenjam the devel API is not reduced to this:

```
# returns true/false, if false, the `composer.job_invocation` contains the annoted active record with the errors
JobInvocationComposer.for_feature(:katello_install_package, host, :package => 'zsh').trigger

# raises exception, the difference really is just `save` vs. `save!`
JobInvocationComposer.for_feature(:katello_install_package, host, :package => 'zsh').trigger!
```

stbenjam:
If there's no objection, I think we can work off one PR, #149. 

FWIW, my comments were addressed and I've tested the developer API. Happy with how it works. :+1: :+1:

app/models/remote_execution_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  validate :label, :name, :presence => true, :unique => true
+
+  belongs_to :job_template
+
+  extend FriendlyId
+  friendly_id :label
+
+  def provided_input_names
+    self.provided_inputs.to_s.split(',').map(&:chomp)
```
ares: why `chomp` here? did you mean `strip` maybe?

ares:
Closing in favor of #149

PR #149, opened 2016-02-04
Fixes #10757 - preparation for Katello remote actions integration

iNecas:
Depends on https://github.com/theforeman/foreman_remote_execution/pull/148
opening separate PR for discussing the katello-related issues. The general developer API comments should go to https://github.com/theforeman/foreman_remote_execution/pull/148.

- [x] - errata support
- [x] - tests for input sets import

### DIFF
--- a/app/controllers/api/v2/remote_execution_features_controller.rb
+++ b/app/controllers/api/v2/remote_execution_features_controller.rb
@@ -1,0 +1,38 @@
+module Api
+  module V2
+    class RemoteExecutionFeaturesController < ::Api::V2::BaseController
+      include ::Api::Version2
+
+      before_filter :find_resource, :only => %w{show update}
+
+      api :GET, '/remote_execution_features/', N_('List remote execution features')
+      def index
+        @remote_execution_features = RemoteExecutionFeature.all
+      end
+
+      api :GET, '/remote_execution_features/:id', N_('Show remote execution feature')
+      param :id, :identifier, :required => true
+      def show
+      end
+
+      def_param_group :remote_execution_feature do
+        param :remote_execution_feature, Hash, :required => true, :action_aware => true do
+          param :job_template_id, :identifier, :required => true, :desc => N_('Job template id to be used for the feature')
+        end
+      end
+
+      api :PUT, '/remote_execution_features/:id', N_('Update a job template')
+      param :id, :identifier, :required => true
+      param_group :remote_execution_feature
+      def update
+        process_response @remote_execution_feature.update_attributes(params[:remote_execution_feature])
+      end
+
+      private
+
+      def parent_scope
+        resource_class.where(nil)
+      end
+    end
+  end
+end
--- a/app/controllers/api/v2/remote_execution_features_controller.rb
+++ b/app/controllers/api/v2/remote_execution_features_controller.rb
@@ -7,7 +7,7 @@ module Api
 
       api :GET, '/remote_execution_features/', N_('List remote execution features')
       def index
-        @remote_execution_features = RemoteExecutionFeature.all
+        @remote_execution_features = resource_scope
       end
 
       api :GET, '/remote_execution_features/:id', N_('Show remote execution feature')
@@ -17,7 +17,7 @@ module Api
 
       def_param_group :remote_execution_feature do
         param :remote_execution_feature, Hash, :required => true, :action_aware => true do
-          param :job_template_id, :identifier, :required => true, :desc => N_('Job template id to be used for the feature')
+          param :job_template_id, :identifier, :required => true, :desc => N_('Job template ID to be used for the feature')
         end
       end
 
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -177,6 +177,66 @@ class JobInvocationComposer
     end
   end
 
+  class ParamsForFeature
+    attr_reader :feature_label, :feature, :provided_inputs
+
+    def initialize(feature_label, hosts, provided_inputs = {})
+      @feature = RemoteExecutionFeature.feature(feature_label)
+      @provided_inputs = provided_inputs
+      if hosts.is_a? Bookmark
+        @host_bookmark = hosts
+      elsif hosts.is_a? Host::Base
+        @host_objects = [hosts]
+      elsif hosts.is_a? String
+        @host_scoped_search = hosts
+      else
+        @host_objects = hosts
+      end
+    end
+
+    def params
+      { :job_category => job_template.job_category,
+        :targeting => targeting_params,
+        :triggering => {},
+        :template_invocations => template_invocations_params,
+        :description_format => job_template.generate_description_format }.with_indifferent_access
+    end
+
+    private
+
+    def targeting_params
+      ret = {}
+      ret['targeting_type'] = Targeting::STATIC_TYPE
+      ret['search_query'] = @host_scoped_search if @host_scoped_search
+      ret['search_query'] = Targeting.build_query_from_hosts(@host_objects) if @host_objects
+      ret['bookmark_id'] = @host_bookmark.id if @host_bookmark
+      ret['user_id'] = User.current.id
+      ret
+    end
+
+    def template_invocations_params
+      [ { 'template_id' => job_template.id,
+          'input_values' => input_values_params } ]
+    end
+
+    def input_values_params
+      @provided_inputs.map do |key, value|
+        input = job_template.template_inputs.find_by_name!(key)
+        { 'template_input_id' => input.id, 'value' => value }
+      end
+    end
+
+    def job_template
+      template = JobTemplate.authorized(:view_job_templates).find_by_id(feature.job_template_id)
+      unless template
+        raise Foreman::Exception.new(N_('The template %{template_name} mapped to feature %{feature_name} is not accessible by the user'),
+                                     :template_name => mapping.template.name,
+                                     :feature_name => feature.name)
+      end
+      template
+    end
+  end
+
   attr_accessor :params, :job_invocation, :host_ids, :search_query
   delegate :job_category, :pattern_template_invocations, :template_invocations, :targeting, :triggering, :to => :job_invocation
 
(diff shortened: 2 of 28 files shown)


iNecas:
Tests for importing input sets added. The PR shoudl be ready for a final review

app/models/job_invocation_composer.rb
```diff
@@ -191,6 +263,21 @@ def compose
     self
   end
 
+  def trigger(raise_on_error = false)
+    generate_description
```
adamruzicka: Generating the description here fails to correctly retrieve the input values and generates description without them.

iNecas: Nice catch, the problem was with using the `pluck` method here https://github.com/theforeman/foreman_remote_execution/pull/149/files#diff-d3ff599b762e4ff95c9b2776d2be1079R159 - using the map instead (and working in memory) seems to fix the issue

app/models/input_template_renderer.rb
```diff
@@ -4,7 +4,7 @@ class UndefinedInput < ::Foreman::Exception
 
   class RenderError < ::Foreman::Exception
   end
-
+  include Rails.application.routes.url_helpers
```
adamruzicka: The included helpers are not used anywhere in this class

iNecas: You're right, this was added as part of the PXE-less reprovision try, where I needed to redner the `foreman_url` inside the template,removed

app/controllers/api/v2/remote_execution_features_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      api :GET, '/remote_execution_features/:id', N_('Show remote execution feature')
+      param :id, :identifier, :required => true
+      def show
+      end
+
+      def_param_group :remote_execution_feature do
+        param :remote_execution_feature, Hash, :required => true, :action_aware => true do
+          param :job_template_id, :identifier, :required => true, :desc => N_('Job template id to be used for the feature')
```
stbenjam: ID in caps

stbenjam:
I tested this, along with the Katello PR and I like the way it works at a lot.  From my perspective, this is good to go

app/models/remote_execution_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def provided_input_names
+    self.provided_inputs.to_s.split(',').map(&:chomp)
+  end
+
+  def provided_input_names=(values)
+    self.provided_inputs = Array(values).join(',')
+  end
+
+  class << self
```
ares: is there some reason (autoloader maybe) for this syntax instead of `def self.feature(label)` which I'd prefer?

app/views/remote_execution_features/_form.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <%= field(f, :label) { @remote_execution_feature.label } %>
+  </div>
+  <div class="row">
+    <%= field(f, :description) { @remote_execution_feature.description } %>
+  </div>
+  <div class="row">
+    <%= field(f, :provided_inputs) { @remote_execution_feature.provided_inputs } %>
+  </div>
+  <div class="row">
+    <%= selectable_f f, :job_template_id, JobTemplate.all.map { |t| [ t.name, t.id ] }, { :include_blank => true }, :class => 'input_type_selector' %>
```
ares: missing authorization for :view_job_templates (the question is what to display if some is already associated but user does not have permissions to see it, maybe display warning?

db/migrate/20160118124600_create_remote_execution_features.rb
```diff
@@ -0,0 +1,14 @@
+class CreateRemoteExecutionFeatures < ActiveRecord::Migration
+  def change
+    create_table :remote_execution_features do |t|
+      t.string :label, :index => true, :null => false
+      t.string :name
+      t.string :description
+      t.text :provided_inputs, :null => true
+      t.integer :job_template_id, :null => true
```
ares: do you want to be explicit about `null`? true is default so it's not required here and on line above

lib/foreman_remote_execution/engine.rb
```diff
@@ -45,6 +45,8 @@ class Engine < ::Rails::Engine
                                             :'api/v2/job_templates' => [:update],
                                             :'api/v2/template_inputs' => [:create, :update, :destroy],
                                             :'api/v2/foreign_input_sets' => [:create, :update, :destroy]}, :resource_type => 'JobTemplate'
+          permission :edit_remote_execution_features, { :remote_execution_features => [:index, :show, :update],
```
ares: not needed now but would it make sense to split into two permissions, one for viewing, second for editing (the form would either display select box or just text information)? if you agree, creating a tracking issue would be enough

app/controllers/api/v2/remote_execution_features_controller.rb
```diff
@@ -0,0 +1,38 @@
+module Api
+  module V2
+    class RemoteExecutionFeaturesController < ::Api::V2::BaseController
+      include ::Api::Version2
+
+      before_filter :find_resource, :only => %w{show update}
+
+      api :GET, '/remote_execution_features/', N_('List remote execution features')
+      def index
+        @remote_execution_features = RemoteExecutionFeature.all
```
ares: missing authorization

app/models/remote_execution_feature.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  validate :label, :name, :presence => true, :unique => true
+
+  belongs_to :job_template
+
+  extend FriendlyId
+  friendly_id :label
+
+  def provided_input_names
+    self.provided_inputs.to_s.split(',').map(&:chomp)
```
ares: why chomp here? did you mean strip maybe?

ares:
For some reason with this PR applied, administer menu is missing divider between Roles and Bookmarks
administer

lib/foreman_remote_execution/engine.rb
iNecas: I don't think it makes sense to have the view permission here right now

iNecas: And I would prefer to have the issue opened by someone, that misses it :)

iNecas:
@ares I can't reproduce the menu issue, what foreman sha are you running?

app/models/remote_execution_feature.rb
iNecas: fixed

app/controllers/api/v2/remote_execution_features_controller.rb
iNecas: fixed

app/views/remote_execution_features/_form.html.erb
iNecas: fixed

app/models/remote_execution_feature.rb
iNecas: fixed

db/migrate/20160118124600_create_remote_execution_features.rb
iNecas: Fixed, also added `:null => false` for `:name`

iNecas:
All comments I wanted to fix are fixed now, the commit squashed and rebased. I'm waiting for https://github.com/theforeman/foreman_remote_execution/pull/150 to get in to update the import to re-assign the default seed in case the remote execution feature has no other template assigned

lib/foreman_remote_execution/engine.rb
ares: ok fine, it's just inconsistent with everything else we have (edit never grants view permission), I'll file it and we'll see if someone finds it confusing

ares:
> @ares I can't reproduce the menu issue, what foreman sha are you running?

seems unrelated, I see it even with master now, probably some local issue

ares:
APJ

iNecas:
Rebased against the latest master and updated the import to sync the input sets and features

app/models/job_template.rb
```diff
@@ -72,8 +73,10 @@ def self.import(contents, options = {})
 
       template = existing || self.new
       template.sync_inputs(metadata.delete('template_inputs'))
-      template.assign_attributes(metadata.merge(:template => contents.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options))
+      template.sync_foreign_input_sets(metadata.delete('foreign_input_sets'))
+      template.assign_attributes(metadata.merge(:template => contents.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options).except('feature'))
       template.assign_taxonomies if template.new_record?
+      template.sync_feature(metadata.delete('feature'))
```
stbenjam: I think you'll have to do the delete before `assign_attributes`. When we do the assign_attributes(metadata), feature will be in there but not an attribute of job_template.

app/models/job_template_effective_user.rb
```diff
@@ -4,8 +4,6 @@ class JobTemplateEffectiveUser < ActiveRecord::Base
 
   before_validation :set_defaults
 
-  belongs_to :job_template
```
stbenjam: Why?

iNecas: Becuase it was there twice, so I left there just one https://github.com/iNecas/foreman_remote_execution/blob/katello-integration/app/models/job_template_effective_user.rb#L3

app/models/job_template.rb
iNecas: I'm using `except` there https://github.com/theforeman/foreman_remote_execution/pull/149/files#diff-f1f60a704cf99b190843f9a7bc38e0cbR77

iNecas:
Tests are green :)

app/models/job_template.rb
stbenjam: Hm, I see - I'd think this could be done differently so the feature wouldn't need persisting. It'll cascade save and save the job template, but I don't want that done in `import`. (There's a reason for it when I want to write the community-templates plugin, I need import *not* to save, so it can do some diff stuff first.)

Anywho, I filed http://projects.theforeman.org/issues/13746 for that.

stbenjam:
Woo!

PR #150, opened 2016-02-05
fixes #13579 - sync templates during seed

stbenjam:
This syncs job templates during `db:seed` (and bypasses the foreman idempotency check here only), so they're always up to date without needing to create individual migrations.

These sync methods will also be able to be used to add support for syncing job templates from git using the foreman_templates plugin.

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -59,20 +59,29 @@ class JobTemplate < ::Template
   end
   self.table_name = 'templates'
 
-  # Import a template from ERB, with YAML metadata in the first comment
-  def self.import(template, options = {})
-    metadata = parse_metadata(template)
-    return if metadata.blank? || metadata.delete('kind') != 'job_template' || self.find_by_name(metadata['name'])
+  # Import a template from ERB, with YAML metadata in the first comment.  It
+  # will overwrite (sync) an existing template if options[:update] is true.
+  def self.import(contents, options = {})
+    transaction do
+      metadata = parse_metadata(contents)
+      return if metadata.blank? || metadata.delete('kind') != 'job_template'
 
-    inputs = metadata.delete('template_inputs')
+      # Don't update if the template already exists, unless we're told to
+      existing = self.find_by_name(metadata['name'])
+      return if !options.delete(:update) && existing
 
-    template = self.create(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options))
-    template.assign_taxonomies
+      template = existing || self.new
+      template.sync_inputs(metadata.delete('template_inputs'))
+      template.assign_attributes(metadata.merge(:template => contents.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options))
+      template.assign_taxonomies if template.new_record?
 
-    inputs.each do |input|
-      template.template_inputs << TemplateInput.create(input)
+      template
     end
+  end
 
+  def self.import!(template, options = {})
+    template = import(template, options)
+    template.save! if template
     template
   end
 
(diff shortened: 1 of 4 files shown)


app/models/job_template.rb
```diff
@@ -76,6 +79,39 @@ def self.import(template, options = {})
...
+    metadata = JobTemplate.parse_metadata(template)
+    return if metadata.blank? || metadata.delete('kind') != 'job_template'
+
+    new_input_hash = {}
+
+    metadata.delete('template_inputs').each do |input|
+      new_input_hash[input['name']] = input
+    end
+
+    self.assign_attributes(metadata.merge(:template => template.gsub(/<%\#.+?.-?%>\n?/m, '')).merge(options))
```
iNecas: Can we extract this and the part in `import` method into one place? Maybe there are more things that could be shared between `import` and `sync`. The import could be understand just as a sync with empty template.

iNecas:
One thing I've noticed while testing was: when some input gets updated, the order of the input in the template changes: this might cause not very pleasant usability issues. Maybe the fix is introducing some editable priority/ordering param, that would also allow reordering the inputs later. So I'm ok with solving that by different PR

app/models/setting/remote_execution.rb
```diff
@@ -22,7 +22,10 @@ def self.load_defaults
                  'root'),
         self.set('remote_execution_effective_user_method',
                  N_('What command should be used to switch to the effective user. One of %s') % SSHExecutionProvider::EFFECTIVE_USER_METHODS.inspect,
-                 'sudo')
+                 'sudo'),
+        self.set('remote_execution_sync_templates',
+                    N_('Whether we should sync templates from disk when running db:seed.'),
```
adamruzicka: nitpick, wrong indentation

adamruzicka:
@iNecas: created [issue for that](http://projects.theforeman.org/issues/13610)

iNecas:
ACK (pending the indentation fix)

ares:
[test]

ares:
seems like foreman tasks are missing locales dependency, @iNecas will look at that I believe

iNecas:
[test]

iNecas:
The seed tests are failing now http://ci.theforeman.org/job/test_plugin_pull_request/2294/database=mysql,label=fast,ruby=2.0.0/

test/unit/job_template_test.rb
```diff
@@ -109,6 +109,67 @@
...
+      ping -c <%= input("count") %> <%= input("hostname") %>
+      END_TEMPLATE
+    end
+
+    it 'will not overwrite by default' do
+      refute existing.import!(updated)
+    end
+
+    let(:synced_template) do
+      existing.import!(updated, :update => true)
```
iNecas: `undefined method `import!' for #<JobTemplate:0x0055e91020d178>`

stbenjam: @iNecas  Updated.  Local tests won't run on my Katello for some reason, so if you're really anxious to get this merged (seems like you're watching closely?), then if you want to run job_template_test again locally, by all means....

stbenjam: Nevermind, I fixed it, and there's some other issues.  give me a bit, and I'll fix them.

iNecas: I will try to sent you a patch to fix the issues:)

stbenjam: It's fine, I'm fixing it please, it's my pr.

stbenjam:
```
# Running:

.................

Finished in 1.460330s, 11.6412 runs/s, 20.5433 assertions/s.
```

test/unit/job_template_test.rb
iNecas: Well, if your test-suite was not running locally, it would be easier for me to fix it. Anyway, some interesting escaping challenges were there :) TY

iNecas:
Thanks @stbenjam

PR #152, opened 2016-02-09
fixes #13133, #13632 - improve package action template

stbenjam:
This change renames the 'die' function to 'exit_with_message' as some
users found it particularly terrifying to see this as the first line in
the template preview. It also adds documentation to the function.

We can extract this to a helper snippet later if we have more or it
becomes useful elsewhere.

Secondly, it causes the package action template to fail more gracefully
when a host has no operating system, or an unsupported one.

### DIFF
--- a/app/views/templates/package_action.erb
+++ b/app/views/templates/package_action.erb
@@ -26,7 +26,17 @@ template_inputs:
   advanced: true
 %>
 
-die() {
+<%
+  supported_families = ['Redhat', 'Debian']
+  render_error(N_('Unsupported or no operating system found for this host.')) unless @host.operatingsystem && supported_families.include?(@host.operatingsystem.family)
+%>
+
+# Helper function that exits with a particular message and code.
+#
+# Usage:
+#   exit_with_message "Could not do a thing" 2
+#
+function exit_with_message() {
   echo "${1}, exiting..."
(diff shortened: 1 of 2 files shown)


app/views/templates/package_action.erb
```diff
@@ -26,7 +26,17 @@ template_inputs:
...
-die() {
+<%
+  supported_families = ['Redhat', 'Debian']
+  render_error(N_('Unsupported or no operating system found for this host.')) unless @host.operatingsystem && supported_families.include?(@host.operatingsystem.family)
+%>
+
+# Helper function that exits with a particular message and code.
+#
+# Usage:
+#   exit_with_message 2 "Could not do a thing"
```
adamruzicka: Wrong order, first arg should be the message

app/views/templates/package_action.erb
```diff
@@ -26,7 +26,17 @@ template_inputs:
...
+#
+exit_with_message() {
```
adamruzicka: Wouldn't it be even more user friendly to use the `function exit_with_message()` syntax?

adamruzicka:
I inlined two small comments, the rest looks good to me

iNecas:
ACK, tested and works great!

PR #153, opened 2016-02-10
Fixes #13648 - Use humanized_input for task details

adamruzicka:


stbenjam:
Oh, I see, this fixes my problem with localization :smile:

stbenjam:
Tested and works great, thanks so much @adamruzicka 

APJ

adamruzicka:
You're welcome, thanks for reviewing it

PR #154, opened 2016-02-11
Fixes #13666 - Moving deface templates to app/views

daviddavis:
A couple notes:

- I'm following foreman_salt's [example as to where to put the templates](https://git.io/vgK4B) but am happy to move them.
- I didn't test this. Either you can give me some guidance as to how to get to these templates or test this yourself.


stbenjam:
Thanks @daviddavis! Tested and works fine, the only thing I would say though is Salt isolates it's things in the `ForemanSalt` namespace, which remote execution doesn't do.

I wonder if `app/views/overrides` or something similar would be more appropriate, mybe @ares, @adamruzicka or @iNecas has an opinion

iNecas:
`app/views/overrides` would be a bit less surprising

daviddavis:
@stbenjam @iNecas updated

stbenjam:
Thanks, backported to 0.2-stable as well as 460d22bf212ad2ebe32e5548097cdc8a804ed895

stbenjam:
And 0.2.3 released, we'll be good to go for whenever Foreman wants to update Rails.

PR #155, opened 2016-02-12
Fixes #13712 - seed power action templates

iNecas:


app/views/templates/power_action.erb
```diff
@@ -0,0 +1,22 @@
...
+provider_type: SSH
+template_inputs:
+- name: action
+  description: Action to perform on the service
+  input_type: user
+  options: "restart\nshutdown"
+  required: true
+%>
+
+echo <%= input('action') %> host && sleep 3
```
stbenjam: Why a sleep?

iNecas: To give it a time to get the last output from the connection: we need the `echo` message to be received by the proxy (so that we know the disconnection is expected) before we loose the connection (and we can say that the disconnection error was expected, and not fail the whole script)

ares:
Please, create a separate issue for this

iNecas:
Here we go http://projects.theforeman.org/issues/13712

ares:
Thanks, could you please change the commit message to `Fixes`? I think `refs` only links commit but does not close the issue.

ares:
Otherwise :+1:

ares:
commit message links to different issue 13491 -> 13712

iNecas:
Thanks for the catch: an issue between chair and :computer:, updated

PR #156, opened 2016-02-12
Fixes #13689 - don't use authorized scope in provisioning template

iNecas:
The user is not set in context of provisioning, therefore the public
keys of the global proxies were not populated properly into the provisioned
hosts.

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -38,7 +38,7 @@ module ForemanRemoteExecution
       get_interface_by_flag(:execution)
     end
 
-    def remote_execution_proxies(provider)
+    def remote_execution_proxies(provider, authorized = true)
       proxies = {}
       proxies[:subnet]   = execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface && execution_interface.subnet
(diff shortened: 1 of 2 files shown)


stbenjam:
Oh, awesome. This explains a problem a user was having recently, and I could not figure it out.  

ACK

ares:
Shouldn't we pass the user to determine the proxies then? It allows user that has no permissions on a particular proxy to use it.

iNecas:
Setting the user by owner might not be trivial (as the host can be owned by user or usergroup). What about asking for the unscoped proxies in https://github.com/iNecas/foreman_remote_execution/blob/fix-ssh-keys-seed-global-proxy/app/models/concerns/foreman_remote_execution/host_extensions.rb#L60 by passing some optional param to `remote_execution_proxies` method

stbenjam:
That sounds ok to me, I guess.  Although I'm not sure I find it very interesting from a permissions perspective to have such a restriction. I would think the only thing anyone cares about is if a user can execute this on a particular host, not from which proxies it gets executed.

If a subnet has proxies A, B, C, and D, I'd want them all to be used.

iNecas:
I've updated the PR based on our discussion (+ added a test case for this)

ares:
APJ

ares:
Merging, thanks @iNecas

PR #158, opened 2016-02-18
Fixes #13786 - Distinguish between planned and running

adamruzicka:
Old: can't figure out how many tasks are actually running when using concurrency control
old

New:
new


stbenjam:
Looks so much better!

stbenjam:
ACK

adamruzicka:
@stbenjam thanks!

PR #160, opened 2016-02-19
Fixes #13811 -  Refactor RunHostJob and RunProxyCommand to support more providers

dLobatog:
The current code for RunHostJob and RunProxyCommand is effectively just
the code to run SSH. Most of the code can be reused in other providers,
by creating subclasses RunSSHJob and RunProxySSHCommand which only
contain the specific bits for the SSH provider.


dLobatog:
https://github.com/theforeman/foreman_ansible/pull/18 is where I'm using it

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -95,6 +64,23 @@ def verify_permissions(host, template_invocation)
...
         true
       end
+
+      def save_template_invocation(host, template_invocation)
+        template_invocation.host_id = host.id
+        template_invocation.run_host_job_task_id = task.id
+        template_invocation.save!
+      end
+
+      def validate_template_invocation(template_invocation)
```
adamruzicka: It would be nice if methods which can `raise` would have an exclamation mark in their name

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -95,6 +64,23 @@ def verify_permissions(host, template_invocation)
...
+        raise _('Could not use any template used in the job invocation') if template_invocation.blank?
+      end
+
+      def validate_proxy(proxy)
```
adamruzicka: ditto

app/lib/actions/remote_execution/run_ssh_job.rb
```diff
@@ -0,0 +1,36 @@
...
+                    proxy,
+                    hostname,
+                    script(template_invocation, host),
+                    provider.proxy_command_options(template_invocation, host))
+      end
+
+      def find_ip_or_hostname(host)
+        %w(execution primary provision).each do |flag|
+          if host.send("#{flag}_interface") && host.send("#{flag}_interface").ip.present?
+            return host.execution_interface.ip
```
adamruzicka: I believe you wanted to return `host.send("#{flag}_interface").ip`

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -71,6 +77,16 @@ def determine_proxy(template_invocation, host, load_balancer)
 
         proxy
       end
+
+      def provider_job_class(provider)
+        case provider
+        when 'SSH'
+          RunSSHJob
```
dLobatog: This should be extended via plugins too:

```
when 'Ansible' 
  RunAnsibleJob
```

iNecas: What about putting it `here https://github.com/theforeman/foreman_remote_execution/blob/v0.3.0/app/models/remote_execution_provider.rb#L22` to make it provider-specific?

app/lib/actions/remote_execution/run_ssh_job.rb
```diff
@@ -0,0 +1,36 @@
+module Actions
+  module RemoteExecution
+    class RunSSHJob < RunHostJob
```
iNecas: What about putting `RunSSHJob` and `RunProxySSHCommand` into `RemoteExecution::SSH` namespace, so that we have them grouped by provider.

app/lib/actions/remote_execution/run_hosts_job.rb
dLobatog: Sorry, I'm not sure what you mean. This method provides a matching between the `provider_type` attribute of a template and the job class to be triggered. How would I put this matching in the provider class?

dLobatog:
Latest update moved RunSSHJob and RunProxySSHCommand to RemoteExecution::SSH::RunJob and RunProxyCommand

app/lib/actions/remote_execution/run_hosts_job.rb
iNecas: instead of `provider_type` which returns string that you would need to match here, you can use `template.provider` https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_template.rb#L113 so that you get the provider object, that can define the method.

iNecas:
I'm thinking about moving the provider-specific action one level higher, on the `RunHostsJob` level. In this case, the only thing needed would be overriding a method that would provide the `RunHostJob` class. The reason why it might be a better interface is a case, where we would want to rely the execution of the job on set of hosts to external service completely (not using the built-in load-balancing mechanism etc.). We don't need to go this way right now, but this interface would be more flexible to this. I expect most of the `RunHostsJob` would be shared via common parent on the provider actions

dLobatog:
@iNecas Sorry, I don't think I understood the whole thing. You want `RunHostsJob` not to call `RunHostJob` but the provider action? That's what's happening here - it calls `SSH::RunJob` or `Ansible::RunJob`.  `RunHostJob` is the parent of these classes, and contains the non-provider specific stuff. 

If a new provider wanted to call an action that's not a `RunHostJob`, that's still possible now, just make `NewProvider::RunJob` not inherit from `RunHostJob` and that's it. 

Is that what you were looking for?

iNecas:
I was referring more the `RunHostsJob`, so the `RunHostsJob` would be provider specific (and it would then probably call `SSH:RunJob` vs `Ansible::RunJob`. Right now, both implementations would share most of the `RunHostsJob` logic, but if we wanted to rely completely on the technology to perform the execution on a set of hosts (for example with Ansible, it would allow the multi-host orchestration). Is it more clear now?

PR #161, opened 2016-02-22
Fixes #13263 - Distinguish between no proxies found and available, BZ1310687

adamruzicka:


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -24,11 +24,16 @@ def plan(job_invocation, host, template_invocation, proxy)
 
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
 
-        settings = { :global_proxy   => 'remote_execution_global_proxy',
-                     :fallback_proxy => 'remote_execution_fallback_proxy' }
+        if proxy.blank?
+          offline_count = options[:offline_count] || 0
+          raise _('All %{number} usable proxies are down.') % { :number => offline_count } if offline_count > 0
```
stbenjam: Should use `n_()` here to handle the case where it's 1, or more than 1.

adamruzicka: fixed

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -24,11 +24,16 @@ def plan(job_invocation, host, template_invocation, proxy)
 
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
 
-        settings = { :global_proxy   => 'remote_execution_global_proxy',
-                     :fallback_proxy => 'remote_execution_fallback_proxy' }
+        if proxy.blank?
+          offline_count = options[:offline_count] || 0
+          raise N_('All %{number} usable proxies are down.') % { :number => offline_count } if offline_count > 0
```
stbenjam: Ah I meant lower case `n`, it handles plurals, so you can give:

`n("message when there's one", "message when > 1", count)`

Here's an example:

https://github.com/theforeman/foreman/blob/develop/app/helpers/layout_helper.rb#L485

Otherwise the way it was would render "All 1 usable proxies are down" which isn't correct English (and probably not correct in a bunch of languages).  It should read either "The only usable proxy is down" or "All X usuable proxies are down."

adamruzicka: Oh, I thought that was a typo, sorry. Fixed again

adamruzicka: Oh, sorry, thought that was a typo. Should it be like this? Can't test it right now

stbenjam: Yup, looks good - although better to still use the hash interpolation instead of `%s` to give translators context.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -24,11 +24,18 @@ def plan(job_invocation, host, template_invocation, proxy)
 
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
 
-        settings = { :global_proxy   => 'remote_execution_global_proxy',
-                     :fallback_proxy => 'remote_execution_fallback_proxy' }
+        if proxy.blank?
+          offline_count = options[:offline_count] || 0
+          raise n_('The only usable proxy is down',
+                   'All %{number} usable proxies are down.',
+                   offline_count) % { :number => offline_count } if offline_count > 0
```
stbenjam: So, I was thinking about this, and I think we might make the error message more helpful if we told the user which proxies we tried to contact.  I also don't really like 'usable' either, maybe 'applicable' is better?

What do you think about this? "No applicable proxy could be contacted.  The follow proxies are down: %{offline}"

adamruzicka: And %{offline} would be a comma-separated list of proxy names? It would become a mess if someone had a lot of proxies, but generally I'm not opposed to providing some insight into what's going on under the hood.

Yeah, you're right 'No applicable proxy' sounds better

stbenjam: Indeed.  I think the error only shows up in the task output, right? So I don't think it's such a big problem if it's huge, it's not being shown in a flash or something small.

But, you could limit it to showing the first 5, and then say something like "and 17 more."

adamruzicka: It didn't look as bad as I expected with many failed proxies, so I didn't put the "and N more." thing there.

stbenjam:
Tested and works great, ACK

adamruzicka:
@stbenjam Thanks for testing!

PR #162, opened 2016-02-22
Fixes #13310 - Provide deep_clone and deep_clone!

adamruzicka:

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -93,13 +93,16 @@ class JobInvocation < ActiveRecord::Base
 
   def deep_clone
     JobInvocationComposer.from_job_invocation(self).job_invocation.tap do |invocation|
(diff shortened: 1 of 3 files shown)


app/models/job_invocation.rb
```diff
@@ -93,12 +93,18 @@ def queued?
...
       invocation.description_format = self.description_format
       invocation.description = self.description
       invocation.pattern_template_invocations = self.pattern_template_invocations.map(&:deep_clone)
-      invocation.save!
+    end
+  end
+
+  def deep_clone!
+    deep_clone.tap do |clone|
+      clone.task_group.save!
```
stbenjam: Wouldn't rails already automatically save this association?

adamruzicka: Oh, it would. Good point

stbenjam:
Looks good to me, ACK

ares:
Thanks @adamruzicka, merging!

PR #163, opened 2016-02-22
Fixes #13839 - Show provider humanized name on job templates form

dLobatog:
Currently it shows the provider class name, which the user doesn't need
to know about.

Before:
screenshot from 2016-02-22 21-26-14

After:
screenshot from 2016-02-22 21-25-40


stbenjam:
Thanks, this looks a lot nicer.

PR #164, opened 2016-02-22
Fixes #12105 - Parse ANSI terminal codes into colorful HTML

dLobatog:
There are client-side libraries that can do this, and I'd argue that's
the way it should be done. However, given the state of affairs with
gemified assets (we don't know if we're going to move to npm or
something like that) and the fact I already had this code ready, this
could be a temporary solution to display nicer console output.

Taken from https://github.com/theforeman/foreman_remote_execution/pull/134/files

### DIFF
--- a/app/helpers/job_invocation_output_helper.rb
+++ b/app/helpers/job_invocation_output_helper.rb
@@ -1,0 +1,42 @@
+module JobInvocationOutputHelper
+  CONSOLE_COLOR = {
+      '31' => 'red',
+      '32' => 'lightgreen',
+      '33' => 'orange',
+      '34' => 'deepskyblue',
+      '35' => 'mediumpurple',
+      '36' => 'cyan',
+      '37' => 'grey',
+      '91' => 'red',
+      '92' => 'lightgreen',
+      '93' => 'yellow',
+      '94' => 'lightblue',
+      '95' => 'violet',
+      '96' => 'turquoise',
+      '0'  => 'default',
+    }
+  CONSOLE_COLOR.default = 'default'
+
+  def colorize_line(line)
+    line = line.gsub(/\e\[.*?m/) do |seq|
+      color = seq[/(\d+)m/,1]
+      "{{{format color:#{color}}}}"
+    end
+
+    current_color = 'default'
+    out = %{<span style="color: #{@current_color}">}
+    parts = line.split(/({{{format.*?}}})/)
+    parts.each do |console_line|
+      if console_line.include?('{{{format')
+        if (color_index = console_line[/color:(\d+)/, 1]).present?
+          current_color = CONSOLE_COLOR[color_index]
+          out << %{</span><span style="color: #{current_color}">}
+        end
+      else
+        out << h(console_line)
+      end
+    end
+    out << %{</span>}
+    out
+  end
+end
--- a/app/views/template_invocations/_output_line_set.html.erb
+++ b/app/views/template_invocations/_output_line_set.html.erb
@@ -2,6 +2,6 @@
   <%= content_tag :div, :class => 'line ' + output_line_set['output_type'], :data => { :timestamp => output_line_set['timestamp'] } do %>
 
     <%= content_tag(:span, (@line_counter += 1).to_s.rjust(4).gsub(' ', '&nbsp;').html_safe + ':', :class => 'counter', :title => (output_line_set['timestamp'] && Time.at(output_line_set['timestamp']))) %>
-    <%= content_tag(:div, (line.empty? ? '&nbsp;' : h(line)).html_safe, :class => 'content') %>
+    <%= content_tag(:div, (line.empty? ? '&nbsp;' : colorize_line(line)).html_safe, :class => 'content') %>
   <% end %>
 <% end %>


stbenjam:
A couple thoughts:

- For this to work with SSH, we need to set TERM in the SSH session to support ANSI colors. I would guess we need to do something like add  :term => 'xterm-color' or similar here: https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh/connector.rb#L25, but I couldn't get it to work

- This is too small a subset of colors, I'd prefer some library to render the full set of ANSI codes - including bright, underline, etc.   A library would probably be better

dLobatog:
@stbenjam True that, we'd need to tweak that for SSH. I originally had the problem with Ansible output looking terrible but now that you mention it I didn't try with SSH. 

Regarding the small sample of colors, I meant it to be just a temporary solution to avoid pulling in more gems for JS stuff (we're working on switching from gemified assets to npm/bower/something). 

Let me know if that's fine and if not I'll move it to foreman_ansible instead, and submit an actual fix with js whenever we figure out how to add js assets.

stbenjam:
I would definitely like it in the plugin, but I think if it's here we need to make it work with the default SSH provider.  Maybe @iNecas has an idea how to do that since he wrote the SSH connector (see https://github.com/theforeman/foreman_remote_execution/pull/164#issuecomment-187774251).

hammer also needs some logic to make sure when displaying the output the ANSI codes are filtered out on unsupported terminals, but I guess that's probably not a huge deal.

iNecas:
@stbenjam I will look into the ssh part - we're doing some filtering on the proxy side (we had some issues with non-utf chars previously)

ares:
@iNecas what's the status here?

iNecas:
@ares not progress here

ares:
@iNecas would it make sense to take this as a temporary solution then? It is available for nearly a year. I applied the patch manually and it still works well.

dLobatog:
@ares I've rebased the PR in case you still want to add it

ares:
@stbenjam thanks! I think the test failure is unrelated, @iNecas if you agree, I'll merge it

app/views/template_invocations/_output_line_set.html.erb
```diff
@@ -2,6 +2,6 @@
   <%= content_tag :div, :class => 'line ' + output_line_set['output_type'], :data => { :timestamp => output_line_set['timestamp'] } do %>
 
     <%= content_tag(:span, (@line_counter += 1).to_s.rjust(4).gsub(' ', '&nbsp;').html_safe + ':', :class => 'counter', :title => (output_line_set['timestamp'] && Time.at(output_line_set['timestamp']))) %>
-    <%= content_tag(:div, (line.empty? ? '&nbsp;' : h(line)).html_safe, :class => 'content') %>
+    <%= content_tag(:div, (line.empty? ? '&nbsp;' : h(colorize_line(line))).html_safe, :class => 'content') %>
```
iNecas: It should be colorize_line(h(line)), otherwise you would see 
rex-ansi

iNecas:
Still haven't seen this actually working

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+      if line.include?('{{{format')
+        if color_index = line[/color:(\d+)/, 1]
+          @current_color = case color_index
+                           when '31'
+                             'red'
+                           when '32'
+                             'green'
+                           when '33'
+                             'orange'
+                           when '0'
```
iNecas: When testing with `ls -al --color=always`, the lightblue for code 34 looked very nice and I suggest to include that as well
res-ansi-2

adamruzicka:
@dLobatog what's the status here?

dLobatog:
screenshot from 2017-12-27 09-36-23


cc @adamruzicka

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+                           when '31'
+                             'red'
+                           when '32'
+                             'green'
+                           when '33'
+                             'orange'
+                           when '34'
+                             'lightblue'
+                           when '0'
+                             'default'
```
adamruzicka: Is the `when '0'` really necessary?

ohadlevy:
btw: if you want to change the colors on the browser instead of in rails, you can use something like https://github.com/rburns/ansi-to-html

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+  def colorize_line(line)
+    line = line.gsub(/\e\[.*?m/) { |seq| color = seq[/(\d+)m/,1]; "{{{format color:#{color}}}}" }
+
+    @current_color ||= 'default'
+    out = %{<span style="color: #{@current_color}">}
+    parts = line.split(/({{{format.*?}}})/)
+    parts.each do |line|
+      if line.include?('{{{format')
+        if color_index = line[/color:(\d+)/, 1]
+          @current_color = case color_index
```
adamruzicka: How about putting the mapping into a deeply frozen hash constant and just accessing it here to split it up a bit?

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+    out = %{<span style="color: #{@current_color}">}
+    parts = line.split(/({{{format.*?}}})/)
+    parts.each do |console_line|
+      if console_line.include?('{{{format')
+        if (color_index = console_line[/color:(\d+)/, 1]).present?
+          @current_color = console_color[color_index]
+          out << %{</span><span style="color: #{@current_color}">}
+        end
+      else
+        out << console_line
```
iNecas: missing `h` escaping: risk of XSS

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,36 @@
+module JobInvocationOutputHelper
+  def console_color
+    return @console_color if @console_color.present?
+    @console_color = {
```
iNecas: I would recommend defining them as CONSTANT

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+  CONSOLE_COLOR.default = 'default'
+
+  def colorize_line(line)
+    binding.pry
+    line = line.gsub(/\e\[.*?m/) do |seq|
+      color = seq[/(\d+)m/,1]
+      "{{{format color:#{color}}}}"
+    end
+
+    current_color = 'default'
```
dLobatog: @adamruzicka I think this is the part that's causing the problem you're seeing where our own output is colorized. The real issue though, is that at this point we have no way of discriminating where the `line` is coming from. From `foreman_remote_execution` POV, it's all `continuous_output`. 

I made `current_color` NOT take the previous value (so after a green line another would follow if there's no color specified) - now all lines are `default` unless a color is specified.

https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L106

https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks_core/continuous_output.rb

app/helpers/job_invocation_output_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+    }
+  CONSOLE_COLOR.default = 'default'
+
+  def colorize_line(line)
+    line = line.gsub(/\e\[.*?m/) do |seq|
+      color = seq[/(\d+)m/,1]
+      "{{{format color:#{color}}}}"
+    end
+
+    current_color = 'default'
```
dLobatog: @adamruzicka I think this is the part that's causing the problem you're seeing where our own output is colorized. The real issue though, is that at this point we have no way of discriminating where the `line` is coming from. From `foreman_remote_execution` POV, it's all `continuous_output`. 

I made `current_color` NOT take the previous value (so after a green line another would follow if there's no color specified) - now all lines are `default` unless a color is specified.

https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L106

https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks_core/continuous_output.rb

ares:
I think both reviews were satisfied, I'd kindly ask for more color codes since commands like ls uses blue, cyan etc. I sent a PR - https://github.com/dLobatog/foreman_remote_execution/pull/2 with such codes and css colors mapping, I'd merge if you accept.

ohadlevy:
for future work, it would be interesting to try and use the xterm react
component https://github.com/patternfly/patternfly-react/pull/160 once its
merged.

PR #165, opened 2016-02-26
fixes #13922 - import/export job templates with metadata

stbenjam:
This lets you export the template in our import format (metadata in a YAML comment).  The idea here is to facilitate easier sharing / creation of new templates, moving us a step closer to creating a vibrant community of job templates in the community-templates repo.

### DIFF
--- a/app/controllers/api/v2/job_templates_controller.rb
+++ b/app/controllers/api/v2/job_templates_controller.rb
@@ -7,7 +7,7 @@ module Api
       include ::Foreman::Controller::ProvisioningTemplates
 
       before_filter :find_optional_nested_object
-      before_filter :find_resource, :only => %w{show update destroy clone}
+      before_filter :find_resource, :only => %w{show update destroy clone export}
 
       before_filter :handle_template_upload, :only => [:create, :update]
 
@@ -20,6 +20,23 @@ module Api
       param_group :search_and_pagination, ::Api::V2::BaseController
       def index
         @job_templates = resource_scope_for_index
+      end
+
+      api :POST, '/job_templates/import', N_('Import a job template from ERB')
+      param :template, String, :required => true, :desc => N_('Template ERB')
+      param :overwrite, :bool, :required => false, :desc => N_('Overwrite template if it already exists')
+      def import
+        options = params[:overwrite] ? { :update => true } : { :build_new => true }
+
+        @job_template = JobTemplate.import(params[:template], options)
+        @job_template ||= JobTemplate.new
+        process_response @job_template.save
+      end
+
+      api :GET, '/job_templates/:id/export', N_('Export a job template to ERB')
+      param :id, :identifier, :required => true
+      def export
+        send_data @job_template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @job_template.filename
       end
 
       api :GET, '/job_templates/:id', N_('Show job template details')
@@ -110,6 +127,10 @@ module Api
         case params[:action]
         when 'clone'
           :create
+        when 'import'
+          :create
+        when 'export'
+          :view
         else
           super
         end
--- a/app/controllers/job_templates_controller.rb
+++ b/app/controllers/job_templates_controller.rb
@@ -25,7 +25,7 @@ class JobTemplatesController < ::TemplatesController
   end
 
   def import
-    contents = params[:imported_template][:template].respond_to?(:read) ? params[:imported_template][:template].read : nil
+    contents = params.fetch(:imported_template, {}).fetch(:template, nil).try(:read)
 
     @template = JobTemplate.import(contents, :update => Foreman::Cast.to_bool(params[:imported_template][:overwrite]))
     if @template && @template.save
@@ -56,9 +56,7 @@ class JobTemplatesController < ::TemplatesController
 
   def action_permission
     case params[:action]
-      when 'auto_complete_job_category'
-        :view_job_templates
-      when 'export'
+      when 'auto_complete_job_category', 'export'
         :view_job_templates
       else
         super
--- a/app/controllers/job_templates_controller.rb
+++ b/app/controllers/job_templates_controller.rb
@@ -24,6 +24,26 @@ class JobTemplatesController < ::TemplatesController
     end
   end
 
+  def import
+    contents = params[:imported_template][:template].respond_to?(:read) ? params[:imported_template][:template].read : nil
+
+    @template = JobTemplate.import(contents, :update => Foreman::Cast.to_bool(params[:imported_template][:overwrite]))
+    if @template && @template.save
+      flash[:notice] = _('Job template imported successfully.')
+      redirect_to job_templates_path(:search => "name = \"#{@template.name}\"")
+    else
+      @template ||= JobTemplate.import(contents, :build_new => true)
+      @template.valid?
+      flash[:warning] = _('Unable to save template. Correct highlighted errors')
+      render :action => 'new'
+    end
+  end
+
+  def export
+    find_resource unless @template.present?
+    send_data @template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @template.filename
+  end
+
   private
 
   def find_resource
@@ -37,6 +57,8 @@ class JobTemplatesController < ::TemplatesController
   def action_permission
     case params[:action]
       when 'auto_complete_job_category'
+        :view_job_templates
+      when 'export'
         :view_job_templates
       else
         super
--- a/app/models/concerns/foreman_remote_execution/exportable.rb
+++ b/app/models/concerns/foreman_remote_execution/exportable.rb
@@ -16,12 +16,11 @@ module ForemanRemoteExecution
       self.class.exportable_attributes.keys.inject({}) do |hash, attribute|
         value = export_attr(attribute, self.class.exportable_attributes[attribute], include_blank)
 
-        if include_blank
+        # Rails considers false blank, but if a boolean value is explicitly set false, we want to ensure we export it.
+        if include_blank || value.present? || value == false
           hash.update(attribute => value)
         else
-          # Rails considers false blank, but if a boolean value is explicitly
-          # set false, we want to ensure we export it.
-          (value.blank? && value != false) ? hash : hash.update(attribute => value)
+          hash
         end
       end.stringify_keys
     end
--- a/app/models/concerns/foreman_remote_execution/exportable.rb
+++ b/app/models/concerns/foreman_remote_execution/exportable.rb
@@ -12,12 +12,17 @@ module ForemanRemoteExecution
   module Exportable
     extend ActiveSupport::Concern
 
-    def to_export
+    def to_export(include_blank = true)
       self.class.exportable_attributes.keys.inject({}) do |hash, attribute|
-        value = export_attr(attribute, self.class.exportable_attributes[attribute])
+        value = export_attr(attribute, self.class.exportable_attributes[attribute], include_blank)
 
-        # Rails considers false blank, but if a boolean value is explicitly set false, we want to ensure we export it.
-        (value.blank? && value != false) ? hash : hash.update(attribute => value)
+        if include_blank
+          hash.update(attribute => value)
+        else
+          # Rails considers false blank, but if a boolean value is explicitly
+          # set false, we want to ensure we export it.
+          (value.blank? && value != false) ? hash : hash.update(attribute => value)
+        end
       end.stringify_keys
     end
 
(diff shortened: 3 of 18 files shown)


stbenjam:
For some reason, Foreman's spinner doesn't go away after downloading the Export from the job templates page.  It must not realize the transaction completed when downloading w/ `send_data`?

app/models/concerns/foreman_remote_execution/exportable.rb
```diff
@@ -0,0 +1,54 @@
...
+# attributes and associations only.  If a specified assocation also includes
+# this concern, then it will likewise be exported. That behavior can be
+# overriden by specifying a custom export lambda.
+#
+# Example:
+#   attr_exportable :name, :address, :company => ->(company) { company.name }
+#
+module ForemanRemoteExecution
+  module Exportable
+    extend ActiveSupport::Concern
```
stbenjam: I didn't find a good way to do the kind of selective and recursive exporting of all the associations on JobTemplates, so I made this. The idea here is that you mark say, `template_inputs` as exportable from `JobTemplate` and because `TemplateInput` itself is also exportable, it exports the selected attributes from it too.  Maybe this is crazy :hankey:, it's late.

app/models/concerns/foreman_remote_execution/exportable.rb
```diff
@@ -0,0 +1,54 @@
...
+# attributes and associations only.  If a specified assocation also includes
+# this concern, then it will likewise be exported. Custom attributes can be
+# specified as an options hast consisting of a custom export lambda.
+#
+# Example:
+#   attr_exportable :name, :address, :company => ->(user) { user.company.name }
+#
+module ForemanRemoteExecution
+  module Exportable
+    extend ActiveSupport::Concern
```
stbenjam: I didn't find a good way to do the kind of selective and recursive exporting of all the associations on JobTemplates, so I made this. The idea here is that you mark say, `template_inputs` as exportable from `JobTemplate` and because `TemplateInput` itself is also exportable, it exports the selected attributes from it too.  Maybe this is crazy :hankey:, it's late.

stbenjam:
[test]

app/models/concerns/foreman_remote_execution/exportable.rb
iNecas: I have no issues with this. Would there be any use case in other parts of Foreman for this? (I've got a feeling that there would be, although, we are using mainly the `rabl` to do this job)

app/models/concerns/foreman_remote_execution/exportable.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    module ClassMethods
+      attr_reader :exportable_attributes, :custom_exports
+
+      # Takes an array of exportable attributes, or a custom exports hash.  The
+      # custom exports hash should be a key/proc pair used to export the
+      # particular attribute.
+      def attr_exportable(*args)
+        @custom_exports = args.last.is_a?(Hash) ? args.pop : {}
+        @exportable_attributes ||= []
```
iNecas: What about using just `:exporteble_attributes`, as hash with symbols when referring to a method and lambda when using the custom one (attr_exportable inteface would stay the same), but the data structure would look like this

```
{ :name => :name
  :company => -> (user) { user.company.name } }
```

It would read a bit better (at least for me)

app/models/concerns/foreman_remote_execution/exportable.rb
stbenjam: >  Would there be any use case in other parts of Foreman for this?

It's reusable for sure, but I can't think of any other uses in core at the moment.  

I also tried to find a gem like you mentioned or something that does this, and couldn't. Rails has `as_json` which can take some options but it didn't handle everything I needed.

> We are using mainly the rabl to do this job

Hmm, that's an interesting idea.  Not RABL, but  we could use a view to render the job template export, which might be simpler than this Exportable concern.


```erb
<%%\#
kind: job_template
name: <%= @template.name %>
template_inputs: <%= template_inputs.each do |input|
  <% render :partial => 'template_inputs'
<% end %>

etc.

%%>
```

iNecas: I'm not sure the rendering to yaml would work nicely due to white-space issues: I know we do that for could-init templates) but in this case I probably prefer the `to_export` approach

app/models/concerns/foreman_remote_execution/exportable.rb
stbenjam: Yea that makes sense, updated to only use one hash

app/controllers/job_templates_controller.rb
```diff
@@ -24,6 +24,26 @@ def preview
     end
   end
 
+  def import
+    contents = params[:imported_template][:template].respond_to?(:read) ? params[:imported_template][:template].read : nil
```
ares: maybe a bit safer and more readable way
`contents = params.fetch(:imported_template, {}).fetch(:template, nil).try(:ready)`

app/controllers/job_templates_controller.rb
```diff
@@ -24,6 +24,26 @@ def preview
...
+    else
+      @template ||= JobTemplate.import(contents, :build_new => true)
+      @template.valid?
+      flash[:warning] = _('Unable to save template. Correct highlighted errors')
+      render :action => 'new'
+    end
+  end
+
+  def export
+    find_resource unless @template.present?
```
ares: seems to already be handled by https://github.com/theforeman/foreman_remote_execution/pull/165/files#diff-227ae74b586f585ebb653d96ff2ee7e6R10 is there a reason to try to load second time here if it didn't succeed?

app/controllers/job_templates_controller.rb
```diff
@@ -38,6 +58,8 @@ def action_permission
     case params[:action]
       when 'auto_complete_job_category'
         :view_job_templates
+      when 'export'
+        :view_job_templates
```
ares: nitpick, could you please deduplicate by

```ruby
when 'auto_complete_job_category', 'export'
  :view_job_templates
else
```

app/models/concerns/foreman_remote_execution/exportable.rb
ares: we do have something similar in core https://github.com/theforeman/foreman/blob/94508562b8c10b8b53dcb0badaa8abe31887d059/app/models/concerns/enc_output.rb

I'm not saying we must reuse it, just something to consider, we could improve core code to be more generic

app/models/job_template.rb
```diff
@@ -180,7 +197,6 @@ def sync_foreign_input_sets(input_sets)
   def sync_feature(feature_name)
     if feature_name && (feature = RemoteExecutionFeature.feature(feature_name))
       feature.job_template ||= self
-      feature.save!
```
ares: does this get saved now with template because it's unsaved association?

app/views/job_templates/_import_job_template_modal.html.erb
```diff
@@ -0,0 +1,18 @@
+<!-- modal window -->
+<div class="modal fade" id="importJobTemplateModal" role="dialog" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" onclick="close_import_job_template_modal(); return false;"><span aria-hidden="true">&times;</span><span class="sr-only"><%= _('Close') %></span></button>
+        <h4 class="modal-title">Import Job Template</h4>
+      </div>
+      <div class="modal-body" style="overflow-y: auto; max-height: 500px;">
```
ares: `style="overflow-y: auto; max-height: 500px;"` gets duplicated with app/views/job_invocations/_preview_hosts_modal.html.erb maybe we could extract it to external CSS file?

app/controllers/api/v2/job_templates_controller.rb
```diff
@@ -22,6 +22,23 @@ def index
...
+
+        @job_template = JobTemplate.import(params[:template], options)
+        @job_template ||= JobTemplate.new
+        process_response @job_template.save
+      end
+
+      api :GET, '/job_templates/:id/export', N_('Export a job template to ERB')
+      param :id, :identifier, :required => true
+      def export
+        send_data @job_template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @job_template.filename
```
ares: `@template.filename + '.txt'` helps my browser to get "a.erb.txt" instead of "a.bin" with unknown extension, not entirely sure why, headers seems to be correct in both cases

ares:
I've tested it and it works just great! :green_heart:

app/models/job_template.rb
stbenjam: Yes, the `save!` should've never been here, it causes a cascade save forcing the template to be saved.

ares: makes sense, thanks

app/controllers/api/v2/job_templates_controller.rb
stbenjam: Which browser? Firefox handled it correctly for me

ares: chrome 49.0.2623.87 (64-bit), probably not a big deal then

app/controllers/job_templates_controller.rb
stbenjam: That's the API controller.  For some reason, the methods in the plain controller do find_resource in the first method call (see preview).  I assume because this one is inheriting from Foreman's templates controller and we didn't want to accidentally override any of their filters?

app/controllers/api/v2/job_templates_controller.rb
stbenjam: Just tried on 49.0.2623.110-1 on Fedora 23, and it got the right filename. Maybe it was a bug in that particular version of Chrome?

stbenjam:
[test] Transient

app/models/concerns/foreman_remote_execution/exportable.rb
stbenjam: http://projects.theforeman.org/issues/14509

stbenjam: And https://github.com/theforeman/foreman/pull/3399

app/controllers/api/v2/job_templates_controller.rb
ares: not a big deal anyway, dismiss

app/controllers/job_templates_controller.rb
ares: ah right, the reason is even documented https://github.com/theforeman/foreman/blob/develop/app/controllers/templates_controller.rb#L82

sorry, dimiss

app/models/concerns/foreman_remote_execution/exportable.rb
ares: if we want to stay compatible with Foreman 1.11 we have to define it here anyway right?

stbenjam: Yes, although 1.12 is not so far away.  Did we decide on what our commitment is supporting older Foreman versions now that we're out of the super active development phase?  I have another PR that is 1.12-only (the new button helpers).

stbenjam: This will need to be included here to avoid a cherry-picking nightmare for Satellite, as it's been approved for inclusion in Sat 6.2.  And it's just a nice feature, so we can do at least one more release for 1.11 that has this.

There should be no conflict as our concerns are namespaced if the foreman PR gets merged into 1.12. We can remove the concern from us when we drop 1.11 support.  Sound OK?

stbenjam:
Updated

app/models/concerns/foreman_remote_execution/exportable.rb
```diff
@@ -12,37 +12,42 @@ module ForemanRemoteExecution
...
 
-    def to_export
+    def to_export(include_blank = true)
       self.class.exportable_attributes.keys.inject({}) do |hash, attribute|
-        value = export_attr(attribute, self.class.exportable_attributes[attribute])
+        value = export_attr(attribute, self.class.exportable_attributes[attribute], include_blank)
 
-        # Rails considers false blank, but if a boolean value is explicitly set false, we want to ensure we export it.
-        (value.blank? && value != false) ? hash : hash.update(attribute => value)
+        if include_blank
```
ares: I think this whole if is equivalent to

```ruby
hash.update(attribute => value) if include_blank || value.present? || value == false
#or
hash.update(attribute => value) if include_blank || value.present? || value.is_a?(FalseClass)
```

but note that `'    '.present? # => false` as well as `'   '.blank? # => true` which I'm not sure is a problem

ares:
Except last nit I think it's ready to merge, could you please squash? Somehow I don't see the squash option on this PR, also I think it would use just last commit for commit message.

app/models/concerns/foreman_remote_execution/exportable.rb
stbenjam: Not quite, you have to return the hash in the inject, but the conditional is cleaner, i'll use it thanks

stbenjam:
> Could you please squash? Somehow I don't see the squash option on this PR, also I think it would use just last commit for commit message.

It's not really obvious :-1:  But it shows up as an option when you press the green button, and you can adjust the commit title and message.

squash

app/models/concerns/foreman_remote_execution/exportable.rb
ares: yeah you could either wrap it in #tap block or just return hash just after the conditional update... anyway whatever works for you... it would be great if you could update the same in Foreman PR, anyway that does not block this PR

ares:
yeah, I just found it, thanks... I'll wait for jenkins to confirm and then I'll try squashing via github

stbenjam:
Cool, thanks! I also updated the foreman PR.

ares:
Thanks @stbenjam, I just merged great patch!

PR #166, opened 2016-03-03
Fixes #14015 - do not raise exception on invocation creation

ares:


ares:
For some reason, it seems that with this patch, it does not schedule the invocation.

adamruzicka:
Looks good to me, ACK

ares:
Thanks (and btw my previous comment was fixed already) :-) Merging.

PR #168, opened 2016-03-31
[Do not merge] Fixes #14419 - Save recalculated global status

ares:


ares:
it might got fixed without this patch if `refresh_statuses` gets update - https://github.com/theforeman/foreman/pull/3380#issuecomment-203953246

ares:
won't be needed, closing

PR #170, opened 2016-04-06
fixes #14506 - use new_link helpers, and use button css

stbenjam:
Button links are broken in a few places due to https://groups.google.com/forum/#!topic/foreman-dev/QB9ZeK_m0Qw.

Foreman now provides a new_link helper for creating new objects. We need to use
that on the Job Templates and Invocations pages.  Additionally, other places
like Host "Run Job", we need to explicitly set the btn btn-default CSS.

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -96,7 +96,7 @@ module RemoteExecutionHelper
   def job_invocations_buttons
     [
       documentation_button_rex('3.2ExecutingaJob'),
-      display_link_if_authorized(_('Run Job'), hash_for_new_job_invocation_path)
+      new_link(_('Run Job'))
     ]
   end
 
--- a/app/views/job_templates/index.html.erb
+++ b/app/views/job_templates/index.html.erb
@@ -6,8 +6,7 @@
 <% title _("Job Templates") %>
 <% title_actions(documentation_button_rex('3.1JobTemplates'),
                  link_to_function(_('Import'), 'show_import_job_template_modal();', :class => 'btn btn-default'),
-                 display_link_if_authorized(_("New Job Template"),
-                                            hash_for_new_job_template_path)) %>
+                 new_link(_("New Job Template"))) %>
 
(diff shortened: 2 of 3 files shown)


stbenjam:
Maybe we should do this in a portable way instead - like add the CSS instead of using the new_link helper?

Or will we begin dropping support for older Foremans faster?

iNecas:
No need for css hacks, if the new_link is the recommended way for doing so. I would suggest still to wait few weeks before dropping the 1.11, let's say until branch warning landing to foreman-dev?

iNecas:
Thanks for the fix @stbenjam !

app/views/job_templates/index.html.erb
```diff
@@ -4,8 +4,7 @@
 
 <% title _("Job Templates") %>
 <% title_actions(documentation_button_rex('3.1JobTemplates'),
-                 display_link_if_authorized(_("New Job Template"),
-                                            hash_for_new_job_template_path)) %>
+                 new_link(_("New Job Template"), hash_for_new_job_template_path)) %>
```
tbrisker: no need to pass the hash here, it get automatically calculated according to the controller

app/helpers/remote_execution_helper.rb
```diff
@@ -96,7 +96,7 @@ def remote_execution_provider_for(template_invocation)
   def job_invocations_buttons
     [
       documentation_button_rex('3.2ExecutingaJob'),
-      display_link_if_authorized(_('Run Job'), hash_for_new_job_invocation_path)
+      new_link(_('Run Job'), hash_for_new_job_invocation_path)
```
tbrisker: no need to pass the hash here, it get automatically calculated according to the controller

tbrisker: Or pass `{:controller => whatever}` if override is needed

stbenjam:
Fixed, thanks @tbrisker

stbenjam:
The commit message is fine, @theforeman-bot is referencing an old commit message for some reason.

ares:
Thanks @stbenjam, merging!

PR #172, opened 2016-04-19
Fixes #14684 - Save end time for recurring

adamruzicka:
- [x] Requires https://github.com/theforeman/foreman-tasks/pull/183

### DIFF
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -62,6 +62,7 @@ class JobInvocationComposer
       return {} unless ui_params.key?(:triggering)
       trig = ui_params[:triggering]
       keys = (1..5).map { |i| "end_time(#{i}i)" }
+      return trig unless trig.key?(:end_time) && trig[:end_time].keys == keys
       trig.merge(:end_time => Time.local(*trig[:end_time].values_at(*keys)))
     end
   end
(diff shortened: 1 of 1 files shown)


ares:
It does not seem to work for me, when I set end time, the action results in 200 but it redirects me back to the invocation form without any error. Also it creates the job, it creates recurring action but it does not create the task so the job is queued forever. I don't see any error in the log.

ares:
After short debugging session we realized that if there's no window for job to start, task is not created causing all of this problems. So a validation should be added that checks that $start_time < $end_time

adamruzicka:
[test]

adamruzicka:
Wasn't there a specific need for doing it this way? Something like it was needed to force validation of all of the template invocations (which the all? approach wouldn't do)?

ares:
ah right, `.all?` is lazy :bulb: thanks!

adamruzicka:
Great, glad we didn't miss it

iNecas:
Foreman-tasks 0.7.17 with required change was just released

adamruzicka:
Let's see how the [test] goes

adamruzicka:
[test]

adamruzicka:
[test]

ares:
Works fine, thanks @adamruzicka, merging!

ares:
oops I left double comment in commit message, sorry :-)

PR #173, opened 2016-04-21
WIP Fixing tests

ares:
This should be compatible with Foreman 1.11 as well, so I'm not setting 1.12 flag.

PR #177, opened 2016-04-27
Refs #14842 - 'validate' should be 'validates'

dLobatog:
On remote_execution_feature, there's a line calling `validate` and then
calls some Rails validations. It has a typo and should say `validates`.
Passenger does not start because of this, with an error

' Message from application: Unknown key: :presence. Valid keys are: :on,
  :if, :unless, :prepend. Perhaps you meant to call `validates` instead
  of `validate`? (ArgumentError) '

Fix is as simple as correcting the typo.

### DIFF
--- a/test/unit/remote_execution_feature_test.rb
+++ b/test/unit/remote_execution_feature_test.rb
@@ -1,6 +1,10 @@
 require 'test_plugin_helper'
 
 describe RemoteExecutionFeature do
+  should validate_presence_of(:name)
+  should validate_presence_of(:label)
+  should validate_uniqueness_of(:name)
+  should validate_uniqueness_of(:label)
 
   let(:install_feature) do
     RemoteExecutionFeature.register(:katello_install_package, N_('Katello: Install package'),


test/test_plugin_helper.rb
```diff
@@ -1,7 +1,7 @@
 # This calls the main test_helper in Foreman-core
 require 'test_helper'
-
 require 'dynflow/testing'
+require 'database_cleaner'
```
dLobatog: I also took the ocassion to fix this. If you're running tests individually (not with `rake test`), these tests will not work  because of DatabaseCleaner in line 11. This is only loaded on [integration tests](https://github.com/theforeman/foreman/blob/develop/test/integration_test_helper.rb#L9) now in core , so we have to require it manually.

app/models/remote_execution_feature.rb
```diff
@@ -1,7 +1,7 @@
 class RemoteExecutionFeature < ActiveRecord::Base
   attr_accessible :label, :name, :provided_input_names, :description, :job_template_id
 
-  validate :label, :name, :presence => true, :unique => true
+  validates :label, :name, :presence => true, :uniqueness => true
```
ares: thanks, this is already addressed in a PR fixing tests and 4.2 compatibility

dLobatog: :disappointed_relieved: should I close this one then?

ares: no, I think the database_cleaner is still valuable patch, just remove the `validates` change I'll use the ticket number it that combined PR

dLobatog: Aight, I'll keep the unit tests though, as I don't see it anywhere else (they would've prevented this)

stbenjam:
My commit had the database cleaner patch, it's in https://github.com/theforeman/foreman_remote_execution/pull/173/files#diff-b7be47364acf6629bcdcf5d0ab5f9ffbR3

The unit tests are valuable though.

stbenjam:
@dLobatog Please rebase and change the commit to `refs`, as #173 actually fixed it.

dLobatog:
Updated :smile:

ares:
Thanks @dLobatog, merging!

PR #178, opened 2016-04-27
fixes #14850 - drop foreman 1.11 support

stbenjam:


lib/foreman_remote_execution/version.rb
```diff
@@ -1,3 +1,3 @@
 module ForemanRemoteExecution
-  VERSION = '0.3.0'
+  VERSION = '1.0.0'
```
stbenjam: Did we settle on this being 1.0 or 0.4?

I wasn't 100% sure if we were going to call this 1.0 and then start following http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies going forward.

iNecas: I'm ok with 1.0

ares: I'm ok with both as long as we have redmine versions :-)

adamruzicka: ^ This

stbenjam:
This should be good to go now that we've branched for 0.3-stable.

ares:
The support was dropped by #189, merging, thanks @stbenjam

PR #179, opened 2016-04-27
Release 0.3.1

stbenjam:


iNecas:
I have tried with nightlies only, but it seems to work (me seriously thinks about using the german locale: it definitely shows our limitations to handle long-names :)

iNecas:
:+1:

stbenjam:
Thanks! Will make sure it works in production mode on 1.11 when I do the packaging PR's of course :)

PR #181, opened 2016-05-02
Fixes #14684 - Save end time for recurring

adamruzicka:
Requires https://github.com/theforeman/foreman-tasks/commit/a637787a5d3963e2a16e57e905dc31f19ce07680


ares:
tests from 0.3-stable were triggered on develop, here's the run against 1.11-stable http://ci.theforeman.org/job/test_plugin_pull_request/2578/

adamruzicka:
@ares Thanks for rerunning them there. Should I be worried (and/or do something) about them failing on develop?

ares:
No, we decided 0.3 branch won't be compatible with Foreman 1.12. Thanks @adamruzicka, merging.

PR #183, opened 2016-05-09
Fixes #14875 - Add host search definition

ares:


adamruzicka:
Works nicely, APJ

ares:
Thanks for review, merging.

PR #185, opened 2016-05-11
fixes #15017 - trigger save correctly with feature

stbenjam:
Setting the template on the feature instead of the feature on the
template causes the cascade save in Rails to not save it.

### DIFF
--- a/test/unit/job_template_test.rb
+++ b/test/unit/job_template_test.rb
@@ -73,6 +73,10 @@ describe JobTemplate do
   end
 
   context 'importing a new template' do
+    let(:remote_execution_feature) do
+      FactoryGirl.create(:remote_execution_feature)
+    end
+
     let(:template) do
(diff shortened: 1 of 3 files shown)


iNecas:
Tests are failing

test/unit/job_template_test.rb
```diff
@@ -129,6 +134,11 @@
       template_with_input_sets.template_inputs_with_foreign.map(&:name).must_equal ['service_name']
     end
 
+    it 'imports feature' do
+      remote_execution_feature.reload
+      remote_execution_feature.job_template.must_equal template
```
iNecas: For the test failure - the template is loaded lazily, so you need to reference it somewhere before checking the `job_templates` in the feature.

stbenjam:
@iNecas Green

iNecas:
Thanks @stbenjam Merged

PR #188, opened 2016-05-19
fixes #15017 - trigger save correctly with feature

ares:
Setting the template on the feature instead of the feature on the
template causes the cascade save in Rails to not save it.


ares:
cherry-pick to 0.3-stable, tests are expected to fail...

stbenjam:
ACK

ares:
thanks, merged

PR #189, opened 2016-05-25
fixes #15179 - Select box for settings

sean797:
Requires theforeman/foreman@ad603e4

### DIFF
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -22,7 +22,10 @@ class Setting::RemoteExecution < Setting
                  'root'),
         self.set('remote_execution_effective_user_method',
                  N_('What command should be used to switch to the effective user. One of %s') % SSHExecutionProvider::EFFECTIVE_USER_METHODS.inspect,
-                 'sudo'),
+                 'sudo',
+                 'remote_execution_effective_user_method',
+                 nil,
+                 { :collection => Proc.new {Hash[SSHExecutionProvider::EFFECTIVE_USER_METHODS.map{|method| [method, method]}]} }),
         self.set('remote_execution_sync_templates',
                  N_('Whether we should sync templates from disk when running db:seed.'),
                  true)


stbenjam:
The commit message is good, @theforeman-bot seems to be looking at the older commit.

app/models/setting/remote_execution.rb
```diff
@@ -22,7 +22,10 @@ def self.load_defaults
                  'root'),
         self.set('remote_execution_effective_user_method',
                  N_('What command should be used to switch to the effective user. One of %s') % SSHExecutionProvider::EFFECTIVE_USER_METHODS.inspect,
-                 'sudo'),
+                 'sudo',
+                 'remote_execution_effective_user_method',
+                 nil,
+                 { :collection => Hash[SSHExecutionProvider::EFFECTIVE_USER_METHODS.map{|method| [method, method]}]}),
```
stbenjam: This needs to be a proc, as Foreman changed how this worked shortly after, see https://github.com/theforeman/foreman/commit/1f823285f5c8231fc6afab0c7c2064690fde0d39

sean797: Using a proc for that seems pointless? It's not like it's going to change dynamically like Organisations/Locations or templates.

stbenjam: It's explained in the issue http://projects.theforeman.org/issues/14963.  This may be static, but everything needs to be wrapped in proc, your change doesn't work against the latest foreman develop. It generates an exception when you load the page.

sean797: I see what you mean, just rebased my dev environment :+1:

stbenjam: Thanks, looks good.  The proc is somewhat useless here I agree, I wish it was optional but oh well.

stbenjam:
Thanks @sean797 !

PR #191, opened 2016-07-07
Fixes #15600 - fix deprecated after: :finisher_hook

iNecas:

PR #192, opened 2016-07-07
Fixes #15605 - fix failing params tests

iNecas:


iNecas:
Now the tests are failing due to issues fixed in https://github.com/theforeman/foreman_remote_execution/pull/191, I will get thie commit into the https://github.com/theforeman/foreman_remote_execution/pull/191 PR so that both reasons for failed tests are fixed at once

PR #196, opened 2016-07-26
Refs #15214 - Added setting ssh port for host and simple tests

dhlavac:
Added settings - ssh port and simple test (return value , set up port from settings , set up port from params)

### DIFF
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -28,7 +28,10 @@ class Setting::RemoteExecution < Setting
                  { :collection => Proc.new {Hash[SSHExecutionProvider::EFFECTIVE_USER_METHODS.map{|method| [method, method]}]} }),
         self.set('remote_execution_sync_templates',
                  N_('Whether we should sync templates from disk when running db:seed.'),
-                 true)
+                 true),
+        self.set('remote_execution_ssh_port',
+                N_('Port to use for SSH communication. Default port 22.'),
+                '22')
       ].each { |s| self.create! s.update(:category => 'Setting::RemoteExecution') }
     end
 
--- a/app/models/ssh_execution_provider.rb
+++ b/app/models/ssh_execution_provider.rb
@@ -6,7 +6,8 @@ class SSHExecutionProvider < RemoteExecutionProvider
     def proxy_command_options(template_invocation, host)
       super.merge(:ssh_user => ssh_user(host),
                   :effective_user => effective_user(template_invocation),
-                  :effective_user_method => effective_user_method(host))
+                  :effective_user_method => effective_user_method(host),
+                  :ssh_port => ssh_port(host))
     end
 
     def humanized_name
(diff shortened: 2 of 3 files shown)


stbenjam:
[test]

test/unit/remote_execution_provider_test.rb
```diff
@@ -89,5 +89,28 @@
...
+      end
+
+      it 'has ssh port changed in settings and check return type' do
+        proxy_options[:ssh_port].must_be_kind_of Integer
+        proxy_options[:ssh_port].must_equal 66
+      end
+    end
+
+    describe 'ssh port from params' do
+      it 'takes ssh port number from paramsand check return type' do
```
stbenjam: Space missing: `params and`*

app/models/ssh_execution_provider.rb
```diff
@@ -23,6 +24,10 @@ def ssh_user(host)
       host.params['remote_execution_ssh_user']
     end
 
+    def ssh_port(host)
+      (host.params['remote_execution_ssh_port'] || Setting[:remote_execution_ssh_port]).to_i
```
stbenjam: If the user accidentally enters a  string value this returns 0, which could be confusing/hard to troubleshoot.

Using something like this might be more desirable, so it raises if the user has entered something nonsensical:

```ruby
Integer(host.params['remote_execution_ssh_port'] || Setting[:remote_execution_ssh_port])
```

stbenjam:
Thanks @dhlavac, code looks great! We need to make changes to smart_remote_execution_ssh too, so it uses the new ssh_port option.

dhlavac:
Thanks for notice, I make changes .

stbenjam:
[test]

TuningYourCode:
Made the needed changes, because i needed them. Looking forward to get them into master :)

app/models/setting/remote_execution.rb
```diff
@@ -28,7 +28,10 @@ def self.load_defaults
                  { :collection => Proc.new {Hash[SSHExecutionProvider::EFFECTIVE_USER_METHODS.map{|method| [method, method]}]} }),
         self.set('remote_execution_sync_templates',
                  N_('Whether we should sync templates from disk when running db:seed.'),
-                 true)
+                 true),
+        self.set('remote_execution_ssh_port',
+                N_('Port to use for SSH communication. Default port 22.'),
```
iNecas: Could you make a not about ability to override this in host param in the description, similarly as it's described in remote_execution_ssh_user https://github.com/dhlavac/foreman_remote_execution/blob/4418fbcc758442eb58111297d842060d013a5285/app/models/setting/remote_execution.rb#L18 ?

iNecas:
I've tested it works nicely with @TuningYourCode change. Pending one comment.

iNecas:
Since the only issue was with a description for settings, I updated that locally and merged as b6283bd. Thanks @dhlavac and @TuningYourCode, new packages will be released shortly

iNecas:
For some reason (probably connected to the code transformations while moving some parts from smart proxy rex), we missed the changes in master, pushing the changes again as 14187c2

PR #198, opened 2016-08-12
Fixes #16067 - state the API key for dynflow_task explicitly

iNecas:
This fixes the regression introduce by
 http://projects.theforeman.org/issues/15778, as the output
 of the API for job-invocation show changed.


stbenjam:
ack

PR #199, opened 2016-08-15
Fixes #16107 - Allow user to set own keys parameter

dLobatog:
Currently Foreman is failing to render the REX snippet with this error:
"The snippet 'remote_execution_ssh_keys' threw an error: undefined
method `join' for #String:0x007f4775c597e8"

From what I see, the parameter is expected to be an array or similar.
The docs say nothing about it -
https://theforeman.org/plugins/foreman_remote_execution/0.3/index.html -
and there's no way for the user to set the parameter as Array as a
global param or a host param.

The 'expected' Array comes from the plugin itself who queries all
proxies for the parameter, but it doesn't respect if there's a parameter
that already exists.

I propose making the method in REX return a string so that we can:
- remove the `join` call from the template
- allow people to set their own SSH keys via global params/host params

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -26,8 +26,9 @@ module ForemanRemoteExecution
 
     def params_with_remote_execution
       params = params_without_remote_execution
-      keys = remote_execution_ssh_keys
-      params['remote_execution_ssh_keys'] = keys unless keys.blank?
+      if (keys = remote_execution_ssh_keys).present?
+        params['remote_execution_ssh_keys'] = params['remote_execution_ssh_keys'].to_s + keys.join('\n')
+      end
       [:remote_execution_ssh_user, :remote_execution_effective_user_method].each do |key|
         params[key.to_s] = Setting[key] unless params.key?(key.to_s)
       end


stbenjam:
`remote_execution_ssh_keys` is an internal parameter, like the many others Foreman provides.  Foreman doesn't give users a way to override the interfaces, environment, and puppet-related parameters it supplies, why should this one be different?

If an SSH key is applicable for rex, we know about it.  What purpose is there for a user to supply their own? My position on this hasn't changed: if users want to supply SSH keys, Foreman should provide a different mechanism to do it other than using REX's.

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -26,8 +26,9 @@ def execution_status_label(options = {})
 
     def params_with_remote_execution
       params = params_without_remote_execution
-      keys = remote_execution_ssh_keys
-      params['remote_execution_ssh_keys'] = keys unless keys.blank?
+      if (keys = remote_execution_ssh_keys).present?
+        params['remote_execution_ssh_keys'] = params['remote_execution_ssh_keys'].to_s + keys.join('\n')
```
stbenjam: Isn't this creating a recursion? params is this same method.

stbenjam:
That being said, this fixes the problem if a user does supply a parameter of that name, but it's not like Foreman doesn't have other parameters that will cause problems. What happens to the network snippet if a user tries to override say the network hash?

app/models/concerns/foreman_remote_execution/host_extensions.rb
iNecas: I guess we should join also the user-provided keys with "\n"

dLobatog: The user provided keys come in a string, you can't `join` them like the array, it's why I wrote the patch. user provided keys already have '\n' as global, host, or hostgroup params are strings

iNecas: `([params['remote_execution_ssh_keys'].to_s] + keys).join('\n')`, but let's talk to @stbenjam about this even being a good idea or not…

dLobatog:
OK I think I understood it a bit better now. 

Since the ENC is serving an array.. we shouldn't allow users to create or have a parameter with this name, correct?  If so, I can change this PR to validate any param that's created with that purpose.

https://github.com/theforeman/community-templates/pull/292/files#diff-de90342192dacee6a086bfd0c10bf6dcR29

Also I would add a warning to the template as initially I thought I could just put my keys in a param and that template would serve them. I still think it's more reasonable to have a generic template to deploy SSH keys (for REX and user-provided) but we can go with fixing the existing case for the moment. 

Let me know if  that sounds better and I'll update the PR

stbenjam:
> Also I would add a warning to the template as initially I thought I could just put my keys in a param and that template would serve them.

You weren't alone in thinking that, if you want to do some validation on the parameter in the snippet itself to make sure it's an array first, you could do that and it'd avoid the error people get.

Secondly, if the plugin is installed, we could merge the user-specified keys (in a \n separated string) with ours and return an array. Although I would prefer if people wouldn't do that, I can live with it I guess.

iNecas:
I guess the conclusion of this + the discussion on the issue is we're not going to get this change in. Closing, If anyone things otherwise and has more to add here, please reopen/comment

iNecas:
I see people running into this every now and then, which makes revisit this and give the `join as an array` a try. @dLobatog do you want to update this PR with the change, as @stbenjam suggests, or should I open separate PR on my own. I can do both ways

PR #202, opened 2016-09-02
Fixes #16426 - Add JobTemplateImporter shim class for foreman_templates

GregSutcliffe:
This is pretty much verbatim copy from the `update_job_template` method already present in foreman_templates itself. The only difference is that it deletes `model` from the metadata before passing it on to the real import call.

Not quite sure if tests are wanted / best place to implement them, guidance welcome.

### DIFF
--- a/app/models/job_template_importer.rb
+++ b/app/models/job_template_importer.rb
@@ -1,0 +1,24 @@
+# This class is a shim to handle the importing of templates via the
+# foreman_templates plugin. It expects a method like
+#     def import(name, text, metadata)
+# but REx already has an import! method, so this class provides the
+# translation layer.
+
+class JobTemplateImporter
+  def self.import!(name, text, _metadata)
+    template = JobTemplate.import(
+      text.sub(/^name: .*$/, "name: #{name}").sub(/^model: .*$/, ''),
+      :update => true
+    )
+
+    c_or_u = template.new_record? ? 'Created' : 'Updated'
+    id_string = ('id' + template.id) rescue ''
+
+    result = "  #{c_or_u} Template #{id_string}:#{name}"
+    { :old => template.template_was,
+      :new => template.template,
+      :status => template.save,
+      :result => result
+    }
+  end
+end
--- a/test/unit/job_template_importer_test.rb
+++ b/test/unit/job_template_importer_test.rb
@@ -1,0 +1,48 @@
+require 'test_plugin_helper'
+
+describe JobTemplateImporter do
+  context 'importing a new template' do
+    # JobTemplate tests handle most of this, we just check that the shim
+    # correctly loads a template returns a hash
+    let(:remote_execution_feature) do
+      FactoryGirl.create(:remote_execution_feature)
+    end
+
+    let(:result) do
+      name = "Community Service Restart"
+      text = <<-END_TEMPLATE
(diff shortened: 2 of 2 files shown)


app/models/job_template_importer.rb
```diff
@@ -0,0 +1,26 @@
+class JobTemplateImporter
+  def self.import!(name, text, metadata)
+    return {
+      :status => false,
+      :result => 'Skipping job template import, remote execution plugin is not installed.'
+    } unless defined?(JobTemplate)
```
stbenjam: As this is now in the plugin itself, "this should never happen" :tm:  :-)

app/models/job_template_importer.rb
```diff
@@ -0,0 +1,26 @@
...
+    template = JobTemplate.import(
+      text.sub(/^name: .*$/, "name: #{name}").sub(/^model: .*$/, ''),
+      :update => true
+    )
+
+    c_or_u = template.new_record? ? 'Created' : 'Updated'
+    id_string = ('id' + template.id) rescue ''
+
+    if template.template != template.template_was
+      diff = Diffy::Diff.new(
```
stbenjam: You'll need to add this to our gemspec

GregSutcliffe: that's fair, I can probably refactor that

app/models/job_template_importer.rb
GregSutcliffe: well i did say it was a verbatim copy. I'll update :)

stbenjam:
I do think I would prefer a simple test of the shim to make sure it keeps working, we try to keep test coverage on everything as much as possible. I could see JobTemplate model changing and someone overlooking this.

Also a comment at the top about why it's here would be useful, with a pointer to the foreman-templates plugin.

GregSutcliffe:
@stbenjam sure, happy to. I'll try to sort that out shortly.

stbenjam:
Should be just a copy and paste job from https://github.com/theforeman/foreman_remote_execution/blob/master/test/unit/job_template_test.rb#L75, but just checking the hash

app/models/job_template_importer.rb
GregSutcliffe: Deleted

app/models/job_template_importer.rb
GregSutcliffe: Ok, refactored this, in combination with https://github.com/theforeman/foreman_templates/commit/9efafedf039173a18d1d62c264d2b42902d8c398 it should work cleanly now

GregSutcliffe:
@stbenjam added a simple test for the hash structure.

stbenjam:
Thanks, sorry for the delay.  It looks good to me.

I'm ok to merge once [test]s are green. Unfortunately master is currently red (http://ci.theforeman.org/job/test_plugin_matrix/2026/), so don't want to merge anything until it's green again.

adamruzicka:
The tests are green, merging. Thanks @GregSutcliffe !

GregSutcliffe:
Ty @adamruzicka, let me know when it hits release and I'll mark the duplicated code in foreman_templates for removal.

adamruzicka:
@GregSutcliffe I'll try to remember that

PR #205, opened 2016-09-13
Fixes #15311 free form job command cannot exceed 255 chars

vision2910:
remote execution is unable to handle commands longer than 255 chars. In a normal scenario executed command could exceed 255 chars. 

Added a migration to remove the limit for command length.

### DIFF
--- a/db/migrate/20160909090711_remove_char_limit_for_remote_job_invocation.rb
+++ b/db/migrate/20160909090711_remove_char_limit_for_remote_job_invocation.rb
@@ -1,0 +1,10 @@
+class RemoveCharLimitForRemoteJobInvocation < ActiveRecord::Migration
+  def up
+    change_column :template_invocation_input_values, :value, :string, :null => false, :limit => nil
+    change_column :job_invocations, :description, :string, :limit => nil
+  end
+  def down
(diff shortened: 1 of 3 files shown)


db/migrate/20160909090711_remove_char_limit_for_remote_job_invocation.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveCharLimitForRemoteJobInvocation < ActiveRecord::Migration
+  def up
+    change_column :template_invocation_input_values, :value, :string, :null => false, :limit => nil
+    change_column :job_invocations, :description, :string, :limit => nil
```
stbenjam: These should at least be type `:text` instead.

As far as extending the length of job description, why? I think it's not particularly related, descriptions should be fairly short and truncated.

ares: We should also make sure that outputting unlimited text values is trimming it in e.g. job detail page.

iNecas:
I've opened new PR addressing the comments in this review here https://github.com/theforeman/foreman_remote_execution/pull/206

PR #206, opened 2016-09-26
Fixes #15311 allow large inputs

iNecas:
Addresses comments in https://github.com/theforeman/foreman_remote_execution/pull/205

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -167,6 +167,11 @@ module RemoteExecutionHelper
     end
   end
 
+  def invocation_description(invocation)
+    description = invocation.description.try(:capitalize) || invocation.job_category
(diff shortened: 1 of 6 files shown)


stbenjam:
This looks reasonable to me, but tests are failing on master: http://ci.theforeman.org/job/test_plugin_matrix/2026/

adamruzicka:
Works as expected, but can mess up the UI and flow out of the screen.
job invocation detail overflow

On the job invocation index page we should truncate the descriptions the same way we do it on tasks index
job invocation index overflow

iNecas:
I've added some truncation in place to make sure we don't go too far, @adamruzicka or @stbenjam, might re-review?

adamruzicka:
Looks good and works as expected, ACK

iNecas:
Thanks folks!

PR #208, opened 2016-10-24
Fixes #17066 - escape the life output properly

iNecas:


adamruzicka:
Works as expected, APJ

PR #211, opened 2016-11-23
Fixes #17456 - The register method will execute update only when ther…

xgazdi:
Hi all, 
  I wrote small patch which prevents executing hooks each time the foreman / Passenger is being started. It is achieved by comparing the current feature in database with the feature which should be updated. There is compared each of the values which is being saved including the options[:description] which is array, but I believe the size of this array will not exceed 5, so it should be safe comparing the arrays like this.


iNecas:
Thanks @xgazdi, I've taken the idea and updated it to go more Rails way + added some tests to cover the branch, see https://github.com/theforeman/foreman_remote_execution/pull/212, If you don't mind, I would close this PR in favor of the mentioned one

xgazdi:
@iNecas, OK I am fine with ;-)

PR #213, opened 2016-12-07
Fixes #17603 - setting for using interfaces for ip calculation

iNecas:
Before this patch, we tried to use data from interfaces and the ip
address to connect to the server.

It turns out, in production the data about the interfaces IP address
might not be up to date, which might lead to the execution not targeting
the system it should.

Therefore, I'm introducing settings for influencing how the calculation
is performed and by default (for production), preferring the fqdn
over the interfaces.


iNecas:
Manual mention bot: asking for opinions from @stbenjam, @ares, @adamruzicka

iNecas:
[test]

stbenjam:
What's the use case for remote_execution_fallback_to_all_interfaces? We already prefer execution/primary/provision in order, right?

I just worry about how many settings we're adding, it can be a little confusing.

stbenjam:
Also, I would call `remote_execution_use_interfaces` something else. Interfaces actually have possible DNS names, too.  I'd call it `remote_execution_connect_by_ip` or something with IP in the name.

ares:
It's slightly confusing when it says that host FQDN can be preferred over interface. Note that host FQDN is == primary interfaces FQDN. I guess the remote_execution_fallback_to_all_interfaces setting is there if we want to use some interface that has none of these flags as a fallback. While this is quite edge case I can see a value in it since only one interface can be marked as rex. That said, I'd like to also see PR to the rex manual describing both new settings. Probably under https://theforeman.org/plugins/foreman_remote_execution/0.3/index.html#4.1Determiningthesmartproxyforhost and https://theforeman.org/plugins/foreman_remote_execution/0.3/index.html#4.2Globalsettings

iNecas:
Updated. I agree the `connect_by_ip` describes better what it opens. I also agree that less options is better than more, therefore I've removed the option for all interfaces for now, as it can be just more confusing to the people. We can always add it later if the demand is there, but lets keep it simpler if possible. Ready for another round of reviews

iNecas:
I would update the manual after this gets in: I need to do it anyway to reflect all the changes that happened recently.

app/models/ssh_execution_provider.rb
```diff
@@ -18,27 +18,51 @@ def supports_effective_user?
...
         raise _('Effective user method "%{current_value}" is not one of %{valid_methods}') %
                   { :current_value => method, :valid_methods => EFFECTIVE_USER_METHODS}
       end
       method
     end
+
+    private
+
+    def effective_interfaces(host)
+      return [] unless host_setting(host, :remote_execution_connect_by_ip)
```
stbenjam: I'm not sure we should skip this - each interface can have a name, too.  There are users with extraordinarily complex network setups, and hosts with an fqdn per network. 

Given a host `web01.example.com`, it might have interfaces on app, db, and mgmt networks.  Let's say `eth2`, flagged as the execution interface, is on the management network.  `eth2` might have have the hostname `web01.mgmt.example.com` 

I think the expectation would be that if connect_by_ip is false, we should try to connect to `web01.mgmt.example.com` rather than `host.fqdn`.

iNecas: Good point. Please have a look at the latest version. I hope it should have everything right now.

app/models/ssh_execution_provider.rb
```diff
@@ -18,27 +18,58 @@ def supports_effective_user?
...
+        ip_interface = interfaces.find { |i| i.ip.present? }
+      end
+      if ip_interface
+        ip_interface.ip
+      else
+        fqdn_interface = interfaces.find { |i| i.fqdn.present? }
+        if fqdn_interface
+          fqdn_interface.fqdn
+        else
+          host.fqnd
```
stbenjam: Typo: host.fqdn

stbenjam: (Although this `else` could probably go away, a host must always have a primary interface, and host.fqdn is delegated to it.)

stbenjam:
Code looks OK to me - just one comment

app/models/ssh_execution_provider.rb
iNecas: Right, replaces it just with a plain error message, good catch!

iNecas:
[test] Should be ready after Jenkins finishes

iNecas:
Jeknins failed on sqlite due to intermittent issues: not related to the change. The last comment got addressed. Getting this in.

iNecas:
Thanks @stbenjam for review

PR #214, opened 2016-12-20
fixes #17803 - incorrect usage on scope search for job invocation

ohadlevy:


app/models/job_invocation.rb
```diff
@@ -48,7 +48,7 @@ class JobInvocation < ActiveRecord::Base
 
   scope :with_task, -> { references(:task) }
 
-  scoped_search :in => :recurring_logic, :on => 'id', :rename => 'recurring_logic.id', :auto_complete => true
+  scoped_search :in => :recurring_logic, :on => 'id', :rename => 'recurring_logic.id', :complete_value => true
```
ohadlevy: tbh: I dont know if it should be true (e.g. is there a value to show the auto completer id?)

ares: I don't like it either but I think it's required because of this https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_invocation_task_group.rb#L14

ohadlevy: i have no problem with the search condition, just that it shouldn't be part of the auto completer, e.g. you don't need to see it in the drop down.

ares: I don't think it's exposed anywhere in UI but I see your point. I think it's safe to remove it. I'll merge this PR as it is to see that tests becomes green and then open separate issue.

ares:
Thanks @ohadlevy, merging.

PR #216, opened 2017-01-06
Fixes #17315 - Handle output coming from .bashrc

adamruzicka:
This fix contains a not-so-pretty monkey patch for how Net::SCP's state machine handles the communication. Probably needs some discussion


adamruzicka:
Closing in favor of #218

PR #218, opened 2017-01-09
 Fixes #17315 - Handle output coming from .bashrc, v2

adamruzicka:
Alternative to #216

Instead of using SCP to copy the script we open an ssh connection and write the script into its stdin. On the other side of the tunnel we use `tee` to read from the stdin and write it to a file and `chmod` to set the permissions on the script.

### DIFF
--- a/lib/foreman_remote_execution_core/script_runner.rb
+++ b/lib/foreman_remote_execution_core/script_runner.rb
@@ -152,7 +152,7 @@ module ForemanRemoteExecutionCore
       output = ""
       exit_status = nil
       channel = session.open_channel do |ch|
-        ch.on_data { |data| output.concat(data) }
+        ch.on_data { |_, data| output.concat(data) }
         ch.on_extended_data { |_, _, data| output.concat(data) }
         ch.on_request("exit-status") { |_, data| exit_status = data.read_long }
         # on signal: sending the signal value (such as 'TERM')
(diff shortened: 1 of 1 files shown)


adamruzicka:
[test]

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -215,29 +223,33 @@ def ensure_local_directory(path)
...
-      scp = Net::SCP.new(session)
-      upload_channel = scp.upload(local_path, remote_path)
-      upload_channel.wait
-    ensure
-      if upload_channel
-        upload_channel.close
-        upload_channel.wait
+    def upload_data(data, path, permissions = 555)
+      ensure_remote_directory File.dirname(path)
+      command = "#{su_prefix} tee '#{path}' >/dev/null && #{su_prefix} chmod '#{permissions}' '#{path}'"
```
iNecas: wouldn't `cat > {path}` would be probably a bit simpler

adamruzicka: Running `sudo cat > $path` would fail if the actual user (not the sudo one) didn't have write permissions to `$path`, because the redirection is handled by the shell and not the process with elevated permissions

iNecas: Ah, right, in that case, just put a small comment there why we're going this way

iNecas:
Thanks @adamruzicka

PR #220, opened 2017-01-10
Fixes #18011 - allow add specific action to Run Job in host detail

ares:


adamruzicka:
[test]

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -8,11 +8,27 @@ module HostsHelperExtensions
...
+
+      if host_features.present?
+        action_buttons(schedule_job_button(*args), *host_features)
+      else
+        schedule_job_button(*args)
+      end
+    end
+
+    def schedule_job_button(*args)
+      link_to(_('Schedule Job'), new_job_invocation_path(:host_ids => [args.first.id]), :id => :run_button, :class => 'btn btn-default')
```
iNecas: I wonder if 'Schedule Execution Job' would be too long?

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -8,11 +8,27 @@ module HostsHelperExtensions
     end
 
     def multiple_actions_with_remote_execution
-      multiple_actions_without_remote_execution + [[_('Run Job'), new_job_invocation_path, false]]
+      multiple_actions_without_remote_execution + [ [_('Schedule Job'), new_job_invocation_path, false] ]
+    end
+
+    def schedule_job_multi_button(*args)
+      host_features = RemoteExecutionFeature.where(:host_action_button => true).map do |feature|
```
iNecas: what about using named scope `with_host_action_button` instead. And in adding a test into remote_execution_feature test to cover registration and retrieving of the action buttons?

BTW. what should be the ordering of the buttons?

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
ares: sch1
sch2

ares: I think I like it more without "Execution"

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
ares: Good point, moved to named scope.

I'd keep ordering aside for now. If we need to have control over the order, we could add priority numbers to features later.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
iNecas: What about "Schedule Remote Job"? My concern is the Job name itself is not well descriptive. For some reason, comparing `Run Job` and `Schedule Job`, the `Run Job` felt more like it has something to do with Rex. The `schedule` seems more generic to me, and I feel specifying what kind of job we mean might help with user understanding. On the other hand, I'm not going to bikeshed on this.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
iNecas: What about ordering by name: in that case, there is at least some determinism.

test/unit/remote_execution_feature_test.rb
```diff
@@ -43,4 +43,28 @@
...
+      feature.must_be :persisted?
+      feature.label.must_equal 'new_feature_that_does_not_exist'
+      feature.name.must_equal 'name'
+      refute feature.host_action_button
+    end
+
+    it "creates a feature with host action flag" do
+      feature = RemoteExecutionFeature.register('new_feature_that_does_not_exist_button', 'name', :host_action_button => true)
+      feature.must_be :persisted?
+      assert feature.host_action_button
```
iNecas: I know you don't like testing ActiveRecord features, such as named scopes much, but it would make me happy, if we used the RemoteExecutionFeature.with_host_action_button to check if the feature is there. If nothing else, it will prove, that the named scope is defined right.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
ares: Renamed, that hopefully avoid any bikeshedding :-)

test/unit/remote_execution_feature_test.rb
ares: I added separate test just to test the scope, I hope it makes you at least [this happy](https://www.youtube.com/watch?v=y6Sxv-sUYtM) :-)

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -8,11 +8,27 @@ module HostsHelperExtensions
     end
 
     def multiple_actions_with_remote_execution
-      multiple_actions_without_remote_execution + [[_('Run Job'), new_job_invocation_path, false]]
+      multiple_actions_without_remote_execution + [ [_('Schedule Job'), new_job_invocation_path, false] ]
```
iNecas: I guess we should use `Schedule Remote Job` here as well

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
ares: sorry, somehow I missed this, I added explicit order to the helper, sorting by label

iNecas:
After tests are passing: I see some issue with

```
  1) Error:
RemoteExecutionFeature::.register#test_0004_updates a feature if it exists:
ActiveRecord::RecordInvalid: Validation failed: Name can't be blank
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:11:in `block in transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:5:in `transaction'
    /home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/test/unit/remote_execution_feature_test.rb:69:in `block (3 levels) in <top (required)>'
```

ares:
hm, tests failures on 2.2 seems unrelated to me but I have no idea what's the cause

ares:
anyway the original test failure is fixed and the PR is rebased

test/unit/remote_execution_feature_test.rb
```diff
@@ -53,4 +53,34 @@
...
+
+    it "created feature with host action flag can be found using named scope" do
+      feature = RemoteExecutionFeature.register('new_feature_that_does_not_exist_button', 'name', :host_action_button => true)
+      assert_includes RemoteExecutionFeature.with_host_action_button, feature
+    end
+
+
+    it "updates a feature if it exists" do
+      existing = FactoryGirl.create(:feature, :name => 'existing_feature_withou_action_button')
+      feature = RemoteExecutionFeature.register(existing.label, existing.name, :host_action_button => true)
```
iNecas: Use remote_execution_feature https://github.com/ares/foreman_remote_execution/blob/2707f7353b94561ab891d9058e30cc7ec92b7b19/test/factories/foreman_remote_execution_factories.rb#L63, tests are failling

ares: omg

ares:
@iNecas tests are now failing on sqlite 2.2 and 2.3 only, @adamruzicka maybe knows the cause?

db/migrate/20170110145641_add_host_action_button_to_remote_execution_feature.rb
```diff
@@ -0,0 +1,5 @@
+class AddHostActionButtonToRemoteExecutionFeature < ActiveRecord::Migration
+  def change
+    add_column :remote_execution_features, :host_action_button, :bool, :null => false, :default => false
```
iNecas: I've noticed this in schema.rb

```
# Could not dump table "remote_execution_features" because of following NoMethodError
# undefined method `[]' for nil:NilClass
```

This meant `bundle exec rake db:schema:load RAILS_ENV=test` was not populating the `:remote_execution_features` table. Changing `:bool` to `:boolean` seemed to do the trick

ares:
:green_apple:

iNecas:
Thanks @ares

PR #222, opened 2017-01-15
Fixes #17921 - optimized task queries

ShimShtein:
Task queries were optimized for not bringing too much unneeded
objects. For more optimizations a proper performance analysis
is needed

### DIFF
--- a/app/models/targeting.rb
+++ b/app/models/targeting.rb
@@ -36,9 +36,13 @@ class Targeting < ActiveRecord::Base
     raise ::Foreman::Exception, _('Cannot resolve hosts without a bookmark or search query') if bookmark.nil? && search_query.blank?
 
     self.search_query = bookmark.query if dynamic? && bookmark.present?
-    self.save!
-    self.touch(:resolved_at)
-    self.hosts = User.as(user.login) { Host.authorized(RESOLVE_PERMISSION, Host).search_for(search_query) }
+    self.resolved_at = Time.zone.now
+    self.validate!
+    # avoid validation of hosts objects - tey will be loaded for no reason.
+    host_ids = User.as(user.login) { Host.authorized(RESOLVE_PERMISSION, Host).search_for(search_query).pluck(:id) }
+    # this can be optimized even more, by introducing bulk insert
+    self.targeting_hosts.build(host_ids.map { |id| { :host_id => id } })
+    self.save(:validate => false)
   end
 
   def dynamic?
--- a/test/benchmark/run_hosts_job_benchmark.rb
+++ b/test/benchmark/run_hosts_job_benchmark.rb
@@ -1,0 +1,70 @@
+require "benchmark/benchmark_helper"
+require 'dynflow/testing'
+
+# Add plugin to FactoryGirl's paths
+FactoryGirl.definition_file_paths << File.expand_path('../../factories', __FILE__)
+FactoryGirl.definition_file_paths << "#{ForemanTasks::Engine.root}/test/factories"
+FactoryGirl.reload
+
+module Actions
+  module RemoteExecution
+    class RunHostsJob < Actions::ActionWithSubPlans
+      # stub function that calls other actions.
+      def trigger(*args)
+        nil
+      end
+    end
+  end
+end
+
+module Support
+  class DummyDynflowAction < Dynflow::Action
+  end
+end
+
+def generate_hosts(total)
+  FactoryGirl.create_list(:host, total, :comment => "benchmark-#{Foreman.uuid}")
+end
+
+Rails.logger.level = Logger::ERROR
+
+class ActionTester
+  include Dynflow::Testing::Factories
+
+  def initialize(task)
+    @task = task
+  end
+
+  def run_action(*args)
+    action = create_action(Actions::RemoteExecution::RunHostsJob)
+    @task.update_attributes(:external_id => action.execution_plan_id)
+    plan_action(action, *args)
+  end
+end
+
+puts 'generating admin user'
+admin = FactoryGirl.build(:user, :admin)
+admin.save(:validate => false)
+User.current = admin
+targeting = FactoryGirl.create(:targeting, :search_query => "comment = benchmark-#{Foreman.uuid}", :user => User.current)
+template_invocation = FactoryGirl.build(:template_invocation, :job_invocation => nil)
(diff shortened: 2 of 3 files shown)


app/models/targeting.rb
```diff
@@ -36,9 +36,13 @@ def resolve_hosts!
...
     self.search_query = bookmark.query if dynamic? && bookmark.present?
-    self.save!
-    self.touch(:resolved_at)
-    self.hosts = User.as(user.login) { Host.authorized(RESOLVE_PERMISSION, Host).search_for(search_query) }
+    self.resolved_at = Time.zone.now
+    self.validate!
+    # avoid validation of hosts objects - tey will be loaded for no reason.
+    host_ids = User.as(user.login) { Host.authorized(RESOLVE_PERMISSION, Host).search_for(search_query).pluck(:id) }
+    # this can be optimized even more, by introducing bulk insert
+    self.targeting_hosts.build(host_ids.map { |id| { :host_id => id } })
```
ohadlevy: would ```self.targeting_host_ids = host_ids``` work too?

app/models/targeting.rb
```diff
@@ -36,9 +36,13 @@ def resolve_hosts!
     raise ::Foreman::Exception, _('Cannot resolve hosts without a bookmark or search query') if bookmark.nil? && search_query.blank?
 
     self.search_query = bookmark.query if dynamic? && bookmark.present?
-    self.save!
-    self.touch(:resolved_at)
-    self.hosts = User.as(user.login) { Host.authorized(RESOLVE_PERMISSION, Host).search_for(search_query) }
+    self.resolved_at = Time.zone.now
+    self.validate!
+    # avoid validation of hosts objects - tey will be loaded for no reason.
```
ohadlevy: afaiu you can define not to validate on the AR relationship definitions, ideally fix it there.

iNecas:
What about adding some test with benchmarking?

app/models/targeting.rb
ShimShtein: No, since `TargetingHost` is a model by itself, it would have a different id.
You mean `self.host_ids = host_ids` probably (assign the array to the property that should hold ids only). I have tried it, ActiveRecord tries to validate that those hosts exist and puts them into `self.hosts` property under the hood, so it queries the database with `SELECT hosts.* from hosts where id in (...)` which I want to avoid, since there is no use for hosts here.

app/models/targeting.rb
ShimShtein: Avoiding the validation would be a change of the model (maybe a good change, I can't tell right now... @inecas?). 
Here I want to temporarily disable host's existence validation, since it loads a host object into memory, and I want to avoid it here.
Besides, I have just queried the database, of course the hosts are there! I don't need this validation in this specific case.

iNecas: What if we changed the validation on the TargetingHost to use `:host_id` instead of `:host`?

ShimShtein: It's the standard `:presence => true` validation. I am not sure you want to remove this constraint.

iNecas: I could live with `validates :host_id`

ShimShtein:
Benchmark tests added. To run them use:
```

RAILS_ENV=production ruby -Itest ../foreman_remote_execution/test/benchmark/run_hosts_job_benchmark.rb

RAILS_ENV=production ruby -Itest ../foreman_remote_execution/test/benchmark/targeting_benchmark.rb
```
from foreman folder.

app/models/targeting.rb
ShimShtein: Doesn't help too. Plus `host_ids = [ 1,2,3 ]` is its own DB action. I prefer the whole action to be a single DB action if it's possible.

ShimShtein:
## Benchmarking results on my machine:
### after changes:

```
Warming up --------------------------------------
rex-targeting-resolve-hosts
                         1.000  i/100ms
Calculating -------------------------------------
rex-targeting-resolve-hosts
                        229.277  (±14.8%) i/s -      2.197k in   9.999769s
Memory stats
Total objects allocated: 42017977
Total heap pages allocated: 6942
```

### Before changes

```
Warming up --------------------------------------
rex-targeting-resolve-hosts
                         1.000  i/100ms
Calculating -------------------------------------
rex-targeting-resolve-hosts
                        179.630  (±10.6%) i/s -      1.753k in  10.000234s
Memory stats
Total objects allocated: 42724693
Total heap pages allocated: 6929
```

## Hosts job benchmarking

### after changes:

```
Warming up --------------------------------------
   rex-run-hosts-job     1.000  i/100ms
Calculating -------------------------------------
   rex-run-hosts-job     94.729  (±12.7%) i/s -    920.000  in  10.006169s
Memory stats
Total objects allocated: 45232408
Total heap pages allocated: 6847
```

### Before change (to the hosts_job file only)

```
Warming up --------------------------------------
   rex-run-hosts-job     1.000  i/100ms
Calculating -------------------------------------
   rex-run-hosts-job     94.087  (±10.6%) i/s -    919.000  in  10.007119s
Memory stats
Total objects allocated: 45223668
Total heap pages allocated: 6847
```

iNecas:
The difference doesn't seem that significant from this numbers, unless I'm reading something wrong? I guess we need;

1) verify with production setup (including Katello)
2) looking into other potential places for improvement.

ShimShtein:
Basically agreed to both.

From the results, I can see that the first change (targeting) is pretty efficient - total time decrease from ~180 hosts per second to ~230 hosts per second. Plus the memory footprint went down by 700K objects.

The second change (run_hosts_job) is indeed pretty useless (it has improved memory usage by 10K objects, but it's marginally an accomplishment).

I will try to see if there is something that I am missing in run_hosts_job usecase, otherwise I will remove the unnecessary optimization.

iNecas:
[test]

app/models/targeting.rb
iNecas: Could you be more specific? Would that validation load the host record? What if the validatoin was not there at all, would we still need to disable validation here?

ShimShtein: > Would that validation load the host record? 

Yes, presence validation loads hosts records, I saw that in the logs. Didn't debug to the actual rails method though...

> What if the validatoin was not there at all, would we still need to disable validation here?

If the validation was not set on the model, we could use any other method (although setting host_ids directly would cause an extra query).

iNecas:
Works nicely and the benchmark proves improvement (although I expected a bit higher numbers). We need to have eyes opened for further non-efficiencies. Pulling this in for now. Thanks @ShimShtein

PR #223, opened 2017-01-16
Fixes #18098 - remove doubled export

ares:


app/controllers/job_templates_controller.rb
```diff
@@ -41,11 +41,6 @@ def import
     end
   end
 
-  def export
```
iNecas: should be removed from `routes.rb`: this is the reason why the tests are failling on permission check

ares:
Good catch, I removed it for UI, hopefully tests will be green now.

ares:
@iNecas mind to take another look?

iNecas:
Jenkins is happy => me is happy

PR #224, opened 2017-01-24
Fixes #18221 - support other template types

ares:


iNecas:
I have nothing against. Just give me some background on where this fits for your usecase outside of JobTemplates? I would be more than happy, if we converged this concept into core

ares:
I was playing with using the InputTemplateRenderer in core, if you want to try yourself, see the https://gist.github.com/ares/f0a4f4bcee1e0f2e9e16e9c4b6ec6325 (you would need to have rex installed in your Foreman). There's no plan yet, I'll first gather feedback whether inputs are something that core would want.

ares:
[test]

iNecas:
Thanks @ares

PR #227, opened 2017-02-09
Fixes #18650 - Plan job executions in batches

adamruzicka:
requires ~~[the foreman-tasks PR](https://github.com/theforeman/foreman-tasks/pull/229)~~ [foreman-tasks/236](https://github.com/theforeman/foreman-tasks/pull/236) and transitively [the dynflow PR](https://github.com/Dynflow/dynflow/pull/214)

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -25,13 +25,25 @@ module Actions
         job_invocation = JobInvocation.find(input[:job_invocation_id])
         proxy_selector = RemoteExecutionProxySelector.new
 
-        # composer creates just "pattern" for template_invocations because target is evaluated
-        # during actual run (here) so we build template invocations from these patterns
-        job_invocation.targeting.hosts.map do |host|
+        ::Host.where(:id => current_batch).map do |host|
+          # composer creates just "pattern" for template_invocations because target is evaluated
+          # during actual run (here) so we build template invocations from these patterns
           template_invocation = job_invocation.pattern_template_invocation_for_host(host).deep_clone
           template_invocation.host_id = host.id
           trigger(RunHostJob, job_invocation, host, template_invocation, proxy_selector)
         end
+      end
+
+      def entries(from, size)
+        host_ids.slice(from, size)
+      end
+
+      def total_count
+        host_ids.count
+      end
+
+      def host_ids
+        JobInvocation.find(input[:job_invocation_id]).targeting.host_ids
       end
 
       def set_up_concurrency_control(invocation)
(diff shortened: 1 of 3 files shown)


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -25,18 +25,30 @@ def create_sub_plans
         job_invocation = JobInvocation.find(input[:job_invocation_id])
         proxy_selector = RemoteExecutionProxySelector.new
 
-        # composer creates just "pattern" for template_invocations because target is evaluated
-        # during actual run (here) so we build template invocations from these patterns
-        job_invocation.targeting.hosts.map do |host|
+        ::Host.where(:id => current_batch).map do |host|
```
ohadlevy: must we use the full host object?

adamruzicka: I'm afraid we do. The host object is passed to the `RunHostJob` action where the permissions are verified for it.

I'm planning to look into permissions in rex later today, so I'll file another issue and open a PR if I find a better way to do it since it's not really in the scope of this issue.

iNecas: I saw `current_batch` is not defined for `RunHostsJob` when testing this, `include Dynflow::Action::WithBulkSubPlans` helped

iNecas: :+1: for @adamruzicka approach: one issue at a time

adamruzicka: Right, forgot about it. I fixed it in the [foreman-tasks PR](https://github.com/theforeman/foreman-tasks/pull/229). Since `RunHostsJob` inherits from `Actions::ActionWithSubPlans` we'll get two fixes for the price of one

adamruzicka:
@iNecas applied your patch, looks much better now, thank you

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -25,15 +27,27 @@ def create_sub_plans
...
+          # composer creates just "pattern" for template_invocations because target is evaluated
+          # during actual run (here) so we build template invocations from these patterns
           template_invocation = job_invocation.pattern_template_invocation_for_host(host).deep_clone
           template_invocation.host_id = host.id
           trigger(RunHostJob, job_invocation, host, template_invocation, proxy_selector)
         end
       end
 
+      def batch(from, size)
+        hosts.slice(from, size)
```
iNecas: Would it be possible to prevent running the full sql on the whole set, and use limit sql capabilities for this? From what I've seen, the `slice` is first evaluating the arel and then just slicing the original result, which might be quite expensive operation.

adamruzicka: `limit` and `offset` seems to do the trick

iNecas:
I'm approving the changes, that last piece is to merge https://github.com/theforeman/foreman-tasks/pull/236 and add dependency on new foreman-tasks once released

iNecas:
[test]

adamruzicka:
[test]

adamruzicka:
[test]

iNecas:
Thanks @adamruzicka

PR #229, opened 2017-02-16
Fixes #18515 - Show scheduling info on job invocation overview tab

adamruzicka:


adamruzicka:
Now requires https://github.com/theforeman/foreman-tasks/pull/237

iNecas:
@adamruzicka would you mind addressing https://bugzilla.redhat.com/show_bug.cgi?id=1426773 here as well?

adamruzicka:
Not at all, I'll see what can be done about it.

adamruzicka:
How about this?

Screenshot

adamruzicka:
CC @rohoover

Rohoover:
@adamruzicka 

Sorry for the delayed response. My main issue with this page, is that what is displayed is not in the same order of how it was inputed. Also - the status message in the right hand corner is sort of lost in a void. I can work for a visual solution if desired.

iNecas:
Thanks @adamruzicka !

PR #230, opened 2017-02-16
Fixes #18510 - Display warning when search query matches no host

adamruzicka:


app/helpers/remote_execution_helper.rb
```diff
@@ -184,8 +184,14 @@ def preview_box(template_invocation, target)
     if (preview = renderer.preview)
       content_tag :pre, preview
     else
-      alert :class => 'alert-block alert-danger base in fade has-error',
-            :text => renderer.error_message.html_safe
+      if target.nil?
+        alert :text => _('Rendering of the preview failed because no host matches the search query.'),
```
iNecas: I would not mention the rendering failed (to avoid the negative taste of that world), What about:

    Could not render the preview because no host matches the search query

iNecas:
[test]

app/helpers/remote_execution_helper.rb
adamruzicka: Sounds good, updated

iNecas:
Thanks @adamruzicka

PR #231, opened 2017-02-21
Refs #18098 - keep the export in routes

iNecas:
We originally remove that due to permissions check, but
the proper fix is to keep it there.


ares:
APJ

ares:
Tests error unrelated (ActionView::Template::Error: Cannot allocate memory - node). Thanks @iNecas, merging.

PR #233, opened 2017-02-23
Fixes #18651 - Provide a way to test REX at scale

adamruzicka:
This commit adds an alternative implementation for the ScriptRunner
class. Instead of doing ssh connections it discards any input its
given and gives back fake output.

It is controlled by setting the following environment variables
- ~~`REX_DEBUG`~~ `REX_SIMULATE` - setting to `1`, `yes` or `true` enables the use of the fake runner
- ~~`REX_DEBUG_PATH`~~ `REX_SIMULATE_PATH` - full path to a file which should be given back as output by the runner, currently one line per runner's refresh
- ~~`REX_DEBUG_FAIL_CHANCE`~~ `REX_SIMULATE_FAIL_CHANCE` - number from 0 to 100, each host has a ~~`REX_DEBUG_FAIL_CHANCE`~~ `REX_SIMULATE_FAIL_CHANCE` of exiting with ~~`REX_DEBUG_EXIT`~~ `REX_SIMULATE_EXIT`, useful for simulating random failures
- ~~`REX_DEBUG_EXIT`~~ `REX_SIMULATE_EXIT` - see the previous

Because it doesn't really open the outgoing connections, it doesn't need hosts with valid IP addresses.  This means one can spin up arbitrary number of hosts for testing purposes with the following snippet
```shell
bundle exec rails console <<-END
group = Hostgroup.find_or_create_by(:name => 'fakes')
group.save
300.times do
  h = Host.new
  h.hostgroup_id = group.id
  h.save!
end
END
```

All that is left to do is run the `smart_proxy_dynflow_core` (or `smart-proxy` or even `foreman`, depending on configuration) with properly set environment variables and use targeting `hostgroup = fakes`

### DIFF
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -12,6 +12,10 @@ module ForemanRemoteExecutionCore
     %w(yes true 1).include? ENV.fetch('REX_SIMULATE', '').downcase
   end
 
+  def self.runner_class
+    @runner_class ||= simulate? ? FakeScriptRunner : ScriptRunner
+  end
+
   if ForemanTasksCore.dynflow_present?
     require 'foreman_tasks_core/runner'
     if simulate?
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -8,9 +8,18 @@ module ForemanRemoteExecutionCore
                     :remote_working_dir    => '/var/tmp',
                     :local_working_dir     => '/var/tmp')
 
+  def self.debug?
+    %w(yes true 1).include? ENV.fetch('REX_DEBUG', '')
+  end
+
   if ForemanTasksCore.dynflow_present?
     require 'foreman_tasks_core/runner'
-    require 'foreman_remote_execution_core/script_runner'
+    if debug?
+      # Load the fake implementation of the script runner if debug is enabled
+      require 'foreman_remote_execution_core/fake_script_runner'
+    else
+      require 'foreman_remote_execution_core/script_runner'
+    end
     require 'foreman_remote_execution_core/actions'
   end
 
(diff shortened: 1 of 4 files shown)


iNecas:
I suggest setting some sane defualts (especially about the fake file), so that one doesn't need to create his own when wants to give this a try: just turn the fake executor on should be enough to start with

lib/foreman_remote_execution_core.rb
```diff
@@ -8,9 +8,18 @@ module ForemanRemoteExecutionCore
                     :remote_working_dir    => '/var/tmp',
                     :local_working_dir     => '/var/tmp')
 
+  def self.debug?
+    %w(yes true 1).include? ENV.fetch('REX_DEBUG', '')
```
iNecas: `REX_DEBUG` name is misleading: `REX_SIMULATE` or `REX_FAKE` would more describe what's going on there

adamruzicka: Right, fixed

lib/foreman_remote_execution_core/fake_script_runner.rb
```diff
@@ -0,0 +1,82 @@
+module ForemanRemoteExecutionCore
+  class ScriptRunner < ForemanTasksCore::Runner::Base
```
iNecas: Using the same name for two implementations might not be the best idea (if nothing else, it would be hard to test), what about putting the decision what runner to use into https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution_core/actions.rb#L7 ?

adamruzicka: I wanted to make it decide which implementation to use at load time to avoid checking the env variables over and over again. But sure, I can do that

iNecas: I guess we hide this behind `ForemanRemoteExecutionCore.runner_class`, that could be memoized to avoid the checking every time.

iNecas:
Works as expected and it's extremely useful. Thanks @adamruzicka !

PR #235, opened 2017-03-08
Fixes #18567 - Filter hosts by task result

adamruzicka:


adamruzicka:
### Showing all
showing all
### Showing failed
showing failed

ohadlevy:
I personally don't like the checkboxes - why can't we use scope search here?

even if we can't use scope search, why not override the complete_for and search_for methods for this controller?

adamruzicka:
The way it is now is by no means the final version. I need to investigate how (and if) this could be done using scoped search or some other way.

adamruzicka:
We can now use scoped search in the form of `job_task.result = $result` to search in the hosts tab in job invocation details. This also brings the ability to search for hosts in Hosts > All Hosts by providing queries like `job_invocation.id = 15 and job_task.result = error` to see all the hosts from job invocation with id 15 which ended with error

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -11,9 +11,31 @@ module HostExtensions
       has_many :targeting_hosts, :dependent => :destroy, :foreign_key => 'host_id'
       has_many :template_invocations, :dependent => :destroy, :foreign_key => 'host_id'
       has_one :execution_status_object, :class_name => 'HostStatus::ExecutionStatus', :foreign_key => 'host_id'
+      has_many :run_host_job_tasks, :through => :template_invocations
+
+      scoped_search :relation => :run_host_job_tasks, :on => :result, :rename => 'job_task.result',
+                    :ext_method => :search_by_job_invocation,
+                    :only_explicit => true,
+                    :complete_value => { :error => 'error', :success => 'success',
+                                         :pending => 'pending', :warning => 'warning',
```
iNecas: Error/warning would be better to be hidden behind 'failed' as that's how we should the results in hosts list (there is no error or warning word there)

iNecas:
+ please add some tests to cover the new searches

adamruzicka:
@iNecas updated

adamruzicka:
The test failure seems to be unrelated

adamruzicka:
[test]

iNecas:
Works nicely, thanks @adamruzicka

PR #241, opened 2017-04-18
Fixes #17410 - effective user from job invocation on rerun

iNecas:
It also fixes issue with keeping effective user value when the form is
re-rendered.

### DIFF
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -72,7 +72,7 @@
 
                     <div class="advanced hidden">
                       <% if job_template.effective_user.overridable? %>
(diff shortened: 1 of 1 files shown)


adamruzicka:
It indeed makes the effective user "stick" when one refreshes the page or does a rerun, however it still gets lost when selecting another job category. Is it out of scope of the original issue?

iNecas:
I would say it's outside of scope

adamruzicka:
Ok, created [issue 19316](https://projects.theforeman.org/issues/19316) for it. With that out of the way, let's get this in

adamruzicka:
Thanks @iNecas

PR #244, opened 2017-04-25
Fixes #19277 - Allow importing using foreman_templates

adamruzicka:
Summarization of changes (relative to JobTemplate):
- use `::import_raw!` instead of original `::import!`
- use `::import!` for importing templates from `foreman_templates`
- use `::import_parsed` to import already parsed template


app/models/job_template.rb
```diff
@@ -65,30 +65,41 @@ def self.base_class
   # Import a template from ERB, with YAML metadata in the first comment.  It
   # will overwrite (sync) an existing template if options[:update] is true.
```
ares: could you perhaps add a comment that this method is used by foreman_templates and the API should not change? I don't know if it helps but it's probably worth of trying

app/models/job_template.rb
```diff
@@ -65,30 +65,41 @@ def self.base_class
...
+      metadata = metadata.dup
+      return if metadata.blank? || metadata.delete('kind') != 'job_template' ||
+                (metadata.key? 'model' && metadata.delete('model') != self.to_s)
+      metadata['name'] = name
       # Don't look for existing if we should always create a new template
-      existing = self.find_by_name(metadata['name']) unless options.delete(:build_new)
+      existing = self.find_by_name(name) unless options.delete(:build_new)
       # Don't update if the template already exists, unless we're told to
       return if !options.delete(:update) && existing
+      metadata.delete('associate')
```
ares: if this comes from foreman_templates, should it be possibly removed in `.import`?

adamruzicka:
@ares Thank you for testing this, moved the 'assign' deletion somewhere else and added a comment about keeping the api stable

iNecas:
rebase needed

adamruzicka:
rebased

test/unit/template_invocation_task_result_map_test.rb
```diff
@@ -0,0 +1,25 @@
+require 'test_plugin_helper'
+
+describe TemplateInvocation do
```
iNecas: Seems like accidental file that doesn't belong to this PR? Tests are failing on `TemplateInvocation::TaskResultMap` not being defined.

test/unit/job_template_importer_test.rb
```diff
@@ -10,29 +10,26 @@ class JobTemplateImporterTest < ActiveSupport::TestCase
...
+        'provider_type' => 'SSH',
+        'feature' => remote_execution_feature.label,
+        'template_inputs' => [
+          { 'name' => 'service_name', 'input_type' => 'user', 'required' => true },
+          { 'name' => 'verbose', 'input_type' => 'user' }
+        ]
+      }
       text = <<-END_TEMPLATE
 <%#
-model: JobTemplateImporter
```
iNecas: We had JobTemplateImporter, while it was replaced with JobTemplate, was it just wrong at the beginning or some additiona change is related to this?

adamruzicka: I think `JobTemplateImporter` was chosen as a model just because `JobTemplate` already had `::import!` defined (which foreman-templates uses). This PR renames the import related methods a bit, this is described in the PR description.

iNecas:
Thanks for explanation, merging now

PR #245, opened 2017-05-10
Fixes #19401 - Compatibility fix for t?csh

adamruzicka:
- `tcsh` doesn't have `$(command)`
- `csh` doesn't have `$(command)` and apparently `$?`

### DIFF
--- a/lib/foreman_remote_execution_core/script_runner.rb
+++ b/lib/foreman_remote_execution_core/script_runner.rb
@@ -27,8 +27,10 @@ module ForemanRemoteExecutionCore
 
       # pipe the output to tee while capturing the exit code in a file
       script = <<-SCRIPT
-          (#{su_prefix}#{remote_script} < /dev/null; echo $status>#{exit_code_path}) | /usr/bin/tee #{output_path}
-          exit `cat #{exit_code_path}`
+          sh <<WRAPPER
+          (#{su_prefix}#{remote_script} < /dev/null; echo \\$?>#{exit_code_path}) | /usr/bin/tee #{output_path}
+          exit \\$(cat #{exit_code_path})
+          WRAPPER
       SCRIPT
 
       logger.debug("executing script:\n#{script.lines.map { |line| "  | #{line}" }.join}")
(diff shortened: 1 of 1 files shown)


lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -27,8 +27,8 @@ def start
 
       # pipe the output to tee while capturing the exit code in a file
       script = <<-SCRIPT
-          (#{su_prefix}#{remote_script} < /dev/null; echo $?>#{exit_code_path}) | /usr/bin/tee #{output_path}
```
iNecas: `sh` doesn't have `$status` (at least when I tried that). When trying csh, I've seen `$?` working. Which version were you testing with?

```
$ csh --version
tcsh 6.19.00 (Astron) 2015-05-21 (x86_64-unknown-linux) options wide,nls,dl,al,kan,sm,rh,color,filec
$ echo $?
0
```

adamruzicka:
```shell
$ dpkg -l | grep csh
ii  csh                     20110502-2.2ubuntu1    amd64        Shell with C-like syntax
ii  tcsh                    6.20.00-7              amd64        TENEX C Shell, an enhanced version of Berkeley csh
```
It seems as if `csh` is just a symlink to `tcsh` on your machine. Now when I'm thinking about it `$status` in `sh` worked for me because it is probably a symlink to `bash` on my machine...

How about we won't try to support various shells and just wrap the whole script in `/bin/sh -c '#{script}'`? `/bin/sh` should be present and behave roughly the same way everywhere

iNecas: While looking at the csh (for me) unusual syntax, it seems the `/bin/sh -c` option is probably safest. To avoid the escaping, perhaps using the send-data way to transfer this executable script, the same way we actually send the data.

adamruzicka:
Send-data way didn't really work because of [this issue](https://github.com/net-ssh/net-ssh/issues/460), wrapping the control command in a heredoc seems to work

iNecas:
Tested and works nicely. Thanks @adamruzicka

PR #246, opened 2017-05-12
Fixes #19525 - fix user reparation in tests

iNecas:
A change in Foreman's #19463 revealed an issue in remote execution
where we were not saving the user properly when testing permissions.
When auditing was present, it ensured the user was saved, while without
it, the saving of test user fails on password validation.

This commit uses Foreman's helper method to prepare the test user
which has the user preparation set properly.

### DIFF
--- a/test/unit/job_invocation_composer_test.rb
+++ b/test/unit/job_invocation_composer_test.rb
@@ -4,22 +4,9 @@ RemoteExecutionProvider.register(:Mcollective, OpenStruct)
 
 class JobInvocationComposerTest < ActiveSupport::TestCase
   before do
-    permission1 = Permission.find_by(name: 'view_job_templates')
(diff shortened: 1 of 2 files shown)


iNecas:
This should fix issues with running tests of master against develop branch http://ci.theforeman.org/job/test_plugin_pull_request/3683/

iNecas:
There were additional tests failling due to `assert_equal nil` in combination with backtrace silencing, that lead to `undefined method 'split' on nil`

iNecas:
New tests running http://ci.theforeman.org/job/test_plugin_pull_request/3685/

iNecas:
One intermittent test failure on Foreman side, but the jenkins show it should help to get this thing passing

adamruzicka:
Thanks @iNecas !

PR #247, opened 2017-05-17
Fixes #19575 - don't use sudo prefix for script preparation

iNecas:
For the infrastructure files, we use the remote user: the effective
user is for the scope of the actual script


adamruzicka:
Test failure unrelated, merging

adamruzicka:
Thanks @iNecas !

PR #249, opened 2017-05-22
Fixes #17514 - Provide asynchronous variant of ssh provider

adamruzicka:
requires https://github.com/theforeman/foreman-tasks/pull/250 and https://github.com/theforeman/smart_proxy_dynflow/pull/33

## How does it work?
TODO: Extract to manual later
TODO: Maybe some sequence diagrams?

With regular script runner the ssh connection is kept open and data is periodically (every second) read from it. The polling script runner takes a different approach. It copies over to the remote machine the user's script (just `SCRIPT` from now on) and two additional helper scripts (`callback` and `retrieve`). Over the same connection it runs the so-called `control` script, runs the `retrieve` script (to check the command started successfully) and closes the SSH connection. This way, the `SCRIPT` runs on the remote server unsupervised and no persistent connection is maintained. This also means for each run (except the first one) of the `retrieve` a new connection has to be established.

### so-called `control`
Spawns a new `shell` in background which basically just runs `SCRIPT; callback`, redirecting `SCRIPT`'s output into `output_file` and writing `shell`'s pid into a `pid_file`.

### `callback`
It tries to do a POST request using curl to `smart_proxy_dynflow_core`'s API, authenticating with the task's uuid as username and token as password using HTTP basic authentication. This tells the `smart_proxy_dynflow_core` to check on a task using the `retrieve` script.

### `retrieve`
Checks the status of `shell` using its' pid from `pid_file`, reads new parts of `SCRIPT`'s output from `output_file` and updates `position` file (keeps track of number of bytes read from `output_file`). Outputs information whether script is running or not and the new parts of output.

### about token
`PollingScriptRunner` requests an one-time password from `ForemanTasksCore::OtpManager` when it initializes. Similarly `smart_proxy_dynflow_core`'s API asks the `OtpManager` if a task_uuid x token combination is valid.

### dependencies on remote hosts
To work properly, this needs some utilities to reside on the remote host. Those are mostly common tools, but they may be missing in containers.
- GNU coreutils (tail, expr) - doesn't work with busybox tail
- pgrep
- which
- curl

## Configuration
This feature can be enabled on a per-proxy basis by setting `:async_ssh` to `true` in `/etc/smart_proxy_dynflow_core/settings.d/remote_execution_ssh.yml` (TODO: Check if the path is correct). The interval for checking on the remote jobs can be set in the same file under the `runner_refresh_interval` key.

### DIFF
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -9,7 +9,7 @@ module ForemanRemoteExecutionCore
                     :local_working_dir       => '/var/tmp',
                     :kerberos_auth           => false,
                     :async_ssh               => false,
-                    :runner_refresh_interval => 60)
+                    :runner_refresh_interval => nil)
 
   def self.simulate?
     %w(yes true 1).include? ENV.fetch('REX_SIMULATE', '').downcase
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -3,12 +3,13 @@ require 'foreman_tasks_core'
 module ForemanRemoteExecutionCore
   extend ForemanTasksCore::SettingsLoader
   register_settings([:remote_execution_ssh, :smart_proxy_remote_execution_ssh_core],
-                    :ssh_identity_key_file => '~/.ssh/id_rsa_foreman_proxy',
-                    :ssh_user              => 'root',
-                    :remote_working_dir    => '/var/tmp',
-                    :local_working_dir     => '/var/tmp',
-                    :kerberos_auth         => false,
-                    :async_ssh             => false)
+                    :ssh_identity_key_file   => '~/.ssh/id_rsa_foreman_proxy',
+                    :ssh_user                => 'root',
+                    :remote_working_dir      => '/var/tmp',
+                    :local_working_dir       => '/var/tmp',
+                    :kerberos_auth           => false,
+                    :async_ssh               => false,
+                    :runner_refresh_interval => 60)
 
   def self.simulate?
     %w(yes true 1).include? ENV.fetch('REX_SIMULATE', '').downcase
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -24,6 +24,7 @@ module ForemanRemoteExecutionCore
       require 'foreman_remote_execution_core/fake_script_runner'
     else
       require 'foreman_remote_execution_core/script_runner'
+      require 'foreman_remote_execution_core/polling_script_runner'
     end
     require 'foreman_remote_execution_core/actions'
   end
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -7,14 +7,21 @@ module ForemanRemoteExecutionCore
                     :ssh_user              => 'root',
                     :remote_working_dir    => '/var/tmp',
                     :local_working_dir     => '/var/tmp',
-                    :kerberos_auth         => false)
+                    :kerberos_auth         => false,
+                    :async_ssh             => false)
 
   def self.simulate?
     %w(yes true 1).include? ENV.fetch('REX_SIMULATE', '').downcase
   end
 
   def self.runner_class
-    @runner_class ||= simulate? ? FakeScriptRunner : ScriptRunner
+    @runner_class ||= if simulate?
+                        FakeScriptRunner
+                      elsif settings[:async_ssh]
+                        PollingScriptRunner
+                      else
+                        ScriptRunner
+                      end
   end
 
   if ForemanTasksCore.dynflow_present?
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -34,6 +34,7 @@ module ForemanRemoteExecutionCore
       require 'foreman_remote_execution_core/script_runner'
       require 'foreman_remote_execution_core/polling_script_runner'
     end
+    require 'foreman_remote_execution_core/dispatcher'
     require 'foreman_remote_execution_core/actions'
   end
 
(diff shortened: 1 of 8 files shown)


lib/foreman_remote_execution_core/polling_script_runner.rb
```diff
@@ -0,0 +1,133 @@
...
+    def initialize(options = {})
+      super(options)
+      @callback_host = options[:callback_host]
+      @task_id = options[:uuid]
+      @step_id = options[:step_id]
+      try_obtain_otp!
+    end
+
+    def refresh_interval
+      60
```
iNecas: I guess this should be configurable

adamruzicka: At job invocation/template level?

iNecas: Even on proxy settings level would be fine

mccun934: even if it was just a global setting for the plugin, that would be better than hard coded at 60sec

lib/foreman_remote_execution_core/polling_script_runner.rb
```diff
@@ -0,0 +1,138 @@
...
+    def destroy_session
+      if @session
+        @logger.debug("Closing session with #{@ssh_user}@#{@host}")
+        @session.close
+        @session = nil
+      end
+    end
+
+    def try_obtain_otp!
+      @otp = SmartProxyDynflowCore::OtpManager.get_otp(@uuid) if defined? SmartProxyDynflowCore
```
iNecas: What should happen when the `SmartProxyDynflowCore` is not present? Shouldn't the `otp` logic go to `foreman_tasks_core` instead?

lib/foreman_remote_execution_core/polling_script_runner.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    def external_event(event)
+      publish_data(event.data['output'], 'stdout')
+      publish_exit_status(event.data['exit_code'].to_i)
+      super
+    end
+
+    def close
+      super
+      SmartProxyDynflowCore::OtpManager.drop_otp(@uuid, @otp) if @otp
```
iNecas: In the `try_obtain_otp` we are checking if the `SmartProxyDynflowCore` is present. We should be consistent with the expactations, probably extracting this logic to `foreman-tasks-core`

iNecas:
Went though the initial review, will continue on Monday, ideally in combination with some explanations offline, that should transfer to some documentation notes on how it actually works.

lib/foreman_remote_execution_core.rb
```diff
@@ -24,7 +32,9 @@ def self.runner_class
       require 'foreman_remote_execution_core/fake_script_runner'
     else
       require 'foreman_remote_execution_core/script_runner'
+      require 'foreman_remote_execution_core/polling_script_runner'
     end
+    require 'foreman_remote_execution_core/dispatcher'
```
iNecas: Either the path is wrong or there is some file missing in this PR

lib/foreman_remote_execution_core/polling_script_runner.rb
```diff
@@ -0,0 +1,129 @@
...
+      else
+        ':' # Shell synonym for "do nothing"
+      end
+    end
+
+    def callback_script
+      <<-SCRIPT.gsub(/^ +\| /, '')
+      | #!/bin/sh
+      | exit_code=$(cat "#{@exit_code_path}")
+      | url="#{@callback_host}/tasks/#{@task_id}/done"
```
adamruzicka: Also, this URL changes depending on whether smart_proxy_dynflow_core runs inside smart-proxy or as a standalone process...

lib/foreman_remote_execution_core.rb
```diff
@@ -3,18 +3,26 @@
...
-                    :remote_working_dir    => '/var/tmp',
-                    :local_working_dir     => '/var/tmp',
-                    :kerberos_auth         => false)
+                    :ssh_identity_key_file   => '~/.ssh/id_rsa_foreman_proxy',
+                    :ssh_user                => 'root',
+                    :remote_working_dir      => '/var/tmp',
+                    :local_working_dir       => '/var/tmp',
+                    :kerberos_auth           => false,
+                    :async_ssh               => false,
+                    :runner_refresh_interval => 60)
```
iNecas: The `runner_refresh_interval` should be different when using `async_ssh` and when not. With non_async run, I would suggest keeping the 1s interval, so that the life output works as expected.

lib/foreman_remote_execution_core/dispatcher.rb
```diff
@@ -0,0 +1,17 @@
+require 'foreman_tasks_core/runner/dispatcher'
+
+module ForemanRemoteExecutionCore
+  class Dispatcher < ::ForemanTasksCore::Runner::Dispatcher
+
+    def refresh_interval
+      # For the regular runner we don't want to let the user configure
+      #   the refresh interval so live output keeps working
+      @refresh_interval ||= if ForemanRemoteExecutionCore.settings[:async_ssh]
+                              ForemanRemoteExecutionCore.settings[:runner_refresh_interval].to_i
```
iNecas: Overriding the value without taking into account the config value, if specified, doesn't sound good. We could perhaps set the default for `runner_refresh_interval` in settings to `nil` and then apply the defaults here, if not specified. The difference would be we wouldn't overrride the value, if it was specified by the end user in settings

lib/foreman_remote_execution_core.rb
```diff
@@ -3,18 +3,26 @@
...
-                    :remote_working_dir    => '/var/tmp',
-                    :local_working_dir     => '/var/tmp',
-                    :kerberos_auth         => false)
+                    :ssh_identity_key_file   => '~/.ssh/id_rsa_foreman_proxy',
+                    :ssh_user                => 'root',
+                    :remote_working_dir      => '/var/tmp',
+                    :local_working_dir       => '/var/tmp',
+                    :kerberos_auth           => false,
+                    :async_ssh               => false,
+                    :runner_refresh_interval => nil)
```
iNecas: Let's just put a comment here about why we put `nil` here (to set the default based on the async_ssh true/false` later)

iNecas:
One last comment about comment, let's get it in after that

iNecas:
Test failures unrelated, thanks @adamruzicka

PR #250, opened 2017-05-25
Fixes #11936 - Allow authenticating with kerberos

adamruzicka:
For now it requires `net-ssh-krb` to be required into the `smart-proxy` or `smart_proxy_dynflow_core`

Packaging PRs:
- [DEB](https://github.com/theforeman/foreman-packaging/pull/1664)
- [RPM](https://github.com/theforeman/foreman-packaging/pull/1665)

### DIFF
--- a/lib/foreman_remote_execution_core/script_runner.rb
+++ b/lib/foreman_remote_execution_core/script_runner.rb
@@ -285,7 +285,7 @@ module ForemanRemoteExecutionCore
 
     def available_authentication_methods
       methods = %w(pubkey) # Always use pubkey auth as fallback
-      if settings.kerberos_auth
+      if settings[:kerberos_auth]
         if defined? Net::SSH::Kerberos
           methods.unshift('gssapi-with-mic')
         else
--- a/lib/foreman_remote_execution_core/script_runner.rb
+++ b/lib/foreman_remote_execution_core/script_runner.rb
@@ -2,6 +2,7 @@ require 'net/ssh'
 begin
   require 'net/ssh/krb'
 rescue LoadError
+  $stderr.puts 'Failed to load net/ssh/krb, kerberos authentication won\'t be available'
 end
 
 module ForemanRemoteExecutionCore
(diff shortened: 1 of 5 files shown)


iNecas:
I guess there should be also a PR to https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/settings.d/remote_execution_ssh.yml.example, to mention the config option

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -279,5 +282,17 @@ def check_expecting_disconnect
         @expecting_disconnect = true
       end
     end
+
+    def available_authentication_methods
+      methods = %w(pubkey) # Always use pubkey auth as fallback
+      if settings.kerberos_auth
```
iNecas: I'm getting `NoMethodError - undefined method `kerberos_auth' for #<Hash:0x005609203c17f0>`, tried with smart proxy 1.15 as well as develop, with latest versions of plugins. Am I missing something?

adamruzicka:
Had it prepared, forgot to open the [PR](https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/33).

iNecas:
After workarounding the issue with hash[], it was working nicely. Ready to merge once this issue is clarified

lib/foreman_remote_execution_core/script_runner.rb
adamruzicka: Do you run smart_proxy_dynflow_core as a separate process or as a part of smart proxy? From what I remember I expected `settings` to be an `OpenStruct` whereas in your case it is a `Hash`. Will investigate tomorrow

adamruzicka:
Comments addressed

iNecas:
Seems like Jenkis is having some issues, let's wait till it's operating again

adamruzicka:
[test]

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -1,4 +1,9 @@
 require 'net/ssh'
+begin
+  require 'net/ssh/krb'
+rescue LoadError
+  $stderr.puts 'Failed to load net/ssh/krb, kerberos authentication won\'t be available'
```
iNecas: One last thing: for the users that don't plan to use keberos authentication, this produces an error message every time they start the service. I guess we could just be silent here and include the cause of the error ('net/ssh/krb' missing, in the log message when `Net::SSH::Kerberos` is not authenticated.

iNecas:
Thanks @adamruzicka

iNecas:
I guess we also need an installer update

adamruzicka:
It's in the works https://github.com/theforeman/puppet-foreman_proxy/pull/351

PR #252, opened 2017-05-30
Fixes #18356 - Ensure UTF-8 encoding

adamruzicka:
Let's override the default locale with `C.UTF-8`. Users wanting to use their own locales can set them in their scripts, as long as the encoding is `UTF-8`.

Requires https://github.com/theforeman/foreman-tasks/pull/254

### DIFF
--- a/lib/foreman_remote_execution_core/script_runner.rb
+++ b/lib/foreman_remote_execution_core/script_runner.rb
@@ -28,6 +28,7 @@ module ForemanRemoteExecutionCore
       # pipe the output to tee while capturing the exit code in a file
       script = <<-SCRIPT.gsub(/^\s+\| /, '')
       | sh <<WRAPPER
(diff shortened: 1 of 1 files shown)


lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -28,6 +28,7 @@ def start
       # pipe the output to tee while capturing the exit code in a file
       script = <<-SCRIPT.gsub(/^\s+\| /, '')
       | sh <<WRAPPER
+      | LC_ALL=C.UTF-8
```
iNecas: I guess this is not needed anymore

iNecas:
Tested and worked well, one request

adamruzicka:
Comments addressed

iNecas:
Thanks @adamruzicka

PR #254, opened 2017-06-14
Fixes #20016 - Allow jobs to timeout

adamruzicka:
requires https://github.com/theforeman/foreman-tasks/pull/258

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -53,7 +53,7 @@ module Api
           param :bookmark_id, Integer, :required => false
           param :search_query, Integer, :required => false
           param :description_format, String, :required => false, :desc => N_('Override the description format from the template for this invocation only')
-          param :timeout_interval, Integer, :required => false, :desc => N_('Override the timeout interval from the template for this invocation only')
+          param :execution_timeout_interval, Integer, :required => false, :desc => N_('Override the timeout interval from the template for this invocation only')
         end
       end
 
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -53,6 +53,7 @@ module Api
           param :bookmark_id, Integer, :required => false
           param :search_query, Integer, :required => false
           param :description_format, String, :required => false, :desc => N_('Override the description format from the template for this invocation only')
+          param :timeout_interval, Integer, :required => false, :desc => N_('Override the timeout interval from the template for this invocation only')
         end
       end
 
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -78,7 +78,7 @@
                     </div>
 
                     <div class="advanced hidden">
-                      <%= number_f job_template_fields, :timeout_interval, :value => f.object.timeout_interval || job_template.timeout_interval, :label => _('Timeout to kill'), :help_inline => N_('Time in seconds from the start on the remote host after which the job should be killed.') %>
+                      <%= number_f job_template_fields, :execution_timeout_interval, :value => f.object.execution_timeout_interval || job_template.execution_timeout_interval, :label => _('Timeout to kill'), :help_inline => N_('Time in seconds from the start on the remote host after which the job should be killed.') %>
                     </div>
                   <% end %>
                 </fieldset>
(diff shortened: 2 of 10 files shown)


app/views/job_invocations/_form.html.erb
```diff
@@ -76,6 +76,10 @@
                       <% end %>
                       <%= render :partial => 'description_fields', :locals => { :f => f, :job_template => job_template, :disabled => job_template != selected_templates_per_provider[provider_type] } %>
                     </div>
+
+                    <div class="advanced hidden">
+                      <%= number_f job_template_fields, :timeout_interval, :value => f.object.timeout_interval || job_template.timeout_interval, :label => _('Timeout interval'), :help_inline => N_("Time in seconds after which the job should be killed.") %>
```
iNecas: what about calling it `Timeout to kill` so that it's clear what kind of timeout we have in mind (and be clear about it's consequences)?

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -52,5 +52,8 @@
     <% if job_invocation.task && job_invocation.task.delayed? || job_invocation.task.recurring? %>
       <b><%= _("Scheduled to start at") %></b>: <%= job_invocation.task.start_at.try(:in_time_zone) %><br>
     <% end %>
+    <% unless job_invocation.timeout_interval.nil? %>
+      <b><%= _("Set to timeout after") %></b>: <%= job_invocation.timeout_interval %> <%= _('seconds') %><br>
```
iNecas: What about `"Timeout to kill after"`, without the `set`

app/views/job_invocations/_form.html.erb
iNecas: Also, we should be explicit in the description, that the time counts since the execution starts on remote host, not from the time of the job_invocation starting

adamruzicka:
Comments addressed

iNecas:
Tested and works well: let's wait until the foreman-tasks is released, so that we can bump a version requirement on foreman-tasks

iNecas:
Rubocop doesn't seem to be happy with something + foreman-tasks-core-0.1.3 was released with the required changes: please add the dependency on it from the foreman_remote_execution_core

adamruzicka:
Seems like Jenkins is still having some issues

ares:
[test]

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -53,6 +53,7 @@ def show; end
           param :bookmark_id, Integer, :required => false
           param :search_query, Integer, :required => false
           param :description_format, String, :required => false, :desc => N_('Override the description format from the template for this invocation only')
+          param :timeout_interval, Integer, :required => false, :desc => N_('Override the timeout interval from the template for this invocation only')
```
iNecas: One last thing: what about being more explicit with `execution_timeout_interval`? To make it clear it's not a connection timeout interval (and in case we wanted to expose the connection timeout later)

iNecas:
Running tests manually, until the master tests are green http://ci.theforeman.org/job/test_plugin_pull_request/3871/

adamruzicka:
Rebased, might need to trigger the tests again

iNecas:
Here we go http://ci.theforeman.org/job/test_plugin_pull_request/3873/

foreman_remote_execution_core.gemspec
```diff
@@ -17,6 +17,6 @@ DESC
   s.files = Dir['lib/foreman_remote_execution_core/**/*'] +
             ['lib/foreman_remote_execution_core.rb', 'LICENSE']
 
-  s.add_runtime_dependency('foreman-tasks-core', '~> 0.1.0')
+  s.add_runtime_dependency('foreman-tasks-core', '>= 0.1.3', '< 2')
```
iNecas: Nitpick: shouldn't this be `< 1`? Or perhaps we don't even need to keep the top restriction

adamruzicka: Dropped it altogether

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -8,6 +8,8 @@
 
 module ForemanRemoteExecutionCore
   class ScriptRunner < ForemanTasksCore::Runner::Base
+    attr_reader :execution_timeout_interval
```
iNecas: It seems we still need to override the `timeout_interval` method, to return the `@execution_timeout_interval` value

adamruzicka:
Requires https://github.com/theforeman/foreman-tasks/pull/265

iNecas:
Works well, especially with https://github.com/theforeman/foreman-tasks/pull/265 :)

PR #255, opened 2017-06-15
Fixes #19920 - Remove job invocations when deleting tasks

adamruzicka:


iNecas:
This approach actually doesn't work, as the task cleanup does the deletion in bulk: bypassing the dependent destroy flags.

What do you think about turning the job into "archived" mode, where we would still have a notion, that the task was completed, but we don't have the full data: this would also let the user to get the data back if needed, using the backup of tasks, that is going to be added sooner or later?

Alternatively, we would need to enhance the tasks:cleanup task to remove the jobs that were orphaned by the task cleanup (I expect we could use the standard destroy there, as the amount of job invocations should be lower by order of magnitude to the amount of tasks in the system). But I would probably still prefer the archived state here, and handle the full deletion of the job invocation in separate issue.

iNecas:
Thanks @adamruzicka

PR #256, opened 2017-06-19
Fixes #20018 - Reap processes after error when opening connection

adamruzicka:
This is a workaround until this gets fixed in `net/ssh`

### DIFF
--- a/lib/foreman_remote_execution_core/script_runner.rb
+++ b/lib/foreman_remote_execution_core/script_runner.rb
@@ -97,9 +97,30 @@ module ForemanRemoteExecutionCore
 
     def session
       @session ||= begin
-                     @logger.debug("opening session to #{@ssh_user}@#{@host}")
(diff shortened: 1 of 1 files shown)


lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -97,11 +97,32 @@ def close
...
+      yield
+    rescue => e
+      reap
+      raise e
+    end
+
+    def reap
+      reaped = 0
+      # While there are any dead child processes without waiting for the ones that are running
+      while Process.waitpid(-1, Process::WNOHANG)
```
iNecas: I'm a bit worried about this causing some race-conditions, when it would fall into a time where some process finished, but `io.close` was not called just yet. As a conseuqnece, the `io.close` would for example not set the `$?` variable when calling the proper `io.close`. I can't think of any other way than fixing it directly in net ssh

PR #257, opened 2017-06-20
Is there any possibility to user FreeIPA account instead of creating ssh keys?

zerocool1992:
Any ideas hot make this?


adamruzicka:
Please don't misuse github PRs for asking questions. Feel free to reach us by any means described in the [support part of Foreman's docs](https://theforeman.org/support.html) and we'll be happy to help you through those channels.

PR #259, opened 2017-07-10
Fixes #20260 - Do not duplicate hosts belonging to host collections

adamruzicka:


app/models/targeting.rb
```diff
@@ -38,8 +38,9 @@ def resolve_hosts!
     self.search_query = bookmark.query if dynamic? && bookmark.present?
     self.resolved_at = Time.zone.now
     self.validate!
-    # avoid validation of hosts objects - tey will be loaded for no reason.
-    host_ids = User.as(user.login) { Host.authorized(RESOLVE_PERMISSION, Host).search_for(search_query).pluck(:id) }
+    # avoid validation of hosts objects - they will be loaded for no reason.
+    #   pluck(:id) returns duplicate results for HostCollections
```
mmoll: commented code? :)

iNecas: Since it's part of the explanation on why `pluck(:id)` is there, I'm ok with that :)

iNecas:
Tests running at http://ci.theforeman.org/job/test_plugin_pull_request/3876/

iNecas:
The test failures are not related. Thanks @adamruzicka

PR #260, opened 2017-07-10
Fixes #20264 - less version restriction on foreman-tasks

iNecas:


adamruzicka:
Merged, thanks @iNecas !

PR #262, opened 2017-07-12
Fixes #20107 - Resolve targeting only once

adamruzicka:


iNecas:
It seems the rubocop is not heapy about something there

iNecas:
Works as expected. Bonus points for additional tests!

PR #263, opened 2017-07-13
Fixes #19937 - Better job cancel handling

adamruzicka:
requires https://github.com/theforeman/foreman-tasks/pull/267

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -124,7 +124,7 @@ module RemoteExecutionHelper
                          :method => :post)
       buttons << link_to(_('Abort Job'), abort_foreman_tasks_task_path(task),
                          :class => 'btn btn-danger',
-                         :title => _('Try to abort the job'),
+                         :title => _('Try to abort the job without waiting for the results from the remote hosts'),
                          :disabled => !task.cancellable?,
                          :method => :post)
     end
@@ -143,6 +143,11 @@ module RemoteExecutionHelper
       buttons << link_to(_('Cancel Job'), cancel_foreman_tasks_task_path(task),
                          :class => 'btn btn-danger',
                          :title => _('Try to cancel the job on a host'),
+                         :disabled => !task.cancellable?,
+                         :method => :post)
+      buttons << link_to(_('Abort Job'), abort_foreman_tasks_task_path(task),
+                         :class => 'btn btn-danger',
+                         :title => _('Try to abort the job on a host without waiting for its result'),
                          :disabled => !task.cancellable?,
                          :method => :post)
     end
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -122,6 +122,11 @@ module RemoteExecutionHelper
                          :title => _('Try to cancel the job'),
                          :disabled => !task.cancellable?,
                          :method => :post)
+      buttons << link_to(_('Abort Job'), abort_foreman_tasks_task_path(task),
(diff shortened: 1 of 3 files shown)


app/helpers/remote_execution_helper.rb
```diff
@@ -122,6 +122,11 @@ def job_invocation_task_buttons(task)
                          :title => _('Try to cancel the job'),
                          :disabled => !task.cancellable?,
                          :method => :post)
+      buttons << link_to(_('Abort Job'), abort_foreman_tasks_task_path(task),
+                         :class => 'btn btn-danger',
+                         :title => _('Try to abort the job'),
```
iNecas: The title is not very descriptive, what about `Abort the the job: trying to cancel without waiting for the results from the remote hosts)` or something like that?

iNecas:
I was still able to get into non-cancellable state (see screenshots).

What I did:

1. spawn a remote execution job
2. restart the rails server as well as the smart proxy while the job is running

non-cancellable-task-1
non-cancellable-task-2

Although the job is marked as failed, the host jobs are marked as running.

What was the reproducer you were running it against?

adamruzicka:
I went with create a job, wait for the tasks to show up on the proxy, stop the proxy, try to cancel the job. Guess I got sidetracked with events not being propagated to sub plans.

If we want to tackle the rails restart too, we need to revisit resuming. Current "if everything failed in plan start over else fail and dont care about subplans any more" behavior obviously doesn't work here.

iNecas:
What if we just started polling the sub-tasks in that case?

adamruzicka:
The issue you described is now handled in https://github.com/theforeman/foreman-tasks/pull/267

adamruzicka:
The issue you described is now handled in https://github.com/theforeman/foreman-tasks/pull/267

iNecas:
Thanks, I can confirm it works for me after the change. The last thing I would suggest is improving the error messaging on the abort, currently I get something like this. The user probably doesn't care about the exception class, neither it's true it failed to initialize.

job-abort

adamruzicka:
screenshot_20170801_123031 is this good enough or should I tweak the message a bit more?

iNecas:
That would work for me

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -89,20 +90,28 @@ def continuous_output_providers
...
       def fill_continuous_output(continuous_output)
-        fill_planning_errors_to_continuous_output(continuous_output) unless exit_status
+        final_timestamp = (continuous_output.last_timestamp || task.ended_at).to_f + 1
+
+        if task.state == 'stopped' && task.result == 'cancelled'
+          continuous_output.add_output(_('Job cancelled by user'), 'debug', final_timestamp)
+        else
+          fill_planning_errors_to_continuous_output(continuous_output) unless exit_status
+        end
+
```
iNecas: Please add this code after the line 103-105, to get the error messages about task cancelling to the end of the output:

```ruby
        delegated_output.fetch('result', []).each do |raw_output|
          continuous_output.add_raw_output(raw_output)
        end

        final_timestamp = (continuous_output.last_timestamp || task.ended_at).to_f + 1

        if task.state == 'stopped' && task.result == 'cancelled'
          continuous_output.add_output(_('Job cancelled by user'), 'debug', final_timestamp)
        else
          fill_planning_errors_to_continuous_output(continuous_output) unless exit_status
        end
```

iNecas:
Thanks @adamruzicka

PR #266, opened 2017-07-31
Fixes #19936 - Don't call Host::Base.authorized

ShimShtein:
It will fail if the user has filters defined in roles.


iNecas:
@ares as the original author of this code + permissions boss, would you mind reviewing?

ares:
:+1:. I'm not entirely sure what changed since I remember this worked before but this makes more sense now. We never used REX on discovered hosts anyway.

iNecas:
Testd and works well, test failures unrelated. Thanks @ShimShtein

PR #267, opened 2017-08-01
Fixes #20466 - Use polling in RunHostsJob

adamruzicka:
The default poll interval set in Dynflow is 10 seconds. This might cause a slight responsiveness drop for smaller jobs, but is probably way too frequent for jobs running on thousands of hosts. Maybe computing the value from the number of hosts (which we know before we start polling) would be the way to go.


iNecas:
Thanks @adamruzicka

PR #268, opened 2017-08-14
Fixes #17411 - Remove cause for error in recurring logics

adamruzicka:


adamruzicka:
Both commits include a tests which tests the thing being fixed. Isn't it enough?

iNecas:
Sry, my bad: the tests are there (not sure where I was looking). Tested and works well. Merging now, thanks @adamruzicka

PR #269, opened 2017-08-17
Fixes #20627 - Set up concurrency control properly

adamruzicka:


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -40,17 +40,17 @@ def batch(from, size)
...
-      def hosts
-        JobInvocation.find(input[:job_invocation_id]).targeting.hosts.order(:name, :id)
+      def hosts(job_invocation = nil)
+        (job_invocation || JobInvocation.find(input[:job_invocation_id])).targeting.hosts.order(:name, :id)
       end
 
       def set_up_concurrency_control(invocation)
         limit_concurrency_level invocation.concurrency_level unless invocation.concurrency_level.nil?
-        distribute_over_time invocation.time_span unless invocation.time_span.nil?
+        distribute_over_time(invocation.time_span, total_count(invocation)) unless invocation.time_span.nil?
```
iNecas: I know reusing the `total_count` is temping, but in this case, I would probably go just with simple `invocation.targeting.hosts.count`, to keep the rest of the code cleaner.

adamruzicka: It indeed looks much cleaner when sacrificing a bit of DRYness

adamruzicka:
Good catch,should be good now

iNecas:
I'm happy with the code, but 👮 has come comments. Tested and works well. Waiting jenkins.

iNecas:
One thing that it's still there is the incomplete info on time-span in host details (tested with 60 seconds)
rex-time-span-info

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -47,7 +47,7 @@
       <b><%= _("Concurrency level limited to") %></b>: <%= job_invocation.concurrency_level %> <%= _('tasks at a time') %><br>
     <% end %>
     <% if job_invocation.time_span %>
-      <b><%= _("Set to distribute over") %></b>: <%= job_invocation.concurrency_level %> <%= _('minutes') %><br>
+      <b><%= _("Set to distribute over") %></b>: <%= job_invocation.time_span %> <%= _('minutes') %><br>
```
iNecas: We talk about seconds in the invocation form and minutes here, which one is correct?

iNecas:
Thanks @adamruzicka

PR #272, opened 2017-08-24
Prevent from SSH connection close during reboot

lzap:
This might be better than:

https://github.com/theforeman/community-templates/pull/415

### DIFF
--- a/app/views/templates/ssh/power_action.erb
+++ b/app/views/templates/ssh/power_action.erb
@@ -14,8 +14,12 @@ template_inputs:
 
 echo <%= input('action') %> host && sleep 3
 <%= case input('action')
+      when 'restart_at'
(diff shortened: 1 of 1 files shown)


iNecas:
Followup on discussion https://github.com/theforeman/community-templates/pull/415#issuecomment-324870571, this has implications on the existing functionality without a prove that it actually helps with the issues you're seeing. Therefore I suggest waiting until it's confirmed it is really caused by this (which I don't think is the case)

lzap:
Yeah it also assumes that `atd` is installed and running. I amended this to be an option rather than default action. They confirmed improvements in failures.

iNecas:
Closing, as it turned out the original issue was somewhere else and there is already a mechanism to handle the restarts gracefully.

PR #275, opened 2017-09-26
Fixes #21113 - require templates asset

iNecas:


iNecas:
After https://github.com/theforeman/foreman/pull/4950#pullrequestreview-72160391, we don't need this at all

PR #276, opened 2017-10-03
Fixes #21178 - correct strong parameters on Nic::Interface

iNecas:
The current setting of the `parameter_filter` was not allowing the
:execution parameter on the interface correctly.


iNecas:
The tests are green Since this change is not breaking (not more that it was before), I would be for getting this in now and it will get fixed fully once the foreman part is merged as well. What do you think @adamruzicka

adamruzicka:
Ok, let's do that

adamruzicka:
Thanks @iNecas

PR #281, opened 2017-10-09
Fixes #21242 - revert using prepend instaed of alias_method_chain

iNecas:
This is only intended for the 1.15 and 1.16, the nightlies should
keep the change and the proper fix should be done in Katello
to make it work again.

### DIFF
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -1,7 +1,14 @@
 module ForemanRemoteExecution
   module HostsHelperExtensions
-    def multiple_actions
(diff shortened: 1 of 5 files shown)


adamruzicka:
Is there a way to make the tests pass?

iNecas:
Running against 1.16 stable (http://ci.theforeman.org/job/test_plugin_pull_request/4193/) and 1.15-stable (http://ci.theforeman.org/job/test_plugin_pull_request/4194/)

evgeni:
The patch on top of 1.3.4 makes it work fine again with Foreman 1.15 and Katello 3.4, thanks!

iNecas:
1.3.5 released and packaging PR opened here https://github.com/theforeman/foreman-packaging/pull/1852

PR #282, opened 2017-10-11
Fixes #21272 - Unscoped templates

adamruzicka:


ares:
I wonder if user should see such template invocations in the first place. How is this issue triggered?

adamruzicka:
Someone encountered this after upgrading satellite. Job Templates didn't have any orgs/locs assigned and accessing a job from before the upgrade would raise the undefined method error

iNecas:
[test]

ares:
Ok, sounds reasonable

iNecas:
Test failures unrelated (filed an issue and proposed fix for that in infra http://projects.theforeman.org/issues/21318). Thanks @adamruzicka

PR #283, opened 2017-10-12
Fixes #21308, #21310 - Handle power actions on remote host better

adamruzicka:
- d770b64 fixes the missing constant
- 4dd59b9 makes the `#with_disconnect_handling` method rescue `IOError`, which is raised when the remote host reboots (or at least on ruby-2.4.1)


iNecas:
[test]

adamruzicka:
The tests are green

iNecas:
Thanks @adamruzicka

PR #284, opened 2017-10-13
Refs #21168 - raise => false only for Rails 5

iNecas:
Rails 4 doesn't raise any issue when callback is not found and it
doesn't even know the raise option. Due to this, it considers the
additional option being a sign of need for re-definition of the
original callback, instead of just making it as skipped.

This patch sets the `raise => false` only for Rails 5, where it's
needed.


adamruzicka:
@evgeni thanks for testing it

This now has two ACKs and green tests, I don't see any reason for delaying it any longer.

adamruzicka:
Thanks @iNecas !

iNecas:
1.3.6 released. Thanks for reviews and testing. I hope we just got over with this one :)

evgeni:
@iNecas thanks! will you do the needed packaging dance in theforeman/foreman-packaging?

iNecas:
https://github.com/theforeman/foreman-packaging/pull/1859

PR #286, opened 2017-10-18
Fixes #18967 - Display output for queued jobs properly

adamruzicka:


iNecas:
Thanks @adamruzicka

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -71,6 +71,10 @@ def create
       param :host_id, :identifier, :required => true
       param :since, String, :required => false
       def output
+        if @nested_obj.task.delayed?
```
iNecas: Actually, it seems that the task keeps set as `delayed?` even after the task actually finished. Therefore, it means it would not never return the real result, even after finished.

adamruzicka: Ohh, `delayed?` actually means something else. I forgot it means "was the task scheduled to execute in the future?" instead of "is it waiting to be executed in the future?"

PR #287, opened 2017-10-19
Fixes #20496 - Add API for canceling jobs

adamruzicka:
- [x] tests

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -101,10 +101,11 @@ module Api
         unless Authorizer.new(User.current, :collection => [@job_invocation.task]).can? :edit_foreman_tasks
           render :json => { :error => _('The user is not authorized to cancel the job') },
                  :status => 403
+          return
         end
 
         if @job_invocation.task.cancellable?
-          result = @job_invocation.cancel(params.fetch(['force'], false))
+          result = @job_invocation.cancel(params.fetch('force', false))
           render :json => { :cancelled => result, :id => @job_invocation.id }
         else
           render :json => { :message => _('The job could not be cancelled.') },
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -119,7 +119,7 @@ module Api
         when 'output'
           :view
         when 'cancel'
-          :view
+          :edit
         else
           super
         end
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -94,7 +94,7 @@ module Api
         end
       end
 
-      api :POST, '/job_invocations/:id/cancel'
+      api :POST, '/job_invocations/:id/cancel', N_('Cancel job invocation')
       param :id, :identifier, :required => true
       param :force, :bool
       def cancel
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -119,6 +119,8 @@ module Api
         case params[:action]
         when 'output'
           :view
+        when 'cancel'
+          :cancel
         else
           super
         end
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -98,12 +98,6 @@ module Api
       param :id, :identifier, :required => true
       param :force, :bool
       def cancel
-        unless Authorizer.new(User.current, :collection => [@job_invocation.task]).can? :edit_foreman_tasks
-          render :json => { :error => _('The user is not authorized to cancel the job') },
-                 :status => 403
-          return
-        end
-
         if @job_invocation.task.cancellable?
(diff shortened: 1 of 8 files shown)


app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -85,12 +85,32 @@ def output
         end
       end
 
+      api :POST, '/job_invocations/:id/cancel'
+      param :id, :identifier, :required => true
+      param :force, :bool
```
iNecas: Please add a description of what that means.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -85,12 +85,32 @@ def output
...
+      end
+
       private
 
       def action_permission
         case params[:action]
         when 'output'
           :view
+        when 'cancel'
+          :view
```
iNecas: Should this be `:view`?

adamruzicka: The user needs to be able to view the job invocation and edit tasks. I can create a dedicated permission for editing job invocations if needed, so far we don't have one.

iNecas: +1 for addiing it.

iNecas:
Tests are failing

adamruzicka:
The tests are finally green

lib/foreman_remote_execution/engine.rb
```diff
@@ -56,6 +56,7 @@ class Engine < ::Rails::Engine
                                                 'api/v2/job_invocations' => [:create] }, :resource_type => 'JobInvocation'
           permission :view_job_invocations, { :job_invocations => [:index, :show, :auto_complete_search], :template_invocations => [:show],
                                               'api/v2/job_invocations' => [:index, :show, :output] }, :resource_type => 'JobInvocation'
+          permission :cancel_job_invocations, { 'api/v2/job_invocations' => [:cancel] }, :resource_type => 'JobInvocation'
```
iNecas: You're adding `:cancel_job_invocation` here, while adding `:edit_job_invocations` permission later to the `MANAGER_PERMISSIONS`.  I would probably stick with `:edit_job_invocations` here

adamruzicka: There is a double-save happening when a job invocation is being created by the composer (probably due to associations). This means a user has to have both `:create` and `:edit` permissions to be able to run jobs. I would like to go with `:cancel` before we get this sorted out

iNecas: I see, we should then unify the way how the UI and API uses the permission, by using `:cancel_job_invocations` for the UI as well

app/views/template_invocations/show.html.erb
```diff
@@ -9,7 +9,7 @@
                      link_to_function(_('Toggle STDERR'), '$("div.line.stderr").toggle()', :class => 'btn btn-default'),
                      link_to_function(_('Toggle STDOUT'), '$("div.line.stdout").toggle()', :class => 'btn btn-default'),
                      link_to_function(_('Toggle DEBUG'), '$("div.line.debug").toggle()', :class => 'btn btn-default')) %>
-    <%= button_group(template_invocation_task_buttons(@template_invocation_task)) %>
+    <%= button_group(template_invocation_task_buttons(@template_invocation_task, @template_invocation.job_invocation)) %>
```
iNecas: I don't see the `job_invocation_task_buttons` to have updated the list of arguments. Am I missing something here, or this PR is? :)

iNecas: Taking back: I was looking at wrong method. NM

app/helpers/remote_execution_helper.rb
```diff
@@ -116,7 +116,7 @@ def job_invocation_task_buttons(task)
                          :class => 'btn btn-default',
                          :title => _('See the last task details'))
     end
-    if authorized_for(:permission => :edit_foreman_tasks, :auth_object => task, :authorizer => task_authorizer)
+    if authorized_for(:permission => :cancel_job_invocations, :auth_object => job_invocation)
       buttons << link_to(_('Cancel Job'), cancel_foreman_tasks_task_path(task),
```
iNecas: It doesn't work as expected: I had an user that had only rex manager rigths: I got an error

```
Please request one of the required permissions listed below from a Foreman administrator:

    edit_foreman_tasks

```
We probably need a separate end-point for the job cancellation, and not use the task end point for it.

adamruzicka:
@iNecas could you please take another look?

app/controllers/job_invocations_controller.rb
```diff
@@ -72,6 +72,26 @@ def preview_hosts
...
+                         _('Trying to cancel the job')
+                       end
+    else
+      flash[:warning] = if params[:force]
+                          _('The job cannot be aborted at the moment.')
+                        else
+                          _('The job cannot be cancelled at the moment.')
+                        end
+    end
+    redirect_to :back
```
iNecas: This doesn't work in rails 5, see https://github.com/theforeman/foreman-tasks/pull/308/files on how to transition that

iNecas:
Works as expected. The tests were passing before and I don't see the latest updates to cause any additional failure. Merging now

PR #289, opened 2017-10-25
Fixes #21457 - Allow password based auth

adamruzicka:


iNecas:
Rubocop is not happy

iNecas:
Also, waiting for future and recurring logic to work with this.

adamruzicka:
Starting with ed37691 this PR requires https://github.com/theforeman/foreman-tasks/pull/297 .

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -38,6 +41,10 @@ def create_sub_plans
         end
       end
 
+      def finalize
+        input.delete(:secrets)
```
iNecas: Is this still needed, with `HideSecrets` in place?

adamruzicka: The `HideSecrets` middleware only prevents showing the secrets in the webui. Anyone with access to the database could still look there to find the secrets. For this reason we should delete them by hand when they're not needed anymore

iNecas: Oh, I"ve seen https://github.com/theforeman/foreman-tasks/pull/297/files#diff-03dde90ec7e9caa322015dee15f4e956R103 also doing some wiping, but now I see it is run_host_job vs. run_hosts_job. Nm then

adamruzicka:
Update:
- [ ] (imagine this box half checked) successful run ditches the secrets properly: however, if the execution fails they stay in the db, but at least they're encrypted now
- [x] password and passphrase are now stored as attributes of job invocation and are encrypted before hitting the db
- [ ] encrypted parameters: to be done

ares:
Note that http://projects.theforeman.org/issues/3378 mentiones smart variables. I don't think we should further extend parameters. They support hidden values already. IMHO this is good to be shipped without parameters encryption. The encryption does not provide much, the value could be easily decrypted and read.

app/models/concerns/foreman_remote_execution/encrypted_value_extensions.rb
```diff
@@ -0,0 +1,22 @@
+module ForemanRemoteExecution
```
iNecas: Any reason we can't use the `Encryptable` module from the Foreman? https://github.com/theforeman/foreman/blob/b5eefd928921ddab7fce2b03882e1b8afa5f37f5/app/models/concerns/encryptable.rb#L2

iNecas:
I believe there is an added value for the encryption, as on can't at least decode the credentials from db dump. Anyway, I would be good with just using the secret params for that, and I don't insist on this being blocker for this PR. However, I what I would like to see before we proceed is:

1. global setting for default credentials
2. the setting overridable by host params, similarly as we have for `remote_execution_ssh_user` and others, for the sake of consistency

adamruzicka:
Added encrypted global setting and possibility to override it with host params

ares:
needs rebasing :-/

adamruzicka:
@ares rebased

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -47,10 +48,17 @@ def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExe
         raise _('Failed rendering template: %s') % renderer.error_message unless script
 
         provider = template_invocation.template.provider
-        hostname = provider.find_ip_or_hostname(host)
+
+        secrets = { :ssh_password => job_invocation.password || provider.ssh_password(host),
+                    :key_passphrase => job_invocation.key_passphrase || provider.ssh_key_passphrase(host) }
```
iNecas: This gives me `Failed to initialize: Dynflow::ExecutionPlan::Steps::Error - private method `ssh_password' called for SSHExecutionProvider:Class`

iNecas:
I've released the foreman-tasks 0.11.1 with the required changes. I've tested this (including future and recurring execution) and everything works well. Please bump the version requirement here and rebase, so that we can get the green tests.

iNecas:
Test failure related

iNecas:
Thanks @adamruzicka

ares:
:clap: :+1: :1st_place_medal:

PR #290, opened 2017-10-25
Fixes #16991 - Allow setting ssh log level

adamruzicka:


iNecas:
Rubocop is not happy

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -146,6 +146,7 @@ def ssh_options
       ssh_options[:paranoid] = true
       ssh_options[:auth_methods] = available_authentication_methods
       ssh_options[:user_known_hosts_file] = prepare_known_hosts if @host_public_key
+      ssh_options[:verbose] = settings[:ssh_log_level]
```
iNecas: While testing, I've noticed that the ssh uses a `Logger` around `STDERR`, if not provided with different `:logger` option: therefore, I was not seeing the log messages from SSH in the logs.

I see that using the existing logger we have directly is probably not what we want, as even if we are interested into
`info`, or even `debug` log messages from proxy, we might not need the `debug` messages from the ssh.

What about having a logger wrapper around our own logger, to pass the messages to the logger only if the level is >= as `ssl_log_level`. This would require the standard logger to be configured more verbosely that the ssl logger, to get the messages propagated: I guess we could just refuse to start if the `:ssl_log_level > logger.level`.

lib/foreman_remote_execution_core.rb
```diff
@@ -10,12 +10,37 @@ module ForemanRemoteExecutionCore
...
+  end
+
+  def self.validate_ssh_log_level!
+    wanted_level = @settings[:ssh_log_level].to_s
+    unless SSH_LOG_LEVELS.include? wanted_level
+      raise "Wrong value '#{@settings[:ssh_log_level]}' for ssh_log_level, must be one of #{SSH_LOG_LEVELS.join(', ')}"
+    end
+
+    # regular log levels correspond to upcased ssh logger levels
+    ssh, regular = [wanted_level, SmartProxyDynflowCore::SETTINGS.log_level.to_s.downcase].map do |wanted|
```
iNecas: This is breaking on my foreman server side when trying to run the server with this code:

```
/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/lib/foreman_remote_execution_core.rb:35:in `validate_ssh_log_level!': uninitialized constant ForemanRemoteExecutionCore::SmartProxyDynflowCore (NameError)
	from /home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/lib/foreman_remote_execution_core.rb:24:in `validate_settings!'
	from /home/inecas/Projects/ws/foreman-rex/foreman-tasks/lib/foreman_tasks_core/settings_loader.rb:41:in `initialize_settings'
	from /home/inecas/Projects/ws/foreman-rex/foreman-tasks/lib/foreman_tasks_core/settings_loader.rb:31:in `setup_settings'
	from /home/inecas/Projects/ws/foreman-rex/foreman-tasks/lib/foreman_tasks/engine.rb:128:in `block (2 levels) in <class:Engine>'
	from /home/inecas/Projects/ws/foreman-rex/foreman-tasks/lib/foreman_tasks/engine.rb:124:in `each_key'
	from /home/inecas/Projects/ws/foreman-rex/foreman-tasks/lib/foreman_tasks/engine.rb:124:in `block in <class:Engine>'

```

iNecas:
Run rubocop and passes. Tested and works well

iNecas:
Just to be sure, I've triggered a job to run the tests: to make sure we don't break rubocop with this PR http://ci.theforeman.org/job/test_plugin_pull_request/4759/

iNecas:
Everything seems ok now, test failures unrelated. Thanks @adamruzicka

PR #292, opened 2017-10-26
Fixes #21477 - Deal with rubocop 0.51

adamruzicka:


iNecas:
Running the tests here http://ci.theforeman.org/job/test_plugin_pull_request/4245/

adamruzicka:
The tests are running here http://ci.theforeman.org/job/test_plugin_pull_request/4247/

adamruzicka:
The tests are green

app/models/concerns/foreman_remote_execution/errors_flattener.rb
```diff
@@ -6,7 +6,8 @@ def flattened_validation_exception
 
     def flattened_errors
       errors = Hash.new { |h, k| h[k] = [] }
-      self.errors.each_key do |key|
+      # rubocop:disable Performance/HashEachMethods
```
iNecas: Please put here a comment about the reason why we can't do that (`ActiveModel::Errors` missing `each_key` method)

app/models/targeting.rb
```diff
@@ -10,7 +10,9 @@ class Targeting < ActiveRecord::Base
 
   has_many :targeting_hosts, :dependent => :destroy
   has_many :hosts, :through => :targeting_hosts
+  # rubocop:disable Rails/HasManyOrHasOneDependent
```
iNecas: Do you have any insights on why this has happened? Is there a risk this would affect behavior in other places that we have the `:dependent` specified? Why is it raising this? Why it complains it wants to crate something, when we are actually deleting

```
ActiveRecord::RecordNotSaved: You cannot call create unless the parent is saved
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:11:in `block in transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:5:in `transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:11:in `block in transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:5:in `transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:11:in `block in transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:5:in `transaction'
    /var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.2/plugin/app/models/concerns/foreman_remote_execution/foreman_tasks_cleaner_extensions.rb:29:in `delete_job_invocations'
    /var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.2/plugin/app/models/concerns/foreman_remote_execution/foreman_tasks_cleaner_extensions.rb:25:in `delete_tasks'
    foreman-tasks (0.10.4) lib/foreman_tasks/cleaner.rb:104:in `delete'
    /var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.2/plugin/app/models/concerns/foreman_remote_execution/foreman_tasks_cleaner_extensions.rb:12:in `delete' (ActiveRecord::RecordNotSaved)
/usr/local/rvm/gems/ruby-2.2.5@test_plugin_pull_request-1/gems/activerecord-4.2.9/lib/active_record/associations/collection_association.rb:497    
```

adamruzicka: When I tried to set it to `:dependent => :nullify` I got an error about modifying an already destroyed record, so probably `JobInvocation` got destroyed, triggered destroy callbacks on its `Targeting` which tried to modify already destroyed `JobInvocation`. It is quite possible the same thing happens with `:dependent => :destroy`.

I'll take some time tomorrow to go over all the changes and check it is safe

adamruzicka:
Ok, this should be good to go

app/models/targeting.rb
iNecas: Any update on this?

adamruzicka:
I couldn't reproduce the targeting issue this morning, giving it another go with `:dependent => :destroy`, tests running here http://ci.theforeman.org/job/test_plugin_pull_request/4258/

adamruzicka:
So, the issue is not present anymore in rails 5.0.6. From what I've seen it seems like rails can't reliably handle having `:dependent => :destroy` on both sides of the assocaition. Tests are running here http://ci.theforeman.org/job/test_plugin_pull_request/4259/ , hopefully for the last time

iNecas:
The failures are independent, merging now. Thanks @adamruzicka

PR #296, opened 2017-12-22
Fixes #22061 - Associate taxonomy based on template metadata

ranjan:


app/models/job_template.rb
```diff
@@ -93,20 +93,36 @@ def import_parsed(name, text, metadata, options = {})
...
       # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     end
   end
 
+  def map_metadata(metadata, taxonomy)
+    if metadata[taxonomy]
+      case param
+      when 'locations'
+        metadata[taxonomy].map do |loc|
+          Location.all.map { |db| db.name =~ /^#{loc}/ ? db : nil }
```
adamruzicka: I assume here we want to find a single (not more) `Location` for each location name  (`metadata['location']`). This makes me think `metadata[taxonomy].map { |loc_name| Location.find(:name => loc_name) }.compact` would be more appropriate. Also why are you matching using a prefix?

app/models/job_template.rb
```diff
@@ -93,20 +93,36 @@ def import_parsed(name, text, metadata, options = {})
...
+        template.assign_attributes(metadata.merge(:template => text.gsub(/<%\#.+?.-?%>\n?/m, '').strip).merge(options).except('organizations', 'locations'))
+        template.assign_taxonomies(metadata) if template.new_record?
         template
       end
       # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     end
   end
 
+  def map_metadata(metadata, taxonomy)
+    if metadata[taxonomy]
```
adamruzicka: Maybe an early return would be better.

adamruzicka:
@ranjan any progress here?

adamruzicka:
@ranjan There was no progress here for more than two years, closing. Feel free to reopen

PR #298, opened 2017-12-30
Fixes #22099 - reuse core exporting

ares:
Introduced in Foreman 1.12, see http://projects.theforeman.org/issues/14509 for more details


app/models/job_template.rb
```diff
@@ -1,10 +1,10 @@
 class JobTemplate < ::Template
-  include ForemanRemoteExecution::Exportable
+  include ::Exportable
 
   class NonUniqueInputsError < Foreman::Exception
   end
 
-  attr_exportable :name, :job_category, :description_format, :snippet, :template_inputs,
+  attr_exportable :job_category, :description_format, :template_inputs,
```
ares: note for reviewer: these are now inherited from Template along with `model`

adamruzicka:
Shouldn't we keep the old tests just to be sure it behaves the same way as before?

ares:
I think that would be a duplication of tests from core.

adamruzicka:
@ares could you please rebase this?

ares:
@adamruzicka sure, rebased

ares:
rebased to avoid test failures (only postgres on 2.4, probably was not related)

ares:
@adamruzicka this got stuck somewhere, mind to take another look?

adamruzicka:
Sorry for the delay. I looked at it again now and it works like a charm, merged. Thanks @ares !

PR #297, opened 2017-12-24
Foreman 1.17 updates

mmoll:

PR #299, opened 2018-01-07
Fixes #22168 - Fix ActiveModel::ForbiddenAttributesError

adamruzicka:
Due to a change in Foreman core if we call triggering_params in REX, we receive
a Hash instead of ActiveController::Params. Merging a Hash into unpermitted
Params leads to the merged Hash not being #permitted? and causing
ActiveModel::ForbiddenAttributesError when trying to mass-assign the
triggering_params when creating a new job.

Related to theforeman/foreman commit 1ccf4df


iNecas:
This fixes the issue for me. I'm merging regardless of the CI right now, as there are other things that need to be fixed in tests in https://github.com/theforeman/foreman_remote_execution/pull/297, but this has influence on usability of the rex itself, therfore pulling it in

PR #300, opened 2018-01-10
Fixes #21888 - Associate default templates to Orgs and Locs during seed

adamruzicka:


iNecas:
I don't think this is the right fix for the issue, see migrations, what were done in http://projects.theforeman.org/issues/20708, The right fix should IMO be to add the the migration to convert the JobTemplates's`taxable_taxonomies`, similarly as we did in https://github.com/theforeman/foreman/blob/bc038098bc2f96986442ae2fffac4e4746933cec/db/migrate/20170828114310_fix_taxable_taxonomies_template_types.rb

Opened an alternative PR for this issue https://github.com/theforeman/foreman_remote_execution/pull/302/files

adamruzicka:
Closing in favor of https://github.com/theforeman/foreman_remote_execution/pull/302

PR #301, opened 2018-01-10
Fixes #22237 - notification on success run

ares:
see how it works
record

ready for later customization, job_invocation would return different notification builder class set per feature

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -66,6 +66,12 @@ module Actions
         super unless event == Dynflow::Action::Skip
       end
 
+      def finalize
+        # creating the success notification should be the very last thing this tasks do
+        job_invocation = JobInvocation.find(input[:job_invocation_id])
+        job_invocation.build_notification.deliver!
+      end
+
       def humanized_input
         input.fetch(:job_invocation, {}).fetch(:description, '')
       end
--- a/db/seeds.d/50-notification_blueprints.rb
+++ b/db/seeds.d/50-notification_blueprints.rb
@@ -1,0 +1,18 @@
+blueprints = [
+  {
+    group: N_('Jobs'),
+    name: 'rex_job_succeeded',
+    message: N_("A job '%{subject}' has finished successfully"),
+    level: 'success',
+    actions:
+    {
+      links:
+      [
+        path_method: :job_invocation_path,
+        title: N_('Job Details')
+      ]
+    }
+  }
+]
+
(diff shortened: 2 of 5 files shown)


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -66,6 +66,12 @@ def run(event = nil)
         super unless event == Dynflow::Action::Skip
       end
 
+      def finalize(*args)
```
adamruzicka: Nitpick: I don't think `finalize` can take any arguments

adamruzicka:
Looks mostly good to me, left one nitpick inline, however I'm hitting one issue which doesn't appear on your gif where the notification dropdown menu shows just 'Details' instead of 'Job Details'

menu detail

db/seeds.d/50-notification_blueprints.rb
```diff
@@ -0,0 +1,18 @@
+blueprints = [
+  {
+    group: N_('Jobs'),
+    name: 'rex_job_succeeded',
+    message: N_('%{subject} has finished successfully'),
```
iNecas: Nitpick: IMO the %{subject} should be quoted to better distinguish between static and dynamic part of the notification.

iNecas:
Nice work in general. Mind adding some tests for this. We don't have the test file for this currently, but https://github.com/theforeman/foreman_remote_execution/blob/master/test/unit/actions/run_hosts_job_test.rb could be used for inspiration: I'm at your hand if help needed on this part.

ares:
@iNecas added test and fixed both comments

iNecas:
Running the tests here http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4564/

ares:
Fixed the rubocop, rerun at http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4570/, expected to fail on colorful output broken cops which is fixed by #310

test/unit/actions/run_hosts_job_test.rb
```diff
@@ -119,5 +119,14 @@ class RunHostsJobTest < ActiveSupport::TestCase
         planned.input[:concurrency_control][:level].wont_be_empty
       end
     end
+
+    describe 'notifications' do
+      it 'creates notification on sucess run' do
+        FactoryBot.create(:notification_blueprint, :name => 'rex_job_succeeded')
+        assert_difference 'NotificationRecipient.where(:user_id => targeting.user.id).count' do
```
iNecas: Nice: wasn't aware of this assertion.

iNecas:
Thanks @ares

PR #302, opened 2018-01-12
Fixes #21888 - convert the taxable_taxonomies for JobTemplate

iNecas:
In #20708, we were converting the taxable_taxonomies association for
Templates. ProvisioningTemplate and PTable were converted, but not
JobTemplate. This commits completes the migration for remote execution.

### DIFF
--- a/db/migrate/20180112125015_fix_taxable_taxonomies_job_template.rb
+++ b/db/migrate/20180112125015_fix_taxable_taxonomies_job_template.rb
@@ -1,0 +1,14 @@
+class FixTaxableTaxonomiesJobTemplate < ActiveRecord::Migration[4.2]
+  def up
+    # we need to count on the fact that the user might already has assigned the templates
+    # to taxonomies after the previous update
+    already_present = TaxableTaxonomy.where(:taxable_type => 'JobTemplate').pluck(:taxable_id)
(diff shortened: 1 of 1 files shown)


ares:
LGTM

db/migrate/20180112125015_fix_taxable_taxonomies_job_template.rb
```diff
@@ -0,0 +1,9 @@
+class FixTaxableTaxonomiesJobTemplate < ActiveRecord::Migration
```
mmoll: needs a version added now.

iNecas: Updated, tested originally with Foreman 1.15, therefore missed this.

iNecas:
While more testing, I've hit a case, where I had already templates assigned to the JobTemplates using the `:taxable_type => 'JobTemplates'`, which lead to violating the unique constraint on `[:taxable_type, :taxable_id]`. Added additional logic to convert only the templates that were not already converted to avoid this situation. Please re-review

adamruzicka:
Looks good to me. @ares, @mmoll anything I missed here?

ares:
ACK, it fixes the problem for me. I'm not sure what's the policy on this repo atm re broken tests so leaving unmerged, but :+1: from me

PR #305, opened 2018-01-23
Fixes #22395 - Added DEFAULT_REFRESH_INTERVAL constant

witlessbird:


lib/foreman_remote_execution_core/fake_script_runner.rb
```diff
@@ -1,6 +1,7 @@
 module ForemanRemoteExecutionCore
   class FakeScriptRunner < ForemanTasksCore::Runner::Base
-
+    DEFAULT_REFRESH_INTERVAL = 60
```
adamruzicka: Shouldn't the `FakeScriptRunner` use the same default interval as a regular `ScriptRunner`, i.e. 1 second?

iNecas: +1 on 1 second

witlessbird:
* changed the default interval to 1 second.

iNecas:
Thanks @witlessbird

PR #306, opened 2018-01-25
Fixes #22421 - Correct Nic migration for Rails 5

tbrisker:
This also has an added benefit of running much faster, especially for
installations that have a large number of interfaces.


iNecas:
Nice indeed. Thanks @tbrisker!

PR #309, opened 2018-01-31
Fixes #22187 - shows correct job stats

dmitri-d:


witlessbird:
Please note that this doesn't fix job stats on the index page -- those will still include hosts that have been deleted.

adamruzicka:
In case the deleted host was the only one in the job I'm getting
```
Showing /home/aruzicka/git/foreman-remote-execution/app/views/job_invocations/_tab_overview.html.erb where line #2 raised:

NaN

/home/aruzicka/git/foreman-remote-execution/app/models/job_invocation.rb:100:in `round'
/home/aruzicka/git/foreman-remote-execution/app/models/job_invocation.rb:100:in `progress'
/home/aruzicka/git/foreman-remote-execution/app/models/job_invocation.rb:197:in `progress_report'
/home/aruzicka/git/foreman-remote-execution/app/helpers/remote_execution_helper.rb:16:in `job_invocation_chart'
/home/aruzicka/git/foreman-remote-execution/app/views/job_invocations/_tab_overview.html.erb:2:in `_8538d9537a9cca7ac2b4bf39ef0ddaf2'
```

in case I don't delete the host and just try to look at the job from another Org/Loc, I'm getting the same error but with "Infinity" instead of "NaN"

witlessbird:
@iNecas, @adamruzicka: any chance we could chat about this irl? In particular, I'd like to hear how foreman_tasks interacts with dynflow, including when dynflow is running in a separate process. I wonder if it would be possible for dynflow to emit events on task completion, and for foreman_tasks to catch these events (and then update an append-only log of completed tasks).

iNecas:
For you first question: the dynflow action runs in context of the Rails environment. So it's possible to do whatever we want in respond to the events.

We actually used the approach of updating the stats based on the events. The was one bulk action that was responsible for planning the sub-tasks, and once the sub-task finished, it send event back to the parent task to recalculate the stats. (see https://github.com/Dynflow/dynflow/blob/5749d43578360e115712f46e3acdd6c5e629435d/lib/dynflow/action/with_sub_plans.rb#L20).

The problem with this approach was that the event queue got bloated with this messages (especially when having a job with 10 000s of hosts), and the calculated data were getting behind the actual state of the completion + were keeping the workers from doing other stuff + there was an issue where the events were not persisted and there was a potential loss, that we needed to count on. There are multiple things that can improve the situation (optimizations, multiple queues support, persisted storage for the events).

On the other hand, it was possible to eliminate the high load of messages by switching to polling, where, instead of getting called on every sub-task finished, we periodically check if the sub-stask as finished or not (https://github.com/Dynflow/dynflow/blob/5749d43578360e115712f46e3acdd6c5e629435d/lib/dynflow/action/with_polling_sub_plans.rb#L63-L69). This approach offloaded the amount of events + made the flow more resilient to restarts, where we would risk loosing the events.

I think we need to work more on improving the eventing + throughput before we would move this direction again.

bkearney:
Hey @iNecas, @adamruzicka what is the status of this?

iNecas:
Need revisiting

bkearney:
@iNecas should we close this and the bug out?

iNecas:
I don't think so

bkearney:
@adamruzicka Dmtri is no longer working on foreman. Can this be fixed by someone still in the community?

adamruzicka:
We have this on our team's backlog

ares:
@adamruzicka should this be closed now?

adamruzicka:
It should. In the end we decided to take a slightly different approach than suggested here, but thank you @dmitri-d for this

PR #310, opened 2018-02-01
Fixes #22478 - put assets into rex namespace

ares:


iNecas:
Tests running here http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4565/

ares:
Pushing also rubocop fixes introduced by color output PR, tests rerun at http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4569/

app/helpers/job_invocation_output_helper.rb
```diff
@@ -1,20 +1,20 @@
...
+    '36' => 'cyan',
+    '37' => 'grey',
+    '91' => 'red',
+    '92' => 'lightgreen',
+    '93' => 'yellow',
+    '94' => 'lightblue',
+    '95' => 'violet',
+    '96' => 'turquoise',
+    '0'  => 'default',
+  }.freeze
```
dLobatog: This won't work because of the '.default' unfortunately.. I've fixed this in #130 with https://github.com/theforeman/foreman_remote_execution/pull/130/files#r165339121

iNecas: Tests faiing with:

```
can't modify frozen Hash
/var/lib/workspace/workspace/test_plugin_pull_request/database/mysql/label/fast/ruby/2.4/plugin/app/helpers/job_invocation_output_helper.rb:18:in `default='
```

ares: I think there's a better way while keeping the default being set and not copying the default value out of it, let's see if the updated version works

ares:
rubocop pass locally, tests running at http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4573/

iNecas:
Thanks @ares

mmoll:
> tests running at http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/4573/

this was 💔

PR #311, opened 2018-02-01
Fixes #22483 - override host_param instead of host_params

ares:
to reproduce, try to preview remote_execution_ssh_keys snippet

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -45,15 +45,17 @@ module ForemanRemoteExecution
       @execution_status_label ||= get_status(HostStatus::ExecutionStatus).to_label(options)
     end
 
-    def host_params
-      params = super
-      keys = remote_execution_ssh_keys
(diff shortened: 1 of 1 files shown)


iNecas:
Other places also need updating, otherwise I guess this would not work https://github.com/theforeman/foreman_remote_execution/blob/9a8264b4a9e26c44dddb00934898e8a9529629cc/app/models/ssh_execution_provider.rb#L21 and https://github.com/theforeman/foreman_remote_execution/blob/9a8264b4a9e26c44dddb00934898e8a9529629cc/app/models/ssh_execution_provider.rb#L21?

Alternatively, we could override the `host_params_hash` to include the values.

iNecas:
Btw. using the `host_param` method from other places will also have quite nice performance benefits, as we will not need to evaluate the `SmartVariables`

adamruzicka:
[test]

iNecas:
This needs to be re-worked, I talked to @ares and we will re-open the re-worked PR, as @ares is out of capacity these days.

iNecas:
Closing in favor of https://github.com/theforeman/foreman_remote_execution/pull/349

PR #313, opened 2018-02-06
Fixes #22519 - add hover effect to tables

ares:


iNecas:
Thanks @ares and @adamruzicka

PR #314, opened 2018-02-13

dLobatog:
It's a bit difficult to notice some sections in the job invocations
page. Similarly, the piechart is strangely aligned to the right with a
big space in between.
To try to overcome these issues and homogenize the interface, we can use
Patternfly cards
https://rawgit.com/patternfly/patternfly/master/tests/cards.html

This is how it used to look like:
http://theforeman.org/plugins/foreman_remote_execution/0.1/job_detail_1.png
After this change, it looks like: http://i.imgur.com/3kKigMr.png

Reopened from https://github.com/theforeman/foreman_remote_execution/pull/130 so that js linting & testing runs here


dLobatog:
@iNecas Closing this as it appears Travis didn't pick it up.

PR #315, opened 2018-02-13

dLobatog:
Testing if hound picks up JS linting


webpack/index.js
```diff
@@ -0,0 +1,23 @@
+import URI from 'urijs';
+import { mount, registerReducer } from 'foremanReact/common/MountingService';
```
dLobatog: According to the linting rules of the repo, this should be 

```
/home/vagrant/foreman_remote_execution/webpack/index.js
  2:1  error  'foremanReact' should be listed in the project's dependencies. Run 'npm i -S foremanReact' to add it  import/no-extraneous-dependencies
  3:1  error  'foremanReact' should be listed in the project's dependencies. Run 'npm i -S foremanReact' to add it  import/no-extraneous-dependencies
  6:8  error  Absolute imports should come before relative imports                                                  import/first
```

sharvit:
@dLobatog I see your point.
Is it possible that your local linter version is different than what hound uses?

PR #316, opened 2018-02-16
Fixes #22605 - JobInvocation API accepts 'feature' param

dLobatog:
The API for JobInvocation could accept a 'feature' param which finds the
REX feature and creates the JobInvocation for said feature with all the
right parameters.
This saves other plugins from having to find the template ID associated
with the feature they want to run.

Here's a screencast showcasing this feature - clicking on "run playbook" sends a POST request with the feature & required inputs/host ids

https://gfycat.com/gifs/detail/SpottedOfficialCarpenterant

### DIFF
--- a/test/functional/api/v2/job_invocations_controller_test.rb
+++ b/test/functional/api/v2/job_invocations_controller_test.rb
@@ -23,45 +23,68 @@ module Api
         assert_equal template['job_category'], @invocation.job_category
       end
 
-      test 'should create valid with job_template_id' do
-        attrs = { :job_category => @template.job_category, :name => 'RandomName', :job_template_id => @template.id,
-                  :targeting_type => 'static_query', :search_query => 'foobar'}
-        post :create, params: { :job_invocation => attrs }
+      context 'creation' do
+        setup do
+          @attrs = { :job_category => @template.job_category,
+                     :name => 'RandomName',
+                     :job_template_id => @template.id,
+                     :targeting_type => 'static_query',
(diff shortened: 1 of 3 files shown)


dLobatog:
@iNecas I've added a .hound.yml file as it's commenting according to the default rubocop rules, not this repos'

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -23,45 +23,60 @@ class JobInvocationsControllerTest < ActionController::TestCase
...
+          invocation = ActiveSupport::JSON.decode(@response.body)
+          assert_equal invocation['mode'], 'future'
+          assert_response :success
+        end
 
-        post :create, params: { :job_invocation => attrs }
-        invocation = ActiveSupport::JSON.decode(@response.body)
-        assert_equal invocation['mode'], 'future'
-        assert_response :success
+        test 'should create with feature' do
```
dLobatog: This is the only test I added, I just refactored the test to take advantage of a global `@attrs`

iNecas:
[test] Seems like 2.3 tests timed out, probably not related, but still…

dLobatog:
@iNecas :green_salad:

iNecas:
Thanks @dLobatog

ares:
the feature parameter was not documented so hammer does not know about it, fix sent in separate PR - https://github.com/theforeman/foreman_remote_execution/pull/322

PR #317, opened 2018-02-16
Fixes #22604 - Accept input values as GET params in UI

dLobatog:
The UI should accept template inputs through the URL to display
the JobInvocation new form with pre-filled fields.

For example,
https://centos7-devel.lobatolan.home/job_invocations/new?feature=
ansible_run_insights_plan&inputs[plan_id]=3&inputs[organization_id]=20
would find the 'ansible_run_insights_plan' feature and 'plan_id' /
'organization_id' as pre-filled fields.

A small screencast demonstrating this feature around 1:10 -  https://webmshare.com/1WQee

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -3,6 +3,7 @@ class JobInvocationsController < ApplicationController
   include ::ForemanTasks::Concerns::Parameters::Triggering
 
   def new
+    return @composer = prepare_composer if params[:feature].present?
     ui_params = {
       :host_ids => params[:host_ids],
       :targeting => {
@@ -87,7 +88,12 @@ class JobInvocationsController < ApplicationController
 
   def prepare_composer
     if params[:feature].present?
-      JobInvocationComposer.for_feature(params[:feature], params[:host_ids], {})
+      inputs = params[:inputs].permit!.to_hash if params.include?(:inputs)
+      JobInvocationComposer.for_feature(
+        params[:feature],
+        params[:host_ids],
+        inputs
+      )
     else
       # triggering_params is a Hash
       #   when a hash is merged into ActionController::Parameters,
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -254,6 +254,7 @@ class JobInvocationComposer
     end
 
     def input_values_params
+      return {} if @provided_inputs.blank?
       @provided_inputs.map do |key, value|
         input = job_template.template_inputs_with_foreign.find { |i| i.name == key.to_s }
         unless input
--- a/test/functional/job_invocations_controller_test.rb
+++ b/test/functional/job_invocations_controller_test.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'test_plugin_helper'
+
+class JobInvocationsControllerTest < ActionController::TestCase
+  test 'should parse inputs coming from the URL params' do
+    template = FactoryBot.create(:job_template, :with_input)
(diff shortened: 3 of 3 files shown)


app/controllers/job_invocations_controller.rb
```diff
@@ -10,14 +10,34 @@ def new
         :bookmark_id => params[:bookmark_id]
       }
     }
-
-    if (template = JobTemplate.find_by(id: params[:template_id]))
+    template = (
+      JobTemplate.find_by(id: params[:template_id]) ||
+      RemoteExecutionFeature.find_by(label: params[:feature]).job_template
```
ares: maybe I'm missing something, but should it be handled by https://github.com/dLobatog/foreman_remote_execution/blob/7ea7ad9a909aca2461a18c030292f00d506a94f7/app/controllers/job_invocations_controller.rb#L110 ?

dLobatog: :man_facepalming:

app/controllers/job_invocations_controller.rb
```diff
@@ -87,7 +88,11 @@ def action_permission
 
   def prepare_composer
     if params[:feature].present?
-      JobInvocationComposer.for_feature(params[:feature], params[:host_ids], {})
+      JobInvocationComposer.for_feature(
+        params[:feature],
+        params[:host_ids],
+        params[:inputs].permit!.to_hash
```
ares: this will not work if inputs are not sent, which is the case for REX features which are marked as host_action_button (ran from host detail page from the schedule job drop down button), here's the trace

```
2018-02-23T11:44:23 9991c3b7 [app] [F]   
2018-02-23T11:44:23 9991c3b7 [app] [F] NoMethodError (undefined method `permit!' for nil:NilClass):
2018-02-23T11:44:23 9991c3b7 [app] [F]   
2018-02-23T11:44:23 9991c3b7 [app] [F] /home/ares/Projekty/Zdrojaky/foreman_remote_execution/app/controllers/job_invocations_controller.rb:94:in `prepare_composer'
 | /home/ares/Projekty/Zdrojaky/foreman_remote_execution/app/controllers/job_invocations_controller.rb:34:in `create'
 | app/controllers/concerns/application_shared.rb:15:in `set_timezone'
 | app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 | app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
 | lib/middleware/catch_json_parse_errors.rb:8:in `call'
 | lib/middleware/session_safe_logging.rb:17:in `call'
 | lib/middleware/tagged_logging.rb:18:in `call'
```

since tests were green, I'd like to ask also for adding a tests with no inputs

ares:
screenshot_20180223_160709

this button - all rex features which are registered with `:host_action_button => true` are show here and can be triggered immediately from host page

app/controllers/job_invocations_controller.rb
```diff
@@ -87,7 +88,13 @@ def action_permission
 
   def prepare_composer
     if params[:feature].present?
-      JobInvocationComposer.for_feature(params[:feature], params[:host_ids], {})
+      inputs = params[:inputs].permit!.to_hash if params.include?(:inputs)
+      inputs ||= nil
```
ares: I think this is redundant, inputs is already registered for ruby even if the `if` condition is not evaluated to true, but if you want to keep it for readability, I think it would be better to move it one line up and simplify it to `inputs = nil`

ares:
Thanks @dLobatog, works as expected, merging!

PR #318, opened 2018-02-16
Fixes #22609 - customizable notifications per feature

ares:
This allows defining a custom notification builder for feature, if you want to see example usage, see https://github.com/theforeman/foreman_openscap/pull/326

the fact we now store the feature makes it easy in future to add more customization, e.g. job invocation page, based on it

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -49,6 +49,8 @@ class JobInvocation < ApplicationRecord
   belongs_to :triggering, :class_name => 'ForemanTasks::Triggering'
   has_one :recurring_logic, :through => :triggering, :class_name => 'ForemanTasks::RecurringLogic'
 
+  belongs_to :remote_execution_feature
+
   scope :with_task, -> { references(:task) }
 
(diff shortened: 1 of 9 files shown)


app/models/job_invocation.rb
```diff
@@ -87,7 +89,17 @@ def notification_recipients_ids
   end
 
   def build_notification
-    UINotifications::RemoteExecutionJobs::BaseJobFinish.new(self)
+    # self.remote_execution_feature
```
adamruzicka: Was this left here intentionally?

ares: nope, I'll remove when you'll add more comments :-)

ares:
cops issues resolved, all should be green this time

adamruzicka:
Mr. J is happy with sqlite and postgresql, but mysql timed out. I'd say this is good to go :shipit:

ares:
I swear it was green before also on PG :-)

adamruzicka:
Merged, thanks @ares !

PR #319, opened 2018-02-20
Fixes #22630 - job template for reboot to use shutdown -r +1

pmoravec:
Signed-off-by: Pavel Moravec <pmoravec@redhat.com>


iNecas:
Tested and works well and since it seems it's more reliable than the current approach, ACK from me: please open it against the templates though, as @dLobatog suggests.

pmoravec:
Created https://github.com/theforeman/community-templates/pull/456 .

Since I can't move the http://projects.theforeman.org/issues/22630 from Foreman Remote Execution to the Foreman project, neither I can close it, I created new ticket http://projects.theforeman.org/issues/22679.

PR #320, opened 2018-02-20
Fixes #22633 - "humanize" setting names

dLobatog:
REX settings are displayed in the Administer > Settings with their
variable names. e.g: 'remote_execution_effective_user_method'.
Instead, they should just show a name that does not look a variable
(snakecase, namespaced..).

### DIFF
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -2,6 +2,7 @@ class Setting::RemoteExecution < Setting
 
   ::Setting::BLANK_ATTRS.concat %w{remote_execution_ssh_password remote_execution_ssh_key_passphrase}
 
+  # rubocop:disable AbcSize
   # rubocop:disable Metrics/MethodLength
   def self.load_defaults
     # Check the table exists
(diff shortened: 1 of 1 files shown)


dLobatog:
All the hound failures show up because there is no `.hound.yml`  so it just uses the default rubocop rules. I have submitted PRs with this file in #317 #316 #130

app/models/setting/remote_execution.rb
```diff
@@ -12,54 +13,63 @@ def self.load_defaults
       [
         self.set('remote_execution_fallback_proxy',
                  N_('Search the host for any proxy with Remote Execution, useful when the host has no subnet or the subnet does not have an execution proxy'),
-                 false),
+                 false,
+                 N_('Fallback Any Proxy')),
```
adamruzicka: This feels a bit odd compared to `Fallback Without Proxy`, maybe something like `Fallback to Any Proxy` would be better?

dLobatog: Sounds better indeed, my brain was a bit fried after writing all these descriptions :laughing: 

@adamruzicka Can you do that change on merge? Just clone the branch, make the change, add it, `git commit --amend` it , cherry-pick that change to master and push. It's just so we can save another 2h long Jenkins job

adamruzicka: No, problem, will do

adamruzicka:
Merged as https://github.com/theforeman/foreman_remote_execution/commit/0ac4ff1b7fa3c7a8aece5d56bd6faa1ff264bc37, thanks @dLobatog !

PR #321, opened 2018-02-23
Fixes #22670 - Allow search query on JobInvocation API

dLobatog:
The API currently only accepts a list of host IDs as parameters for the
JobInvocation CREATE method when using a feature.
It should also be able to accept a search query.

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -49,7 +49,7 @@ module Api
           end
 
           param :bookmark_id, Integer, :required => false
-          param :search_query, Integer, :required => false
+          param :search_query, String, :required => false
           param :description_format, String, :required => false, :desc => N_('Override the description format from the template for this invocation only')
           param :execution_timeout_interval, Integer, :required => false, :desc => N_('Override the timeout interval from the template for this invocation only')
         end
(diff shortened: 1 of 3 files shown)


ares:
LGTM, the api does not work since params are not permitted and using feature is broken from hammer too but that is not something to be addressed by this PR, just blocks me from testing, if tests are green though, I think it could be merged

ares:
hm, tests are red though, @dLobatog please take a look

app/models/job_invocation_composer.rb
```diff
@@ -221,6 +221,10 @@ def initialize(feature_label, hosts, provided_inputs = {})
         @host_bookmark = hosts
       elsif hosts.is_a? Host::Base
         @host_objects = [hosts]
+      elsif hosts.is_a? Array
+        @host_objects = hosts.map do |id|
+          Host::Managed.authorized.friendly.find(id)
```
dLobatog: This is the key part - it should be able to find arrays of host names/ids. I switched from using the resource finder because it is only available in the controller.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -123,16 +123,13 @@ def job_invocation_params
         job_invocation_params.merge!(job_invocation_params.delete(:ssh)) if job_invocation_params.key?(:ssh)
         job_invocation_params[:inputs] ||= {}
         job_invocation_params[:inputs].permit!
-        job_invocation_params
+        @job_invocation_params = job_invocation_params
```
dLobatog: I added this, even if it's unrelated to the PR because this method is inefficiently running every time instead of returning the instance variable (it was not set anywhere)

dLobatog:
All tests are green, pg & 2.4 couldn't finish because REX + core tests tend to go over 2 hours. http://ci.theforeman.org/job/test_plugin_pull_request/4720/ https://github.com/theforeman/foreman-infra/issues/490

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -64,26 +64,27 @@ class JobInvocationsControllerTest < ActionController::TestCase
...
+        context 'with_feature' do
+          setup do
+            @feature = FactoryBot.create(:remote_execution_feature,
+                                         :job_template => @template)
+            @attrs = {
+              feature: @feature.label
+            }
+          end
+
+          test 'host ids as array' do
```
iNecas: Could you add a specific case for array of ids? I know Katello is using ids, not fqdns: I tested and it seems to work, but I would rather have this covered by tests.

iNecas:
Test failures unrelated. Thanks @dLobatog

PR #322, opened 2018-02-23
Refs #22605 - fix params documentation

ares:


app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -23,7 +23,7 @@ def show; end
       # rubocop:disable Metrics/BlockLength
       def_param_group :job_invocation do
         param :job_invocation, Hash, :required => true, :action_aware => true do
-          param :job_template_id, String, :required => true, :desc => N_('The job template to use')
+          param :job_template_id, String, :required => false, :desc => N_('The job template to use')
```
iNecas: We should probably describe in desc, that the attribute is required unless `feature` specified.

ares:
I updated the description, would you mind to take another look?

iNecas:
OTOH I don't see anything that should influence the tests. Merging now. Thanks @ares

PR #323, opened 2018-02-26
Fixes #22688 - Use UUID column type for storing task UUIDs

adamruzicka:
Required after https://github.com/theforeman/foreman-tasks/pull/315 goes in

### DIFF
--- a/db/migrate/20180226095631_change_task_id_to_uuid.rb
+++ b/db/migrate/20180226095631_change_task_id_to_uuid.rb
@@ -1,0 +1,31 @@
+class ChangeTaskIdToUuid < ActiveRecord::Migration[4.2]
+  def up
+    if on_postgresql?
+      change_table :job_invocations do |t|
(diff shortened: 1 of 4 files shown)


iNecas:
Test failures related

adamruzicka:
To be precise, this is required for REX to work with tasks with UUID column type and at the same time it requires tasks with UUID column type to work at all Waiting for new foreman-tasks release

iNecas:
foreman-tasks-0.12.0 just released, please update the dependency on it here

adamruzicka:
Tests running here http://ci.theforeman.org/job/test_plugin_pull_request/4748/

iNecas:
Seems even there were some intermittent issues, starting again in http://ci.theforeman.org/job/test_plugin_pull_request/4752/

iNecas:
The test failures unrelated. Merging, as otherwise rex is not compatible with tasks

PR #326, opened 2018-03-05
Refs #22609 - rename the field

ares:
I wonder how it could have worked for me when I tested it.


iNecas:
Ups. Thanks @ares, I don't think we need to wait for CI with this patch :)

PR #328, opened 2018-03-08
Fixes #22807 - Do not join on columns of possibly different types

adamruzicka:
Steps to reproduce:
1. Have a postgres DB
2. enable `foreman-tasks` plugin, disable REX
3. `bundle exec rake db:create && bundle exec rake db:migrate`
4. Enable REX
5. Try to run `bundle exec rake db:migrate` again, without this fix it should fail with the error described in the issue

When we tried to do `.includes([:sub_tasks => :locks])` we tried to join `JobInvocation` with `ForemanTasks::Task` but the column used to match the records were of different types. If we don't do this join, we sacrifice a bit of performance by not loading all the needed data at once, but retrieving later should be done using `SELECT ... WHERE` which should work.

### DIFF
--- a/db/migrate/20160113162007_expand_all_template_invocations.rb
+++ b/db/migrate/20160113162007_expand_all_template_invocations.rb
@@ -14,7 +14,7 @@ class ExpandAllTemplateInvocations < ActiveRecord::Migration[4.2]
     FakeTemplateInvocation.update_all 'host_id = NULL'
 
     # expand all pattern template invocations and link RunHostJob
-    JobInvocation.joins(:targeting).where("#{Targeting.table_name}.resolved_at IS NOT NULL").includes([ :pattern_template_invocations, :targeting, :sub_tasks => :locks ]).each do |job_invocation|
+    JobInvocation.joins(:targeting).where("#{Targeting.table_name}.resolved_at IS NOT NULL").includes([:pattern_template_invocations, :targeting]).each do |job_invocation|
       job_invocation.pattern_template_invocations.each do |pattern_template_invocation|
         job_invocation.targeting.hosts.each do |host|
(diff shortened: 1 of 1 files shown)


iNecas:
makes sense, thanks @adamruzicka

lpramuk:
Tested successfully on production setup:
```
## foreman-installer -h
...
    --[no-]enable-foreman-plugin-tasks Enable 'foreman_plugin_tasks' puppet module (default: true)
...
```
>>> tasks plugin is enabled
```
# foreman-rake db:migrate 
```
>>> just to have any pending migrations executed
```
# foreman-installer  --enable-foreman-plugin-remote-execution --enable-foreman-proxy-plugin-remote-execution-ssh

# foreman-installer  --no-enable-foreman-plugin-remote-execution --no-enable-foreman-proxy-plugin-remote-execution-ssh

# foreman-rake db:migrate 
...
== 20160113162007 ExpandAllTemplateInvocations: migrating =====================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedFunction: ERROR:  operator does not exist: uuid = character varying
LINE 1: ...ations_join" ON "tasks_job_invocations_join"."id" = "job_inv...
```
>>> reproduced
```
# wget -O- https://github.com/theforeman/foreman_remote_execution/pull/328.diff | patch -N -p1 -d /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_remote_execution-1.4.5

# foreman-rake db:migrate 
...
== 20160113162007 ExpandAllTemplateInvocations: migrating =====================
== 20160113162007 ExpandAllTemplateInvocations: migrated (0.0450s) ============
```
>>> verified

PR #330, opened 2018-03-13
Fixes #20661 - use separate queue for rex jobs

iNecas:
Requires https://github.com/Dynflow/dynflow/pull/277, https://github.com/theforeman/foreman-tasks/pull/319 and https://github.com/theforeman/foreman/pull/5320

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -7,6 +7,10 @@ module Actions
       middleware.do_not_use Dynflow::Middleware::Common::Transaction
       middleware.use Actions::Middleware::HideSecrets
 
+      def queue
+        ForemanRemoteExecution::DYNFLOW_QUEUE
+      end
+
       def resource_locks
         :link
       end
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -2,7 +2,7 @@ class Setting::RemoteExecution < Setting
 
   ::Setting::BLANK_ATTRS.concat %w{remote_execution_ssh_password remote_execution_ssh_key_passphrase}
 
-  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
   def self.load_defaults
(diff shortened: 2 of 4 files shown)


app/models/setting/remote_execution.rb
```diff
@@ -63,7 +63,11 @@ def self.load_defaults
                  nil,
                  N_('Default SSH key passphrase'),
                  nil,
-                 { :encrypted => true })
+                 { :encrypted => true }),
+        self.set('remote_execution_workers_pool_size',
```
adamruzicka: It should be probably state somwhere that the user has to restart httpd+foreman-tasks services in production deployment in order for the changes to come into effect.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -7,6 +7,10 @@ class RunHostJob < Actions::EntryAction
       middleware.do_not_use Dynflow::Middleware::Common::Transaction
       middleware.use Actions::Middleware::HideSecrets
 
+      def queue
+        :remote_execution
```
adamruzicka: How about replacing all the occurences of `:remote_execution` with a constant?

app/models/setting/remote_execution.rb
iNecas: Good call, added the info into the description

app/lib/actions/remote_execution/run_host_job.rb
iNecas: Good idea, updated.

iNecas:
Updated

adamruzicka:
Could you please bump the Dynflow dependency again?

iNecas:
Updated to `~> 1.0`, as if we're following the semantic versioning, 1.y bumps should still be fine

adamruzicka:
[test]

iNecas:
Running tests manually http://ci.theforeman.org/job/test_plugin_pull_request/4853/

iNecas:
This should be ready for final review

adamruzicka:
Thanks @iNecas !

PR #331, opened 2018-03-20
Fixes #22375 - uses new way of importing templates

ares:
In Foreman 1.18, templates importing was introduced in core. Job
templates only need to specify custom importing logic for custom
attributes. No special integration with foreman_templates is required.
This also simplifies the code. import_raw method can keep its prototype
while internal implementation can be also simplified.

---

I expect some tests to fail, but at least for a preview. Note that this should get in before 1.18 is released.


iNecas:
If this requires 1.18, please bump https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L34

adamruzicka:
Jenkins is quite unhappy about this

ares:
and it found real issues, rex_tests++

this should now also address the duplicate input that you reported, we called import_custom_data twice, didn't set the default flag correctly and didn't set the template name for new objects

I kept the changes in separate commit for easier review, should be green now

iNecas:
I always like the PRs that remove 6x code more than they add. Thanks @ares

PR #332, opened 2018-03-21
[WIP] Fixes #22960 - Fix Uninitialized Constant error when running from Foreman

adamruzicka:


lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -153,7 +153,8 @@ def ssh_options
       ssh_options[:user_known_hosts_file] = prepare_known_hosts if @host_public_key
       ssh_options[:number_of_password_prompts] = 1
       ssh_options[:verbose] = settings[:ssh_log_level]
-      ssh_options[:logger] = ForemanRemoteExecutionCore::LogFilter.new(SmartProxyDynflowCore::Log.instance)
+      logger = defined?(SmartProxyDynflowCore) ? SmartProxyDynflowCore::Log.instance : @logger
```
iNecas: I wonder if this condition shouldn't actually go to https://github.com/theforeman/foreman-tasks/blob/4c482f45f371edd0e71b5dc2d479c09ebdc52382/lib/foreman_tasks_core/runner/base.rb#L15 and here, we would use just `self.logger`? Also, we're using `@logger` instance on several places, which makes me think if we can actually rely on the fact that the `@logger` will be always initialized, given the definition of the `logger` method in the base. Therefore, it might be worth just changing the other occurrences of the `@logger` to use `logger` instead

adamruzicka: The logger should get set by `FTC::Runner::Dispatcher`, which inherits the logger from its world, meaning we don't need to do this at all.

adamruzicka:
This doesn't work -.-

iNecas:
What exactly doesn't work and how should we fix it?

adamruzicka:
The logger as provided by Dynflow has slightly different interface compared to what net-ssh expects.

adamruzicka:
We don't support running jobs directly as of bfd1029, closing.

PR #333, opened 2018-03-22
Fixes #22993 - refactor audits definitions

ares:
while this is not entirely necessary, it would be good to be consistent with other models from plugins and core and have the audited definition first (it must preceed Taxonomix inclusion)

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -1,4 +1,5 @@
 class JobTemplate < ::Template
+  audited
   include ::Exportable
 
   class NonUniqueInputsError < Foreman::Exception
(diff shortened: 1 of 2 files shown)


app/models/job_template.rb
```diff
@@ -1,4 +1,5 @@
 class JobTemplate < ::Template
+  audited :allow_mass_assignment => true
```
tbrisker: `allow_mass_assignment` has been removed and is no longer needed.

iNecas:
After the question of @tbrisker is answered/addressed. No other questions

ares:
Sorry it took me a while. Ticket moved to REX project, removed mass assignment and removed empty space from the array so also hound is happy. @iNecas this should get into release that will be shipped with 1.18.

tbrisker:
[test]

ares:
3 combinations are green, I guess we can consider this as green

```
HostgroupJSTest::edit form::puppet classes are not available in the environment::Puppet classes tab.test_0002_it marks selected classes as unavailable
HostgroupJSTest::edit form::puppet classes are not available in the environment::Puppet classes tab.test_0001_it shows a warning
```

iNecas:
Thanks @ares

PR #335, opened 2018-03-26
Fixes #23016 - Fix setting refresh flag when retrieving output via API

adamruzicka:


iNecas:
@lpramuk mind tresting?

iNecas:
I've tested and works well. Btw. requires https://github.com/theforeman/foreman-tasks/pull/326/commits, so we will need to wait until that's merged and bump dependency here.

adamruzicka:
[test]

The dependency was merged

iNecas:
Thanks. Test failures unrelated.

PR #338, opened 2018-03-28
Fixes #22934 - work directories are being cleaned up now

witlessbird:


lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -120,7 +121,12 @@ def with_disconnect_handling
     end
 
     def close
+      run_sync("rm -rf #{remote_command_dir}") unless @session.nil? || @session.closed?
```
adamruzicka: We should put the path into quotes, so the users won't wipe their systems if they set remote_working_dir to something with a space

witlessbird:
* Put remote working directory name in quotes for its removal during cleanup

witlessbird:
* fixed style-related issues

iNecas:
Tested and works well. Would you mind adding rex settings to determine, wether to use this behavior (default to `true`), It still might be useful to be able to turn off the cleaning temporary, especially for debugging purposes.

adamruzicka:
Tested and works well with the default runner. Agree with making this configurable

When using `PollingScriptRunner`, it leaves the directory behind because the session gets destroyed at the end of `#refresh`. I guess the directories would need to be deleted somewhere around here[1]. It would be nice if both runner behaved the same.

[1] - https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution_core/polling_script_runner.rb#L47

witlessbird:
* added a setting controlling the cleanup
* added cleanup in polling script runner

Please note I cannot test the changes atm, as I have issues with my test environment (pls. see http://projects.theforeman.org/issues/23156).

witlessbird:
* rebased

app/models/remote_execution_provider.rb
```diff
@@ -44,6 +44,11 @@ def effective_user_method(host)
       method
     end
 
+    def cleanup_working_dirs?(host)
+      setting = host_setting(host, :remote_execution_cleanup_working_dirs)
+      ['true', 'True', 'TRUE', '1'].include?(setting)
```
adamruzicka: When the value gets taken from global settings it is a `TrueClass` or `FalseClass`, not `String` and this check then returns `false`. Setting the value on a host level seems to work correctly

witlessbird:
* fixed ```RemoteExecutionProvider.cleanup_working_dirs?``` to correctly handle boolean-valued setting.

adamruzicka:
Needs a rebase

witlessbird:
* rebased

app/models/setting/remote_execution.rb
```diff
@@ -67,7 +67,11 @@ def self.load_defaults
         self.set('remote_execution_workers_pool_size',
                  N_('Amount of workers in the pool to handle the execution of the remote execution jobs. Restart of the dynflowd/foreman-tasks service is required.'),
                  5,
-                 N_('Workers pool size'))
+                 N_('Workers pool size')),
+        self.set('remote_execution_cleanup_working_dirs',
+                 N_('When enabled, working directories will be removed after task completion.'),
```
iNecas: As in other settings options that can be overridden via params, we should mention something like 'You may override this per host by setting a parameter called remote_execution_cleanup_working_dirs.`

witlessbird:
* fixed failing test
* updated help note for ```remote_execution_cleanup_working_dirs``` parameter.

As a side note, we probably should replace "may"s with "can"s in those notes. The former is about permissibility of an action (although in more formal settings), while the latter is about an ability to do something (but is ok to use in "permission" sense these days too).

adamruzicka:
Thanks @witlessbird, @iNecas !

PR #339, opened 2018-03-29
Fixes #22974 - store feature coming via form

ares:
The problem here is that the hidden is under `job_invocation` key so feature was always ignored when user went through the form.


iNecas:
Makes sense. Thanks @ares

PR #340, opened 2018-04-02
Fixes #23089 - drop dynflow dependency

jlsherrill:


foreman_remote_execution.gemspec
```diff
@@ -24,7 +24,6 @@ Gem::Specification.new do |s|
   s.extra_rdoc_files = Dir['README*', 'LICENSE']
 
   s.add_dependency 'deface'
-  s.add_dependency 'dynflow', '~> 0.8.26'
   s.add_dependency 'foreman_remote_execution_core'
   s.add_dependency 'foreman-tasks', '~> 0.12'
```
stbenjam: I believe you'll need to change this, too.

jlsherrill: I don't think its strictly necessary is it?

stbenjam: Oh right, I hate `~>`

jlsherrill:
kicked off tests here:  http://ci.theforeman.org/job/test_plugin_pull_request/4852/

adamruzicka:
I'm slightly against this change. We sometimes use the dependency to state "we need at least this version of Dynflow because it gives us feature X, which we directly use in the plugin". Loosening the constraint to Foreman's `>=1.0.0 & < 2.0.0` wouldn't really help with it. Or am I missing something?

mmoll:
The background was to assume that _rex doesn't use dynflow directly, but only via -tasks.

iNecas:
:-1: I agree with @adamruzicka, it's not just about the tasks providing that, but sometimes, rex needs a specific feature of Dynflow that others don't need. Keeping this dependency is just cleaner way how to make sure we have the right version

foreman_remote_execution.gemspec
iNecas: If the goal is to mineralize the amount of work, when dynflow raises major version, I'm for using `>= 0.8.26`, just to indicate what is the minimal version, and leaving the `~>` style in foreman only

dLobatog:
@jlsherrill since there's some debate about removing this, can you update the PR to have the same dependency as foreman (> 1.0 < 2.0) instead of removing it? It would unblock tests in any projects that depend on REX or tasks - such as Foreman Ansible or Katello.

iNecas:
I wanted to get this updates as part of https://github.com/theforeman/foreman_remote_execution/pull/330, but no issues with getting this in faster.

dLobatog:
Can you make a release with this? Just a gem release is enough, no need for packages -> http://ci.theforeman.org/job/test_plugin_matrix/5069/database=mysql,ruby=2.4,slave=fast/console

iNecas:
rex 1.5.0 pushed to rubygems to unblock the ansible plugin

ekohl:
It'd have been cleaner if the title was modified to mention it just updates the dependency.

PR #342, opened 2018-04-04
Fixes #21731 - sudo can be used with passwords now

witlessbird:


witlessbird:
* rebased

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -213,7 +237,7 @@ def run_sync(command, stdin = nil)
           started = true
         end
       end
-      session.process(0) until started
+      session.loop(0.1) { !started || (use_sudo_password? && !password_sent) }
```
iNecas: When testing, I got the script stuck here, because some commands were actually not using `sudo` (the run_sync is in rex used mostly for preparing the execution environment, while the sudo is being used just to run the actual user script. 

What was the reason to actually change the original line here (as well as in `run_async` method?). It seems that when I reverted this particular changes, the things started working.

witlessbird: I wasn't sure when run_sync is used and mirrored the changes in run_async. I suppose calling ```session.process``` in a loop is sufficient, but I'm curious now why ```session.loop``` gets stuck, as it is *almost* the same.

iNecas: I think it's not about the `loop, but about the actual condition `(use_sudo_password? && !password_sent)` as when `sudo` is not used in the command, it would never be true.

OT: why do you prefer `loop(0.1)` over `process(0)`? Performance?

witlessbird: > I think it's not about the loop, but about the actual condition(use_sudo_password? && !password_sent)as whensudo` is not used in the command, it would never be true.

Hmm. The whole expression ```!started || (use_sudo_password? && !password_sent)``` will evaluate to false (which is when the loop is terminated) once the ```started``` flag switched to ```true``` though?

> OT: why do you prefer loop(0.1) over process(0)? Performance?

It looks to me that loop makes calls ```process(0)``` one more time after the loop terminates. I'm not sure about the exact reasoning for this, but my guess is that eliminates/lowers the chance of losing incoming/outgoing data (which is only received from/sent onto the wire when ```process``` is called).

adamruzicka:
Would it make sense to have this also exposed in the job invocation form so the user can set the sudo password on a per-job basis the same way as we do for ssh password and key passphrase?

witlessbird:
> Would it make sense to have this also exposed in the job invocation form so the user can set the sudo password on a per-job basis the same way as we do for ssh password and key passphrase?

I think it would be consistent to do so. I am curious however, why ssh password/key passphrase are exposed -- they probably don't change much (if at all) and are tied to the host?

iNecas:
Although I agree that this is more of an corner case, still I think it would make sense to provide this option for some cases, where one would for some reason needed to have non-unified infrastructure.

witlessbird:
* updated tests

iNecas:
Rubocop is not happy

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -7,30 +7,116 @@
...
+
+      user_method = if effective_user.nil? || effective_user == ssh_user
+                      NoopUserMethod.new
+                    elsif effective_user_method == 'sudo'
+                      SudoUserMethod.new(options.fetch(:effective_user, nil),
+                                         options.fetch(:ssh_user, 'root'),
+                                         options.fetch(:secrets, {}).fetch(:sudo_password, nil))
+                    elsif effective_user_method == 'su'
+                      SuUserMethod.new(options.fetch(:effective_user, nil),
+                                       options.fetch(:ssh_user, 'root'))
```
adamruzicka: Why not reuse the `effective_user` and `ssh_user` extracted on lines 102 and 103?

app/views/job_invocations/_form.html.erb
```diff
@@ -93,6 +93,7 @@
     <div class="advanced hidden">
       <%= password_f f, :password, :placeholder => '*****', :label => _('Password'), :label_help => N_('Password is stored encrypted in DB until the job finishes. For future or recurring executions, it is removed after the last execution.') %>
       <%= password_f f, :key_passphrase, :placeholder => '*****', :label => _('Private key passphrase'), :label_help => N_('Key passhprase is only applicable for SSH provider. Other providers ignore this field. <br> Passphrase is stored encrypted in DB until the job finishes. For future or recurring executions, it is removed after the last execution.') %>
+      <%= password_f f, :sudo_password, :placeholder => '*****', :label => _('Sudo password'), :label_help => N_('Key passhprase is only applicable for SSH provider. Other providers ignore this field. <br> Passphrase is stored encrypted in DB until the job finishes. For future or recurring executions, it is removed after the last execution.') %>
```
adamruzicka: The label help should be different from the one for private key passphrase

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -7,30 +7,116 @@
...
+                    else
+                      raise "effective_user_method ''#{@effective_user_method}'' not supported"
```
adamruzicka: Double single quotes?

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -167,9 +253,14 @@ def settings
...
       raise 'Async command already in progress' if @started
       @started = false
+      @user_method.reset
+
       session.open_channel do |channel|
         channel.request_pty
-        channel.on_data { |ch, data| publish_data(data, 'stdout') }
+        channel.on_data do |ch, data|
+          publish_data(data, 'stdout')
+          @user_method.on_data(data, ch)
```
adamruzicka: With this change the password prompt shows up in the jobs output. Maybe it would be nice if this method consumed the output if it matches?

app/views/job_invocations/_form.html.erb
witlessbird: Updated the label.

lib/foreman_remote_execution_core/script_runner.rb
witlessbird: Added filtering out of passwords in job output

adamruzicka:
Needs a rebase

witlessbird:
*rebased

iNecas:
@adamruzicka any other comments?

adamruzicka:
Needs another rebase

witlessbird:
*rebased

PR #343, opened 2018-04-06
Refs #21457 - Provide the expected interface in the base model

ekohl:
f9d116105f70041f639f23675884d7e43ad9a1be started to depend on these methods being present in the base model. This ensures it doesn't crash when providers don't implement these methods.

### DIFF
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -72,5 +72,9 @@ class RemoteExecutionProvider
     def host_setting(host, setting)
       host.params[setting.to_s] || Setting[setting]
     end
+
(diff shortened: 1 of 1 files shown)


adamruzicka:
Please make the :dog: happy. Also could you please prefix the arguments with an underscore?

ekohl:
Updated

ekohl:
Btw, I did consider using a default implementation `host_setting(host, :remote_execution_ssh_password)` but decided against it because I couldn't see the bigger picture and if we could rely on that.

iNecas:
Thanks, Mr. J. agrees as well

PR #346, opened 2018-04-17
Fixes #23297 - Webpack assets path changed

dLobatog:
In `app/views/job_invocations/show.html.erb`, we load the webpack
assets currently by calling a path:

*webpack_asset_paths('remoteexecution'

Since #23215 simplified the name of plugin bundles to just be the name
 of the plugin (instead of "name of the plugin minus dashes and
spaces"), it should be changed to:

*webpack_asset_paths('foreman_remote_execution'

### DIFF
--- a/app/views/job_invocations/show.html.erb
+++ b/app/views/job_invocations/show.html.erb
@@ -1,7 +1,7 @@
 <% title @job_invocation.description, trunc_with_tooltip(@job_invocation.description, 120) %>
 <% stylesheet 'foreman_remote_execution/job_invocations' %>
 <% javascript 'charts', 'foreman_remote_execution/template_invocation' %>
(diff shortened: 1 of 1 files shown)


adamruzicka:
Works for me, but Jenkins is unhappy.

iNecas:
The failures don't seem to be related, and since without this, rex is broken against nightly, I'm merging this right away. Also to unblock the packaging. Btw. I don't see http://projects.theforeman.org/issues/23215/ in 1.18 branch. @dLobatog will you coordinate the cherry-picking so that we can rely the changes is there as well soon?

ares:
nightlies (1.5.1) are affected by this, we need a new release that contains this patch, otherwise the job invocation page does not load

iNecas:
1.5.2 is being released now https://github.com/theforeman/foreman-packaging/pull/2490

PR #349, opened 2018-05-03
Fixes #22483 - use host.host_param instead of host.params if possible

iNecas:
We change the way how we extend the host params to be compatible with
the `host_param` helper.

We also use the `host.host_param` if possible, as it has better
performance properties (the value gets memorized and it doesn't
recalculate the smart variables every time it's called.


iNecas:
to reproduce, try to preview remote_execution_ssh_keys snippet

Replaces https://github.com/theforeman/foreman_remote_execution/pull/311

iNecas:
Test failure unrelated.

iNecas:

iNecas:
Test failures unrelated

PR #350, opened 2018-05-07
Fixes #20778 - handle selection of all hosts on all pages.

witlessbird:


witlessbird:
Apologies, I forgot to mention that changes in remote execution rely on changes in Foreman: https://github.com/theforeman/foreman/pull/5539.

> Also, does this cover the situation with "select all" + no search?

I think so: it looks like "select all" is modeled via an empty search string.

app/controllers/job_invocations_controller.rb
```diff
@@ -7,7 +7,7 @@ class JobInvocationsController < ApplicationController
   def new
     return @composer = prepare_composer if params[:feature].present?
     ui_params = {
-      :host_ids => params[:host_ids],
+      :host_ids => params.key?(:search) ? Host.search_for(params[:search]).pluck(:id) : params[:host_ids],
```
iNecas: I think more natural and effective would be just using the search query directly inside the targeting, instead of expanding it here.

```patch
diff --git a/app/controllers/job_invocations_controller.rb b/app/controllers/job_invocations_controller.rb
index 8b78261..5ef03f7 100644
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -7,8 +7,9 @@ class JobInvocationsController < ApplicationController
   def new
     return @composer = prepare_composer if params[:feature].present?
     ui_params = {
-      :host_ids => params.key?(:search) ? Host.search_for(params[:search]).pluck(:id) : params[:host_ids],
+      :host_ids => params[:host_ids],
       :targeting => {
+        :search_query => params[:search],
         :targeting_type => Targeting::STATIC_TYPE,
         :bookmark_id => params[:bookmark_id]
       }
```

witlessbird:
* ```Targetting``` class is now used for handling of search strings

app/models/job_invocation_composer.rb
```diff
@@ -82,6 +82,14 @@ def triggering
       return trig unless trig.key?(:end_time) && trig[:end_time].keys == keys
       trig.merge(:end_time => Time.local(*trig[:end_time].values_at(*keys)))
     end
+
+    def targeting(targeting_params)
+      to_return = targeting_params.merge(:user_id => User.current.id)
+      if !to_return[:search_query].nil? && to_return[:search_query].empty?
+        to_return[:search_query] = "name != ''" # a dummy search query to match all hosts
```
iNecas: Could you put the this search to https://github.com/theforeman/foreman_remote_execution/pull/350/files#diff-5becfffb6454c87c9b39dd84ae3676d2R16, so that it only applies, when the job invocation page is accessed via the bulk action and the 'all hosts' is selected (behind a condition of `params.key?(:search) && params[:search].empty?`. Otherwise, this would be applied also in cases, where it would be confusing.

witlessbird:
* changed empty sting search parameter handling as per above.

FWIW I think responsibility for creating/massaging input data is spread through several classes and is context dependent, which makes it hard to understand/change/fix. I had to experiment a bunch of times to understand the conversion logic for something as trivial as this fix. I could not figure it out by reading the code, which to me is a sign that classes and execution contexts have no clear boundaries/responsibilities.

iNecas:
Thanks @witlessbird

PR #351, opened 2018-05-10
Refs #21731 - revert using `loop` instead of `process`

iNecas:
In e65c15a, the `process` call was replaced with `loop` for no obvious
reason. This change leads potentially for the commands to get stuck.

I've reproduced this by trying `sudo` effective method without `sudo`
being present on the system.


iNecas:
Thanks @pondrejk for review and testing.

iNecas:
Merging as the issues is potentially causing quite a lot of troubles, so I would like to have that in before new release.

PR #354, opened 2018-05-30
Refs #21717 - Added support for enable and disable

ik5:


ares:
This should go through https://github.com/theforeman/community-templates/blob/develop/job_templates/service_action_-_ansible_default.erb and https://github.com/theforeman/community-templates/blob/develop/job_templates/service_action_-_ssh_default.erb first, then should be synced back to REX and foreman_ansible.

ik5:
@ares  oh, thanks, 
so it means that any changes to the templates anywhere starts on community templates, are there any exceptions for this?

ares:
nope, AFAIK all templates goes through that repo

adamruzicka:
Another thing to consider is that while `systemctl enable` works, `service enable` doesn't if I recall so the template will further tweaking.

PR #355, opened 2018-05-31
Fixes #23163 - Fix breadcrumbs in template invocation details

adamruzicka:
breadcrumbs1
breadcrumbs2

### DIFF
--- a/app/controllers/api/v2/template_invocations_controller.rb
+++ b/app/controllers/api/v2/template_invocations_controller.rb
@@ -1,15 +1,13 @@
 module Api
   module V2
     class TemplateInvocationsController < ::Api::V2::BaseController
-      include ::Api::Version2
-      include ::Foreman::Renderer
 
       before_action :find_job_invocation, :only => %w{template_invocations}
 
-      api :GET, '/job_invocations/:id/template_invocations',
+      api :GET, '/job_invocations/:job_invocation_id/template_invocations',
           N_('List template invocations belonging to job invocation')
       param_group :search_and_pagination, ::Api::V2::BaseController
-      param :id, :identifier, :required => true
+      param :job_invocation_id, :identifier, :required => true
       def template_invocations
         @template_invocations = resource_scope.paginate(paginate_options)
         render :layout => 'api/v2/layouts/index_layout'
--- a/app/controllers/api/v2/template_invocations_controller.rb
+++ b/app/controllers/api/v2/template_invocations_controller.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 module Api
   module V2
     class TemplateInvocationsController < ::Api::V2::BaseController
--- a/app/controllers/api/v2/template_invocations_controller.rb
+++ b/app/controllers/api/v2/template_invocations_controller.rb
@@ -1,0 +1,46 @@
+module Api
+  module V2
+    class TemplateInvocationsController < ::Api::V2::BaseController
+      include ::Api::Version2
+      include ::Foreman::Renderer
+
+      before_action :find_job_invocation, :only => %w{template_invocations}
+
+      api :GET, '/job_invocations/:id/template_invocations',
+          N_('List template invocations belonging to job invocation')
+      param_group :search_and_pagination, ::Api::V2::BaseController
+      param :id, :identifier, :required => true
+      def template_invocations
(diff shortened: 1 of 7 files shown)


app/controllers/api/v2/template_invocations_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      include ::Api::Version2
+      include ::Foreman::Renderer
+
+      before_action :find_job_invocation, :only => %w{template_invocations}
+
+      api :GET, '/job_invocations/:id/template_invocations',
+          N_('List template invocations belonging to job invocation')
+      param_group :search_and_pagination, ::Api::V2::BaseController
+      param :id, :identifier, :required => true
+      def template_invocations
```
iNecas: why not calling it `index`, which is the standard name for method listing all the resources in rails?

app/controllers/api/v2/template_invocations_controller.rb
```diff
@@ -0,0 +1,46 @@
+module Api
+  module V2
+    class TemplateInvocationsController < ::Api::V2::BaseController
+      include ::Api::Version2
+      include ::Foreman::Renderer
+
+      before_action :find_job_invocation, :only => %w{template_invocations}
+
+      api :GET, '/job_invocations/:id/template_invocations',
```
iNecas: In nested resources, we would usually do `:job_invocation_id` instead of `:id`, as the `:id` usually refers to the resource of the controller itself (which is `TemplateInvocation` in this case`)

app/controllers/api/v2/template_invocations_controller.rb
iNecas: Perhaps it was moved from `JobInvocationController`?

app/controllers/api/v2/template_invocations_controller.rb
```diff
@@ -0,0 +1,46 @@
+module Api
+  module V2
+    class TemplateInvocationsController < ::Api::V2::BaseController
+      include ::Api::Version2
+      include ::Foreman::Renderer
```
iNecas: Is this needed?

app/controllers/api/v2/template_invocations_controller.rb
adamruzicka: Is it still an `index` when it is by default scoped by another resource? I would expect `index` to return all template invocations (with possible search), not only template invocations belonging to a certain job invocation

app/controllers/api/v2/template_invocations_controller.rb
adamruzicka: Removed

app/controllers/api/v2/template_invocations_controller.rb
adamruzicka: This leads to a slightly confusing situation where the API description states it is `job_invocation_id`, but in the controller it is accessible as `params[:id]`

iNecas:
One thing I forgot to ask yesterday: could you cover it by some controller tests?

app/views/template_invocations/show.html.erb
```diff
@@ -1,4 +1,13 @@
-<% title _('Detail of %s run') % @template_invocation.job_invocation.job_category %>
+<% items = [{ :caption => @template_invocation.job_invocation.description,
+              :url => job_invocation_path(@template_invocation.job_invocation_id) },
+            { :caption => _('Template Invocation for %s') % @template_invocation.host.name }] %>
+
+<% breadcrumbs(:resource_url => template_invocations_api_job_invocation_path(@template_invocation.job_invocation_id),
+               :name_field => 'host_name',
+               :switcher_item_url => '/template_invocations/:id',
```
sharvit: Is it possible to use rails helper to create the `switcher_item_url`?
Example:
https://github.com/theforeman/foreman/blob/833e2aef24d6f3cb6eae90213ea198dd354a5709/app/views/provisioning_templates/edit.html.erb#L2

app/controllers/api/v2/template_invocations_controller.rb
```diff
@@ -0,0 +1,46 @@
+module Api
```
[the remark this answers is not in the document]
iNecas: @houndci-bot neither @rubocop is happy about this

iNecas:
Re-tested and works like a charm. Thanks @adamruzicka

PR #358, opened 2018-06-13
Fixes #21704 - Prevent user from rerunning a job with missing template

adamruzicka:


adamruzicka:
image

In case users try to rerun a job for which they can't see the template, the new job invocation form shows an error box, no job category is selected and the submit button is disabled. After they select a proper category-template combination, the form redraws, error box disappears and submit button gets enabled.

iNecas:
Would it be possible to cover this with some tests. At least one for `rerun_possible?` method.

Also; how would the API behave in this case, or it's not affected at all?

iNecas:
Test failures seem related

iNecas:
Thanks @adamruzicka

PR #360, opened 2018-06-27
Fixes #24085 - Include missing link in breadcrumbs

adamruzicka:
image

CC @Rohoover


iNecas:
Works as expected. Thanks @adamruzicka

PR #361, opened 2018-06-27
Fixes #23394 - Unable to set effective user for job templates

ik5:
There was a problem with the API structure, it didn't work since day one.
The `ssh` sub entry was not recognized by `strong params`, making the `ssh`
object (hash) and children not to be accessible.

This PR fixes the issue, but "breaks" the bad API documentation that existed
before.

### DIFF
--- a/app/controllers/api/v2/job_templates_controller.rb
+++ b/app/controllers/api/v2/job_templates_controller.rb
@@ -57,12 +57,10 @@ module Api
           param :snippet, :bool, :allow_nil => true
           param :audit_comment, String, :allow_nil => true
           param :locked, :bool, :desc => N_('Whether or not the template is locked for editing')
-          param :ssh, Hash, :desc => N_('SSH provider specific options') do
-            param :effective_user, Hash, :desc => N_('Effective user options') do
-              param :value, String, :desc => N_('What user should be used to run the script (using sudo-like mechanisms)'), :allowed_nil => true
-              param :overridable, :bool, :desc => N_('Whether it should be allowed to override the effective user from the invocation form.')
-              param :current_user, :bool, :desc => N_('Whether the current user login should be used as the effective user')
-            end
+          param :effective_user_attributes, Hash, :desc => N_('Effective user options') do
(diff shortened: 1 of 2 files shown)


app/controllers/api/v2/job_templates_controller.rb
```diff
@@ -57,12 +57,10 @@ def show; end
...
           param :audit_comment, String, :allow_nil => true
           param :locked, :bool, :desc => N_('Whether or not the template is locked for editing')
-          param :ssh, Hash, :desc => N_('SSH provider specific options') do
-            param :effective_user, Hash, :desc => N_('Effective user options') do
-              param :value, String, :desc => N_('What user should be used to run the script (using sudo-like mechanisms)'), :allowed_nil => true
-              param :overridable, :bool, :desc => N_('Whether it should be allowed to override the effective user from the invocation form.')
-              param :current_user, :bool, :desc => N_('Whether the current user login should be used as the effective user')
-            end
+          param :effective_user_attributes, Hash, :desc => N_('Effective user options') do
+            param :value, String, :desc => N_('What user should be used to run the script (using sudo-like mechanisms)'), :allowed_nil => true
```
adamruzicka: Having this called `value` makes sense in the API context, but in hammer having an option called `value`, which is not really a value of the object you're manipulating with, feels weird 
```shell
hammer job-template create \
    --file my_template.erb \
    --name my_template \
    --job-category Misc \
    --provider-type SSH \
    --value nobody
```

Maybe renaming this either in the apidoc or in hammer would lead to a better user experience

ik5: I do agree with you, but I don't think it should be on this PR, this one just fixes an issue, another PR can be on better naming things

adamruzicka: Ah, so it was called `value` even before, but it didn't work. Guess I'm ok with that

ik5: the `ssh` hash was inaccessible because it does not exists anywhere else, so I fixed that.

ik5:
done :)

adamruzicka:
Thanks @ik5 !

PR #362, opened 2018-06-29
Fixes #24113 - fix fake executor

iNecas:
This fixes the error with undefined method build, as well as issue with
checking the exit status, as in case of exception, it's actually
'EXCEPTION' string.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -121,7 +121,7 @@ module Actions
       def update_host_status
         host = Host.find(input[:host][:id])
         status = host.execution_status_object || host.build_execution_status_object
-        status.status = exit_status.to_i.zero? ? HostStatus::ExecutionStatus::OK : HostStatus::ExecutionStatus::ERROR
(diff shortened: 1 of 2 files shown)


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -121,7 +121,7 @@ def exit_status
       def update_host_status
         host = Host.find(input[:host][:id])
         status = host.execution_status_object || host.build_execution_status_object
-        status.status = exit_status.zero? ? HostStatus::ExecutionStatus::OK : HostStatus::ExecutionStatus::ERROR
+        status.status = exit_status.to_i.zero? ? HostStatus::ExecutionStatus::OK : HostStatus::ExecutionStatus::ERROR
```
adamruzicka: This wouldn't update the execution status correctly
```
pry(#<Actions::RemoteExecution::RunHostJob>)> exit_status
=> "EXCEPTION"

pry(#<Actions::RemoteExecution::RunHostJob>)> exit_status.to_i
=> 0

iNecas: Good catch: updated

adamruzicka:
Thanks @iNecas !

PR #363, opened 2018-07-04
fixes #24130 - use new rendering engine

timogoebel:
This needs theforeman/foreman#5683.

### DIFF
--- a/app/models/input_template_renderer.rb
+++ b/app/models/input_template_renderer.rb
@@ -5,9 +5,9 @@ class InputTemplateRenderer
   class RenderError < ::Foreman::Exception
   end
 
-  include UnattendedHelper
+  delegate :logger, to: Rails
 
-  attr_accessor :template, :host, :invocation, :input_values, :error_message
+  attr_accessor :template, :host, :invocation, :input_values, :error_message, :templates_stack
 
   # takes template object that should be rendered
   # host and template invocation arguments are optional
@@ -16,8 +16,8 @@ class InputTemplateRenderer
     raise Foreman::Exception, N_('Recursive rendering of templates detected') if templates_stack.include?(template)
 
     @host = host
-    @invocation = invocation
     @template = template
+    @invocation = invocation
     @input_values = input_values
     @preview = preview
     @templates_stack = templates_stack + [template]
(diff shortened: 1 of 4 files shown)


app/lib/foreman_remote_execution/renderer/input_template_scope.rb
```diff
@@ -0,0 +1,13 @@
+module ForemanRemoteExecution
```
timogoebel: We don't need this actually.

iNecas:
Thanks for looking into this. During the first round of testing, I've noticed in editing some templates (such as Package Action - SSH Default), the preview fails with:

```
Warning! Template is missing Missing template job_templates/preview, templates/preview, application/preview with {:locale=>[:en], :formats=>[:html, :text, :js, :css, :ics, :csv, :vcf, :png, :jpeg, :gif, :bmp, :tiff, :svg, :mpeg, :xml, :rss, :atom, :yaml, :multipart_form, :url_encoded_form, :json, :pdf, :zip, :gzip], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :rabl]}. Searched in: * "/home/inecas/Projects/ws/foreman-rex/foreman-rails5/app/views" * "/home/inecas/Projects/ws/foreman-rex/foreman_ansible/app/views" * "/home/inecas/Projects/ws/foreman-rex/foreman_remote_execution/app/views" * "/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/views" * "/home/inecas/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/apipie-rails-0.5.9/app/views"
```

I didn't have a chance to debug this more deeply yet.

timogoebel:
> I didn't have a chance to debug this more deeply yet.

It looks like this has been broken for a while (I suggest the rails upgrade).
There is not a single test for the controller, so it is no surprise this wasn't noticed earlier.

The actual error is that `operatingsystem` is undefined for the scope. I still need to investigate this.
But first, I'll add a couple of controller tests.

timogoebel:
@iNecas: I forgot to pass `@host` as a variable to the renderer, that's why `@host.operatingsystem` showed an error.
The error message you saw is because `render text:` does not work anymore with Rails 5, it has to be `render plain:`. I fixed that as well and added some basic tests for `#preview` in templates controller.

app/models/input_template_renderer.rb
```diff
@@ -16,84 +16,64 @@ def initialize(template, host = nil, invocation = nil, input_values = nil, previ
...
+    @invocation = invocation
     @input_values = input_values
     @preview = preview
     @templates_stack = templates_stack + [template]
   end
 
   def render
     @template.validate_unique_inputs!
-    render_safe(@template.template, ::Foreman::Renderer::ALLOWED_HELPERS + [ :input, :render_template, :preview?, :render_error ], :host => @host)
+    Foreman::Renderer.render_template(
```
iNecas: With the latest version of https://github.com/theforeman/foreman/pull/5683/files, I'm getting:

```
Failed to initialize: RuntimeError - Failed rendering template: error during rendering: undefined method `render_template' for Foreman::Renderer:Module
   2:
Did you mean? render_template_to_tempfile
```

timogoebel:
Rebased and updated to work with the latest PR code.

iNecas:
Tested and works well: ready to merge after the core one is in.

iNecas:
Running tests locally, they also pass successfully.

iNecas:

iNecas:
Thanks @timogoebel for great work.

iNecas:
foreman-rex 1.6.0 has been pushed to rubygems

PR #364, opened 2018-07-05
Fixes #24159 - Non admin user failed to render included template

ranjan:


adamruzicka:
I see why this didn't work before. Could we get a test for it?

ranjan:
@adamruzicka My dev environment for remote-execution is not yet ready, So for now its not possible to add test cases :disappointed:

adamruzicka:
Added a test to this PR, @iNecas could you please take a look?

iNecas:
Thanks @adamruzicka and @ranjan

PR #367, opened 2018-07-19
Fixes #24226 - Do not display same proxy twice on Subnet

dLobatog:
Right now, if you have a proxy with 2 features related with Remote
Execution (e.g: Ansible and SSH), it shows up twice in the Subnet >
Remote Execution tab.

BEFORE:
 https://webmshare.com/jRM01

AFTER:
https://webmshare.com/VPzgv

This commit fixes that so that only 1 proxy shows up.


ares:
LGTM, the duplication is caused by the fact it matches on 2 features. I only wonder if distinct shouldn't be default in `with_features` scope.

PR #368, opened 2018-07-24
Fixes #24368 - Provide output for hosts which are not planned yet

adamruzicka:


iNecas:
Tested and works well. Thanks @adamruzicka

PR #369, opened 2018-07-31
Fixes #14019 - Use template name as default description template

adamruzicka:
Originally we used job_category.

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -136,7 +136,7 @@ class JobTemplate < ::Template
 
   def generate_description_format
     if description_format.blank?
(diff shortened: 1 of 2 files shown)


iNecas:
Test failures related

adamruzicka:
[test]

iNecas:
@adamruzicka you probably meant [test foreman_remote_execution] - after the changes in the pr scanner, the previous way doesn't work anymore.

adamruzicka:
Right, I wasn't sure which way is the right way now. Anyway, thanks for triggering the tests, they're green now

iNecas:
Thanks @adamruzicka

PR #370, opened 2018-07-31
refs #24130 - pass source to template scope

timogoebel:
After https://github.com/theforeman/foreman/commit/b58439fabbeac20b68902021fed366725c26dc92 was merged, we should also pass the source to the scope so snippets still work correctly.

### DIFF
--- a/app/models/input_template_renderer.rb
+++ b/app/models/input_template_renderer.rb
@@ -31,6 +31,7 @@ class InputTemplateRenderer
       host: host
     )
     @scope = Foreman::Renderer.get_scope(
+      source: source,
       host: host,
       klass: renderer_scope,
       template_input_values: @template_input_values,


iNecas:
What's the reproducer for the issue? It seems I can use a snippet in job templates even without this patch.

timogoebel:
> What's the reproducer for the issue? It seems I can use a snippet in job templates even without this patch.

@iNecas: This will only break if the snippets are defined by another source. So right now, this should just be a cosmetic change.

iNecas:
When trying to preview the template with this patch, I'm getting `Warning! Problem with previewing the template: error during rendering: unknown keyword: source. Note that you must save template input changes before you try to preview it.`

timogoebel:
Are you on the latest core develop? The kwarg should be defined, see https://github.com/theforeman/foreman/blob/b58439fabbeac20b68902021fed366725c26dc92/lib/foreman/renderer.rb#L35.
Also tests seem to pass on jenkins.

timogoebel:
@iNecas: Thanks. Rebased.

iNecas:
Thanks @timogoebel

PR #371, opened 2018-08-01
Fixes #20776 - Do not re-render template if it can be loaded

adamruzicka:
When the user visits the job invocation details page, we rendered the
templates for first 20 hosts, which made the page load slowly. With
this change we first try to look into a task for this host, which may
contain the already rendered template.

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -161,7 +161,7 @@ module RemoteExecutionHelper
 
   def preview_box(template_invocation, target)
     renderer = InputTemplateRenderer.new(template_invocation.template, target, template_invocation)
-    if (preview = renderer.preview)
+    if (preview = (load_template_from_task(template_invocation, target) || renderer.preview))
       content_tag :pre, preview
     elsif target.nil?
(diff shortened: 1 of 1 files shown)


app/helpers/remote_execution_helper.rb
```diff
@@ -161,7 +161,7 @@ def invocation_result(invocation, key)
 
   def preview_box(template_invocation, target)
     renderer = InputTemplateRenderer.new(template_invocation.template, target, template_invocation)
-    if (preview = renderer.preview)
+    if (preview = (load_template_from_task(template_invocation, target) || renderer.preview))
```
ares: what are the conditions for input not yet have rendered template? e.g. future execution?

adamruzicka: For future execution the task object would not exist at all (only the parent task exists at that time). This is to prevent a race condition where the `RunHostJob` already exists, but its `ProxyAction` was not planned yet

adamruzicka: I guess I should have attached the benchmarks to the PR

ares:
Thanks @adamruzicka, works well! Failures are unrelated, merging.

PR #372, opened 2018-08-23
Fixes #24698 - RemoteExecutionFeature.register Rails 5.2 fix

iNecas:
Using table_exists? in db:create doesn't work anymore.

### DIFF
--- a/app/models/remote_execution_feature.rb
+++ b/app/models/remote_execution_feature.rb
@@ -23,8 +23,9 @@ class RemoteExecutionFeature < ApplicationRecord
     self.find_by(label: label) || raise(::Foreman::Exception.new(N_('Unknown remote execution feature %s'), label))
   end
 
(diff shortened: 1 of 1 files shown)


iNecas:
I didn't find a way to decrease the measured complexity, so I kindly asked the :policeman: to turn around and leave it alone.

mmoll:
test failures unrelated :green_heart:

adamruzicka:
Thanks @iNecas !

ares:
This introduced a regression in REX and openscap - https://github.com/theforeman/foreman_ansible/blob/master/lib/foreman_ansible/remote_execution.rb#L15-L31

`Foreman.in_setup_db_rake?` returns true in this case... would it be acceptable to skip the check if we pass some option?

iNecas:
I don't understand the context where this is failing: perhaps the best to discuss this while I'm in the office

PR #373, opened 2018-08-28
Fix apidoc wrt puppet_class_name

mdellweg:


mdellweg:
@iNecas can you see, why the test are failing? Is this expected? I do not think, they are related to my change.

PR #374, opened 2018-09-07
Fixes #24866 - introduce caching for macros

ares:


ares:
for use, see https://github.com/theforeman/foreman_ansible/pull/203

basically in other macros we can use cache('some_key') { ... } and the value will be cached for all template invocations, REX will drop all such caches just after all template invocations finishes, one can use as many different keys as needed, we drop all that are prefixed with job_invocation_$id

ares:
I'd appreciate CP to 1.5 branch and new release too

iNecas:
@ares mind adding some test, mainly to touch the code that is being introduced. People that will work on Rails upgrades in the future will thank you in the future.

ares:
pushed with tests and relinked to REX issue (so you can track version properly :-)

iNecas:
whos_awesome

PR #377, opened 2018-09-13
Fixes #24124 - Allow executing jobs on hosts in randomized order

adamruzicka:
- [x] tests

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -69,7 +69,7 @@ module Actions
       end
 
       def hosts
-        job_invocation.targeting.hosts
+        job_invocation.targeting.hosts.order("#{TargetingHost.table_name}.id")
       end
 
       def set_up_concurrency_control(invocation)
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -69,7 +69,7 @@ module Actions
       end
 
       def hosts
-        job_invocation.targeting.hosts.order(:name, :id)
+        job_invocation.targeting.hosts
       end
 
       def set_up_concurrency_control(invocation)
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -103,21 +103,21 @@
       <%= number_f f, :time_span, :label => _('Time span'), :placeholder => 'N', :min => 1, :label_help => N_("Distribute execution over N seconds")  %>
     </div>
 
-    <div class="form-group">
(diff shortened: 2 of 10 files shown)


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -68,7 +68,7 @@ def total_count
       end
 
       def hosts
-        job_invocation.targeting.hosts.order(:name, :id)
+        job_invocation.targeting.hosts
```
iNecas: Will this work with the batches? With randomized order, I guess every batch could include some hosts that were already planned for. One option to fix this could be to exclude the hosts that were already planned from this search, instead of using paging for moving though the hosts list

adamruzicka: I run a test job with randomized order on 500 hosts where each host would write its name into a shared file, it seems to work just fine.

```shell
curl https://paste.pound-python.org/raw/6NcIyuj45nDXLhuXTvjK/ | sort | uniq -d
```

adamruzicka: Guess the question is if we can rely on a db always reading the join table records in the order in which they were saved

adamruzicka:

iNecas:
Removing from 1.20 candidates, as it's touching existing code, and I would like to gain more confidence it's not breaking anything in nightlies: targeting 1.21 seems like a good fit here.

app/lib/actions/remote_execution/run_hosts_job.rb
iNecas: Yes, I would be more confident if param we're basing the ordering on (even if it was the id of the `id` of the `targeting_hosts` table.

adamruzicka:
The tests are green, apart from sqlite being sqlite.

app/views/job_invocations/_form.html.erb
```diff
@@ -110,7 +110,15 @@
         <%= radio_button_f targeting_fields, :targeting_type, :value => Targeting::STATIC_TYPE, :text => _(Targeting::TYPES[Targeting::STATIC_TYPE]) %>
         <%= radio_button_f targeting_fields, :targeting_type, :value => Targeting::DYNAMIC_TYPE, :text => _(Targeting::TYPES[Targeting::DYNAMIC_TYPE]) %>
       </div>
+    </div>
+
+    <div class="form-group advanced hidden">
+      <%= add_label({ :label => _('Execution ordering'), :label_help => _("Execution ordering determines whether the jobs should be executed on hosts in alphabetical order or in randomized order.<br><ul><li><b>Ordered</b> - executes the jobs on hosts in alphabetical order</li><li><b>Randomized</b> - randomizes the order in which jobs are executed on hosts</li></ul>") }, f, :randomized_ordering) %>
```
ares: it would be good to move above tergeting type, now that it's no longer advanced field

ares:
Also a patch to hammer plugin for showing the order is necessary

adamruzicka:
Accompanying hammer PR opened https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/24

ares:
Test failures unrelated, tested also with the hammer PR, works nicely. Let's get this in, merging! Thanks @adamruzicka, please sign up for the next community demo :-)

PR #379, opened 2018-09-18
Fixes #22750 - Rerun failed should rerun also on cancelled hosts

adamruzicka:
If a job is still running the behavior stays the same as it was. However if the
job is finished, rerunning the job will execute it also on hosts which were not
even planned and therefore don't have a template invocation.


iNecas:
Issue from testing: when not having any failed jobs (only cancelled) the `Rerun failed` button is still gray:

rerun-failed

iNecas:
Thanks @adamruzicka

PR #380, opened 2018-09-18
Fixes #22209 - Render empty lines properly in output

adamruzicka:
For some reason the output uses CRLF to mark a newline, writing CRs into
the terminal window led to line numbers not matching the lines. Also we
were splitting the output on newlines and trailing empty fields were
omitted.

To reproduce run the following
```shell
printf "\none\n"
printf "\ntwo\n"
printf "\nthree\n"
```

### DIFF
--- a/app/helpers/job_invocation_output_helper.rb
+++ b/app/helpers/job_invocation_output_helper.rb
@@ -1,5 +1,5 @@
 module JobInvocationOutputHelper
-  COLOR_PATTERN = /\e\[.*?m/.freeze
+  COLOR_PATTERN = /\e\[.*?m/
   CONSOLE_COLOR = {
     '31' => 'red',
     '32' => 'lightgreen',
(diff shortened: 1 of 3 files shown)


adamruzicka:
This still doesn't work fully for ansible and it is related to the output colorization code. WIP

app/helpers/job_invocation_output_helper.rb
```diff
@@ -17,13 +18,13 @@ module JobInvocationOutputHelper
...
   def colorize_line(line)
-    line = line.gsub(/\e\[.*?m/) do |seq|
+    line = line.gsub(COLOR_PATTERN) do |seq|
       color = seq[/(\d+)m/,1]
       "{{{format color:#{color}}}}"
     end
 
     current_color = 'default'
-    out = %{<span style="color: #{@current_color}">}
+    out = %{<span style="color: #{current_color}">}
```
iNecas: Isn't the issue now, that the colorization doesn't survive the the new line?
[rhlogo.zip](https://github.com/theforeman/foreman_remote_execution/files/2396250/rhlogo.zip)

iNecas: rhlogo
The whole hat should be red

adamruzicka:
Fixed
image

adamruzicka:
Test failures don't seem to be related

iNecas:
Test failures unrelated. Thanks @adamruzicka

PR #382, opened 2018-09-19
Fixes #24989 - Modularity Templates for REX

parthaa:


parthaa:
Based of https://github.com/theforeman/community-templates/blob/develop/job_templates/module_action_-_ssh_default.erb

iNecas:
Thanks, it would still be better to get this in using the sync script, but it will take time to fix that. I'm ok with getting this in now. Test failures unrelated

PR #386, opened 2018-09-26
Fixes #23955 - Add macro for accessing user executing the job

adamruzicka:


iNecas:
Tested and works well. Thanks @adamruzicka

PR #387, opened 2018-09-26
Fixes #22510 - Make inputs with options really optional

adamruzicka:
Before this patch even if an input was not marked as required, the
validations wouldn't let the user execute a job where a value for this
job was not provided.


iNecas:
Tested and worked well. Thanks @adamruzicka

ares:
great, now I need to port it to core too :-)

PR #389, opened 2018-10-04
Fixes #25122 - reuse inputs from core

ares:


adamruzicka:
Running a REX job fails with

```
»Failed rendering template: »error during rendering: undefined method `template_input_values' for #<InputTemplateRenderer:0x000056452f1bc9f0>
Did you mean?  templates_stack««
```

adamruzicka:
Jobs now run fine but I'm getting a routing error when visiting the new job form.

```
2018-10-05T12:15:44 [F|app|] ActionController::RoutingError (No route matches [GET] "/javascripts/template_input.js"):
-----B<-----SNIP-----B<-----
2018-10-05T12:15:45 [I|app|]   Rendered /home/aruzicka/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/actionpack-5.2.1/lib/action_dispatch/middleware/templates/rescues/routing_error.html.erb within rescues/layout (771.7ms)
```

johnpmitsch:
This fixed the migration issues in my dev environment :+1:

johnpmitsch:
I am getting this while running the new module stream templates in Katello:
```
and now getting this Failed to initialize: RuntimeError - Failed rendering template: error during rendering: ERF47-0057
[TemplateInput::ValueNotReady]: Input 'pre_script' is not ready for rendering
```

app/models/job_invocation_composer.rb
```diff
@@ -461,14 +461,21 @@ def template_invocation(job_template)
     pattern_template_invocations.find { |invocation| invocation.template == job_template }
   end
 
-  def template_invocation_input_value_for(job_template, input)
+  # TODO replaced by input_value_for
```
iNecas: What's the thing TODO here?

ares:
update the JS and the comment leftover, I reproduced the issue than @johnpmitsch reported, it's related to `render_template` macro that probably does not pass inputs correctly, looking into that

ares:
Nested rendering now works too, the main change is that previously we considered user input ready if there was template invocation ready, now in core we check whether there's incoming parameter. That no longer works for nested renderer instances, because we only pass variables that we receive when calling `render_template`, so optional params would fail with Not Ready error. Therefore we pass also all inputs that we know the template requires, but we set values to nil. That restores the original behavior as much as possible.

I'll look at tests on Monday but tested the functionality with this and ansible and all looks good now.

akofink:
I can also confirm that this fixes the migration issues. :water_buffalo:

ares:
There are still tests that needs to be fixed. Sorry for the delay, it's not as easy as it looked. The biggest problem now is supporting optional inputs the way they used to work before.

ares:
it should turn green now

iNecas:
I've tested this extensively, including foreign input sets via UI, hammer and template syncing, import, export, hit two issues, but both are present on sat 6.4 beta, as well as on master branch without this PR, so they are not related (https://projects.theforeman.org/issues/25161, https://projects.theforeman.org/issues/25162).

ACK, @adamruzicka anything else to add? Otherwise I'm merging

adamruzicka:
Nothing to add, ACK from me as well. The merge button is yours

iNecas:
Thanks @ares, @adamruzicka, @johnpmitsch, @lpramuk and @akofink for participation. Working on new rex release now

iNecas:
rex 1.6.4 pushed to rubygems, packaging PR is here https://github.com/theforeman/foreman-packaging/pull/3075

dLobatog:
Great job here! :+1:

PR #391, opened 2018-11-05
Rubocop

adamruzicka:
PR tests are currently failing due to rubocop complaining


iNecas:
I would vote for disabling `rubocop -a --only Layout/EmptyLineAfterGuardClause`, but if you feel it's better with it, I can live with it

adamruzicka:
Reverted the commit and disabled the cop

iNecas:
Thanks @adamruzicka, test failures unrelated

PR #392, opened 2018-11-07
 Fixes #24138 - Trigger tasks in batches

adamruzicka:
Requires:
- [x] https://github.com/theforeman/smart_proxy_dynflow/pull/55
- [x] https://github.com/theforeman/foreman-tasks/pull/369

### DIFF
--- a/lib/foreman_remote_execution_core/polling_script_runner.rb
+++ b/lib/foreman_remote_execution_core/polling_script_runner.rb
@@ -3,8 +3,8 @@ module ForemanRemoteExecutionCore
 
     DEFAULT_REFRESH_INTERVAL = 60
 
-    def initialize(options, user_method)
-      super(options, user_method)
+    def initialize(options, user_method, suspended_action: nil)
+      super(options, user_method, suspended_action: suspended_action)
       @callback_host = options[:callback_host]
       @task_id = options[:uuid]
       @step_id = options[:step_id]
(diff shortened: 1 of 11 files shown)


bkearney:
@adamruzicka what is the status of this?

adamruzicka:
This is waiting for the foreman-tasks dependency https://github.com/theforeman/foreman-tasks/pull/385

iNecas:
I've released new foreman-tasks/core and bumped dependency here: let's see how the tests go

lib/foreman_remote_execution_core/polling_script_runner.rb
```diff
@@ -3,8 +3,8 @@ class PollingScriptRunner < ScriptRunner
 
     DEFAULT_REFRESH_INTERVAL = 60
 
-    def initialize(options, user_method)
-      super(options, user_method)
+    def initialize(options, user_method, suspended_action: nil)
```
iNecas: @adamruzicka I needed to add this to fix the async mode

iNecas:
During testing, I've noticed that disabling of the batch mode actually was not working, so fixed it here https://github.com/theforeman/foreman-tasks/pull/397/files (+ reduced the batch size to a bit smaller number).

All my tests worked well at the end: I was a bit surprised that turning the `async` mode on proxy influenced the way how ansible is reporting (I would expect that affecting ssh only), but that's not related to this PR.

adamruzicka:
Looks good to me

iNecas:
Changed squashed: merging

iNecas:
Thanks @adamruzicka

PR #394, opened 2018-11-14
Fixes #21595 - Show Static Query in Job invocations overview

apuntamb:


app/views/job_invocations/index.html.erb
```diff
@@ -6,6 +6,7 @@
   <thead>
     <tr>
       <th><%= sort :description, :as => _('Description') %></th>
+      <th class="col-md-1"><%= _('Static Query') %></th>
```
adamruzicka: Because we will display this column even if the user selected "dynamic query", the name would be misleading. Maybe just "Search Query" would be a better name?

adamruzicka:
The failures are because of rubocop issues which are being handled in another PR. Don't worry, there's nothing wrong with this one

apuntamb:
@adamruzicka the tests are now green. Can we merge this now ?

adamruzicka:
Thanks @apuntamb !

PR #397, opened 2018-11-26
Fixes #25534 - Add /api/v2/hosts/:id/ssh_params

iNecas:


mvollmer:
I have trouble getting this to work:

```
$ curl -b _session_id=xxx http://localhost:3000/api/v2/hosts/1/ssh_params
<div id="container">
  <h2>The action &#39;ssh_params&#39; could not be found for Api::V2::HostsController</h2>
</div>
```

So I think the routing is all set up, but the HostsControllerExtension is somehow not effective.

Also, the `:iod` portion doesn't seem to match arbitrary strings:
```
curl -s -b _session_id=xxx http://localhost:3000/api/v2/hosts/x.y.z/ssh_params | grep "No route"
  <h2>No route matches [GET] &quot;/api/v2/hosts/x.y.z/ssh_params&quot;</h2>
```

Other API methods for the HostsController include `{:id=>/[^\/]+/}`, maybe that's necessary here also.

mvollmer:
I don't know what I am doing, but this seems to help:

```diff
diff --git a/app/controllers/concerns/foreman_remote_execution/api/v2/hosts_controller_extensions.rb b/app/controllers/concerns/foreman_remote_execution/api/v2/hosts_controller_extensions.rb
index 3fd5ef6..cf08bce 100644
--- a/app/controllers/concerns/foreman_remote_execution/api/v2/hosts_controller_extensions.rb
+++ b/app/controllers/concerns/foreman_remote_execution/api/v2/hosts_controller_extensions.rb
@@ -17,3 +17,5 @@ module Concerns::ForemanRemoteExecution::Api::V2::HostsControllerExtensions
     end
   end
+
+  prepend_features ::Api::V2::HostsController
 end
```

mvollmer:
Here is a patch to make hostnames with dots work:
```diff
diff --git a/config/routes.rb b/config/routes.rb
index 1969288..26295b9 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -81,5 +81,7 @@ Rails.application.routes.draw do
       resources :remote_execution_features, :only => [:show, :index, :update]
 
-      get 'hosts/:id/ssh_params', to: 'hosts#ssh_params'
+      constraints(:id => /[^\/]+/) do
+        get 'hosts/:id/ssh_params', to: 'hosts#ssh_params'
+      end
```

iNecas:
Thanks for the `constraint` pointers. PR updated.

Regarding `prepend`: for some reason, it works for me, but seems to not work for you or in jenkins: trying to figure out now that's the difference.

Btw. any plugins you would have installed when testing? Are you running that in development, or production?

iNecas:
@mvollmer I think I've found the reason why it failed: https://github.com/theforeman/foreman_remote_execution/pull/397/commits/9eaa4a6833e7b228fa3e0ad6fba07362b074c2d3, I was testing this while having `foreman_discovery` enabled, which lead for me not to be running into that particular issue.

I think everything should work just fine now

app/controllers/concerns/foreman_remote_execution/api/v2/hosts_controller_extensions.rb
```diff
@@ -0,0 +1,19 @@
...
+  def ssh_params
+    render :json => SSHExecutionProvider.ssh_params(@host)
+  end
+
+  protected
+
+  def action_permission
+    case params[:action]
+    when 'ssh_params'
+      'view'
```
iNecas: I wonder if we should have some more expiclit persmission here, as one can get to ssh details though this end-point?

mvollmer: Yes, that's a good question. Are there other ways already to piece together the information returned by this method?  For example, the password might also be available via /api/v2/hosts/:id as a parameter.

```
$ curl -s -b _session_id=9dda32ff969c0feed7c9ccb29ef62382 http://localhost:3000/api/v2/hosts/dev.mvo.lan | jq .parameters
[
  {
    "priority": 70,
    "created_at": "2018-11-29 10:10:37 UTC",
    "updated_at": "2018-11-29 10:10:37 UTC",
    "id": 1,
    "name": "remote_execution_ssh_password",
    "value": "foobar"
  }
]
```

mvollmer:
> Just for my own education, can you explain why 

I tinkered a bit, and if I put some prints before the problematic line like so:
```ruby
      print("1: #{Api::V2::HostsController}\n")
      print("2: #{Api::V2::HostsController}\n")
      Api::V2::HostsController.send(:prepend, Concerns::ForemanRemoteExecution::Api::V2::HostsControllerExtensions)
```
I do in fact get a name error for the _second_ print.  The first `print` outputs `1: HostsController` and the second raises a NameError. I thought I was getting used to Ruby magic by now, but this feels like a little bit too much... :-)

iNecas:
This is more a Rails autoloading magic than Ruby magic: No idea why it's happening, but probably not worth digging too much into it either, as long as we have a working solution

mvollmer:
Here is a tweak related to the proxy selection: https://github.com/mvollmer/foreman_remote_execution/commit/b96804817d1bccbf53b101c2cc1dadb92360fb7e

With this, the client can distinguish these cases:

- proxy == :not_available
  Error "A proxy is required to reach #{host} but all of them are down".
- proxy == :not_defined
  Error "A proxy is required to reach #{host} but none has been configured"
- proxy == :direct
  Use Net::SSH directly
- else
  Use /remote_execution/ssh_session on proxy

mvollmer:
@iNecas, I want to make sure that we do the right thing with ssh "known hosts", and I would need some help.

The script_runner.rb code looks for `:host_public_key` in the options, but I can't find the code that would set it.  I would expect `proxy_command_options` in ssh_execution_provider.rb to do that, but it doesn't.

Also, script_runner.rb refers to `@known_hosts` but I can't find the definition for that.

mvollmer:
@iNecas, is it correct to include `:ssh_key_file` in ssh_params?  This is a setting that can be different on each proxy, no?  Where does Foreman get the value from when doing remote execution without a proxy?

iNecas:
@mvollmer you're right. Foreman doesn't need this value, as the proxy part takes it from its settings.

mvollmer:
> Foreman doesn't need [`:ssh_identity_key_file`], as the proxy part takes it from its settings.

But it is possible to configure REX so that Foreman will attempt a remote execution directly, with the `remote_execution_without_proxy` setting, right?

So what about leaving `:ssh_key_file` in this API but only for `proxy == :direct` and we make sure that the proxy always uses its own value in its code for `/remote_execution/ssh_session`.

mvollmer:
> Also, script_runner.rb refers to @known_hosts but I can't find the definition for that.

Okay, it's not an error to refer to non existing instance variables, the result is just `nil`.  (See how much I have to learn? :-/)

So now I think `@known_hosts_file` is in fact never set to anything, and the line referencing it in script_runner.rb can just be removed.  Given that this line is far away from the second line that sets `ssh_options[:user_known_hosts_file]`, it might well be a forgotten left over.  The git history unfortunately doesn't go far back enough for me to figure this out.  I'll make a PR to remove it.

mvollmer:
> I'll make a PR to remove [@known_hosts_file].

#398

mvollmer:
@iNecas, I opened https://github.com/iNecas/foreman_remote_execution/pull/6 against your branch, so now we have a PR for a PR. :-) I hope that's not too confusing...

iNecas:
Btw. we're thinking about disabling remote_execution_without_proxy option from here, as it just complicates the flow. Would removing this help you a bit with not having to count on this case?

mvollmer:
> Btw. we're thinking about disabling remote_execution_without_proxy option from here, as it just complicates the flow. Would removing this help you a bit with not having to count on this case?

Yes, it would help from a simplicity point of view.

However, transport performance is more of an issue for Cockpit than it is for batch execution.  But of course, adding another hop via localhost should barely be noticable.

iNecas:
After more discussions, we're making this endpoint cockpit-specific - I've moved the commits to https://github.com/theforeman/foreman_remote_execution/pull/408/files, closing this PR in its favor.

PR #406, opened 2019-01-11
Refs #24299 - migration to remove proxy-less setting

iNecas:
We missed this change in the original PR.


db/migrate/20190111153330_remove_remote_execution_without_proxy_setting.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveRemoteExecutionWithoutProxySetting < ActiveRecord::Migration[5.2]
+  def up
+    Setting.where(name: 'remote_execution_without_proxy').delete_all
+  end
+
+  def down
+    Setting.create!(:name => 'remote_execution_without_proxy',
+                    :description => N_('When enabled, the remote execution will try to run \
```
adamruzicka: I don't think the slash should go inside the string

iNecas:
Tests look good

adamruzicka:
Thanks @iNecas !

PR #408, opened 2019-02-21
Fixes #25534 - Seamless cockpit integration

mvollmer:
This adds a "Web Console" button next to the "Schedule Remote Job"
button in the title row of a host page.
    
The button links to a specially configured Cockpit instance.  It's
location is configured with the new remote_execution_cockpit_href
setting.
    
The Cockpit instance will use the new /cockpit_redirect path to get a
token via OAuth style redirection.

### DIFF
--- a/app/controllers/api/v2/remote_execution_features_controller.rb
+++ b/app/controllers/api/v2/remote_execution_features_controller.rb
@@ -28,6 +28,23 @@ module Api
         process_response @remote_execution_feature.update_attributes(remote_execution_feature_params)
       end
 
+      def cockpit_url_for_host(host)
+        tmpl = Setting[:remote_execution_cockpit_href]
+        tmpl && tmpl != "" ? tmpl % { :host => host } : nil
+      end
+
+      def cockpit_redirect
+        return head(:forbidden) unless params[:redirect_uri]
+        redir_url = URI.parse(params[:redirect_uri])
+        cockpit_url = cockpit_url_for_host("")
+        redir_url.query = if redir_url.hostname == URI.join(Setting[:foreman_url], cockpit_url).hostname
+                            "access_token=#{request.session_options[:id]}"
+                          else
+                            "error_description=Sorry"
+                          end
+        redirect_to(redir_url.to_s)
+      end
+
       private
 
       def parent_scope
--- a/app/controllers/api/v2/remote_execution_features_controller.rb
+++ b/app/controllers/api/v2/remote_execution_features_controller.rb
@@ -28,23 +28,6 @@ module Api
         process_response @remote_execution_feature.update_attributes(remote_execution_feature_params)
       end
 
-      def cockpit_url_for_host(host)
-        tmpl = Setting[:remote_execution_cockpit_href]
-        tmpl && tmpl != "" ? tmpl % { :host => host } : nil
-      end
-
-      def cockpit_redirect
-        return head(:forbidden) unless params[:redirect_uri]
-        redir_url = URI.parse(params[:redirect_uri])
-        cockpit_url = cockpit_url_for_host("")
-        redir_url.query = if redir_url.hostname == URI.join(Setting[:foreman_url], cockpit_url).hostname
-                            "access_token=#{request.session_options[:id]}"
-                          else
-                            "error_description=Sorry"
-                          end
-        redirect_to(redir_url.to_s)
-      end
-
       private
 
       def parent_scope
--- a/app/controllers/cockpit_controller.rb
+++ b/app/controllers/cockpit_controller.rb
@@ -1,0 +1,34 @@
+class CockpitController < ApplicationController
+  before_action :find_resource, :only => [:host_ssh_params]
+
+  def host_ssh_params
+    render :json => SSHExecutionProvider.ssh_params(@host)
+  end
+
+  def redirect
+    return invalid_request unless params[:redirect_uri]
+    redir_url = URI.parse(params[:redirect_uri])
+
+    cockpit_url = SSHExecutionProvider.cockpit_url_for_host('')
+    redir_url.query = if redir_url.hostname == URI.join(Setting[:foreman_url], cockpit_url).hostname
+                        "access_token=#{request.session_options[:id]}"
+                      else
+                        "error_description=Sorry"
+                      end
+    redirect_to(redir_url.to_s)
+  end
+
+  private
+
+  def resource_name
+    "host"
+  end
+
+  def controller_permission
+    :hosts
+  end
+
+  def action_permission
+    :cockpit
+  end
+end
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -20,13 +20,9 @@ module ForemanRemoteExecution
       link_to(_('Schedule Remote Job'), new_job_invocation_path(:host_ids => [args.first.id]), :id => :run_button, :class => 'btn btn-default')
     end
 
-    def cockpit_url_for_host(host)
-      tmpl = Setting[:remote_execution_cockpit_href]
-      tmpl && tmpl != "" ? tmpl % { :host => host } : nil
-    end
-
-    def web_console_button(*args)
-      url = cockpit_url_for_host(args.first.name)
+    def web_console_button(host, *args)
+      return unless authorized_for(permission: 'cockpit_hosts', auth_object: host)
+      url = SSHExecutionProvider.cockpit_url_for_host(host.name)
       url ? link_to(_('Web Console'), url, :class => 'btn btn-default') : nil
     end
 
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -20,9 +20,21 @@ module ForemanRemoteExecution
       link_to(_('Schedule Remote Job'), new_job_invocation_path(:host_ids => [args.first.id]), :id => :run_button, :class => 'btn btn-default')
     end
 
+    def cockpit_url_for_host(host)
+      tmpl = Setting[:remote_execution_cockpit_href]
+      tmpl && tmpl != "" ? tmpl % { :host => host } : nil
+    end
+
+    def web_console_button(*args)
+      url = cockpit_url_for_host(args.first.name)
+      url ? link_to(_('Web Console'), url, :class => 'btn btn-default') : nil
+    end
+
     def host_title_actions(*args)
-      title_actions(button_group(schedule_job_multi_button(*args)))
+      title_actions(button_group(schedule_job_multi_button(*args)),
+                    button_group(web_console_button(*args)))
       super(*args)
     end
   end
+
 end
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -1,6 +1,6 @@
 class Setting::RemoteExecution < Setting
 
-  ::Setting::BLANK_ATTRS.concat %w{remote_execution_ssh_password remote_execution_ssh_key_passphrase remote_execution_sudo_password remote_execution_cockpit_href}
+  ::Setting::BLANK_ATTRS.concat %w{remote_execution_ssh_password remote_execution_ssh_key_passphrase remote_execution_sudo_password remote_execution_cockpit_url}
 
   # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
   def self.load_defaults
@@ -68,7 +68,7 @@ class Setting::RemoteExecution < Setting
                  N_('When enabled, working directories will be removed after task completion. You may override this per host by setting a parameter called remote_execution_cleanup_working_dirs.'),
                  true,
                  N_('Cleanup working directories')),
-        self.set('remote_execution_cockpit_href',
+        self.set('remote_execution_cockpit_url',
                  N_('Where to find the Cockpit instance for the Web Console button.  By default, no button is shown.'),
                  nil,
                  N_('Cockpit URL'),
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -1,6 +1,6 @@
 class Setting::RemoteExecution < Setting
 
-  ::Setting::BLANK_ATTRS.concat %w{remote_execution_ssh_password remote_execution_ssh_key_passphrase remote_execution_sudo_password}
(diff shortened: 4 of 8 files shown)


app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -20,9 +20,16 @@ def schedule_job_button(*args)
       link_to(_('Schedule Remote Job'), new_job_invocation_path(:host_ids => [args.first.id]), :id => :run_button, :class => 'btn btn-default')
     end
 
+    def web_console_button(href, *args)
+      link_to(_('Web Console'), href + "/=" + args.first.name, :class => 'btn btn-default')
```
mvollmer: It would be better if href is a template such as "/webcon/=${host}", instead of just a prefix.

app/models/setting/remote_execution.rb
```diff
@@ -67,7 +67,12 @@ def self.load_defaults
         self.set('remote_execution_cleanup_working_dirs',
                  N_('When enabled, working directories will be removed after task completion. You may override this per host by setting a parameter called remote_execution_cleanup_working_dirs.'),
                  true,
-                 N_('Cleanup working directories'))
+                 N_('Cleanup working directories')),
+        self.set('remote_execution_cockpit_href',
+                 N_('Where to find the Cockpit instance for the Web Console button.  By default, no button is shown.'),
```
mvollmer: We might want to make it clear here that a relative URL is just fine, like the "/webcon" one I have been using with the reverse proxy.

config/routes.rb
```diff
@@ -80,6 +80,9 @@
       end
 
       resources :remote_execution_features, :only => [:show, :index, :update]
+
+      get 'cockpit_redirect', to: 'remote_execution_features#cockpit_redirect'
```
mvollmer: I could only make this work as `/api/cockpit_redirect`. not as `/cockpit_redirect` as we had before.  This is not really a API entry point, and it should, for example, behave differently when called with an expired session token.  As part of `/api`, the request will be rejected with 401, but as a regular web ui path, it would redirect to the login page.  The latter is more appropriate for this, I'd say.

mvollmer: When putting this outside the `/api` block, the error is:

```
The action 'cockpit_redirect' could not be found for RemoteExecutionFeaturesController
```

app/controllers/api/v2/remote_execution_features_controller.rb
```diff
@@ -28,6 +28,18 @@ def update
         process_response @remote_execution_feature.update_attributes(remote_execution_feature_params)
       end
 
+      def cockpit_redirect
```
mvollmer: This should probably be in some other class.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
mvollmer: Although this might complicate the check in cockpit_redirect.

app/controllers/api/v2/remote_execution_features_controller.rb
```diff
@@ -28,6 +28,23 @@ def update
         process_response @remote_execution_feature.update_attributes(remote_execution_feature_params)
       end
 
+      def cockpit_url_for_host(host)
+        tmpl = Setting[:remote_execution_cockpit_href]
+        tmpl && tmpl != "" ? tmpl % { :host => host } : nil
```
iNecas: can be simplified like this:

```
Setting[:remote_execution_cockpit_href] % { :host => host } if Setting[:remote_execution_cockpit_href].present?
```

iNecas: Btw. do we need to set the host?

app/controllers/api/v2/remote_execution_features_controller.rb
```diff
@@ -28,6 +28,23 @@ def update
...
+      end
+
+      def cockpit_redirect
+        return head(:forbidden) unless params[:redirect_uri]
+        redir_url = URI.parse(params[:redirect_uri])
+        cockpit_url = cockpit_url_for_host("")
+        redir_url.query = if redir_url.hostname == URI.join(Setting[:foreman_url], cockpit_url).hostname
```
iNecas: we should fail more nicely, when cockpit_url returns nil, probably mentioning to the user they should configure the cockpit url. Currently, one gets ugly error with:

```
bad argument (expected URI object or URI string)
```

app/controllers/api/v2/remote_execution_features_controller.rb
iNecas: I'm preparing a patch to move it to standalone `CockpitController` so that we have all the pieces at one place. Will send it to you once ready.

mvollmer: Please just push to my branch.

app/controllers/api/v2/remote_execution_features_controller.rb
iNecas: Ok, I see we need it :)

app/models/setting/remote_execution.rb
iNecas: would be good to put an example of expected format, so that the user knows that he needs to put '%{host}' in there

config/routes.rb
```diff
@@ -38,6 +38,10 @@
     end
   end
 
+  constraints(:id => %r{[^/]+}) do
+    get 'cockpit/host_ssh_params/:id', to: 'cockpit#host_ssh_params'
```
iNecas: @mvollmer here is the updated endpoint for the host_ssh_params: no long nested behind `api/v2`

config/routes.rb
```diff
@@ -41,6 +41,7 @@
   constraints(:id => %r{[^/]+}) do
     get 'cockpit/host_ssh_params/:id', to: 'cockpit#host_ssh_params'
   end
+  get 'cockpit/redirect', to: 'cockpit#redirect'
```
iNecas: @mvollmer updating the redirect endpoint: you might need to update it in the session script

app/controllers/cockpit_controller.rb
```diff
@@ -5,15 +5,11 @@ def host_ssh_params
     render :json => SSHExecutionProvider.ssh_params(@host)
   end
 
-  def cockpit_url_for_host(host)
```
iNecas: Since we're using this at two places, moved the code the `SSHExecutionProvider`, as I could not think of better place, and adding separate class seemed like an overkill

app/controllers/cockpit_controller.rb
```diff
@@ -5,15 +5,11 @@ def host_ssh_params
...
-    tmpl = Setting[:remote_execution_cockpit_href]
-    tmpl && tmpl != "" ? tmpl % { :host => host } : nil
-  end
-
   def redirect
-    return head(:forbidden) unless params[:redirect_uri]
+    return invalid_request unless params[:redirect_uri]
```
iNecas: changed the error code, as forbidden would more indicate permission issue, while it seems we are validating the params here.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -20,14 +20,10 @@ def schedule_job_button(*args)
...
-    def cockpit_url_for_host(host)
-      tmpl = Setting[:remote_execution_cockpit_href]
-      tmpl && tmpl != "" ? tmpl % { :host => host } : nil
-    end
-
-    def web_console_button(*args)
-      url = cockpit_url_for_host(args.first.name)
-      url ? link_to(_('Web Console'), url, :class => 'btn btn-default') : nil
+    def web_console_button(host, *args)
+      return unless authorized_for(permission: 'cockpit_hosts', auth_object: host)
```
iNecas: Added check for permissions as well

app/models/setting/remote_execution.rb
```diff
@@ -68,7 +68,7 @@ def self.load_defaults
                  N_('When enabled, working directories will be removed after task completion. You may override this per host by setting a parameter called remote_execution_cleanup_working_dirs.'),
                  true,
                  N_('Cleanup working directories')),
-        self.set('remote_execution_cockpit_href',
+        self.set('remote_execution_cockpit_url',
```
iNecas: I've changed `href` to `url`, as it seemed more appropriate to me + matches with other settings, such as `provision_url` or `foreman_url`

iNecas:
@mvollmer I've pushed all the code I've had here. To update your setup, you would need to;

- set `remote_execution_cockpit_url` instead of `remote_execution_cockpit_href` setting
- use `cockpit/redirect` instead of `cockpit_redirect` as the redirect url
- use `cockpit/host_ssh_params/:host` instead of `api/v2/hosts/:host/ssh_params` for getting the ssh params

config/routes.rb
iNecas: I've update this to `cockpit/redirect` and works for me

app/controllers/api/v2/remote_execution_features_controller.rb
iNecas: Done

app/controllers/api/v2/remote_execution_features_controller.rb
iNecas: Updated in my commits

iNecas:

iNecas:
add to PR whitelist

app/models/setting/remote_execution.rb
mvollmer: Yes, I agree.  Initially, I thought that a relative URL like `/webcon/foo` is not in fact a URL, but it is.

app/models/ssh_execution_provider.rb
```diff
@@ -25,6 +25,24 @@ def ssh_key_passphrase(host)
...
+      proxy_selector = ::RemoteExecutionProxySelector.new
+      proxy = proxy_selector.determine_proxy(host, 'SSH')
+      {
+        :hostname => find_ip_or_hostname(host),
+        :proxy => proxy.class == Symbol ? proxy : proxy.url,
+        :ssh_user => ssh_user(host),
+        :ssh_port => ssh_port(host),
+        :ssh_password => ssh_password(host),
+        :ssh_key_passphrase => ssh_key_passphrase(host),
+        :ssh_key_file => File.expand_path(ForemanRemoteExecutionCore.settings.fetch(:ssh_identity_key_file))
```
mvollmer: `:ssh_key_file` should be dropped, we never need it with a SmartProxy.

app/models/ssh_execution_provider.rb
```diff
@@ -25,6 +25,24 @@ def ssh_key_passphrase(host)
...
+      }
+    end
+
+    def cockpit_url_for_host(host)
+      Setting[:remote_execution_cockpit_url] % { :host => host } if Setting[:remote_execution_cockpit_url].present?
```
mvollmer: This might raise an exception and we should catch that, I'd say.

app/models/ssh_execution_provider.rb
adamruzicka: I'm pretty sure `ForemanRemoteExecutionCore` is not available in this context

app/models/ssh_execution_provider.rb
iNecas: what exception we are talking about?

app/models/ssh_execution_provider.rb
iNecas: Dropped

app/models/ssh_execution_provider.rb
```diff
@@ -51,5 +51,15 @@ def ssh_user(host)
...
       Integer(host_setting(host, :remote_execution_ssh_port))
     end
+
+    def ssh_proxy_params(proxy)
+      return proxy if proxy.class == Symbol
+
+      {
+        :url => proxy.url,
+        :ssl_certificate => File.open(Setting[:ssl_certificate]).read,
+        :ssl_priv_key => File.open(Setting[:ssl_priv_key]).read
```
adamruzicka: The opened `File`s need to be closed. After a while the proxy would crash because of reaching the opened file limit. Something like `File.open(Setting[:ssl_priv_key], &:read)` might do the trick

iNecas: Or even simpler `File.read`

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
mvollmer: This is done.

app/models/setting/remote_execution.rb
mvollmer: Do want to keep using %{host}, or should we rather switch to the templating engine that Foreman uses in other places?

app/models/ssh_execution_provider.rb
mvollmer: When `Setting[...]` is malformed for the `%` operator, such as with a typo like `"%{hos}"`

app/models/ssh_execution_provider.rb
mvollmer: Oh, does the Ruby GC not close files?

mvollmer: We should catch exceptions here, and probably only add this information when the proxy uses https.

app/models/setting/remote_execution.rb
iNecas: I think it would be unnecessary complex,'%{host}' is good enough IMO

app/models/ssh_execution_provider.rb
iNecas: Actually, we should perhaps not return this at all in params, but rather configure the session script to read the right files form the filesystem: returning this via http might have potencial recurity risks that we should avoid

app/models/ssh_execution_provider.rb
```diff
@@ -51,15 +51,5 @@ def ssh_user(host)
     def ssh_port(host)
       Integer(host_setting(host, :remote_execution_ssh_port))
     end
-
-    def ssh_proxy_params(proxy)
-      return proxy if proxy.class == Symbol
-
-      {
-        :url => proxy.url,
```
iNecas: I guess we still need the url, don't we?

iNecas:
ok to test

iNecas:
@mvollmer I've rebased this PR, resolved conflicts and let the CI to run: this should be really close to merging (probably after the proxy one is in). Anything else you want to add?

iNecas:
Squashed and rebased

iNecas:
Test failures unrelated. The dependency on proxy side merged. Merging this as well.

PR #409, opened 2019-02-22
Add foreman-cockpit-session tool

mvollmer:


tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+
+# Talking to the Foreman API
+
+def get_token_from_auth_data(auth_data)
+  auth_data.split(" ")[1]
+end
+
+def api_call(path, token)
+  dbg("=> #{path}")
+  uri = URI('http://localhost:3000/api/v2/' + path)
```
mvollmer: This needs to be configurable.

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+def read_size(io)
+  size = 0
+  seen = 0
+
+  while seen < 8
+    t = io.readpartial(1)
+
+    unless t
+      return 0
+    end
```
adamruzicka: `return 0 unless t` would be more ruby-like. This pattern repeats several times in this file

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+
+  if seen == 8
+    raise ArgumentError("Invalid frame: size too long")
+  end
+
+  size
+end
+
+def read_control()
+  io = IO.for_fd(1)
```
adamruzicka: Do we really want a process to read from its own standard output?

In any case there are constants `STDOUT`, `STDIN` and `STDERR` for this

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+  size = read_size(io)
+
+  data = ""
+  while size > 0
+    d = io.readpartial(1)  # XXX - can't we just do readpartial(size)?
```
adamruzicka: Not sure why we can't just `io.read(size)`. We know that a message with that size is coming. And even if it doesn't, we can't really do anything with an incomplete message

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+  end
+
+  dbg("<- #{data.strip}")
+  JSON.parse(data)
+end
+
+# Specific control messages
+
+def send_auth_challenge(challenge)
+    send_control({
```
adamruzicka: In ruby we usually use indentation of two spaces

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+end
+
+def send_auth_response(response)
+    send_control({
+        "command" => "authorize",
+        "response" => response
+    })
+end
+
+def read_auth_reply()
```
adamruzicka: Nitpick: the empty parens `()` are not really needed in ruby

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+end
+
+def api_call(path, token)
+  dbg("=> #{path}")
+  uri = URI('http://localhost:3000/api/v2/' + path)
+  req = Net::HTTP::Get.new(uri)
+  req["Cookie"] = "_session_id=#{token}"
+  res = Net::HTTP.start(uri.hostname, uri.port) { |http| http.request(req) }
+  dbg("<= #{res.code} #{res.body[0,500]}")
+  if res.code == "200"
```
adamruzicka: I would favor a case statement over several `elsif`s
```ruby
case res.code
when '200'
  return JSON.parse(res.body)
...
else
  # Something
end
```

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+send_auth_challenge("*")
+token = get_token_from_auth_data(read_auth_reply())
+params = api_call("hosts/#{host}/ssh_params", token)
+
+if !params
+  exit_with_problem("access-denied", "Host #{host} is not known", nil)
+end
+
+params["command"] = "cockpit-bridge"
+if params["proxy"] == "not_available"
```
adamruzicka: Again a nice candidate for `case`

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+  dbg("-> #{text}")
+  $stdout.write("#{text.length+1}\n\n#{text}")
+  $stdout.flush
+end
+
+def read_size(io)
+  size = 0
+  seen = 0
+
+  while seen < 8
```
adamruzicka: Is there a reason for reading it byte by byte? If not it could probably be done like this

```ruby
def read_size(io)
  data = io.readline.chomp
  raise ArgumentError, 'Invalid frame: size too long' if data.length > 8
  size = data.to_i
  raise ArgumentError, 'Invalid frame: invalid size' if size == 0 && data != '0'
  size
rescue EOFError
  0
end
```

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+    unless t
+      return 0
+    end
+
+    if t == "\n"
+      break
+    end
+
+    if t.to_i.zero? && t != "0"
+      raise ArgumentError("Invalid frame: invalid size")
```
adamruzicka: The syntax for raising exceptions is slightly different, it should be `raise ArgumentError, 'Invalid frame: invalid size'`

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+    end
+
+    size = (size * 10) + t.to_i
+    seen += 1
+  end
+
+  if seen == 8
+    raise ArgumentError("Invalid frame: size too long")
```
adamruzicka: ditto

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+        "command" => "authorize",
+        "response" => response
+    })
+end
+
+def read_auth_reply()
+    cmd = read_control()
+    response = cmd["response"]
+    if cmd["command"] != "authorize" || !response
+      raise ArgumentError("Did not receive a valid authorize command")
```
adamruzicka: ditto

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+  inp_buf = "POST /ssh/session HTTP/1.1\r\nConnection: upgrade\r\nUpgrade: raw\r\nContent-Length: #{data.length}\r\n\r\n#{data}"
+  out_buf = ""
+
+  ws_eof = false
+  bridge_eof = false
+  expecting_http_header = true
+
+  status = ""
+
+  while true
```
adamruzicka: `loop do ... end` is more idiomatic way of expressing `while true`

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+    return nil
+  else
+    $stderr.write("Error talking to foreman: #{res.body}\n")
+    exit 1
+  end
+end
+
+# SSH via the smart proxy
+
+def ssh_with_proxy(urlstr, params)
```
adamruzicka: I'd suggest breaking this function into several smaller one so they're easier to read and understand

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,248 @@
...
+
+# Main
+
+host = ARGV[0]
+
+send_auth_challenge("*")
+token = get_token_from_auth_data(read_auth_reply())
+params = api_call("hosts/#{host}/ssh_params", token)
+
+if !params
```
adamruzicka: `unless $CONDITION` is the more idiomatic way of expressing `if not $CONDITION`

iNecas:
@mvollmer  When testing with proxy-puma, I'm getting `Puma::HttpParserError: Invalid HTTP format, parsing fails`. Not sure where it's coming from exactly, but was wondering, if that's something you've seen as well + any idea on how to fix it in my environment?

iNecas:
Np, seems like the issue was with old session script I was using before: seems like some difference in tolerance between webrick and puma

tools/foreman-cockpit-session
mvollmer: I'll add a simple envionment variable for this.

mvollmer:
> My main question is, we're probably not the only ruby project that might ever want to interface with cockpit. Wouldn't it be better to move the protocol encoding/decoding into a standalone gem?

I am aware of one other: https://github.com/ManageIQ/manageiq/blob/master/tools/cockpit/cockpit-auth-miq. In fact, I used that program as a starting point.

So, yes, I think it would make sense to share code between the two.  Good point.

On the other hand, the protocol is stable and it's not a lot of code, so I wouldn't see it as problematic to not share code immediately.

tools/foreman-cockpit-session
mvollmer: Ouch, using `1` looks like a bug to me, too. (I got that from the ManageIQ version of the script.)

tools/foreman-cockpit-session
mvollmer: Done.

tools/foreman-cockpit-session
mvollmer: I have changed a bunch of them,

tools/foreman-cockpit-session
mvollmer: Fixed by using `$stdin`.

tools/foreman-cockpit-session
mvollmer: Change do use `$stdin.read(size)` and it works just fine.

tools/foreman-cockpit-session
mvollmer: Fixed.

tools/foreman-cockpit-session
mvollmer: Fixed.

tools/foreman-cockpit-session
mvollmer: Fixed.

tools/foreman-cockpit-session
mvollmer: Yes, fixed.

tools/foreman-cockpit-session
mvollmer: Yes, fixed like you propose. We can treat a frame size of zero as an error for control frames, which further simplifies this.

tools/foreman-cockpit-session
mvollmer: Fixed, but not tested.

tools/foreman-cockpit-session
mvollmer: Fixed, but not tested.

tools/foreman-cockpit-session
mvollmer: Fixed, but not tested.

tools/foreman-cockpit-session
mvollmer: Fixed.

tools/foreman-cockpit-session
mvollmer: Done.

Now we send and read the headers in a blocking fashion, which should be okay as long as the proxy sends the response header as soon as it gets the request header.  This is currently only the case when then executed program starts writing to something to stdout before reading from stdin.  This is because the proxy currently takes the first byte on stdout as the indication that the program has been sucessfully started. This is the case for cockpit-bridge, which sends a "init" control message immediately, so we should be good in our specific case.

tools/foreman-cockpit-session
mvollmer: Fixed.

mvollmer:
@adamruzicka, thanks a lot for all the comments! I hope I have addressed them all.

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,254 @@
...
+
+def read_auth_reply
+  cmd = read_control()
+  response = cmd["response"]
+  raise ArgumentError, "Did not receive a valid authorize command" if cmd["command"] != "authorize" || !response
+  response
+end
+
+def exit_with_problem(problem, message, auth_methods)
+  send_control({ "command" => "init",
```
iNecas: We should also log the message out (I guess stderr is the way?)

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,254 @@
...
+  when "200"
+    return JSON.parse(res.body)
+  when "401"
+    exit_with_problem("authentication-failed",
+                      "Token was not valid",
+                      { "password" => "not-tried", "token" => "denied" })
+  when "404"
+    return nil
+  else
+    $stderr.write("Error talking to foreman: #{res.body}\n")
```
iNecas: It would be good to wrap the `$stderr` into `log` method, or something like that. Also, I would recommend `$stderr.puts("..."`)`. over `$stderr.write("...\n")`

iNecas: Also, the ideal would be to create `logger = Logger.new($stderr)` and allow setting log level via setting and add debug logging as much as possible (including all requests and responses) once something doesn't work, it's pretty hard to figure out what's going on.

tools/foreman-cockpit-session
```diff
@@ -0,0 +1,254 @@
...
+    exit 1
+  end
+end
+
+# SSH via the smart proxy
+
+def ssh_write_request_header(sock, params)
+  data = JSON.dump(params) + "\r\n"
+  sock.write("POST /ssh/session HTTP/1.1\r\nConnection: upgrade\r\nUpgrade: raw\r\nContent-Length: #{data.length}\r\n\r\n#{data}")
```
iNecas: In my setup with ssl, I hit `400` from proxy, as `\r\nHost: #{url.hostname}:#{url.port}` is missing, and at least some versions of sinatra require them to be there.

iNecas:
@mvollmer I did comple of updates in this PR based on my testing. Everything should be good to go from my side. Please have a look at the changes in case there is anything you would like to change

iNecas:
ok to test for jenkins

iNecas:
ok to test

iNecas:
Squashed and rebased

iNecas:
I don't see a reason to hold this any longer. Merging now

PR #411, opened 2019-04-26
Fixes #26689 - remove lookup_keys from views

xprazak2:


ares:
do we need to initialize this another way? this probably changed during moving template inputs to core right?

ohadlevy:
> do we need to initialize this another way? this probably changed during moving template inputs to core right?

the original issue was that lookup_keys were always complied into the [application.js](https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/application.js#L14), we simply removed it from being complied again in core. including here additionally was never needed.

iNecas:
Thanks @xprazak2

PR #412, opened 2019-04-26
Fixes #26690 - fix lookup_keys.js assets issue

iNecas:


iNecas:
My tests show this works both in devel and 1.22-stable

ares:
is this dup of https://github.com/theforeman/foreman_remote_execution/pull/411 ? (also the redmine issue)

PR #419, opened 2019-07-22
Fixes #27302 - Make query type option more visible

dgoetz:
This pull request is a follow up on https://projects.theforeman.org/issues/27302 and https://community.theforeman.org/t/changing-the-default-from-static-to-dynamic-query-for-rex/14649

It will look like this with the change:
image

### DIFF
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -103,7 +103,7 @@
       <%= number_f f, :time_span, :label => _('Time span'), :placeholder => 'N', :min => 1, :label_help => N_("Distribute execution over N seconds")  %>
     </div>
 
-    <div class="form-group advanced hidden">
+    <div class="form-group">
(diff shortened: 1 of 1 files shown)


dgoetz:
Error in sqlite3 seams unrelated to me:
```
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
Finished: FAILURE
```

ares:
should this be above the display advance link then? I guess it's now always visible...

dgoetz:
@ares: I thought about this as well and I would have preferred to move it below "Schedule" and perhaps make it only visible when selecting "Schedule future execution" or "Set up recurring execution", but this would require changes in the logic behind and I opted for the easy one. Moving it above "Display advanced fields"  I opted against because it logically more near to the "Schedule" which was already below it. Another option would be moving it next to the filter so below "Search Query", but then I am not sure if the "Resolves to" is misleading as it will probably change for dynamic.

ares:
ok, I played with it a bit and I agree this is probably the best we can do atm, given there's an ack from @adamruzicka I'm taking the liberty and merging this. Thanks @dgoetz!

PR #422, opened 2019-07-30
Pull in templates from community-templates

adamruzicka:


adamruzicka:
The tests are green-ish

app/views/templates/ssh/check_update.erb
```diff
@@ -0,0 +1,33 @@
...
+
+<%= command %>
+<% if command.start_with? 'yum' %>
+rc=$?
+# yum check-update returns 110 even on success
+if [ $rc -eq 110 ]; then
+    (exit 0)
+else
+    (exit $rc)
+fi
```
lzap: Yum on my system returns "100", man page says it means there are updates available. I can't find any info about 110, can you elaborate and link docs?

What the script is supposed to return? Can you explain here and also in the description field or at least in ERB comment for future edits?

app/views/templates/ssh/check_update.erb
```diff
@@ -0,0 +1,33 @@
...
+provider_type: SSH
+%>
+
+<%
+  supported_families = ['Redhat', 'Debian', 'Suse']
+  render_error(N_('Unsupported or no operating system found for this host.')) unless @host.operatingsystem && supported_families.include?(@host.operatingsystem.family)
+
+  command = case @host.operatingsystem.family
+            when 'Redhat'
+              'yum check-update'
```
lzap: I've verified on my Fedora and `dnf` behaves the same as `yum` for `check-update`. So no special detection for `dnf` required for now as yum alias is always installed for now. See my comment about return code tho.

app/views/templates/ssh/check_update.erb
adamruzicka: This must have been a typo.

Remote execution considers jobs which ended with exit code 0 as successful and all others as failed. In here we override the exit code to 0 in case yum check-update returns 100 so the job is marked as successful (because there was no error) in remote execution.

ares:
Template needs to go to community-templates and then are synced to this repo

ares:
@adamruzicka is this 1:1 to community-templates now? I didn't see the force push and idk, if there was some change in community-templates PR

adamruzicka:
@ares Now it is synced

PR #424, opened 2019-07-31
Fixes #27489 - remove global in eslint config

amirfefer:
i18n capabilities and jquery should be imported in react code, ie:

```js
import { translate as __ } from 'foremanReact/common/i18n`
import $ from 'jquery'
```


adamruzicka:
Thank you @amirfefer !

PR #425, opened 2019-08-01
Fixes #27481 - Lay groundwork for provider secrets

adamruzicka:


xprazak2:
Works and the code seems sensible to me, but my knowledge in this area is very limited.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -38,9 +38,11 @@ def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExe
...
         provider = template_invocation.template.provider
 
-        secrets = { :ssh_password => job_invocation.password || provider.ssh_password(host),
-                    :key_passphrase => job_invocation.key_passphrase || provider.ssh_key_passphrase(host),
-                    :sudo_password => job_invocation.sudo_password || provider.sudo_password(host) }
+        secrets = { :ssh_password => job_invocation.password,
+                    :key_passphrase => job_invocation.key_passphrase,
+                    :sudo_password => job_invocation.sudo_password }
+
+        secrets = secrets.merge(provider.secrets(host))
```
lzap: This won't work if I am not mistaken. Previously the behaviour was "A | B" (A or B) and now it's just B because A is merged with B effectively overwriting the keys. What you probably want to do is to iterate over the secrets or use `map` call to keep the same behaviour. From the merge docs:

_If no block is given, entries with duplicate keys are overwritten with the values from each other_hash successively, otherwise the value for each duplicate key is determined by calling the block with the key, its value in the receiver and its value in each other_hash._

If this is what you want to do please explain then. In that case you can probably drop the merge and simply use the provider secrets instead?

app/models/ssh_execution_provider.rb
```diff
@@ -29,6 +28,12 @@ def proxy_operation_name
       'ssh'
     end
 
+    def secrets(host)
+      { :ssh_password => ssh_password(host),
+        :key_passphrase => ssh_key_passphrase(host),
+        :sudo_password => sudo_password(host) }
+    end
```
lzap: Nitpick: Odd formatting.

adamruzicka: Should be better now

app/lib/actions/remote_execution/run_host_job.rb
adamruzicka: An oversight on my side, should be fixed now.

app/models/ssh_execution_provider.rb
lzap: I see you prefer the original formatting, feel free to keep it.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -59,6 +55,14 @@ def finalize(*args)
         check_exit_status
       end
 
+      def secrets(host, job_invocation, provider)
+        job_secrets = { :ssh_password => job_invocation.password,
+                        :key_passphrase => job_invocation.key_passphrase,
+                        :sudo_password => job_invocation.sudo_password }
+
+        job_secrets.merge(provider.secrets(host)) { |_key, job, provider| job || provider }
```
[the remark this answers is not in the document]
ares: I'm sorry, this will need to be fixed

adamruzicka:
Test failures seem unrelated

lzap:
I merged https://github.com/theforeman/foreman_ansible/pull/295 however I quickly realized that I don't have ReX perms yet :man_facepalming: Need to wait a bit more: https://community.theforeman.org/t/nominating-lukas-zapletal-as-a-maintainer-for-remote-execution-repos/15539

ares:
No need to wait then :-) Thanks @adamruzicka and @lzap, merging now!

PR #426, opened 2019-08-19
Fixes #27566 - Fix blank page after job cancel

adamruzicka:


webpack/react_app/components/jobInvocations/index.js
```diff
@@ -40,18 +40,21 @@ const mapStateToProps = (state) => {
 
 JobInvocationContainer.propTypes = {
   startJobInvocationsPolling: PropTypes.func,
-  data: PropTypes.string,
-  jobInvocations: PropTypes.arrayOf(PropTypes.arrayOf(
```
adamruzicka: Apparently `arrayOf` is an unary function, so this was checking that all elements of the array are strings. The following arguments were ignored.

```
Warning: Failed prop type: Invalid prop `jobInvocations[0][1]` of type `number` supplied to `JobInvocationContainer`, expected `string`.
```

I'd say this is ok, given how the data is typechecked when passed to the donut chart https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/charts/DonutChart/index.js#L32

ares:
Thanks @adamruzicka, merging!

PR #429, opened 2019-09-01
Fixes #27746 - Remove taxonomy enabled checks

tbrisker:


ofedoren:
Just noticed, that after https://github.com/theforeman/foreman/pull/7012 this line https://github.com/theforeman/foreman_remote_execution/blob/773e23edb5db3c83ddd378381184973d38f5a120/app/models/concerns/foreman_remote_execution/host_extensions.rb#L74 should be also changed otherwise this causes error (no such method `enabled_taxonomies` for class..) when creating a host.

tbrisker:
Thanks for noticing @ofedoren ! I've removed that check as well as a couple of other references to taxonomies.

adamruzicka:
Thank you @tbrisker ! Also thanks @ofedoren for noticing the `enabled_taxonomies` thing

PR #430, opened 2019-09-04
Refs #27701 - Bump required Foreman version to 1.23

adamruzicka:


tbrisker:
This should also be 1.24

PR #435, opened 2019-09-24
Fixes #27930 - set job invocation buttons initially disabled

laviro:
While creating a new job,
the buttons are first enabled, and after a few seconds turn disabled..
would it make sense to initially disable them ?

after a few moments they will get updated with the correct state anyway,
it feels better this way.

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -86,12 +86,12 @@ module RemoteExecutionHelper
       buttons << button_to(_('Cancel Job'), cancel_job_invocation_path(job_invocation),
                            :class => 'btn btn-danger',
                            :title => _('Try to cancel the job'),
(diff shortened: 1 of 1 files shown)


app/helpers/remote_execution_helper.rb
```diff
@@ -86,12 +86,12 @@ def job_invocation_task_buttons(task)
       buttons << button_to(_('Cancel Job'), cancel_job_invocation_path(job_invocation),
                            :class => 'btn btn-danger',
                            :title => _('Try to cancel the job'),
-                           :disabled => !task.cancellable?,
+                           :disabled => true,
```
ezr-ondrej: Is this right fix? The button makes no sense if it's always disabled.

laviro:
Discussing it with QE & @adamruzicka about this offline,
it seem the right behavior.. closing, thanks everyone !

PR #436, opened 2019-10-01
Fixes #27976 - Add desc field to job templates

ofedoren:
Requires https://github.com/theforeman/foreman/pull/7051.


ares:
needs Foreman 1.24, so we should bump it here https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L37 and update https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/version.rb to 2.0.0 as part of this I think

ofedoren:
@ares, done. I've created another commit with bumping, so it's not a part of the fix.

adamruzicka:
Thank you @ofedorenko and congrats on releasing rex 2.0 :tada:

PR #438, opened 2019-10-16
Extract strings for translation

adamruzicka:


ares:
Thanks @adamruzicka, merging!

PR #439, opened 2019-10-16
Fixes #28072 - Expand the systemd unit

ekohl:
Note: this is untested.

### DIFF
--- a/extra/cockpit/foreman-cockpit.service
+++ b/extra/cockpit/foreman-cockpit.service
@@ -1,7 +1,14 @@
+[Unit]
+Description=Foreman Cockpit
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
 [Service]
 Environment=XDG_CONFIG_DIRS=/etc/foreman/
 Environment=FOREMAN_COCKPIT_SETTINGS=/etc/foreman/cockpit/foreman-cockpit-session.yml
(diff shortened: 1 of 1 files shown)


ekohl:
Since the original issue is already released I've created a new issue.

adamruzicka:
Thanks @ekohl !

extra/cockpit/foreman-cockpit.service
```diff
@@ -1,7 +1,14 @@
...
+Description=Foreman Cockpit
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
 [Service]
 Environment=XDG_CONFIG_DIRS=/etc/foreman/
 Environment=FOREMAN_COCKPIT_SETTINGS=/etc/foreman/cockpit/foreman-cockpit-session.yml
-ExecStart=/usr/libexec/cockpit-ws --no-tls --address 127.0.0.1 --port 9999
+Environment=ADDRESS=127.0.0.1
+Environment=PORT=9999
```
lzap: This should have been FOREMAN_COCKPIT_ADDRESS instead, this is way too generic for a global environment space I think.

lzap: Frankly we should probably move all variables from here into a separate file `/etc/foreman-cockpit.conf`.

lzap: https://github.com/theforeman/foreman_remote_execution/pull/440

PR #440, opened 2019-10-21
Refs #28072 - expand env and add SELinux wrapper

lzap:
This cleans up environmental variables and also adds SELinux wrapper. Untested!

### DIFF
--- a/extra/cockpit/foreman-cockpit.service
+++ b/extra/cockpit/foreman-cockpit.service
@@ -1,14 +1,14 @@
 [Unit]
-Description=Foreman Cockpit
+Description=Foreman authentication service for Cockpit
 Documentation=https://theforeman.org
 After=network.target remote-fs.target nss-lookup.target
 
 [Service]
 Environment=XDG_CONFIG_DIRS=/etc/foreman/
 Environment=FOREMAN_COCKPIT_SETTINGS=/etc/foreman/cockpit/foreman-cockpit-session.yml
-Environment=ADDRESS=127.0.0.1
-Environment=PORT=9999
-ExecStart=/usr/libexec/cockpit-ws --no-tls --address $ADDRESS --port $PORT
+Environment=FOREMAN_COCKPIT_ADDRESS=127.0.0.1
+Environment=FOREMAN_COCKPIT_PORT=9999
+ExecStart=/usr/libexec/foreman-cockpit --no-tls --address $FOREMAN_COCKPIT_ADDRESS --port $FOREMAN_COCKPIT_PORT
 User=foreman
 Group=foreman
 
--- a/extra/cockpit/foreman-cockpit
+++ b/extra/cockpit/foreman-cockpit
@@ -1,0 +1,3 @@
+#!/bin/sh
+# Wrapper script for SELinux transition
+exec /usr/libexec/cockpit-ws "$@"


extra/cockpit/foreman-cockpit.service
```diff
@@ -1,14 +1,12 @@
...
+Documentation=Websocket authentication service for Foreman-Cockpit integration
 After=network.target remote-fs.target nss-lookup.target
 
 [Service]
-Environment=XDG_CONFIG_DIRS=/etc/foreman/
-Environment=FOREMAN_COCKPIT_SETTINGS=/etc/foreman/cockpit/foreman-cockpit-session.yml
-Environment=ADDRESS=127.0.0.1
-Environment=PORT=9999
-ExecStart=/usr/libexec/cockpit-ws --no-tls --address $ADDRESS --port $PORT
+EnvironmentFile=/etc/foreman/foreman-cockpit.conf
```
lzap: Packaging change needed here - this file must be shipped otherwise the unit fails to start.

extra/cockpit/foreman-cockpit.service
```diff
@@ -1,14 +1,12 @@
...
+ExecStart=/usr/libexec/foreman-cockpit --no-tls --address $FOREMAN_COCKPIT_ADDRESS --port $FOREMAN_COCKPIT_PORT
```
lzap: Packaging change needed here - this file must be shipped otherwise the unit fails to start.

extra/cockpit/foreman-cockpit.conf
```diff
@@ -0,0 +1,11 @@
...
+XDG_CONFIG_DIRS=/etc/foreman/
+
+# Configuration file for foreman-session
+FOREMAN_COCKPIT_SETTINGS=/etc/foreman/cockpit/foreman-cockpit-session.yml
+
+# Address to listen on for cockpit-ws
+FOREMAN_COCKPIT_ADDRESS=127.0.0.1
+
+# Port to listen on for cockpit-ws
+FOREMAN_COCKPIT_PORT=9999
```
lzap: And this needs to be Puppet-managed at some point.

ekohl: I'd like to avoid this and prefer systemd drop in files. We have started to use this in the installer more and more. The main benefit is that it avoids differences between Debian and RH systems where one has env files in `/etc/default` and the other in `/etc/sysconfig`. It is also easy to see the env with `systemctl cat $service`.

lzap: Agreed, rebased.

extra/cockpit/foreman-cockpit.service
```diff
@@ -1,14 +1,14 @@
 [Unit]
 Description=Foreman Cockpit
-Documentation=https://theforeman.org
+Documentation=Websocket authentication service for Foreman-Cockpit integration
```
ekohl: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Documentation= states this is invalid.
> A space-separated list of URIs referencing documentation for this unit or its configuration. Accepted are only URIs of the types "http://", "https://", "file:", "info:", "man:".

lzap: I accidentally swapped this with Description actually. Fixing.

lzap:
Rebased, packaging counterpart at https://github.com/theforeman/foreman-packaging/pull/4207 I am currently testing this all together.

lzap:
@adamruzicka can you release 2.0.1 with this change? We need a new RPM with this wrapper for SELinux.

adamruzicka:
Thank you @lzap !

PR #441, opened 2019-10-24
Refs #28072 - remove SELinux wrapper

lzap:
I have found out that `cockpit-ws` is actually the cockpit service, I thoguht it's some kind of helper script that does not have a policy. It has a policy and a domain and a port. Therefore wrapper is not only unnecessary but it causes it to start in an incorrect domain.

Thus we need to remove the wrapper, also we need a different port than 9090 (used by Katello) or 9999 (which is already in use in SELinux as `jboss_management_port_t`). So I suggest 19090 which is not used either by IANA or SELinux.

We need this patch to get into 1.24 in order to get SELinux working, sorry guys.

### DIFF
--- a/extra/cockpit/foreman-cockpit
+++ b/extra/cockpit/foreman-cockpit
@@ -1,3 +1,0 @@
-#!/bin/sh
-# Wrapper script for SELinux transition
-exec /usr/libexec/cockpit-ws "$@"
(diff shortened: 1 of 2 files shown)


lzap:
@ares and @adamruzicka

adamruzicka:
Thanks @lzap !

PR #442, opened 2019-10-29
Fixes #13139 - Provide autocompletion for search query

adamruzicka:


ares:
should we use `autocomplete_f` helper? that would be automatically switched to react implementation once https://github.com/theforeman/foreman/pull/6980 is merged

adamruzicka:
Ideally yes, but it uses `auto_complete_search` helper which renders a `<input>` and I haven't found any sensible way of turning it into a text area

PR #443, opened 2019-11-04
Fixes #25016 - Choose default job template in form

xprazak2:


app/helpers/job_invocations_helper.rb
```diff
@@ -47,4 +47,20 @@ def collapsed_preview(target)
...
+  def resolve_job_template(default_template)
+    resolve_for_composer(default_template) { |form_template| form_template }
+  end
+
+  def resolve_for_composer(default_value, &block)
+    setting_value = Setting['remote_execution_form_job_template']
+    return default_value unless setting_value
+    form_template = JobTemplate.find_by :name => setting_value
+    return default_value unless form_template
+    yield form_template if block_given?
```
adamruzicka: Could it just return `form_template` if no block is given?

xprazak2:
Good catch. It should be fixed now.

xprazak2:
I looked deeper into the missing blank option - unfortunately core supports it only for selects with [groups](https://github.com/theforeman/foreman/blob/develop/app/models/setting.rb#L291)

xprazak2:
Updated, 'Run Command - SSH Default' is now default template for the setting.

adamruzicka:
Thanks @xprazak2 !

PR #445, opened 2019-11-12
Fixes #28247 - Show chart overall result

adamruzicka:
This reimplements functionality that was once provided by patternfly-react[1],
but was removed in commit 937d2d764a449e31276f1291fc528ff7ed625106. This commit
can be reverted when it gets re-added back there.

I tried opening https://github.com/patternfly/patternfly-react/pull/3299 against PF-react, we'll see how it goes there

[1] - https://github.com/patternfly/patternfly-react/


xprazak2:
Before:

chart-before

After:

chart-after

ares:
Thanks @adamruzicka for the patch and @xprazak2 for the review! Merging.

PR #448, opened 2019-11-19
Fixes #28305 - cleanup assets

ezr-ondrej:
Prepare assets for sprockets 4.0 - this is not necessary as we compile the assets properly now, but still better to compile one file only at least for css.


adamruzicka:
Thank you @ezr-ondrej !

PR #450, opened 2019-11-22
Fixes #28333 - don't expose audit_comment in template view

evgeni:


evgeni:
`SQLite3::SQLException: too many terms in compound SELECT` doesn't look related

tbrisker:
thanks @evgeni !

PR #451, opened 2019-11-25
Fixes #21449 - Drop host from known hosts on proxies

adamruzicka:


tux93:
I tested this by applying it on a foreman-1.22.1 server and it seems currently the keys are only removed from known_hosts upon deletion but kept when only re-building a host.

To completely solve #21449 I believe the keys need to also be removed when re-building a host.

adamruzicka:
@tux93 thank you for testing this. As far as I understood, with these changes the keys should be removed even on rebuild. Looks like I'll have to look into it a bit more

adamruzicka:
Tests are green-ish

tux93:
I tried again with the current state and keys are still only removed upon deletion, not rebuild, though that could also be due to the old foreman version on my test system

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
```diff
@@ -0,0 +1,65 @@
...
+                 end
+      "ssh_remove_known_hosts_#{kind}_#{id}"
+    end
+  end
+
+  module Orchestration::SSHManaged
+    extend ActiveSupport::Concern
+
+    included do
+      register_rebuild(:rebuild_ssh, N_("SSH_#{self.to_s.split('::').first}"))
```
adamruzicka: What is this good for? I wasn't able to trigger anything using just this, I had to trigger things from after_validation callback

tux93:
I got around to re-test today and it works now for rebuilding too!

ares:
oh actually not merging, tests failures are related `NameError: undefined local variable or method `queue' for #<Nic::Base:0x000000000beb5180>`

adamruzicka:

adamruzicka:
@tux93 could you please try this one more time?

tux93:
Re-tested with the latest version of the patch and it still works for me

adamruzicka:

adamruzicka:

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def queue_ssh_destroy
+      return unless should_drop_from_known_hosts?
+
+      ssh_destroy
+    end
+
+    def should_drop_from_known_hosts?
+      host, = host_kind_target
+      host && host.build && host.changes.key?('build')
```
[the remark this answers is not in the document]
ezr-ondrej: this is still on, even though I'm not sure I like the new style more :thinking: 
```suggestion
      host&.build && host&.changes&.key?('build')
```

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    def queue_ssh_destroy
+      return unless should_drop_from_known_hosts?
+
+      ssh_destroy
+    end
+
+    def should_drop_from_known_hosts?
+      host, = host_kind_target
```
ezr-ondrej, line 41:
```suggestion
      host, _, _ = host_kind_target
```
seems bit more readable as it states what we are dropping.

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def ssh_destroy
+      logger.debug "Scheduling SSH known_hosts cleanup"
+      queue.create(id: queue_id, name: _("Remove SSH known hosts for %s") % self,
+        priority: 200, action: [self, :drop_from_known_hosts])
+    end
+
+    def queue_ssh_destroy
+      return unless should_drop_from_known_hosts?
```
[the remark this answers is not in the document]
ezr-ondrej: shouldn't we disable this cop? We've decided to disable it in core.. (theforeman/foreman#7518) this is exactly the case where it doesn't deserve the blank line.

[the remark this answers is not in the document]
adamruzicka: I'm all in for this

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
adamruzicka: :police_officer: doesn't like this
```
C: Style/TrailingUnderscoreVariable: Do not use trailing _s in parallel assignment. Prefer host, = host_kind_target.
      host, _, _ = host_kind_target
```

adamruzicka:
We're finally :green_apple:

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
```diff
@@ -0,0 +1,58 @@
...
+        return false
+      end
+      true
+    end
+
+    def ssh_destroy
+      logger.debug "Scheduling SSH known_hosts cleanup"
+
+      host, _kind, target = host_kind_target
+      proxies = %w(SSH Ansible).map { |kind| host.remote_execution_proxies(kind).values }
```
ezr-ondrej, line 27: Why do we care about Ansible here? it just sounds wrong :thinking: 
We can't have Ansible without SSH anyway, or am I missing something?
```suggestion
      proxies = host.remote_execution_proxies('SSH').values
```

adamruzicka: Right, we shouldn't have proxies with ansible without ssh. I just wanted to explicitly mention anything that could put records into the know hosts file

ezr-ondrej: hmm understood, it just feels weird as REX should ideally know nada about Ansible, right?

adamruzicka: You're right, changed.

app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+      true
+    end
+
+    def ssh_destroy
+      logger.debug "Scheduling SSH known_hosts cleanup"
+
+      host, _kind, target = host_kind_target
+      proxies = host.remote_execution_proxies('SSH').values
+      proxies.flatten.uniq.each do |proxy|
```
ezr-ondrej, line 28: no need to flatten now, rith?
```suggestion
      host.remote_execution_proxies('SSH').values.each do |proxy|
```

adamruzicka: We cannot, `host.remote_execution_proxies('SSH')` returns a hash of `Symbol -> Array<SmartProxy>`, so at this point, we're operating on `Array<Array<SmartProxy>>` and we need to flatten (and possibly even `uniq`) it

ezr-ondrej: ah, got it, thaks for expelation :+1:

ezr-ondrej:
WWFJ
hurry-up

ezr-ondrej:
nooo! now we go again! xD I wanted to squash xD

adamruzicka:
I can always dig through reflog and un-squash it if you want it that bad :)

ezr-ondrej:
> I can always dig through reflog and un-squash it if you want it that bad :)

Nah xD It was just about Jenkins being almost done xD and now he goes again :clock1:

adamruzicka:
:green_apple:

PR #452, opened 2019-11-25
Fixes #28348 - Register assets for precompilation

adamruzicka:
Recent commit 2bf43ec changes the way how we register assets for precompilation.
Sadly, this change caused plugin:assets:precompile[$plugin] rake task to ignore
certain assets for remote execution by disabling automatic asset detection.
This commit keeps the automatic detection disabled, but registers a list of
assets which should be precompiled.

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -10,6 +10,16 @@ module ForemanRemoteExecution
     config.autoload_paths += Dir["#{config.root}/app/helpers/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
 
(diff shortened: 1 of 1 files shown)


adamruzicka:
@ezr-ondrej How does this look?

ares:
I guess it's twice because we want it working in packaging and in dev setup, right?

adamruzicka:
In packaging we use the `plugin:assets:precompile[$plugin]`, sadly it doesn't seem to process assets which are not registered in `Foreman::Plugin::Assets` registry and completely ignores `app.config.assets.precompile`

ares:
Thanks @adamruzicka, merging!

PR #453, opened 2019-11-27
fixes #15265 - making reporting of tasks

domitea:
needs tasks PR: theforeman/foreman-tasks#490
template PR: theforeman/foreman#8178

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -60,6 +60,12 @@ module RemoteExecutionHelper
     job_invocation = task.task_groups.find { |group| group.class == JobInvocationTaskGroup }.job_invocation
     task_authorizer = Authorizer.new(User.current, :collection => [task])
     buttons = []
+    if (template = job_report_template) && authorized_for(controller: :report_templates, action: :generate)
+      buttons << link_to(_('Create Report'), generate_report_template_path(template, job_report_template_parameters(job_invocation, template)),
+        class: 'btn btn-default',
+        title: _('Create report for this job'),
+        disabled: task.pending?)
+    end
     if authorized_for(hash_for_new_job_invocation_path)
       buttons << link_to(_('Rerun'), rerun_job_invocation_path(:id => job_invocation.id),
         :class => 'btn btn-default',
@@ -228,6 +234,27 @@ module RemoteExecutionHelper
     return if task.nil?
 
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
+  end
+
+  def job_report_template
+    template = ReportTemplate.where(name: Setting['remote_execution_job_invocation_report_template']).first
+
+    template if template.template_inputs.where(name: 'job_id').exists?
+  end
+
+  def job_report_template_parameters(job_invocation, template)
+    template_input = template.template_inputs.where(name: 'job_id').first
+    raise "#job_report_template_parameters need template that has 'job_id' input" unless template_input
+
+    {
+      report_template_report: {
+        input_values: {
+          "#{template_input.id}": {
+            value: job_invocation.id,
+          },
+        },
+      },
+    }
   end
 
   def targeting_hosts(job_invocation, hosts)
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -76,6 +76,10 @@ class JobInvocation < ApplicationRecord
 
   encrypts :password, :key_passphrase, :effective_user_password
 
+  class Jail < Safemode::Jail
+    allow :sub_task_for_host, :template_invocations_hosts
+  end
+
   def self.search_by_status(key, operator, value)
     conditions = HostStatus::ExecutionStatus::ExecutionTaskStatusMapper.sql_conditions_for(value)
     conditions[0] = "NOT (#{conditions[0]})" if operator == '<>'
--- a/app/models/setting/remote_execution.rb
+++ b/app/models/setting/remote_execution.rb
@@ -73,6 +73,16 @@ class Setting::RemoteExecution < Setting
         _('Form Job Template'),
         nil,
         { :collection => proc { Hash[JobTemplate.unscoped.map { |template| [template.name, template.name] }] } }),
+      self.set('remote_execution_job_invocation_report_template',
+        N_('Select a report template used for generating a report for a particular remote execution job'),
+        'Jobs - Invocation report template',
+        _('Job Invocation Report Template'),
+        nil,
+        { :collection => proc { self.job_invocation_report_templates_select } }),
     ]
+  end
+
+  def self.job_invocation_report_templates_select
+    Hash[ReportTemplate.unscoped.joins(:template_inputs).where(template_inputs: TemplateInput.where(name: 'job_id')).map { |template| [template.name, template.name] }]
   end
 end
--- a/app/services/renderer_methods.rb
+++ b/app/services/renderer_methods.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# macros to fetch information about invoked jobs
+module RendererMethods
+  extend ActiveSupport::Concern
+
+  def find_job_invocation_by_id(job_id, preload: nil)
+    JobInvocation.preload(preload).find_by(id: job_id)
+  rescue ActiveRecord::NotFound => _e
+    raise ::Foreman::Exception.new(N_("Can't find Job Invocation for an id %s"), job_id)
+  end
+end
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -155,6 +155,8 @@ module ForemanRemoteExecution
           ctx.permit :execution
         end
 
+        extend_template_helpers ForemanRemoteExecution::RendererMethods
+
         extend_rabl_template 'api/v2/smart_proxies/main', 'api/v2/smart_proxies/pubkey'
         extend_rabl_template 'api/v2/interfaces/main', 'api/v2/interfaces/execution_flag'
         extend_rabl_template 'api/v2/subnets/show', 'api/v2/subnets/remote_execution_proxies'
--- a/test/helpers/remote_execution_helper_test.rb
+++ b/test/helpers/remote_execution_helper_test.rb
@@ -23,4 +23,20 @@ class RemoteExecutionHelperTest < ActionView::TestCase
                                                         "Exit status: 0"])
     end
   end
+
+  describe 'test correct setting' do
+    it 'should found correct template from setting' do
+      Setting::RemoteExecution.load_defaults
+      template_name = 'Job Invocation Report Template'
+      setting_key = 'remote_execution_job_invocation_report_template'
+      template = FactoryBot.create(:report_template, name: template_name)
+      input = FactoryBot.create(:template_input, name: 'job_id', input_type: 'user')
+      template.template_inputs << input
+      Setting.expects(:[]).with(setting_key).returns(template_name)
+
+      found_template = job_report_template
+
+      assert_equal template.id, found_template.id
+    end
+  end
 end
--- a/test/unit/job_invocation_report_template_test.rb
+++ b/test/unit/job_invocation_report_template_test.rb
@@ -1,0 +1,57 @@
+require 'test_plugin_helper'
+
+class JobReportTemplateTest < ActiveSupport::TestCase
+  class FakeTask < OpenStruct
+    class Jail < Safemode::Jail
+      allow :action_continuous_output, :result, :ended_at
+    end
+  end
+
+  context 'with valid job invocation report template' do
+    let(:job_invocation_template) do
+      file_path = File.read(File.expand_path(Rails.root + "app/views/unattended/report_templates/jobs_-_invocation_report_template.erb"))
+      template = ReportTemplate.import_without_save("Job Invocation Report Template", file_path)
+      template.save!
+      template
+    end
+
+    describe 'template setting' do
+      it 'in settings includes only report templates with job_id input' do
+        FactoryBot.create(:report_template, name: 'Template 1')
+        job_invocation_template
+        templates = Setting::RemoteExecution.job_invocation_report_templates_select
+
+        assert_include templates, 'Job Invocation Report Template'
+      end
+    end
+
+    describe 'task reporting' do
+      let(:fake_outputs) do
+        [
+          { 'output_type' => 'stderr', 'output' => "error", 'timestamp' => Time.new(2020, 12, 1, 0, 0, 0).utc },
+          { 'output_type' => 'stdout', 'output' => "output", 'timestamp' => Time.new(2020, 12, 1, 0, 0, 0).utc },
+          { 'output_type' => 'stdebug', 'output' => "debug", 'timestamp' => Time.new(2020, 12, 1, 0, 0, 0).utc },
+        ]
+      end
+      let(:fake_task) { FakeTask.new(result: 'success', action_continuous_output: fake_outputs) }
+
+      it 'should render task outputs' do
+        job_invocation = FactoryBot.create(:job_invocation, :with_task)
+        JobInvocation.any_instance.expects(:sub_task_for_host).returns(fake_task)
+
+        input = job_invocation_template.template_inputs.first
+        composer_params = { template_id: job_invocation_template.id, input_values: { input.id.to_s => { value: job_invocation.id.to_s } } }
+        result = ReportComposer.new(composer_params).render
+
+        # parsing the CSV result
+        CSV.parse(result.strip, headers: true).each_with_index do |row, i|
+          row_hash = row.to_h
+          assert_equal 'success', row_hash['result']
+          assert_equal fake_outputs[i]['output_type'], row_hash['type']
+          assert_equal fake_outputs[i]['output'], row_hash['message']
+          assert_kind_of Time, Time.zone.parse(row_hash['time']), 'Parsing of time column failed'
+        end
+      end
+    end
+  end
+end
--- a/foreman_remote_execution.gemspec
+++ b/foreman_remote_execution.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'deface'
   s.add_dependency 'dynflow', '>= 1.0.2', '< 2.0.0'
   s.add_dependency 'foreman_remote_execution_core'
-  s.add_dependency 'foreman-tasks', '>= 4.0.0'
+  s.add_dependency 'foreman-tasks', '>= 4.1.0'
 
   s.add_development_dependency 'factory_bot_rails', '~> 4.8.0'
   s.add_development_dependency 'rdoc'


app/helpers/remote_execution_helper.rb
```diff
@@ -236,4 +241,21 @@ def load_template_from_task(template_invocation, target)
     return if task.nil?
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
   end
+
+  def jobs_template_with_input(job_invocation)
+    template = ReportTemplate.where(name: 'Job Invocation report').first
+    input_id = template.template_inputs.where(name: 'job_id').first.id
```
xprazak2: Template might have been renamed or deleted and `template_inputs` is not defined for nil

xprazak2:
Together with [template](https://github.com/theforeman/community-templates/pull/666) I get:

report-csv

xprazak2:

app/helpers/remote_execution_helper.rb
ares: Perhaps this should be configurable through settings, like ansible inventory template. That prevents hardcoding the name. But yeah, we need to be more defensive here.

app/services/foreman_remote_execution/renderer_methods.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# macros to fetch information about invoked jobs
+module ForemanRemoteExecution
+  module RendererMethods
+    extend ActiveSupport::Concern
+
+    def load_hosts_invoked_job(job_id)
```
ares: load_* macros typically allows searching, preloading etc and returns array. How about creating new convention on for macros searching specific instance, we'd use find_* or fetch_*? Also given we don't pass any host argument, we can comit hosts from the name I think.

ares: Now when I think about it more, it should probably return job invocation object and not the hash. The logic loading output should be in separate methods that would be allowed in safe mode. That way, user can access more things in the report when we allow them. Lets talk about this more, perhaps we could implement load_job_invocations macro at the end :-)

app/helpers/remote_execution_helper.rb
```diff
@@ -66,6 +66,11 @@ def job_invocation_task_buttons(task)
     job_invocation = task.task_groups.find { |group| group.class == JobInvocationTaskGroup }.job_invocation
     task_authorizer = Authorizer.new(User.current, :collection => [task])
     buttons = []
+    template, template_parameters = jobs_template_with_input(job_invocation)
+    buttons << link_to(_('Create Report'), generate_report_template_path(template, template_parameters),
```
ares: This is missing authorization check, there are link_to alternatives that do it

ares: also it may be good to extract to helper, it's 5 lines now for a single button

domitea:
Several updates I've made. After discuss with @ares I've made job invocation call more abstract. So with this PR the user has available macros for load job invocations. Also, I had to add jails to several models [JobInvocation and Task (see https://github.com/theforeman/foreman-tasks/pull/490)]. The example template is also updated (https://github.com/theforeman/community-templates/pull/666). Please look at that (@ares and @xprazak2 ) and I hope I will not fall to hell because I didn't made a squash because I going to do them after everything will be ready.

app/services/foreman_remote_execution/renderer_methods.rb
```diff
@@ -0,0 +1,23 @@
...
+    def load_job_invocations(search: '', includes: nil, preload: nil)
+      load_resource(klass: JobInvocation, search: search, permission: 'view_job_invocations', includes: includes, preload: preload)
+    end
+
+    def load_job_invocation_by_id(id, includes: nil, preload: nil)
+      search_term = "id=#{id}"
+      begin
+        # double 'first' call due to BatchEnumerator and ActiveRecord::Relation
+        load_job_invocations(search: search_term, includes: includes, preload: preload).first.first
+      rescue NoMethodError
```
ares: why an unknown id raises NoMethodError? is that due to first batch returns nil? if that's the case, I'd rather check for result explicitly, like

```ruby
if (batch = load_job_invocations(...))
  batch.first
else
  raise N_("...")
end
```

the reason is, this rescue block would catch also other errors, e.g. if `load_job_invocations` would be calling something that doesn't exist

app/services/foreman_remote_execution/renderer_methods.rb
```diff
@@ -0,0 +1,23 @@
...
+        # Catching NoMethodError for better error message for user
+        raise N_("Can't find Job Invocation with given id")
```
ares: we should probably raise Foreman exception to generate ERF code

```ruby
raise ::Foreman::Exception.new(N_("Can't find Job Invocation for an id %s"), id)
```

app/services/foreman_remote_execution/renderer_methods.rb
```diff
@@ -0,0 +1,23 @@
...
+module ForemanRemoteExecution
+  module RendererMethods
+    extend ActiveSupport::Concern
+
+    def load_job_invocations(search: '', includes: nil, preload: nil)
+      load_resource(klass: JobInvocation, search: search, permission: 'view_job_invocations', includes: includes, preload: preload)
+    end
+
+    def load_job_invocation_by_id(id, includes: nil, preload: nil)
+      search_term = "id=#{id}"
```
ares: nitpick, but I'd keep this small string in and would wrap the equal sign with spaces

```
load_job_invocations(search: "id = #{id}", includes: includes, preload: preload).first.first
```

lib/foreman_remote_execution/engine.rb
```diff
@@ -123,6 +123,9 @@ class Engine < ::Rails::Engine
           ctx.permit :execution
         end
 
+        extend_template_helpers ForemanRemoteExecution::RendererMethods
+        allowed_template_helpers :load_job_invocations
```
ares: do we also need `load_job_invocation_by_id`? and do we need to use `allowed_template_helpers` if we use `extend_template_helpers` (I don't recall but I though the first is enough)?

app/helpers/remote_execution_helper.rb
```diff
@@ -236,4 +243,25 @@ def load_template_from_task(template_invocation, target)
     return if task.nil?
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
   end
+
+  def jobs_template_with_input(job_invocation)
+    template_name = Setting['remote_execution_job_invocation_template']
```
ares: I'd combine this and the next line, but not a strong opinion, IMHO one extra variable is overkill for this, it's not used anywhere else

app/helpers/remote_execution_helper.rb
```diff
@@ -236,4 +243,25 @@ def load_template_from_task(template_invocation, target)
...
+    template = ReportTemplate.where(name: template_name).first
+    return if template.nil?
+    template_input = template.template_inputs.where(name: 'job_id').first
+    return if template_input.nil?
+    input_id = template_input.id
```
ares: I think this is unnecessary variable, `template_input.id` is quite readable, the same case as `job_invocation.id`

domitea:
All PR has been updated. Is everything looks good now?

ares:

ares (dismissed):
Tests are probably broken after smart variables removal from core. I'm going to ignore this for now. However, we need to bump dependency on foreman-tasks to 1.0.2 in https://github.com/theforeman/foreman_remote_execution/blob/master/foreman_remote_execution.gemspec#L30 and wait until the tasks PR is in. @domitea could you please update the gemspec?

domitea:
@ares gemspec updated, commits squashed

adamruzicka:
The gemspec still has a conflict

adamruzicka:
@domitea what's the status here?

domitea:
@adamruzicka sorry, too much another stuff to do... Gemespec conflict resolved and I hope it will solve the problem with tests (as @ares suggested)

ezr-ondrej:
Needs rebase now, Hound comments about indentation relevant.

domitea:
Code updated to comments (see https://github.com/theforeman/foreman-tasks/pull/490 and https://github.com/theforeman/community-templates/pull/666 also). I've added field for filtering hosts that are affected by job. So this covers @ares comment in tasks PR. Now the user can select by query what hosts can be in report. Also the template has been updated.

tbrisker:
Tasks pr has been merged, this pr still has rubocop errors that need fixing.

adamruzicka:
Needs a rebase

adamruzicka:
@domitea what is the status here?

domitea:
@adamruzicka @ezr-ondrej Code updated and rebased. I'm sorry for lacking to finnish this, I was working on others PRs

app/services/renderer_methods.rb
```diff
@@ -0,0 +1,22 @@
...
+# macros to fetch information about invoked jobs
+module RendererMethods
+  extend ActiveSupport::Concern
+
+  def find_job_invocation_by_id(job_id, preload: nil)
+    if (job =  JobInvocation.preload(preload).find_by(id: job_id).first)
+      job
+    else
+      raise ::Foreman::Exception.new(N_("Can't find Job Invocation for an id %s"), job_id)
+    end
```
ezr-ondrej, line 12:
```suggestion
    JobInvocation.preload(preload).find(job_id)
  rescue ActiveRecord::NotFound => _e
    # you can use the exception for more informative message
    raise ::Foreman::Exception.new(N_("Can't find Job Invocation for an id %s"), job_id)
```

app/services/renderer_methods.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def load_job_invocations(search, includes: nil, preload: nil)
+    load_resource(klass: JobInvocation, search: search, permission: :view_job_invocations, includes: includes, preload: preload)
+  end
+
+  def search_in_hosts_relation(hosts, search_query)
```
ezr-ondrej, line 19: Why would I want to do this? Why I haven't used the search directly when I was loading the hosts?
And if you think this is useful, shouldn't it be in core?

app/services/renderer_methods.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# macros to fetch information about invoked jobs
+module RendererMethods
+  extend ActiveSupport::Concern
+
+  def find_job_invocation_by_id(job_id, preload: nil)
```
ezr-ondrej, line 7: Actualy why do we use this, instead of using `JobInvocation.find(id)` in the template?

app/services/renderer_methods.rb
```diff
@@ -0,0 +1,22 @@
...
+    if (job =  JobInvocation.preload(preload).find_by(id: job_id).first)
+      job
+    else
+      raise ::Foreman::Exception.new(N_("Can't find Job Invocation for an id %s"), job_id)
+    end
+  end
+
+  def load_job_invocations(search, includes: nil, preload: nil)
```
ezr-ondrej, line 15: This is nice, but if that should be the only macro, I think using `load_resource` in the template would be good enough :man_shrugging:

ezr-ondrej:
Also :police_officer: (rubocop) is unhappy.

app/services/renderer_methods.rb
domitea: This method trying to solve problem where I've got a relation from another object - like hosts from task. I cannot to filter these result by another method. Of course, I can use the `search_for` method directly. However, this method is from `scoped_search` plugin and I don't think that is good idea to open (allow in safemode) this kind of method. 

So that's the reason why I wrote this method. I have no opinion in that way.
At least, I moved this method to core (Good idea!). PR from core: https://github.com/theforeman/foreman/pull/7879

app/services/renderer_methods.rb
domitea: This method has similar reason like the upper one - I like to avoid user to use methods from AR.

app/services/renderer_methods.rb
domitea: I've removed this macro. The main point for this macro was abstraction as suggested @ares in the begging of this PR. However I don't think that this should be use for daily basis so I removed them.

domitea:
@ezr-ondrej I've also add the template to the plugin since the `community-template` are deprecated. Is it in correct place?

app/helpers/remote_execution_helper.rb
```diff
@@ -230,6 +237,26 @@ def load_template_from_task(template_invocation, target)
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
   end
 
+  def jobs_template_with_input(job_invocation)
```
ezr-ondrej, line 240: Could we split this method into two methods and return just one value?
It would be much more readable IMHO.
```suggestion
  def job_report_template
    ReportTemplate.where(name: Setting['remote_execution_job_invocation_template']).first
    return template if template && template.template_inputs.where(name: 'job_id').exists?
  end
  
  def job_report_template_parameters(job_invocation, template)
    template_input = template.template_inputs.where(name: 'job_id').first
    raise "#job_report_template_parameters need template that has 'job_id' input" unless template_input
    {
      report_template_report: {
        input_values: {
          "#{template_input.id}": {
            value: job_invocation.id,
          },
        },
      },
    }
  end
```

app/helpers/remote_execution_helper.rb
```diff
@@ -60,6 +60,13 @@ def job_invocation_task_buttons(task)
     job_invocation = task.task_groups.find { |group| group.class == JobInvocationTaskGroup }.job_invocation
     task_authorizer = Authorizer.new(User.current, :collection => [task])
     buttons = []
+    template, template_parameters = jobs_template_with_input(job_invocation)
+    if template && authorized_for(controller: :report_templates, action: :generate)
+      buttons << link_to(_('Create Report'), generate_report_template_path(template, template_parameters),
```
ezr-ondrej, line 65: I'd much rather see it splited into two method calls.
```suggestion
    if (template = job_report_template ) && authorized_for(controller: :report_templates, action: :generate)
      buttons << link_to(_('Create Report'), generate_report_template_path(template, job_report_template_parameters(job_invocation, template)),
```

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -121,6 +125,16 @@ def exit_status
         delegated_output[:exit_status]
       end
 
+      # Method that produces hash where key is output type (ie. stdout, stderr...) and value is output for given type
+      # So the returned hash look like {'stdout': 'normal output', 'stderr': 'error'}
+      def output_by_type
```
ezr-ondrej, line 130: @adamruzicka could you ack that this method make sense, I'm not sure here.

app/models/job_invocation.rb
```diff
@@ -72,6 +72,10 @@ class JobInvocation < ApplicationRecord
 
   encrypts :password, :key_passphrase, :sudo_password
 
+  class Jail < Safemode::Jail
+    allow :task_id, :sub_task_for_host, :template_invocations_hosts
```
ezr-ondrej, line 76: We are not using the `task_id` anymore in the template, right?
```suggestion
    allow :sub_task_for_host, :template_invocations_hosts
```

app/services/renderer_methods.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# macros to fetch information about invoked jobs
+module RendererMethods
```
ezr-ondrej, line 4: Why do we have this file in `services` folder? :thinking:

lib/foreman_remote_execution/engine.rb
```diff
@@ -134,6 +135,9 @@ class Engine < ::Rails::Engine
           ctx.permit :execution
         end
 
+        extend_template_helpers ForemanRemoteExecution::RendererMethods
+        allowed_template_helpers :find_job_invocation_by_id, :load_job_invocations, :search_in_hosts_relation
```
ezr-ondrej, line 139: The `search_in_hosts_relation` should be enabled in the core PR, right?
```suggestion
        allowed_template_helpers :find_job_invocation_by_id, :load_job_invocations
```

app/views/report_templates/jobs_invocations.erb
```diff
@@ -0,0 +1,35 @@
...
+  input_type: user
+  description: Field for filter hosts of job invocation. Leave blank for all hosts.
+  advanced: false
+  value_type: search
+  resource_type: Host
+  hidden_value: false
+model: ReportTemplate
+-%>
+<%- invocation = load_job_invocation_by_id(input('job_id')) -%>
+<%- search_in_hosts_relation(invocation.template_invocations_hosts, input('hosts')).each do |host| -%>
```
ezr-ondrej, line 24: BTW, would this work instead of the core PR?
```suggestion
<%- load_hosts(search: input('hosts'), where: {id: template_invocations_hosts.pluck(:id)}).each do |host| -%>
```

app/services/renderer_methods.rb
domitea: I 'stole' this place from foreman_ansible. Is somewhere better place for this module? I can move it.

app/views/report_templates/jobs_invocations.erb
domitea: It should probably work. I understand that this makes core PR pointless but also I like to don't bother users with AR methods.  What do you think? What think others (@ares or @adamruzicka) ?

ares: I wouldn't be too concerned about users needing to understand AR here, there's very low chance they would be modifying this template. Also it may be easier for them to understand `load_hosts` which is well documented and used everywhere else. I'd be ok with the change, proven it works with safe mode. I'm not sure `pluck` is allowed.

domitea: I understand that `.map(&:id)` is much more easier. On the other hand, `.map` method (`.pluck` is not allowed) loads whole object to memory. I think that the host is one of complex models in Foreman. That means that suggested code loads whole host model only if `id`. It produces longer time to generate the report for large numbers of hosts. 

I'm little worried about this. What do you think @ares or @ezr-ondrej ?

tbrisker: Perhaps we should add `.ids` method to AR jails in foreman core?

domitea: Great idea @tbrisker!

domitea:
I edited the linked PR (https://github.com/theforeman/foreman/pull/7879) with `search_in_relation` macro to PR where the `.ids` method is allowed in safe mode.

lib/foreman_remote_execution/engine.rb
```diff
@@ -134,6 +135,9 @@ class Engine < ::Rails::Engine
           ctx.permit :execution
         end
 
+        extend_template_helpers ForemanRemoteExecution::RendererMethods
+        allowed_template_helpers :find_job_invocation_by_id, :load_job_invocations
```
ezr-ondrej, line 139: Is this correct?

ezr-ondrej (changes requested):
Rubocop is still unhappy about this (https://github.com/theforeman/foreman_remote_execution/pull/453/checks?check_run_id=1031568173)

domitea:
@ezr-ondrej rubocop solved- I think it's ready to merge.

app/models/setting/remote_execution.rb
```diff
@@ -73,6 +73,12 @@ def self.default_settings
         _('Form Job Template'),
         nil,
         { :collection => proc { Hash[JobTemplate.unscoped.map { |template| [template.name, template.name] }] } }),
+      self.set('remote_execution_job_invocation_template',
+        N_('Select a report template used for generating a report for a particular remote execution job'),
+        'Job Invocation report',
+        _('Job Invocation Template'),
```
ezr-ondrej, line 79:
```suggestion
        _('Job Invocation Report Template'),
```

ezr-ondrej (changes requested):

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -121,6 +125,16 @@ def exit_status
...
+        return {} unless task.main_action && task.main_action&.live_output&.any?
```
adamruzicka, line 131: `task.main_action` should equal to `self` here

app/lib/actions/remote_execution/run_host_job.rb
adamruzicka: It gets to the result in a bit roundabout way, but essentially yes.

Although I think that having
```json
{ "stdout": ["a", "c"], "stderr": ["b"] }
```

is less useful than

```
{
  "outputs": [
    {"type": "stdout", "value": "a"},
    {"type": "stderr", "value": "b"},
    {"type": "stdout", "value": "c"},
  ]
}
```
because you lose information about ordering

app/views/report_templates/jobs_invocations.erb
```diff
@@ -0,0 +1,35 @@
...
+  advanced: false
+  value_type: search
+  resource_type: Host
+  hidden_value: false
+model: ReportTemplate
+-%>
+<%- invocation = load_job_invocation_by_id(input('job_id')) -%>
+<%- load_hosts(search: input('hosts'), where: {id: template_invocations_hosts.ids}).each do |host| -%>
+<%-   task = invocation.sub_task_for_host(host) -%>
+<%-   sub_task_output = task.output_by_type -%>
```
adamruzicka, line 26: If I'm reading this right then `task` here is an instance of `ForemanTasks::Task`, but `output_by_type` is defined on `RunHostJob` so `task.output_by_type` should raise `NoMethodError` and we should use something along the lines of `task.main_action.output_by_type`

app/lib/actions/remote_execution/run_host_job.rb
domitea: It's little worse, with the current method implementation, I only get the last value for given type. So i this case, the result is: `{  "stdout": "c", "stderr": "b" }`. 

I didn't know that this can happen (seems logically right know) but still I've tried to display this values in the report. In worst case, in report that is excel table and value is in cell. Honestly, I didn't get any suitable solution for this. I have some ideas, but I don't think that add "position" number to result is good idea. Also showing one task on host for more times for given outputs is not good idea to me.  

Any opinions in there? @adamruzicka @ezr-ondrej @ares

app/views/report_templates/jobs_invocations.erb
domitea: I have two options to solve this problem: 

1) Add macro for `task.main_action.output_by_type`. I know, more macros is not a good idea but I would like to shield user from the internals (sounds funny for OSS :D )

2) Allow `main_action` in safe mode. Is there any drawback for that? My opinion is not to do that. 
 
What is more suitable for you? @adamruzicka @ezr-ondrej @ares

app/lib/actions/remote_execution/run_host_job.rb
adamruzicka: Keeping just last output from each type is a no go I'm afraid.

Can't we go with the format I proposed (keeping ordering and type) and then deal with the serialization in the template? If the source data allows us to easily transform it then we could quite easily:
1) Discard the type, join all the outputs into one string with newlines - keeps ordering, discards type
2) Join per-type outputs into strings - loses ordering, keeps type
3) Prefix each output with type specifier (`STDOUT | $message`), join them with newlines - keeps ordering and type, but the reported output as a raw string won't be exactly the same as what came out of the job

app/views/report_templates/jobs_invocations.erb
ezr-ondrej: Well, given the @adamruzicka's comment above (format of the outputs) we really should rename the method to `output`. And I guess it's ok to add `action_output` to `Jail` enabled methods and it being safeguard as well if the task doesn't have output.

```ruby
# ForemanTasks::Task
def action_output
  main_action.output if main_action.respond_to?(:output) #or maybe if main_action.is_a?(Actions::Helpers::WithContinuousOutput)
end
```

BUT `main_action` is already in `Jail` enabled methods. If we Jail guard action as well, there shouldn't be problem to do that.

app/lib/actions/remote_execution/run_host_job.rb
ezr-ondrej: AND we could do this dirrectly on `ForemanTasksCore::ContinuousOutput`.
We could do:
```ruby
# class ForemanTasksCore::ContinuousOutput
  def to_json
    sort!
    # format raw_outputs to json
  end
```

And have this method just as a proxy to that.

ezr-ondrej:
What is the status @domitea ?

app/services/renderer_methods.rb
ezr-ondrej: This method is no longer used IIRC

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -7,6 +7,10 @@ class RunHostJob < Actions::EntryAction
       middleware.do_not_use Dynflow::Middleware::Common::Transaction
       middleware.use Actions::Middleware::HideSecrets
 
+      class Jail < Safemode::Jail
+        allow :output_by_type
+      end
+
```
ezr-ondrej, line 13: We don't need this anymore, right?

app/views/report_templates/jobs_invocations.erb
```diff
@@ -0,0 +1,35 @@
...
+  input_type: user
+  description: Field for filter hosts of job invocation. Leave blank for all hosts.
+  advanced: false
+  value_type: search
+  resource_type: Host
+  hidden_value: false
+model: ReportTemplate
+-%>
+<%- invocation = load_job_invocation_by_id(input('job_id')) -%>
+<%- load_hosts(search: input('hosts'), where: {id: template_invocations_hosts.ids}).each do |host| -%>
```
ezr-ondrej, line 24: Does this work? This seems like `UndefinedMethodError`
```suggestion
<%- load_hosts(search: input('hosts'), where: {id: invocation.template_invocations_hosts.ids}).each do |host| -%>
```

ezr-ondrej (changes requested):
This looks quite good already, but I guess it won't work.
Could you fix what have I found and test if it works and I believe we can give it a final look.

ezr-ondrej:
Btw, please add some description to the comment (`git commit --amend` before pushing should do the job)

app/views/report_templates/jobs_invocations.erb
```diff
@@ -0,0 +1,35 @@
...
+  required: false
+  input_type: user
+  description: Field for filter hosts of job invocation. Leave blank for all hosts.
+  advanced: false
+  value_type: search
+  resource_type: Host
+  hidden_value: false
+model: ReportTemplate
+-%>
+<%- invocation = load_job_invocation_by_id(input('job_id')) -%>
```
ezr-ondrej, line 23: There is still mismatch in name of the macro, does this work like this? If so, we can remove the macro from `ForemanRemoteExecution::RendererMethods`. There we have named it `find_job_invocation_by_id`

domitea: Nice catch :+1:  I miss that :(

app/helpers/remote_execution_helper.rb
```diff
@@ -230,6 +236,26 @@ def load_template_from_task(template_invocation, target)
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
   end
 
+  def job_report_template
+    ReportTemplate.where(name: Setting['remote_execution_job_invocation_template']).first
```
ezr-ondrej, line 240:
```suggestion
    template = ReportTemplate.where(name: Setting['remote_execution_job_invocation_template']).first
```

ezr-ondrej (changes requested):
During the testing I've hitted `UndefinedMethodError`

domitea:
I did a more harder testing, because that was unexpected :/ 

I update a template a bit (due to batch) and also I've made a PR to core (https://github.com/theforeman/foreman_remote_execution/pull/453) because `where` clause was not allowed to use. It was missing in `load_hosts` macro. With that PR, it should be available in all macros defined this way.

app/views/report_templates/jobs_invocations.erb
```diff
@@ -0,0 +1,37 @@
...
+  input_type: user
+  description: Field for filter hosts of job invocation. Leave blank for all hosts.
+  advanced: false
+  value_type: search
+  resource_type: Host
+  hidden_value: false
+model: ReportTemplate
+-%>
+<%- invocation = find_job_invocation_by_id(input('job_id')) -%>
+<%- load_hosts(search: input('hosts'), where: {id: invocation.template_invocations_hosts.ids}).each do |batch| -%>
```
ezr-ondrej, line 24:
```suggestion
<%- load_hosts(search: "#{input('hosts')} and job_invocation.id = #{input('job_id')}).each do |batch| -%>
```

ezr-ondrej (changes requested):
I think we can use the search for the host filtration.

domitea:
@ezr-ondrej I've to update the template a little bit. Your suggestion was great, however `scoped_search` doesn't like to find anything on query `  and job_invocation.id = ...` when you're ignoring the hosts field. In that case, I'm adding a `*` instead of `input('hosts')`. The result query looks like `* and job_invocation.id = ...` and it find what we want.

app/views/report_templates/jobs_invocations.erb
```diff
@@ -0,0 +1,39 @@
...
+  hidden_value: false
+model: ReportTemplate
+-%>
+<%- invocation = find_job_invocation_by_id(input('job_id')) -%>
+<%- input_hosts = input('hosts') -%>
+<%- input_hosts = '*' if input_hosts.blank? -%>
+<%- load_hosts(search: "#{input_hosts} and job_invocation.id = #{input('job_id')}").each do |batch| -%>
+<%-   batch.each do |host|  -%>
+<%-     task = invocation.sub_task_for_host(host) -%>
+<%-     task.action_output['outputs'].each do |output| -%>
```
ezr-ondrej, line 29: Why not?
```suggestion
<%-     task.main_action.live_output.each do |output| -%>
```

app/models/setting/remote_execution.rb
```diff
@@ -73,6 +73,12 @@ def self.default_settings
         _('Form Job Template'),
         nil,
         { :collection => proc { Hash[JobTemplate.unscoped.map { |template| [template.name, template.name] }] } }),
+      self.set('remote_execution_job_invocation_template',
+        N_('Select a report template used for generating a report for a particular remote execution job'),
+        'Job Invocation report',
+        _('Job Invocation Report Template'),
+        nil,
+        { :collection => proc { Hash[ReportTemplate.unscoped.map { |template| [template.name, template.name] }] } }),
```
ezr-ondrej, line 81: Should we help the user by? We require the input to be present, so it would make sense to offer only templates that has it.
```suggestion
        { :collection => proc { Hash[ReportTemplate.unscoped.join(:template_inputs).where(TemplateInput.arel_table[:name].eq('job_id')).map { |template| [template.name, template.name] }] } }),
```

ezr-ondrej (changes requested):
We should definitelly include some tests. I realized we have zero coverage for this funcionality :thinking:

app/models/setting/remote_execution.rb
domitea: Great idea!

app/views/report_templates/jobs_invocations.erb
domitea: I think that I understand what are trying for. However I think that `live_output` is not allowed for safemode, yet. It's not problem, I can add jail for `RunHostJob` but I think we will facing to similar problem like https://github.com/theforeman/foreman_remote_execution/pull/453#discussion_r488658734

domitea:
I've updated a select for valid templates in settings. 

But I'm not sure what kind of tests should cover this. I can only imagine test only for valid selection of templates in settings. Any idea @ezr-ondrej ?

ezr-ondrej:
I'd want to see at least these two:
1) template is found correctly
2) render the template with prepared `JobInvocation` and match the expected result.

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,78 @@
...
+        JobInvocation.any_instance.expects(:sub_task_for_host).returns(fake_task)
+
+        template_source = File.read(File.expand_path(File.dirname(__FILE__) + "/../../app/views/report_templates/jobs_invocations.erb"))
+        template = ReportTemplate.import_without_save("Job Invocation Report Template", template_source)
+        template.save!
+        input = template.template_inputs.first
+        params = {template_id: template.id, input_values: {input.id.to_s => {value: job_invocation.id.to_s} } }
+        composer = ReportComposer.new(params)
+        time = Time.now
+        result = composer.render
```
ezr-ondrej, line 55: Why don't we parse the result as CSV here instead of all the splitting?

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,78 @@
...
+        # parsing the result, csv file originally
+        rows = result.split("\n")
+        # row 0 contains header
+        row_1 = rows[1].split(',')
+        row_2 = rows[2].split(',')
+        row_3 = rows[3].split(',')
+
+        row_1[1].must_equal 'stderr'
+        row_1[2].must_equal 'error'
+        assert_operator Time.parse(row_1[3]), :<, time
```
ezr-ondrej, line 66: We are in control of the value, do we passed as `ended_at`, can't we have `Time.new(2020, 11, 30, 0, 0, 0)` or something like that so we can have precise comparison?

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,78 @@
...
+  context 'job invocation report template' do
+
+    describe 'correct template is available' do
+      it 'settings select includes only report templates with name job_id' do
+        FactoryBot.create(:report_template, name: 'Template 1')
+        correct_template = FactoryBot.create(:report_template, name: 'Job Invocation Report Template')
+        FactoryBot.create(:template_input, template: correct_template, input_type: 'user', name: 'job_id')
+        templates = Setting::RemoteExecution.job_invocation_report_templates_select
+
+        templates.must_include 'Job Invocation Report Template'
```
ezr-ondrej, line 19:
```suggestion
        assert_include templates, 'Job Invocation Report Template'
```

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+
+      it 'should found correct template from setting' do
```
ezr-ondrej, line 22: This test would be better of and made bit more sense in a `RemoteExecutionHelperTest`. We would test a `job_report_template` method. It would made bit more sense to me :thinking:

ezr-ondrej (commented):
Just left couple of suggestions for the tests, but looks good.
I'd preffer `assert_<>` over `#must_<>` but it's just my personal choice :)

domitea:
@ezr-ondrej I've updated the PR regarding to your comments. I think it's ready for merge.

test/helpers/remote_execution_helper_test.rb
```diff
@@ -23,4 +23,20 @@ class RemoteExecutionHelperTest < ActionView::TestCase
...
+
+  describe 'test correct setting' do
+    it 'should found correct template from setting' do
+      Setting::RemoteExecution.load_defaults
+      template_name = 'Job Invocation Report Template'
+      setting_key = 'remote_execution_job_invocation_report_template'
+      template = FactoryBot.create(:report_template, name: template_name)
+      input = FactoryBot.create(:template_input, name: 'job_id', input_type: 'user')
+      template.template_inputs << input
+      Setting[setting_key] = template_name
```
ezr-ondrej, line 35: We would be bit better of with, but not a big deal.
```suggestion
      Setting.expects(:[]).with(setting_key).returns(template_name)
```

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,67 @@
...
+        # parsing the result, csv file originally
+        data = CSV.parse(result, headers: true)
+
+        stderr_row = data[0].to_h
+        stdout_row = data[1].to_h
+        stdebug_row = data[2].to_h
+
+        assert_equal stderr_row['type'], 'stderr'
+        assert_equal stderr_row['message'], 'error'
+        assert_operator Time.parse(stderr_row['ended_at']), :<, time
```
ezr-ondrej, line 55: My point was to get the assertion to be more exact... this is almost equal to:
```suggestion
        assert_kind_of Time.parse(stderr_row['ended_at']), Time, 'Parsing of time column failed'
```
What would be ok by me, but we don't need a helper variable (`time`) for that :thinking:

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,67 @@
...
+        input = template.template_inputs.first
+        params = {template_id: template.id, input_values: {input.id.to_s => {value: job_invocation.id.to_s} } }
+        composer = ReportComposer.new(params)
+        time = Time.now
+        result = composer.render
+
+        # parsing the result, csv file originally
+        data = CSV.parse(result, headers: true)
+
+        stderr_row = data[0].to_h
```
ezr-ondrej, line 49: should we also validate the host column? :thinking:

ezr-ondrej (approved):
Left few more ideas for improvements of the tests, but I'd be ok to merge without.
Let me know if you will have time to get to them, I'll plan to test this manually right away.

test/unit/job_invocation_report_template_test.rb
domitea: I would implement this as finding a host by given name from report. However I think that this functionality is tested at the Host tests.

domitea:
@ezr-ondrej I've updated the PR.

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+        stdout_row = data[1].to_h
+        stdebug_row = data[2].to_h
+
+        assert_equal stderr_row['type'], 'stderr'
+        assert_equal stderr_row['message'], 'error'
+        assert_kind_of Time.parse(stderr_row['ended_at']), Time, 'Parsing of time column failed'
```
ezr-ondrej, line 54:
```suggestion
        assert_kind_of Time, Time.parse(stderr_row['ended_at']), 'Parsing of time column failed'
```

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        assert_equal stdout_row['type'], 'stdout'
+        assert_equal stdout_row['message'], 'output'
+        assert_kind_of Time.parse(stdout_row['ended_at']), Time, 'Parsing of time column failed'
```
ezr-ondrej, line 58:
```suggestion
        assert_kind_of Time, Time.parse(stdout_row['ended_at']), 'Parsing of time column failed'
```

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        assert_equal stdebug_row['type'], 'stdebug'
+        assert_equal stdebug_row['message'], 'debug'
+        assert_kind_of Time.parse(stdout_row['ended_at']), Time, 'Parsing of time column failed'
```
ezr-ondrej, line 62:
```suggestion
        assert_kind_of Time, Time.parse(stdout_row['ended_at']), 'Parsing of time column failed'
```

test/unit/job_invocation_report_template_test.rb
ezr-ondrej: ok, it's not as important

ezr-ondrej (changes requested):
I've messed up an order for the `assert_kind_of`, sorry, can you please update this?

domitea:
@ezr-ondrej PR updated. Interestingly, the tests passed in both situations at my setup.

domitea:
@ezr-ondrej I've made some updated in this PR. I've also removed a template and created for them a separate PR (https://github.com/theforeman/foreman/pull/8178) in core

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+          { type: 'stdebug', value: "debug" },
+        ]
+      end
+      let(:fake_task) { FakeTask.new(action_output: { outputs: fake_json_outputs }, ended_at: Time.new(2020, 12, 1, 0, 0, 0)) }
+
+      it 'should process the result of task to template' do
+        job_invocation = FactoryBot.create(:job_invocation, :with_task)
+        JobInvocation.any_instance.expects(:sub_task_for_host).returns(fake_task)
+
+        template_source = File.read(File.expand_path(File.dirname(__FILE__) + "/../../app/views/report_templates/jobs_invocations.erb"))
```
ezr-ondrej, line 37: this no longer works, we should use the template from core.

domitea:
@ezr-ondrej path updated but I still have a failing test on that. However I think it's caused by core PR (https://github.com/theforeman/foreman/pull/8178) that is not merged. So required template is not included at core repo.

ezr-ondrej:
> @ezr-ondrej path updated but I still have a failing test on that. However I think it's caused by core PR ([theforeman/foreman#8178](https://github.com/theforeman/foreman/pull/8178)) that is not merged. So required template is not included at core repo.

Indeed, that PR will need to go first.

adamruzicka:
The changes in tasks will most likely go out in 3.1.0. Could you bump the dependency here?

domitea:
@adamruzicka PR updated.

adamruzicka:
The tests are red and the failures are most likely related

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+          { output_type: 'stdebug', output: "debug" },
+        ]
+      end
+      let(:fake_task) { FakeTask.new(action_output: fake_outputs, ended_at: Time.new(2020, 12, 1, 0, 0, 0)) }
+
+      it 'should process the result of task to template' do
+        job_invocation = FactoryBot.create(:job_invocation, :with_task)
+        JobInvocation.any_instance.expects(:sub_task_for_host).returns(fake_task)
+
+        template_source = File.read(File.expand_path(Rails.root + "app/views/report_templates/jobs_invocations.erb"))
```
ezr-ondrej, line 37: Wrong path to the template.
`app/views/unattended/report_templates/jobs_-_invocation_report_template.erb`

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    describe 'test correct result for report template' do
+      let(:fake_outputs) do
+        [
+          { output_type: 'stderr', output: "error" },
+          { output_type: 'stdout', output: "output" },
+          { output_type: 'stdebug', output: "debug" },
+        ]
+      end
+      let(:fake_task) { FakeTask.new(action_output: fake_outputs, ended_at: Time.new(2020, 12, 1, 0, 0, 0)) }
```
ezr-ondrej, line 31: we are using the `fake_outputs` timestamp now.

test/unit/job_invocation_report_template_test.rb
```diff
@@ -0,0 +1,66 @@
...
+        assert_include templates, 'Job Invocation Report Template'
+      end
+    end
+
+    describe 'test correct result for report template' do
+      let(:fake_outputs) do
+        [
+          { output_type: 'stderr', output: "error" },
+          { output_type: 'stdout', output: "output" },
+          { output_type: 'stdebug', output: "debug" },
```
ezr-ondrej, line 28: This expect string keys and `timestamp` to be one of them.

ezr-ondrej (changes requested):

app/helpers/remote_execution_helper.rb
```diff
@@ -230,6 +236,27 @@ def load_template_from_task(template_invocation, target)
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
   end
 
+  def job_report_template
+    template = ReportTemplate.where(name: Setting['remote_execution_job_invocation_report_template']).first
+
+    return template if template && template.template_inputs.where(name: 'job_id').exists?
```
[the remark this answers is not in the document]
ezr-ondrej: safe navigation does nothing as it is always scope here.

adamruzicka:
Tests are still red

domitea:
@adamruzicka tests are red because they're trying to find a non-existent template. The problem is that complementary PR with the template (https://github.com/theforeman/foreman/pull/8178) is not merged. Thanks to that template is not included in core and test are failing.

adamruzicka:
The template PR in foreman core was just merged

ezr-ondrej:
> The template PR in foreman core was just merged

Thanks @adamruzicka, I've rebased this to restart the tests and let see :)

ezr-ondrej:
Right, now we need https://github.com/theforeman/foreman-tasks/pull/623

adamruzicka:
> Right, now we need [theforeman/foreman-tasks#623](https://github.com/theforeman/foreman-tasks/pull/623)

Needed changes released in foreman-tasks-4.1.0

ezr-ondrej:
@adamruzicka I've updated the tests and we are :green_apple: now.
I've tested it and it works nice-ish (I'm still not sure about usefullness, but it works xD)

app/helpers/remote_execution_helper.rb
```diff
@@ -60,6 +60,12 @@ def job_invocation_task_buttons(task)
     job_invocation = task.task_groups.find { |group| group.class == JobInvocationTaskGroup }.job_invocation
     task_authorizer = Authorizer.new(User.current, :collection => [task])
     buttons = []
+    if (template = job_report_template) && authorized_for(controller: :report_templates, action: :generate)
+      buttons << link_to(_('Create Report'), generate_report_template_path(template, job_report_template_parameters(job_invocation, template)),
+        class: 'btn btn-default',
+        title: _('Create report for this job'),
+        disabled: task.pending?)
```
adamruzicka, line 67: If you go through monitor > report templates, then you can get a report generated for a still-running job, correct? It doesn't really care the job it is still running, it just generates the report from currently available data. Shouldn't we just allow doing the same through the button? I mean we should be consistent, not allow in through one place but not through the other

adamruzicka (commented):
One question inline, apart from it it appears to work nicely.

app/helpers/remote_execution_helper.rb
ezr-ondrej: I don't feel there is much value in doing so, is there ?
Would we want report on a still running task, do we have tasks that takes e.g. higher order of hours to complete, so someone would be reporting on that in the progress?
I believe we should be opinionated and allow user to do just what we believe has value. Otherwise we're confusing user with actions that doesn't make sense, but I'm not really able to say if it is valid usecase or not :thinking:

adamruzicka: Ok then, let's get it in and see what the users think

adamruzicka (approved):

adamruzicka:
Thank you @domitea & @ezr-ondrej !

PR #454, opened 2019-11-27
Fixes #28372 - Fix targeting autocompletion

adamruzicka:
The previous implementation of autocomplete relied on a legacy js which was
recently dropped from Foreman. This PR uses the new react component for
autocomplete. The drawback is the react component render the search box as an
input, not as a textarea. Sadly the underlying libraries don't seem to be
flexible enough to allow using a textarea easily.

Requires:
- [x] https://github.com/theforeman/foreman/pull/7212

### DIFF
--- a/app/views/job_invocations/_form.html.erb
+++ b/app/views/job_invocations/_form.html.erb
@@ -33,7 +33,7 @@
       <% end %>
     </span>
     <%= selectable_f targeting_fields, :bookmark_id, @composer.available_bookmarks.map { |b| [b.name, b.id] }, :selected => @composer.targeting.bookmark_id, :include_blank => true %>
-    <%= textarea_f targeting_fields, :search_query, :value => @composer.displayed_search_query, :rows => 5, :class => 'autocomplete-input form-control', :'data-url' => auto_complete_search_hosts_path %>
+    <%= autocomplete_f targeting_fields, :search_query, :search_query => @composer.displayed_search_query, :full_path => auto_complete_search_hosts_path %>
 
     <div class="form-group ">
       <label class="col-md-2 control-label"><%= _('Resolves to') %></label>


adamruzicka:
Let's have a little poll:
:+1: autocompletion, but a single line input
:-1:  no autocompletion, multiline input

ekohl:
I'd prefer the consistency with the rest of Foreman. Perhaps it makes sense to get the Foreman component to support expanding into a modal (or similar) so you can use long inputs everywhere. Not just here. This would be the same as the input fields for Puppet class parameters which can expand (though they don't have autocompletion).

@theforeman/ui-ux should also weigh in.

adamruzicka:
It is fine when the input is small, things start to break when it gets longer. Attaching a screenshot for reference
image

laviro:
gotcha, maybe we can open a PR to `react-bootstrap-typeahead` which is used in `Patternfly-react` to use an optional `text-area` or at least to adjust the height or width when the input is full.. I will try to open a PR or a new issue there, or worst case we could adjust our own css to set dynamically the dimensions.
Until that change would take place, I believe we should be consistent with Foreman core - even if the implementation isn't perfect for all of the cases. could you please upload a snapshot of how it used to be? thanks :pray:

adamruzicka:
Screenshot with textarea and legacy js autocompletion
image

laviro:
Opened an issue to [react-bootstrap-typeahead](https://github.com/ericgio/react-bootstrap-typeahead/issues/504)

adamruzicka:

ares:
Failing due to rubocop now

adamruzicka:
I know, waited for the rubocop PR to get in

laviro:
> Opened an issue to react-bootstrap-typeahead

Their answer:
> - Hint overflows the text input: This is documented in #473. You should be able to resolve that fairly easily with some CSS.
> - Long hint doesn't align with input text: Also addressed in #473. As noted there, this is a tougher problem to solve, and I'm not sure if it's possible. If this is a real problem for you, you may want to disable the hint (via CSS).
> - Long menu items: It looks like you're using align="left", in which case the menu items will stretch to the length of their contents. If you use the default (align="justify"), the menu will be the width of the input and the menu items will be truncated accordingly. 
> 
> If you need more control over rendering of the menu items, use the renderMenu prop to customize the menu.
> Giving the input a dynamic width or using a textarea instead of an input are not things I plan on adding out of the box. However, v4 (currently in alpha) adds a renderInput method that provides full flexibility over rendering of the input element. I've never tried using a textarea so I have no idea how well that would work, but give it a shot!

adamruzicka:
I'm failing to reproduce

ezr-ondrej:
> This introduces a regression, when I schedule a job from hosts list, I click bulk action, the filter field is not preffiled even thout it says resolved to 1 hosts. When I hit submit, it fails on empty targetting. Prior this patch it works.

Have you updated core lately - it needs a patch in core that Adam submitted, I believe without it, it would behave as you're describing.

ares:
This requires something that was added in Foreman 1.25/2.0 but we still have it in 2.0.5 release, I suppose it won't work without the commit in core. I'll revert this in 2.0-stable, from now on we need to cherrypick commits for 2.0 stable. There's also another fix for autocompletion in master, that relies on core 2.0 changes.

Also the whole issue was introduced in core 2.0 only, so reverting for 1.24 should be reasonable.

PR #458, opened 2020-01-02
Fixes #28616 - Pass needed option to fix concurrency control

adamruzicka:
when batch triggering is enabled

requires:
- [x] https://github.com/theforeman/foreman-tasks/pull/489


adamruzicka:
This has two ACKs, will merge at the end of the day

adamruzicka:
It was a pretty long day, but it is finally in. Thank you @ares & @pondrejk for reviews!

PR #461, opened 2020-01-08
Fixes #28691 - Load symbols when parsing yaml

adamruzicka:
`foreman-cockpit-session` reads a configuration in YAML which uses symbols as
keys. Before this patch, doing so resulted in the 

```
Psych::DisallowedClass: Tried to load unspecified class: Symbol
```
being raised.


ares:
Thanks @adamruzicka, merging. Tests seems to be broken, but not due to this PR.

PR #463, opened 2020-01-14
Refs #28668 - upgrade tfm-builder to v4

sharvit:
This should hopefully fix nightly since we have builder v4 in packaging :crossed_fingers:


tbrisker:
SQL failed due to too many migrations, but that has nothing to do with this PR.

tbrisker:
Thanks @sharvit !

PR #464, opened 2020-01-22
Refs #28668 - remove tfm/env and use tfm/vendor >= 4

sharvit:
It might be the reason https://github.com/theforeman/foreman-packaging/pull/4347 is failing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
-    "@theforeman/env": "^3.3.2",
     "@theforeman/builder": "^4.0.2",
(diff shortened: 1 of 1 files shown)


ekohl:
Out of curiosity, what makes you think that this would help?

sharvit:
> Out of curiosity, what makes you think that this would help?

The tfm/env asks for builder v3 to be there. It should be ignored in packaging AFAIK but the packaging is still a black box for me :/
https://github.com/theforeman/foreman-js/blob/v3.12.0/packages/env/package.json#L26-L28

adamruzicka:
The tests are green, let's get this in

adamruzicka:
Will someone try building the package with this locally or should I do a full release?

PR #465, opened 2020-01-22
Fixes #28826 - Describe connect_by_ip only applies to IPv4

adamruzicka:


app/models/setting/remote_execution.rb
```diff
@@ -39,7 +39,8 @@ def self.default_settings
                N_('SSH Port')),
       self.set('remote_execution_connect_by_ip',
                N_('Should the ip addresses on host interfaces be preferred over the fqdn? '\
-               'It is useful, when DNS not resolving the fqdns properly. You may override this per host by setting a parameter called remote_execution_connect_by_ip.'),
+               'It is useful, when DNS not resolving the fqdns properly. You may override this per host by setting a parameter called remote_execution_connect_by_ip. '\
```
pondrejk: 'It is useful when DNS is not...' imho drop comma and add is. Otherwise ack

adamruzicka: Fixed

PR #468, opened 2020-02-06
Fixes #25144 - Include taxonomies in Job invocation targeting

stejskalleos:
Add Location and Organization to Job invocations, so users will now see in which context job has been run.

**List of changes:**
* Current Organization & Location are displayed in `/job_invocations/new` form
* Organization & Location in which job has been run are now displayed in the Targeting section on Job invocation edit page.

### DIFF
--- a/app/models/targeting.rb
+++ b/app/models/targeting.rb
@@ -11,6 +11,8 @@ class Targeting < ApplicationRecord
 
   belongs_to :user
   belongs_to :bookmark
+  belongs_to :organization, optional: true
+  belongs_to :location, optional: true
 
   has_many :targeting_hosts, :dependent => :destroy
   has_many :hosts, -> { order TargetingHost.table_name + '.id' }, :through => :targeting_hosts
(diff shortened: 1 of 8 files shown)


lzap:
Hey, is there a related hammer CLI patch?

app/models/targeting.rb
```diff
@@ -11,6 +11,8 @@ class Targeting < ApplicationRecord
 
   belongs_to :user
   belongs_to :bookmark
+  belongs_to :organization, optional: true
+  belongs_to :location, optional: true
```
adamruzicka: I know I'm coming a bit late with this, but just now I finally had time to read through this properly.

Why did you choose to (optionally) link targeting to an organization and a location directly instead of using the `Taxonomix` module?

It goes against the pattern established in Foreman where things are not linked directly to orgs & locs, but instead can belong to multiple. Also it lacks expressiveness a bit. If I have any-any context when triggering a job, it does not mean it will consider hosts from really all organizations and locations, because it is still limited by the organizations and locations the user is assigned to and this should probably be somehow recorded.

PR #470, opened 2020-02-12
Fixes #28997 - remove %{puppet_options} from the format_description

amirfefer:
puppet_options user input is optional input, and after moving the `Run Puppet once` button to REX. it runs `Puppet Run Once - SSH Default` task without `puppet_options` so it doesn't populated and therefore presented in the UI as `%{puppet_options}`

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -202,19 +202,14 @@ class JobInvocation < ApplicationRecord
   end
 
   def generate_description
-    key_re = /%\{([^\}]+)\}/
     template_invocation = pattern_template_invocations.first
-    hash_base = Hash.new { |hash, key| hash[key] = "%{#{key}}" }
-    input_hash = template_invocation.input_values.reduce(hash_base) do |h, v|
-      h.update(v.template_input.name => v.value)
+    input_hash = template_invocation.input_values.reduce({}) do |h, v|
+      h.update("%{#{v.template_input.name}}" => v.value)
     end
-    input_hash.update(:job_category => job_category)
-    input_hash.update(:template_name => template_invocation.template.name)
-    description_format.scan(key_re) { |key| input_hash[key.first] }
-    self.description = description_format
-    input_hash.each do |k, v|
-      self.description.gsub!(Regexp.new("%\{#{k}\}"), v || '')
-    end
+    input_hash.update("%{job_category}" => job_category)
+    input_hash.update("%{template_name}" => template_invocation.template.name)
+    input_hash.default = "''"
+    self.description = description_format.gsub(/%{[^}]+}/) { |key| input_hash[key] }
     self.description = self.description[0..(JobInvocation.columns_hash['description'].limit - 1)]
   end
 
--- a/test/unit/job_invocation_test.rb
+++ b/test/unit/job_invocation_test.rb
@@ -91,7 +91,7 @@ class JobInvocationTest < ActiveSupport::TestCase
       it 'handles missing keys correctly' do
         job_invocation.description_format = '%{job_category} - %{missing_key}'
         job_invocation.generate_description
-        job_invocation.description.must_equal "#{job_invocation.job_category} - %{missing_key}"
+        job_invocation.description.must_equal "#{job_invocation.job_category} - ''"
       end
 
       it 'truncates generated description to 255 characters' do


ares:
Would a better fix be that we interpolate that value to empty string if the input is not present? Plus we could change the string a bit. That's what I'd expect, meaning if there was some option, I'd see:

`Run Puppet once with options "--once"`

and this otherwise

`Run Puppet once with options ""`

amirfefer:
The button `Run Puppet once` arrives from this [feature registration ](https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L200), is there an option the send an empty `puppet_options` input through it?

ares:
yes, the value can easily be `nil` if paramter is omitted thourhg API and empty string if it's just hidden in the form (as advanced input).

app/models/job_invocation.rb
```diff
@@ -204,17 +204,13 @@ def output(host)
...
     template_invocation = pattern_template_invocations.first
-    hash_base = Hash.new { |hash, key| hash[key] = "%{#{key}}" }
-    input_hash = template_invocation.input_values.reduce(hash_base) do |h, v|
+    input_hash = template_invocation.input_values.reduce({}) do |h, v|
       h.update(v.template_input.name => v.value)
     end
     input_hash.update(:job_category => job_category)
     input_hash.update(:template_name => template_invocation.template.name)
-    description_format.scan(key_re) { |key| input_hash[key.first] }
+    description_format.gsub!(/(%\{)([^\}]+)(\})/) { |key| input_hash.key?(Regexp.last_match[2]) ? input_hash[Regexp.last_match[2]] : "''" }
```
amirfefer: when an input doesn't exist, `""`  will be added to the `description`
for example, `Run Puppet Once with ${puppet_options}` returns as `Run Puppet Once with ""`

app/views/templates/ssh/puppet_run_once.erb
```diff
@@ -3,7 +3,7 @@ kind: job_template
 name: Puppet Run Once - SSH Default
 model: JobTemplate
 job_category: Puppet
-description_format: 'Run Puppet once with "%{puppet_options}"'
+description_format: 'Run Puppet once'
```
ares: is this still necessary? also note that changes to job templates are supposed to go through [community-templates](https://github.com/theforeman/community-templates/) repo

amirfefer: yes, it's not necessary, thanks

amirfefer:
updated

amirfefer:
rebased

tbrisker:
Also, looks like test failures are related

app/models/job_invocation.rb
```diff
@@ -204,17 +204,13 @@ def output(host)
   def generate_description
     key_re = /%\{([^\}]+)\}/
     template_invocation = pattern_template_invocations.first
-    hash_base = Hash.new { |hash, key| hash[key] = "%{#{key}}" }
-    input_hash = template_invocation.input_values.reduce(hash_base) do |h, v|
+    input_hash = template_invocation.input_values.reduce({}) do |h, v|
       h.update(v.template_input.name => v.value)
     end
     input_hash.update(:job_category => job_category)
     input_hash.update(:template_name => template_invocation.template.name)
```
tbrisker: I think this should fix the tests:
```suggestion
    input_hash.update("job_category" => job_category)
    input_hash.update("template_name" => template_invocation.template.name)
```

app/models/job_invocation.rb
```diff
@@ -204,17 +204,13 @@ def output(host)
...
-    description_format.scan(key_re) { |key| input_hash[key.first] }
+    description_format.gsub!(/(%\{)([^\}]+)(\})/) { |key| input_hash.key?(Regexp.last_match[2]) ? input_hash[Regexp.last_match[2]] : "''" }
     self.description = description_format
```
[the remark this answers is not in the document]
amirfefer: rubocop doesn't like $1 
```
Style/PerlBackrefs: Avoid the use of Perl-style backrefs.
```

amirfefer:
Thanks @tbrisker!
There's an odd regression while creating a new job, happens also in master branch:
image

tbrisker:
Test failure still related, looks like there is a test looking for `%{missing_key}` that needs to be updated

app/models/job_invocation.rb
```diff
@@ -204,17 +204,12 @@ def output(host)
...
-    input_hash.update(:job_category => job_category)
-    input_hash.update(:template_name => template_invocation.template.name)
-    description_format.scan(key_re) { |key| input_hash[key.first] }
-    self.description = description_format
-    input_hash.each do |k, v|
-      self.description.gsub!(Regexp.new("%\{#{k}\}"), v || '')
-    end
+    input_hash.update("job_category" => job_category)
+    input_hash.update("template_name" => template_invocation.template.name)
+    self.description = description_format.gsub(/%{([^}]+)}/) { |key| input_hash.fetch(Regexp.last_match[1], "''") }
```
tbrisker: I just found out a cool feature - [`Hash#default=`](https://ruby-doc.org/core-2.5.1/Hash.html#method-i-default-3D), so we can do:
```suggestion
    input_hash.default = "''"
    self.description = description_format.gsub(/%{([^}]+)}/) { |key| input_hash[Regexp.last_match[1]] }
```

app/models/job_invocation.rb
```diff
@@ -204,17 +204,12 @@ def output(host)
   def generate_description
     key_re = /%\{([^\}]+)\}/
     template_invocation = pattern_template_invocations.first
-    hash_base = Hash.new { |hash, key| hash[key] = "%{#{key}}" }
-    input_hash = template_invocation.input_values.reduce(hash_base) do |h, v|
+    input_hash = template_invocation.input_values.reduce({}) do |h, v|
       h.update(v.template_input.name => v.value)
```
tbrisker: If we set the keys here to the replacement format like:
```suggestion
      h.update("%{#{v.template_input.name}}" => v.value)
```
Then below we can do 
```ruby
self.description = description_format.gsub(/%{[^}]+}/) { |key| input_hash[key] }

File: app/models/job_invocation.rb [hunk 7]
```diff
@@ -204,17 +204,12 @@ def output(host)
   def generate_description
     key_re = /%\{([^\}]+)\}/
```
tbrisker: this isn't used anymore

amirfefer:
Thanks @tbrisker  ! 
Fixed some tests, and addressed your requests :)

amirfefer:
Rebased

PR #471, opened 2020-02-18
Fixes #29048 - cant create job with end time

MariaAga:
In some environments when you create a job invocation with end_time the end_time hash keys would be unsorted which caused the creation to fail.

### DIFF
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -79,8 +79,9 @@ class JobInvocationComposer
       return {} unless ui_params.key?(:triggering)
       trig = ui_params[:triggering]
       keys = (1..5).map { |i| "end_time(#{i}i)" }
-      return trig unless trig.key?(:end_time) && trig[:end_time].keys == keys
-      trig.merge(:end_time => Time.local(*trig[:end_time].values_at(*keys)))
+      values = trig.fetch(:end_time, {}).values_at(*keys)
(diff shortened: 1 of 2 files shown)


MariaAga:
Added a test

MariaAga:
I think the test failures are unrelated now

app/models/job_invocation_composer.rb
```diff
@@ -79,7 +79,7 @@ def triggering
       return {} unless ui_params.key?(:triggering)
       trig = ui_params[:triggering]
       keys = (1..5).map { |i| "end_time(#{i}i)" }
-      return trig unless trig.key?(:end_time) && trig[:end_time].keys == keys
+      return trig unless trig.key?(:end_time) && trig[:end_time].keys.sort == keys
       trig.merge(:end_time => Time.local(*trig[:end_time].values_at(*keys)))
```
adamruzicka: I know I originally suggested the sorting as an easy way out, but now I'm having second thougths about it. How about something like this
```ruby
values = trig.fetch(:end_time, {}).values_at(*keys)
return trig if values.any?(&:nil?)
trig.merge(:end_time => Time.local(*values))
```

MariaAga: Yeah it makes more sense, thanks

adamruzicka:
Thank you @MariaAga !

PR #472, opened 2020-02-18
Fixes #25144 - Display taxonomies on Job invocation pages

stejskalleos:
Add Location and Organization to Job invocations, so users will now see in which context job has been run.

**List of changes:**

- Current Organization & Location are displayed in `/job_invocations/new` form
- Organization & Location in which job has been run are now displayed in the Targeting section on Job invocation edit page.

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -58,6 +58,8 @@ class JobInvocationsController < ApplicationController
                                      .where(:template_invocations => { :job_invocation_id => @job_invocation.id})
     end
     @hosts = resource_base_search_and_page
+    @job_organization = Taxonomy.find_by(id: @job_invocation.task.input[:current_organization_id])
+    @job_location = Taxonomy.find_by(id: @job_invocation.task.input[:current_location_id])
   end
 
   def index
--- a/app/views/job_invocations/_card_target_hosts.html.erb
+++ b/app/views/job_invocations/_card_target_hosts.html.erb
@@ -23,7 +23,6 @@
     <p>
       <%= _('Organization') %>:
       <strong>
-        <%=  %>
         <%= show_job_organization(@job_organization) %>
       </strong>
     </p>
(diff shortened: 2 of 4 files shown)


app/views/job_invocations/_card_target_hosts.html.erb
```diff
@@ -20,6 +20,19 @@
       <% key = job_invocation.targeting.randomized_ordering ? Targeting::RANDOMIZED : Targeting::ORDERED %>
       <%= _('Execution order') %>: <strong><%= Targeting::ORDERINGS[key].downcase %></strong>
     </p>
+    <p>
+      <%= _('Organization') %>:
+      <strong>
+        <%=  %>
```
adamruzicka: This should go away

app/views/job_invocations/_form.html.erb
```diff
@@ -49,6 +49,20 @@
...
+      </label>
+      <div class="col-md-4">
+        <p class="form-control-static">
+          <%= show_job_organization(Organization.current) %>
+          <br>
+          <%= show_job_location(Location.current) %>
+        </p>
+      </div>
+    </div>
+
```
adamruzicka: Do we want to show it in the form? Currently I don't think it adds much value, the user can always look at the org/loc switcher in the top bar.

stejskalleos: That's true, but sometimes you will quickly forget that you changed loc/org five minutes ago and this could be such a small "reminder" before the user hit the run button.

app/controllers/job_invocations_controller.rb
```diff
@@ -58,6 +58,8 @@ def show
                                      .where(:template_invocations => { :job_invocation_id => @job_invocation.id})
     end
     @hosts = resource_base_search_and_page
+    @job_organization = Taxonomy.find_by(id: @job_invocation.task.input[:current_organization_id])
+    @job_location = Taxonomy.find_by(id: @job_invocation.task.input[:current_location_id])
```
adamruzicka: How about
```ruby
@job_organization = Organization.find(@job_invocation.task.input[:current_organization_id])
@job_location = Location.find(@job_invocation.task.input[:current_location_id])
```

adamruzicka: Ah, that behaves a bit differently when the record cannot be found, thanks for reminding me.

app/views/job_invocations/_form.html.erb
stejskalleos: I reverted it back, didn't realize that _form is also used in rerun action.

adamruzicka:

adamruzicka:
Thank you @stejskalleos !

PR #475, opened 2020-03-13
Adding a new slot on job_invocation detail page

fellipeh:
Creating these Slot, we provide to all developers a way to create how many tab/tab content as they required.

We on OAMG-LEAPP [1] team need to show all in-place upgrade reports to the sysadmin. And we think the best place to show it is on Job Invocation Details page. We will have a plugin [2] to fill these slots after these PR and these https://github.com/theforeman/foreman/pull/7509  been approved and merged.

[1] - https://github.com/oamg/leapp   &  https://github.com/oamg/leapp-repository
[2] - https://github.com/oamg/foreman_leapp

Signed-off-by: Fellipe Henrique <fpedrosa@redhat.com>


fellipeh:
After discussion with Ondrej, it's not possible to use slot/fill to change tab.. so, closing these PR...

PR #477, opened 2020-03-14
Fixes #29347 - enable and fix more cops

mmoll:


.rubocop.yml
```diff
@@ -2,114 +2,139 @@ require:
...
-  Enabled: true
-  Exclude:
-    - 'lib/foreman_remote_execution_core/**'
-    - 'node_modules/**'
+Bundler/OrderedGems:
+  Enabled: false
+
+Layout/ArgumentAlignment:
+  EnforcedStyle: with_fixed_indentation
+  IndentationWidth: 2
```
adamruzicka: Is this something that's being enabled foreman-wide?

mmoll: This is enabled in core. Basically I copied over the contents of core's rubocop.yml...

adamruzicka: I'm not very fond of this one, but if let's stick with what core uses

mmoll: I concur with that, but hope after enabling/configuring a few more multi line cops in the future it will become nicer overall. :nail_care:

adamruzicka:
Thank you @mmoll !

PR #479, opened 2020-03-23
Fixes #29392 - Make RunHostJob action subscribable

stejskalleos:


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -47,8 +47,13 @@ def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExe
...
                                  .merge(additional_options)
 
-        plan_delegated_action(proxy, provider.proxy_action_class, action_options)
-        plan_self
+        proxy_action = plan_delegated_action(proxy, provider.proxy_action_class, action_options)
+        plan_self(proxy_action_output: proxy_action)
+      end
+
+      def run
+        self.input[:proxy_action_output]
```
adamruzicka: What does this line do?

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -47,8 +47,13 @@ def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExe
...
+        self.output.update(success: true)
```
adamruzicka: Just because we are in a run phase, it doesn't mean the previous step was a success. Maybe using `exit_status.to_s == '0'` instead of hardcoded `true` would be better

adamruzicka:
Since we now don't need ordering in run phase, is this still needed?

stejskalleos:
Not needed anymore, we don't use run phases.

PR #482, opened 2020-03-25
Fixes #21358 - cancelled state for delayed jobs

adamruzicka:


adamruzicka:

adamruzicka:

adamruzicka:

adamruzicka:
Keeping previous status probably makes more sense

adamruzicka:
Oh, ok, I see now. If a delayed action is cancelled before it is started, statuses of hosts are not touched at all. And if it already starts running on a host and then gets cancelled, then it is considered to be a regular failure.

ares:
So cancelling running job would make the host red? Does that also apply for recurring/delayed runs?

adamruzicka:
It all depends on when you cancel the job. Under the hood it only depends on whether there is already a task for the given host in that job invocation. If it already exists, then the host will be red, otherwise it its status will not be changed

ares:
And I assume there is a task for something that's scheduled to run in future. Which may be a problem right? Perhaps we could extend `relevant?` method to return false in case the task was cancelled.

adamruzicka:
When something is planned for the future, then there is only the parent task (RunHostsJob). Per-host tasks get created at execution time so that shouldn't be a problem

ares:
Oh, ok got it now, I didn't realize what task we're talking about

app/views/job_invocations/_tab_hosts.html.erb
```diff
@@ -25,11 +25,12 @@
     </thead>
 
     <tbody>
+    <% parent_task = @job_invocation.task %>
```
ezr-ondrej, line 28: Do we need this variable?

app/views/job_invocations/_host_status_td.html.erb
```diff
@@ -1 +1 @@
-<td class="host_status" id="<%= dom_id(host) %>-status" data-refresh_required="<%= task.nil? || task.pending? ? 'true' : '' %>" data-id="<%= host.id %>"><%= template_invocation_status(task) %></td>
+<td class="host_status" id="<%= dom_id(host) %>-status" data-refresh_required="<%= task.nil? || task.pending? ? 'true' : '' %>" data-id="<%= host.id %>"><%= template_invocation_status(task, parent_task) %></td>
```
ezr-ondrej, line 1: could we just do this?
```suggestion
<td class="host_status" id="<%= dom_id(host) %>-status" data-refresh_required="<%= task.nil? || task.pending? ? 'true' : '' %>" data-id="<%= host.id %>"><%= template_invocation_status(task, job_invocation.task) %></td>
```

app/views/job_invocations/_tab_hosts.html.erb
ezr-ondrej: if we insist on the variable, we would need to add it to the
https://github.com/theforeman/foreman_remote_execution/blob/734acd451fac04ab0ebb9c348271bc66c69df3b8/app/views/job_invocations/show.js.erb#L14

ezr-ondrej:
Idea for some followup: it would be great to add it to the job_invocations table as well.
This is good for now though :+1:

adamruzicka:
By job invocations table you mean the job invocation index page?

ezr-ondrej:
> By job invocations table you mean the job invocation index page?

Yes, exactly :) but I think we should be redesigning it to something task-like right?

adamruzicka:
I won't deny the page could use a slight refresh

PR #485, opened 2020-04-02
Fixes #29465 - hide hidden values in job invocation

ares:


ares:
requires https://github.com/theforeman/foreman/pull/7559

adamruzicka:
Could you also take a look at how we generate the description? If someone forgets to set the description format, the input would get shown in job's name
image

adamruzicka:
Thank you @ares !

PR #490, opened 2020-05-13
Fixes #29783 - Allow overriding proxy selector in jobs

adamruzicka:
A Satellite 6.8 feature requires a support from remote execution to be able to run a job from the proxy co-located on the same machine as foreman/satellite. The feature is downstream only, so we can rely on Katello being there and we can use Katello's code to determine which proxy is the right one. However this creates this weird situation where we have a downstream only feature and rex depends on katello. I'd rather avoid less reliable methods of determining the proxy like hostname comparison.

I suggest we get this in as-is now and shortly after 2.1 is out we introduce a new feature to the smart proxy, which will mark the proxy as being co-located on the same machine as foreman/satellite and once it gets in, we change the marked code here to use that marker instead of whatever katello uses.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -29,14 +29,15 @@ module Actions
 
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
 
+        provider = template_invocation.template.provider
+        proxy_selector = provider.required_proxy_selector_for(template_invocation.template) || proxy_selector
+
         provider_type = template_invocation.template.provider_type.to_s
         proxy = determine_proxy!(proxy_selector, provider_type, host)
 
         renderer = InputTemplateRenderer.new(template_invocation.template, host, template_invocation)
         script = renderer.render
         raise _('Failed rendering template: %s') % renderer.error_message unless script
-
-        provider = template_invocation.template.provider
 
         additional_options = { :hostname => provider.find_ip_or_hostname(host),
                                :script => script,
--- a/app/services/default_proxy_proxy_selector.rb
+++ b/app/services/default_proxy_proxy_selector.rb
@@ -1,6 +1,7 @@
-class InternalProxyProxySelector < ::RemoteExecutionProxySelector
+class DefaultProxyProxySelector < ::RemoteExecutionProxySelector
   def initialize
     # TODO: Remove this once we have a reliable way of determining the internal proxy without katello
+    # Tracked as https://projects.theforeman.org/issues/29840
     raise _('Internal proxy selector can only be used if Katello is enabled') unless defined?(::Katello)
     super
   end
--- a/app/services/default_proxy_proxy_selector.rb
+++ b/app/services/default_proxy_proxy_selector.rb
@@ -3,6 +3,7 @@ class DefaultProxyProxySelector < ::RemoteExecutionProxySelector
     # TODO: Remove this once we have a reliable way of determining the internal proxy without katello
     # Tracked as https://projects.theforeman.org/issues/29840
     raise _('Internal proxy selector can only be used if Katello is enabled') unless defined?(::Katello)
+
     super
   end
 
--- a/app/services/internal_proxy_proxy_selector.rb
+++ b/app/services/internal_proxy_proxy_selector.rb
@@ -1,0 +1,16 @@
+class InternalProxyProxySelector < ::RemoteExecutionProxySelector
+  def initialize
+    # TODO: Remove this once we have a reliable way of determining the internal proxy without katello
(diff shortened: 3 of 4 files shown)


adamruzicka:
@ehelms, @ekohl I would appreciate your opinions on the approach I took and the direction I would like to push this in in the near future.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -29,15 +29,16 @@ def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExe
 
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
 
+        provider = template_invocation.template.provider
+        proxy_selector = provider.required_proxy_selector_for(template_invocation.template) || proxy_selector
```
ekohl, line 33: I think it'd be better to make the default in the class nil if something else is going to override it later. Otherwise it's a bit of an odd surprise to the API consumer.

Or the `ProxySelector` should have a way to filter based on what the template provided.

adamruzicka: This action is only planned from one place and the selector is explicitly provided, so the default is in fact never used. This is done so all sub-tasks in a batch share the same selector which keeps some local state. However, in here we need to override the selector on a per-template (and therefore per-host) basis so it felt wrong to make the decision which selector to use and pass down at the parent task level.

That's the reason why I ended up with a somewhat surprising interface where a function may choose not to use some of its arguments based on another argument.

I considered having `ProxySelector` filter based on a template but that didn't feel all that great at the time. I'd say the selector's job is to select a proxy, it shouldn't need to know anything about templates.

I'll sleep on it and see if I can come up with something better

app/services/internal_proxy_proxy_selector.rb
```diff
@@ -0,0 +1,16 @@
+class InternalProxyProxySelector < ::RemoteExecutionProxySelector
```
ehelms, line 1: Get's us into nomenclature discussion, but I tend to like the `default` or `primary` or possibly even `master` terminology better. Since the smart proxy here is not internal to the system or the deployment (it is still running on a public interface).

app/services/internal_proxy_proxy_selector.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize
+    # TODO: Remove this once we have a reliable way of determining the internal proxy without katello
+    raise _('Internal proxy selector can only be used if Katello is enabled') unless defined?(::Katello)
+    super
+  end
+
+  def available_proxies(host, provider)
+    # TODO: Once we have a internal proxy marker/feature on the proxy, we can
```
ehelms, line 9: Is there a Foreman Redmine to add this we can reference here?

app/services/internal_proxy_proxy_selector.rb
```diff
@@ -0,0 +1,16 @@
+class InternalProxyProxySelector < ::RemoteExecutionProxySelector
+  def initialize
+    # TODO: Remove this once we have a reliable way of determining the internal proxy without katello
+    raise _('Internal proxy selector can only be used if Katello is enabled') unless defined?(::Katello)
```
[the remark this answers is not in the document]
ehelms: A naive approach would be to use the proxy with ID=1 as its assumed in nearly all installation cases that a smart proxy will get installed with the server or that the first one is the primary one. But I'm OK with this being a safer route given the workflow it's attempting to support.

app/services/internal_proxy_proxy_selector.rb
adamruzicka: Not yet, I first wanted to talk about if having such a marker is the way to go.

app/services/internal_proxy_proxy_selector.rb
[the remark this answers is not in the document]
adamruzicka: Given it is only a temporary solution until we have another way of marking the proxy, I'd leave it as is.

app/services/internal_proxy_proxy_selector.rb
adamruzicka: Katello uses `default`, guess I'll stick with it

app/services/default_proxy_proxy_selector.rb
```diff
@@ -0,0 +1,18 @@
+class DefaultProxyProxySelector < ::RemoteExecutionProxySelector
+  def initialize
+    # TODO: Remove this once we have a reliable way of determining the internal proxy without katello
```
ekohl, line 3: Internal proxy is IMHO a concept we should not introduce in Foreman and I want to kill in Katello.

adamruzicka:
Thank you everyone

PR #492, opened 2020-05-15
Fixes #29843 - SSH known_deletion failure is non-fatal

lzap:
We should probably not rollback whole deletion transaction. Since the entry is missing, there is nothing we can do about it.

### DIFF
--- a/app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
+++ b/app/models/concerns/foreman_remote_execution/orchestration/ssh.rb
@@ -13,6 +13,9 @@ module ForemanRemoteExecution
       proxy = ::SmartProxy.find(proxy_id)
       begin
         proxy.drop_host_from_known_hosts(target)
+      rescue RestClient::ResourceNotFound => e
+        # ignore 404 when known_hosts entry is missing or the module was not enabled
+        Foreman::Logging.exception "Proxy failed to delete SSH known_hosts for #{name}, #{ip}", e, :level => :error
       rescue => e
         Rails.logger.warn e.message
         return false


adamruzicka:
I don't think it is the entry. If the entry is not there, the proxy returns 204. It looks like the api is not there at all, which would hint towards them having too old remote execution ssh (the proxy plugin).

If someone runs a combination of incompatible versions, is just logging an error an continuing as if nothing happened the right approach?

ezr-ondrej:
I'd say in this case it's the correct approach. There is no real harm done if this fails, right?
Although we lose error reports in cases this really doesn't work in correct setup.

lzap:
> I'd say in this case it's the correct approach. There is no real harm done if this fails, right?

We actually do the very same in `orchestration/dhcp.rb` in core. This should not harm. Otherwise, people are having hard time recovering from this and will be creating tickets and BZs on how to get out of this situation not being able to delete a host.

adamruzicka:
> There is no real harm done if this fails, right?

Except if they reprovision the host, they will get known hosts key errors if they try to run rex against that host again

lzap:
> Except if they reprovision the host, they will get known hosts key errors if they try to run rex against that host again

Good call, what if I constrain this only to 404? The REST endpoint should return 404 when there is nothing. That way, generic errors would still trigger the rollback.

adamruzicka:
Let's go with that

lzap:
Full disclosure - I haven't tested that. But this should be pretty safe to do, unless I made some typo.

adamruzicka:
Thank you @lzap !

PR #493, opened 2020-05-18
Fixes #28723 - Job Invocations page slow loading for non-admin users

stejskalleos:
**Issue**
`/job_invocations/:id` is extremely slow for non-admin users - loading data for Targeting hosts table takes a lot of time and blocks page render.

**Fix**
The issue is not just at this page, but in the Foreman and in the way how permissions are implemented. So I came with different fix - loading data via API and not blocking page rendering.

The issue with slow data loading is still there, but now users can display the page without waiting for  Targeting hosts data. Plus I created React component and removed some legacy JS.


webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,92 @@
...
+      );
+    }
+
+    return (null);
+  };
+
+  const getHostsData = () => {
+    const apiUrl = `/job_invocations/${jobInvocationId}?format=json`;
+
+    API.get(apiUrl)
```
laviro, line 29: can you move this logic into a redux action?

webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,92 @@
...
+import PropTypes from 'prop-types';
+import HostItem from './HostItem/HostItem.js';
+
+const TargetingHosts = (props) => {
+  const { jobInvocationId } = props;
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState(false);
+  const [hosts, setHosts] = useState([]);
+
+  const renderLoading = () => {
```
laviro, line 12: IMO this can be moved into a separate component, e.g `LoadingState`

app/views/job_invocations/show.json.erb
```diff
@@ -0,0 +1,4 @@
+{
+  "autoRefresh": "<%= @auto_refresh %>",
+  "hosts": <%= targeting_hosts(@job_invocation, @hosts).to_json.html_safe %>
+}
```
laviro, line 4: I think this could be moved to the controller where you can use 
```rb
render json: {
        autoRefresh: ...
        hosts: ...,
      }, status: :ok
```
and than in the API call from the UI, you don't need to set `format=json` in the url

webpack/react_app/components/TargetingHosts/HostStatus/HostStatus.js
```diff
@@ -0,0 +1,31 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+const HostStatus = (props) => {
+  const { status } = props;
+
+  switch (status) {
+    case 'cancelled':
+      return (<Fragment><span className="pficon pficon-warning-triangle-o"></span> cancelled</Fragment>);
```
laviro, line 9: you can use the `Icon` component from patternfly

webpack/react_app/components/TargetingHosts/HostActions/HostActions.js
```diff
@@ -0,0 +1,38 @@
...
+    return (<a href={actions[0].path} className="btn btn-sm btn-default">{actions[0].name}</a>);
+  }
+  return (
+      <div className="btn-group">
+        <span className="btn btn-sm btn-default">
+          <a href={actions[0].path}>
+            {actions[0].name}
+          </a>
+        </span>
+        <a className="btn btn-default btn-sm dropdown-toggle" data-toggle="dropdown" href="#">
```
laviro, line 21: you can use the `Dropdown` component from patternfly-react

webpack/index.js
```diff
@@ -4,11 +4,19 @@ import { mount, registerReducer } from 'foremanReact/common/MountingService';
...
 import rootReducer from './react_app/redux/reducers';
 
 componentRegistry.register({
   name: 'JobInvocationContainer',
   type: JobInvocationContainer,
 });
+componentRegistry.register({ name: 'TargetingHosts', type: TargetingHosts });
+componentRegistry.register({ name: 'HostStatus', type: HostStatus });
+componentRegistry.register({ name: 'HostActions', type: HostActions });
+componentRegistry.register({ name: 'HostItem', type: HostItem });
```
laviro, line 20: nit: maybe we should save all of the components with `{ name, type }` pairs in a `components` file
and than iterate over that array with the register function?

app/views/job_invocations/show.json.erb
stejskalleos: yeah, but `JobInvocationController#show` is shared for two actions: GET html view for user and API call for getting data for this component. So I'd have to do something like this:
```
respond_to do |format|
  format.html do
    ...
  end
   format.text do
    render :json ....
   end
end
``` 
which is IMHO basically same as the how I did it - just add `show.json.erb` and have `format` param in URL.

stejskalleos:
Waiting for https://github.com/theforeman/foreman_remote_execution/pull/497 to fix issue with hound

webpack/react_app/components/TargetingHosts/TargetingHosts.js
stejskalleos: Is it necessary? Honestly don't have much experience (or zero) with Redux, and I don't see any bonus in connecting this small component to Redux. Can we keep it like that for now?

laviro: I think we already using redux in this project, you can even reuse the API middleware from core, and reduce more boilerplate.
If you want we can work on it together offline

stejskalleos:
@laviro thanks for review, PR updated:
* Component connected to Redux
* Issue with Hound have been fixed in https://github.com/theforeman/foreman_remote_execution/pull/499.

mmoll:

webpack/react_app/redux/reducers/TargetingHosts/index.js
```diff
@@ -0,0 +1,23 @@
...
+  hosts: [],
+});
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case TARGETING_HOSTS_SUCCESS:
+      return state
+        .set('loading', false)
+        .set('hosts', action.payload.data.hosts)
+        .set('refresh', action.payload.data.autoRefresh === 'true');
```
MariaAga, line 17: can you use `merge` instead of set here?

webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,73 @@
...
+import { getTargetingHostsAction } from '../../redux/actions/TargetingHosts';
+import {
+  selectLoadingState,
+  selectErrorState,
+  selectHostsState,
+  selectRefreshState,
+} from '../../redux/selectors/TargetingHosts';
+
+const TargetingHosts = props => {
+  const { jobInvocationId } = props.data;
```
MariaAga, line 15: Can you use `const TargetingHosts = ({data}) => {` instead?

webpack/react_app/components/TargetingHosts/HostStatus/HostStatus.js
```diff
@@ -0,0 +1,60 @@
+import React, { Fragment } from 'react';
+import { Icon } from 'patternfly-react';
+import PropTypes from 'prop-types';
+
+const HostStatus = props => {
+  const { status } = props;
```
MariaAga, line 6: Can you use `const HostStatus = ({status}) => {` instead?

webpack/react_app/redux/actions/TargetingHosts/index.js
```diff
@@ -0,0 +1,12 @@
+import API from 'foremanReact/API';
+
+import { TARGETING_HOSTS_SUCCESS, TARGETING_HOSTS_ERROR } from '../../consts';
+
+export const getTargetingHostsAction = apiUrl => async dispatch => {
+  try {
+    const result = await API.get(apiUrl);
```
MariaAga, line 7: can you use the middleware for this? `import { get } from 'foremanReact/redux/API'`

webpack/react_app/components/TargetingHosts/HostActions/HostActions.js
```diff
@@ -0,0 +1,34 @@
...
+    return '';
+  }
+
+  if (actions.length === 1) {
+    return <Button href={actions[0].path}>{actions[0].name}</Button>;
+  }
+  return (
+    <SplitButton href={actions[0].path} title={actions[0].name} id="">
+      {actions.slice(1, actions.length).map(link => (
+        <MenuItem key={link.path} href={link.path}>
```
MariaAga, line 18: This can be switched with `/react_app/components/common/ActionButtons/ActionButtons.js`

webpack/react_app/components/TargetingHosts/HostStatus/HostStatus.js
```diff
@@ -0,0 +1,60 @@
...
+
+  switch (status) {
+    case 'cancelled':
+      return (
+        <Fragment>
```
MariaAga, line 11: Fragment is only needed when there is more than one child element this and the next ones can be removed :)

webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,73 @@
...
+      );
+    }
+  }, [dispatch, hosts, jobInvocationId, refresh, loading]);
+
+  if (error) {
+    return (
+      <div className="alert alert-danger">
+        <span className="pficon pficon-error-circle-o " />
+        <span className="text">
+          There was an error while updating the status, try refreshing the page.
```
MariaAga, line 41: Needs translation

webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  return (
+    <LoadingState loading={loading}>
+      <table className="table table-bordered table-striped table-hover ">
+        <thead>
+          <tr>
+            <th>Host</th>
+            <th>Status</th>
+            <th>Actions</th>
```
MariaAga, line 54: Also missing translations

webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,82 @@
...
+  const refresh = useSelector(state => selectRefreshState(state));
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (refresh === true) {
+      setTimeout(
+        () =>
+          dispatch(
+            getTargetingHostsAction(
+              `/job_invocations/${jobInvocationId}?format=json`
```
MariaAga, line 29: This doesn't include the pagination info, maybe you can get the url with:
```js
  import getURI from 'foremanReact/common/urlHelpers`
  const url = getURI();
  url.addQuery('format', 'json');
  return `${url.pathname()}${url.search()}`;
```
 and then the component won't need the `jobInvocationId`

webpack/react_app/components/TargetingHosts/HostItem/HostItem.js
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  return (
+    <tr>
+      <td className="host_name">{renderName()}</td>
+      <td className="host_status">
+        <HostStatus status={status} />
+      </td>
+      <td className="host_actions">
+        <ActionButtons buttons={[...actions]} />
```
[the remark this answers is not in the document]
stejskalleos: It must be like that because of error:
```
ImmutableError {name: "MyError", message: "The shift method cannot be invoked on an Immutable data structure."}
```

stejskalleos:
Waiting for https://github.com/theforeman/foreman/pull/7717 to be merged.

app/helpers/remote_execution_helper.rb
```diff
@@ -16,41 +16,50 @@ def host_counter(label, count)
...
-          icon_text('error-circle-o', _('failed'), :kind => 'pficon')
+          'error'
         when 'cancelled'
-          icon_text('warning-triangle-o', _('cancelled'), :kind => 'pficon')
+          'cancelled'
         when 'success'
-          icon_text('ok', _('success'), :kind => 'pficon')
+          'success'
         else
           task.result
```
adamruzicka, line 36: This could be simplified to "if result is warning return error, otherwise return the result". The part above could be simplified as well

app/helpers/remote_execution_helper.rb
```diff
@@ -237,4 +246,23 @@ def load_template_from_task(template_invocation, target)
 
     task.execution_plan.actions[1].try(:input).try(:[], 'script')
   end
+
+  def targeting_hosts(job_invocation, hosts)
+    response = []
+
+    hosts.each do |host|
```
adamruzicka, line 253: How about not mutating data in place and using `map` instead?

webpack/react_app/components/TargetingHosts/HostItem/HostItem.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ActionButtons } from 'foremanReact/components/common/ActionButtons/ActionButtons';
+import HostStatus from '../HostStatus/HostStatus';
+
+const HostItem = ({ name, link, status, actions }) => {
+  const renderName = () => {
+    if (link) {
+      return <a href={link}>{name}</a>;
```
ares, line 9: it may be a good thing to have css class/html id added to everything that is clickable, so we can write tests against it

webpack/react_app/components/TargetingHosts/HostStatus/HostStatus.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { Icon } from 'patternfly-react';
+import PropTypes from 'prop-types';
+
+const HostStatus = ({ status }) => {
+  switch (status) {
+    case 'cancelled':
+      return (
+        <div>
+          <Icon type="pf" name="warning-triangle-o" /> cancelled
```
ares, line 10: are these status names extracted for translation somehow?

webpack/react_app/components/TargetingHosts/HostItem/HostItem.js
stejskalleos: I added `id` selector to parent `<tr>` element so other child `<td>` can be easily selected as well.

```
<tr id="targeting-host-zzzz113.example.com">
...
</tr>
```

webpack/react_app/components/TargetingHosts/TargetingHosts.js
```diff
@@ -0,0 +1,77 @@
...
+      dispatch(stopInterval(TARGETING_HOSTS));
+    };
+  }, [dispatch]);
+
+  useEffect(() => {
+    if (autoRefresh === 'false') {
+      dispatch(stopInterval(TARGETING_HOSTS));
+    }
+  }, [autoRefresh, dispatch]);
+
```
MariaAga, line 35: Can this be moved to an `index.js` file and add a test for this file?

webpack/react_app/components/TargetingHosts/components/ApiError.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const ApiError = () => (
+  <div className="alert alert-danger">
```
MariaAga, line 5: Can you use the `<Alert type='error'>` component from Patternfly instead?

webpack/react_app/components/TargetingHosts/components/ApiError.js
```diff
@@ -0,0 +1,15 @@
...
+    <span className="pficon pficon-error-circle-o " />
```
MariaAga, line 6: Can you use the Icon component here?

stejskalleos:
> Just 2 small comments but other than those LGTM :)

Updated, now using `<Alert>` component

webpack/react_app/components/TargetingHosts/TargetingHostsActions.js
```diff
@@ -0,0 +1,15 @@
...
+  const url = getURI().addQuery('format', 'json');
+  return {
+    type: API_OPERATIONS.GET,
+    interval: 1000,
+    payload: {
+      key: TARGETING_HOSTS,
+      url: `${url.pathname()}${url.search()}`,
+    },
+  };
+};
```
laviro, line 15:
```suggestion
export const getData = () => withInterval(get({key: TARGETING_HOSTS, url}), 1000)
```

webpack/react_app/components/TargetingHosts/__tests__/HostItem.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { testComponentSnapshotsWithFixtures } from '@theforeman/test';
+import HostItem from '../components/HostItem';
+
+const fixtures = {
+  'renders HostItem': {
+    name: 'Host1',
+    link: '/host1',
+    status: 'success',
+    actions: [],
```
laviro, line 10: can you move it into a fixtures file, the benefit is that you can reuse those props all over,
saw that you used them in other tests

webpack/react_app/components/TargetingHosts/__tests__/HostItem.test.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+};
+
+jest.mock('foremanReact/components/common/ActionButtons/ActionButtons', () => ({
```
laviro, line 14: can you create the mock file as you created the other mocks? thanks

webpack/react_app/components/TargetingHosts/__tests__/TargetingHosts.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+const fixtures = {
+  'renders TargetingHosts': {
+    status: '',
+    items: [
+      {
+        name: 'host',
+        link: '/link',
+        status: 'success',
+        actions: [],
```
laviro, line 12: those fixtures could be also move to the `TargetingHosts.fixtures.js` file and reused un other tests,
here they could be consumed as: 

```js
'renders TargetingHosts': {
￼    status: '',
￼    items: items,
```

webpack/react_app/components/TargetingHosts/__tests__/TargetingHosts.test.js
```diff
@@ -0,0 +1,40 @@
...
+      },
+    ],
+  },
+  'renders TargetingHosts with loading': {
+    status: '',
+    items: [],
+  },
+};
+
+jest.mock('foremanReact/constants', () => ({
```
laviro, line 33: can you also move this one to a mock file as you created the others? thanks

webpack/react_app/components/TargetingHosts/components/HostItem.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ActionButtons } from 'foremanReact/components/common/ActionButtons/ActionButtons';
+import HostStatus from './HostStatus';
+
+const HostItem = ({ name, link, status, actions }) => {
+  const renderName = () => {
```
laviro, line 7: should it be renamed to `hostLink` ? also it doesn't have to be a function,
can be decleared as 

```js
const hostLink = link ? <a href={link}>{name}</a> : <a href="#" className="disabled">{name}</a>
```

webpack/react_app/components/TargetingHosts/components/HostStatus.js
```diff
@@ -0,0 +1,59 @@
...
+        <div>
+          <Icon type="pf" name="error-circle-o" /> {__('failed')}
+        </div>
+      );
+    case 'error':
+      return (
+        <div>
+          <Icon type="pf" name="error-circle-o" /> {__('failed')}
+        </div>
+      );
```
laviro, line 43: because they are the same I think you can do something like:
```suggestion
    case 'warning':
￼    case 'error':
￼      return (
￼        <div>
￼          <Icon type="pf" name="error-circle-o" /> {__('failed')}
￼        </div>
￼      );
```

adamruzicka:
When we have the table in react, could we change pages in that table without doing full page reloads?

MariaAga:
@adamruzicka Yeah there is a `Pagination` component in React

adamruzicka:
Let's leave it for a followup PR

PR #494, opened 2020-05-18
Fixes #29867 - API - Show job status for each host on JobInvocation detail

stejskalleos:
Hammer PR: https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/32

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -18,11 +18,20 @@ module Api
 
       api :GET, '/job_invocations/:id', N_('Show job invocation')
       param :id, :identifier, :required => true
+      param :host_status, String, required: false, allow_blank: true, desc: N_('Show Job status for the hosts.')
       def show
         @hosts = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
         @template_invocations = @job_invocation.template_invocations
                                                .where(host: @hosts)
-                                               .includes(:input_values)
+                                               .includes([:input_values, :run_host_job_task])
+                                               .to_a
+
+        if params[:host_status]
+          @hosts.each do |host|
+            template_invocation = @template_invocations.find { |t| t.host_id == host.id }
+            host.template_invocation_status = template_invocation_status(template_invocation)
+          end
+        end
       end
 
       def_param_group :job_invocation do
@@ -207,6 +216,17 @@ module Api
       # Do not try to scope JobInvocations by taxonomies
       def parent_scope
         resource_class.where(nil)
+      end
+
+      def template_invocation_status(template_invocation)
+        task = template_invocation.try(:run_host_job_task)
+        parent_task = @job_invocation.task
+
+        return(parent_task.result == 'cancelled' ? 'cancelled' : 'N/A') if task.nil?
+        return task.state if task.state == 'running' || task.state == 'planned'
+        return 'error' if task.result == 'warning'
+
+        task.result
       end
     end
   end
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -23,14 +23,11 @@ module Api
         @hosts = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
         @template_invocations = @job_invocation.template_invocations
                                                .where(host: @hosts)
-                                               .includes([:input_values, :run_host_job_task])
-                                               .to_a
+                                               .includes(:input_values)
 
         if params[:host_status]
-          @hosts.each do |host|
-            template_invocation = @template_invocations.find { |t| t.host_id == host.id }
-            host.template_invocation_status = template_invocation_status(template_invocation)
-          end
+          template_invocations = @template_invocations.includes(:run_host_job_task).to_a
+          @host_statuses = Hash[template_invocations.map { |ti| [ti.host_id, template_invocation_status(ti)] }]
         end
       end
 
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -2,6 +2,8 @@ module ForemanRemoteExecution
   module HostExtensions
     def self.prepended(base)
       base.instance_eval do
+        attr_accessor :template_invocation_status
+
         has_many :targeting_hosts, :dependent => :destroy, :foreign_key => 'host_id'
         has_many :template_invocations, :dependent => :destroy, :foreign_key => 'host_id'
         has_one :execution_status_object, :class_name => 'HostStatus::ExecutionStatus', :foreign_key => 'host_id', :dependent => :destroy
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -2,8 +2,6 @@ module ForemanRemoteExecution
   module HostExtensions
     def self.prepended(base)
       base.instance_eval do
-        attr_accessor :template_invocation_status
-
         has_many :targeting_hosts, :dependent => :destroy, :foreign_key => 'host_id'
         has_many :template_invocations, :dependent => :destroy, :foreign_key => 'host_id'
         has_one :execution_status_object, :class_name => 'HostStatus::ExecutionStatus', :foreign_key => 'host_id', :dependent => :destroy
--- a/app/views/api/v2/job_invocations/main.json.rabl
+++ b/app/views/api/v2/job_invocations/main.json.rabl
@@ -24,7 +24,7 @@ child :targeting do
 
     if params[:host_status]
       node :job_status do |host|
-        host.template_invocation_status
+        @host_statuses[host.id]
       end
     end
   end
--- a/app/views/api/v2/job_invocations/main.json.rabl
+++ b/app/views/api/v2/job_invocations/main.json.rabl
@@ -21,6 +21,12 @@ child :targeting do
 
   child @hosts do
     extends 'api/v2/hosts/base'
+
+    if params[:host_status]
+      node :job_status do |host|
+        host.template_invocation_status
+      end
+    end
   end
 end
 


app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -18,6 +18,7 @@ def index
 
       api :GET, '/job_invocations/:id', N_('Show job invocation')
       param :id, :identifier, :required => true
+      param :host_status, String, required: false, allow_blank: true, desc: N_('Show Job status for the hosts.')
```
ofedoren, line 21: Do we actually need this? I mean, it's just a string in the response. I think it could be just added.

stejskalleos: Yop we need it, because of this: https://projects.theforeman.org/issues/28723.

When non-admin user is getting data for too many hosts, checking permissions and getting data takes a lot of time.

app/helpers/remote_execution_helper.rb
```diff
@@ -38,6 +38,31 @@ def template_invocation_status(task, parent_task)
     end
   end
 
+  def template_invocation_api_status(task, parent_task)
+    if task.nil?
```
adamruzicka, line 42: Could we turn this into a case using something like `task&.status`?

app/helpers/remote_execution_helper.rb
```diff
@@ -38,6 +38,31 @@ def template_invocation_status(task, parent_task)
...
+      if parent_task.result == 'cancelled'
+        'cancelled'
+      else
+        'N/A'
+      end
+    elsif task.state == 'running'
+      'running'
+    elsif task.state == 'planned'
+      'planned'
```
adamruzicka, line 51: this could be reduced to "if task is running or planned, return its state"

app/helpers/remote_execution_helper.rb
```diff
@@ -38,6 +38,31 @@ def template_invocation_status(task, parent_task)
...
+      case task.result
+        when 'warning', 'error'
+          'error'
+        when 'cancelled'
+          'cancelled'
+        when 'success'
+          'success'
+        else
+          task.result
+      end
```
adamruzicka, line 62: This could be reduced to "if result is warning, return error, otherwise just return the result"

app/views/api/v2/job_invocations/main.json.rabl
```diff
@@ -21,6 +21,14 @@ child :targeting do
 
   child :hosts do
     extends 'api/v2/hosts/base'
+
+    if params[:host_status]
+      node(:job_status) do |host|
+        template_invocation = @job_invocation.template_invocations.find { |template_inv| template_inv.host_id == host.id }
+        task = template_invocation.try(:run_host_job_task)
```
adamruzicka, line 28: Are we checking permissions in here?

stejskalleos: The question is: Shouldn't we check permissions on `host` object, instead of `template_invocations`? Because we don't do that right now, see this scenario:
* Non-admin user with `view_hosts` filter, with `search` set to `aaa.example.com`
* Job invocation with three hosts: `aaa.example.com`, `bbb.example.com` & `ccc.example.com`
* Call `/api/job_invocations/934?host_status=true`

You should get results only for `aaa.example.com`, but you will get results for all three hosts.
In GUI works as expected, so the bug is only in API action.

[RM issue for host permission](https://projects.theforeman.org/issues/29972)

adamruzicka:
Doing the request as admin, against a job with 1000 hosts:
| with host statuses? | time [seconds] |
|---|---|
| no | 6.971 |
| yes | 26.551 |

```
2020-06-09T16:15:40 [I|app|68a74f52] Completed 200 OK in 26443ms (Views: 25445.8ms | ActiveRecord: 873.8ms | Allocations: 34919071)
 | 
 | 
2020-06-09T16:15:40 [W|app|68a74f52] user: aruzicka
 | GET /api/v2/job_invocations/158?host_status=true
 | USE eager loading detected
 |   TemplateInvocation => [:run_host_job_task]
 |   Add to your query: .includes([:run_host_job_task])
 | Call stack
 |   /home/aruzicka/projects/foreman/foreman-remote-execution/app/views/api/v2/job_invocations/main.json.rabl:29:in `block (3 levels) in cached_source_1041171285654418227'
 |   /home/aruzicka/projects/foreman/foreman/config/initializers/rabl_init.rb:49:in `render'
 |   /home/aruzicka/projects/foreman/foreman/config/initializers/rabl_init.rb:49:in `render'
 |   /home/aruzicka/projects/foreman/foreman/config/initializers/rabl_init.rb:49:in `render'
 |   /home/aruzicka/projects/foreman/foreman/config/initializers/rabl_init.rb:49:in `render'
 |   /home/aruzicka/projects/foreman/foreman-remote-execution/app/views/api/v2/job_invocations/show.json.rabl:3:in `_80d3a27fe225f92765decfc35e31e919'
 |   /home/aruzicka/projects/foreman/foreman/app/controllers/api/v2/base_controller.rb:164:in `disable_json_root'
 |   /home/aruzicka/projects/foreman/foreman/app/controllers/concerns/foreman/controller/timezone.rb:10:in `set_timezone'
 |   /home/aruzicka/projects/foreman/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 |   /home/aruzicka/projects/foreman/foreman/app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
 |   /home/aruzicka/projects/foreman/foreman/lib/foreman/middleware/telemetry.rb:10:in `call'
 |   /home/aruzicka/projects/foreman/foreman/lib/foreman/middleware/catch_json_parse_errors.rb:9:in `call'
 |   /home/aruzicka/projects/foreman/foreman/lib/foreman/middleware/logging_context_session.rb:22:in `call'
 |   /home/aruzicka/projects/foreman/foreman/lib/foreman/middleware/logging_context_request.rb:11:in `call'
```
Given the increase in the time it takes to process this is quite substantial, could we preload the tasks in the controller if host status is true?

stejskalleos:
Updated, added `.includes(:template, :run_host_job_task)`

app/views/api/v2/job_invocations/main.json.rabl
```diff
@@ -21,6 +21,14 @@ child :targeting do
 
   child :hosts do
     extends 'api/v2/hosts/base'
+
+    if params[:host_status]
+      node(:job_status) do |host|
+        template_invocation = @template_invocations.find { |template_inv| template_inv.host_id == host.id }
```
adamruzicka, line 27: If you preprocess the data to have the right shape in the controller (say a hash mapping host ids to template invocations) then you can get instant access here instead of iterating through all of the template invocations (in the worst case). On my machine I was able to trim the request time from ~19 seconds to ~7

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -18,7 +18,20 @@ def index
...
       def show
+        @hosts = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
+        @template_invocations = @job_invocation.template_invocations
+                                               .where(host: @hosts)
+                                               .includes(:input_values)
+
+        if params[:host_status]
+          @template_invocations.includes(:run_host_job_task)
```
adamruzicka, line 29: This does not modify `@template_invocations` so this line essentially does nothing

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -18,7 +18,20 @@ def index
...
+          @hosts.map do |host|
```
adamruzicka, line 30: Map iterates over elements in the enumerable, applying the block to each of the elements and collecting the results into an array. In here it would generate an array of template invocation statuses, which would be immediatelly garbage collected since they're not being used anywhere. `#each` would be a better fit in this case since your modifying already existing objects

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -18,7 +18,20 @@ def index
...
+            template_invocation = @template_invocations.find_by(host_id: host.id)
```
adamruzicka, line 31: Doesn't this do a database lookup?

app/controllers/api/v2/job_invocations_controller.rb
stejskalleos: No idea why I used `map` instead of `each` ...

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -2,6 +2,8 @@ module ForemanRemoteExecution
   module HostExtensions
     def self.prepended(base)
       base.instance_eval do
+        attr_accessor :template_invocation_status
```
adamruzicka, line 5: Since it is a thing used only in one view in the api, could we pass that to the view in some other way to prevent hosts from always having this attribute?

stejskalleos: Yeah but in what other way? I was thinking about different solution, but couldn't find anything better than this.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -18,7 +18,20 @@ def index
...
       param :id, :identifier, :required => true
+      param :host_status, String, required: false, allow_blank: true, desc: N_('Show Job status for the hosts.')
       def show
+        @hosts = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
+        @template_invocations = @job_invocation.template_invocations
+                                               .where(host: @hosts)
+                                               .includes([:input_values, :run_host_job_task])
+                                               .to_a
+
+        if params[:host_status]
```
adamruzicka, line 29: In here we would create a hash, mapping host ids to template invocation statuses. This hash would be saved as an instance variable (let's say `@host_template_invocation_statuses`) so it would be accessible from the template

app/views/api/v2/job_invocations/main.json.rabl
```diff
@@ -19,16 +19,22 @@ child :targeting do
   attributes :bookmark_id, :search_query, :targeting_type, :user_id, :status, :status_label,
     :randomized_ordering
 
-  child :hosts do
+  child @hosts do
     extends 'api/v2/hosts/base'
+
+    if params[:host_status]
+      node :job_status do |host|
+        host.template_invocation_status
```
adamruzicka, line 27: This would become just `@host_template_invocation_statuses[host.id]`

app/models/concerns/foreman_remote_execution/host_extensions.rb
adamruzicka: If we did what I suggest in the other two comments, we could drop this.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -18,11 +18,22 @@ def index
...
+          template_invocations = @template_invocations.includes(:run_host_job_task).to_a
+          @host_statuses = {}
+
+          @hosts.each do |host|
+            template_invocation = template_invocations.find { |t| t.host_id == host.id }
+            @host_statuses[host.id] = template_invocation_status(template_invocation)
+          end
+        end
```
adamruzicka, line 36: There's a shorthand for this
```ruby
Hash[template_invocations.map { |ti| [ti.host_id, template_invocation_status(ti)] }]
```

stejskalleos: Yeah, that's much better :+1:

adamruzicka:
Looks good, will do another run of testing tomorrow and then we can hopefully get it in

adamruzicka:
Thank you @stejskalleos !

PR #498, opened 2020-05-20
Use GitHub actions as CI

ezr-ondrej:


adamruzicka:
So we will have hound, travis, jenkins and gh actions?

ezr-ondrej:
> So we will have hound, travis, jenkins and gh actions?

The idea is being we have only gh-actions :)
GH-actions can do it all. but definitelly we can rule out travis and hound (see `rubocop` task and `test_js` task). we can leave just those two for now.
Unless we insist on Hound comenting the code, what I presonaly don't like too much :thinking:

adamruzicka:
Could I see the action in action somewhere? I don't see it being run here nor in your fork

ezr-ondrej:
I've enabled it on push to demostrate: https://github.com/ezr-ondrej/foreman_remote_execution/actions/runs/122023055

adamruzicka:
Could we get it green before we merge it?

ezr-ondrej:
I've fixed up tests to go :green_apple: https://github.com/ezr-ondrej/foreman_remote_execution/actions/runs/122422695 This is the same code, just with removed `push` we still run ruby tests, but I believe it could be worth to try it out alongside Jenkis for a while.

adamruzicka:
Thank you @ezr-ondrej !

PR #499, opened 2020-05-22
Fixes #29887 - Add js linter to the CI

stejskalleos:
* Added `npm run lint` to Travis
* Linted js files


.hound.yml
```diff
@@ -12,7 +12,7 @@ jshint:
   enabled: false
 
 eslint:
-  enabled: true
+  enabled: false
```
adamruzicka, line 15: Now that we verified it helps, how about we get rid of this section completely?

ezr-ondrej:
The failures are only on 2.5 and it's some jenkins pecadilos. Going to squash merge. Thanks @stejskalleos !! :)

PR #501, opened 2020-05-29
Fixes #29962 - Make known hosts key removal less error prone

adamruzicka:
That means:
- skip the removal if ip and/or hostname if nil
- skip the removal if the host/interface is a new record
- add more tests


apuntamb:
Hello @ares and @ezr-ondrej ! Can we please get this PR in sooner as it is blocking provisioning for all compute resources (atleast for Azure, EC2 and GCE).

ezr-ondrej:
Thanks @adamruzicka!

PR #502, opened 2020-06-03
Job Actions - user customizable quick links to jobs

stejskalleos:
**Description**
Cu has a number of common jobs that they run. Having a quick link to the most common jobs would improve my workflow and help with initial training.

**Feature**
* New section Administer > Job Actions
* Created actions are available on _All hosts_ page and at _Host detail_ page
* Created actions are private, for current user only.

**Links:**
- [Discourse RFC](https://community.theforeman.org/t/remote-execution-user-customizable-quick-links/19121)
- [Redmine](https://projects.theforeman.org/issues/27463)

### DIFF
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -7,14 +7,21 @@ module ForemanRemoteExecution
     end
 
     def multiple_actions
-      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ]
+      job_actions = JobAction.authorized(:view_job_actions).all.order(:name).map do |action|
+         [_(action.name), new_job_invocation_path(template_id: action.job_template.id), false]
+      end
+
+      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ] + job_actions
     end
 
     def schedule_job_multi_button(*args)
       host_features = rex_host_features(*args)
+      job_actions = JobAction.authorized(:view_job_actions).all.order(:name).map do |action|
+        link_to _(action.name), new_job_invocation_path(template_id: action.job_template.id)
+      end
 
-      if host_features.present?
-        action_buttons(schedule_job_button(*args), *host_features)
+      if host_features.present? || job_actions.present?
+        action_buttons(schedule_job_button(*args), host_features, *job_actions)
       else
         schedule_job_button(*args)
       end
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -7,21 +7,15 @@ module ForemanRemoteExecution
     end
 
     def multiple_actions
-      job_actions = JobAction.authorized(:view_job_actions).all.order(:name).map do |action|
-         [_(action.name), new_job_invocation_path(template_id: action.job_template.id), false]
-      end
-
-      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ] + job_actions
+      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ] +
+              user_job_actions.map { |a| a.push(false) }
     end
 
     def schedule_job_multi_button(*args)
-      host_features = rex_host_features(*args)
-      job_actions = JobAction.authorized(:view_job_actions).all.order(:name).map do |action|
-        link_to _(action.name), new_job_invocation_path(template_id: action.job_template.id)
-      end
+      items = [rex_host_features(*args), user_job_actions.map { |a| link_to(a[0], a[1]) }].flatten.compact
 
-      if host_features.present? || job_actions.present?
-        action_buttons(schedule_job_button(*args), host_features, *job_actions)
+      if items.any?
+        action_buttons(schedule_job_button(*args), *items)
       else
         schedule_job_button(*args)
       end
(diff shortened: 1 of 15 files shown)


app/views/job_actions/_form.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="col-md-8">
+<%= form_for @job_action, html: { id: 'job_action_form' } do |f| %>
+  <%= text_f f, :name %>
+  <%= selectable_f f, :job_template_id, JobTemplate.all.map { |t| [t.name, t.id] },
```
ares, line 4: permission check needed when loading `JobTemplate.all`

ares: also this may be a long list, perhaps grouping per job category would be better, but only if easily possible with select2

app/views/job_actions/index.html.erb
```diff
@@ -0,0 +1,35 @@
+<% title _("Job Actions") %>
+<% title_actions new_link(_("Create Job Action")) %>
+
+<table class="<%= table_css_classes('table-fixed') %>">
+  <thead>
+    <tr>
+      <th class="col-md-2"><%= sort :name, :as => s_("JobAction|Name") %></th>
+      <th class="">Template</th>
```
ares, line 8: translations missing in all `<th>`s

stejskalleos:
Updated & rebased, ready for review.

app/models/job_action.rb
```diff
@@ -0,0 +1,25 @@
+class JobAction < ApplicationRecord
+  include Authorizable
+
+  belongs_to :job_template
+  belongs_to :user
+  before_validation :set_user
+
+  validates :name, :job_template, :user, presence: true
```
ezr-ondrej, line 8: maybe unique name?
If we are keeping user, than unique name per user (in scope of user)

stejskalleos:
> I do not understand the user scoping. It seems like these are shortcuts for the given user who is defining it, but placing it to the administer menu is suggesting (at least for me) these are global shortcuts, and the description would suggest global shortcuts as well IMHO

- Right now it's just for the user itself, but in the future I'd like to extend it more. This is just the POC and we'll see if we got any comments from users.
- What do you suggest instead of Administration section? I was thinking maybe about the configuration section.

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -7,14 +7,15 @@ def host_overview_buttons(host)
     end
 
     def multiple_actions
-      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ]
+      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ] +
+              user_job_actions.map { |a| a.push(false) }
```
adamruzicka, line 11: Could we avoid in-place modification? Something like the following might give more clarity what's going on on the first sight
```ruby
user_job_actions.map { |(name, path)| [name, path, false] }
```

app/views/job_actions/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <tr>
+      <th class="col-md-2"><%= sort :name, :as => s_("JobAction|Name") %></th>
+      <th class=""><%= _('Template') %></th>
+      <th class=""><%= _('Category') %></th>
+      <th class=""><%= _('Provider') %></th>
+      <th class="col-md-1"><%= _('Actions') %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @job_actions.each do |job_action| %>
```
adamruzicka, line 15: Could we get an empty state page if there are no actions?

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -42,6 +43,12 @@ def host_title_actions(*args)
         button_group(web_console_button(*args)))
       super(*args)
     end
-  end
 
+    def user_job_actions
+      JobAction.authorized(:view_job_actions)
+               .where(user: User.current)
+               .order(:name)
+               .map { |a| [a.name, new_job_invocation_path(template_id: a.job_template.id)] }
```
adamruzicka, line 51: Maybe just `a.job_template_id` to prevent loading the entire template when only id is needed?

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -7,14 +7,15 @@ def host_overview_buttons(host)
...
 
     def multiple_actions
-      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ]
+      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ] +
+              user_job_actions.map { |a| a.push(false) }
     end
 
     def schedule_job_multi_button(*args)
-      host_features = rex_host_features(*args)
+      items = [rex_host_features(*args), user_job_actions.map { |a| link_to(a[0], a[1]) }].flatten.compact
```
adamruzicka, line 15: Could we destructure the array instead of taking it as a whole and then accessing elements by indexes?

pondrejk:
>     * What do you suggest instead of Administration section? I was thinking maybe about the configuration section

It kind of resembles recurring executions in a way that it would remember the job params but would be called on demand, so imho it could go next to it under Monitor > Foreman Tasks

Right now it kind of does what the 'Run' button in JobTemplates list does, but I'm sure it will be nice. I'm sure someone already suggested this, but just in case, having something like 'bookmark this job' button somewhere in job invocations would be very nice

PR #503, opened 2020-06-04
RM#29972 - API::JobInvocations - Host permissions

stejskalleos:
Fixed API endpoints:
* `show` - Now `@hosts` are filtered based on user's permissions.
* `output` - Now looking for host in Job Invocation scope plus checking if user is authorized to see host's results.
* `raw_output` - same as for `output`

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -20,6 +20,9 @@ module Api
       param :id, :identifier, :required => true
       def show
         @hosts = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
+        @template_invocations = @job_invocation.template_invocations
+                                               .where(host: @hosts)
+                                               .includes(:input_values)
       end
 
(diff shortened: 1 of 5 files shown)


adamruzicka:
Would it make sense to limit which template invocations the user can see?

stejskalleos:
Added permission check for Template Invocations (in API controller & GUI controller)

stejskalleos:
> `/api/v2/job_invocations/:id` still shows all template invocations without checking permissions

Yeah I missed that :/, Updated list of fixed actions

Action|Object
------------ | -------------
`/api/job_invocations/:id`| `@host, @template_invocations`
`/api/job_invocations/:id/hosts/:host_id`| `@host`
`/api/job_invocations/:id/hosts/:host_id/raw`| `@host`
`/job_invocations/:job_invocation_id/template_invocations`|`@template_invocations`
`/job_invocations/:id`|`preview_hosts`

adamruzicka:
Thank you @stejskalleos !

PR #504, opened 2020-06-04
Fixes #30015 - remove rubocop from .hound.yml

stejskalleos:


ezr-ondrej:
Do we even need hound actually? The only thing that is not checked elsewhere is (s)css. And I don't see much of a value in it :thinking:

stejskalleos:
@ezr-ondrej good point, there are only three .scss files left and chances that somebody will add new are pretty low.
Just want to confirm with @adamruzicka.

adamruzicka:
I'm fine with getting rid of it completely

PR #506, opened 2020-06-08
Fixes #30051 - Set rubocop version to same as for Foreman

stejskalleos:


stejskalleos:

foreman_remote_execution.gemspec
```diff
@@ -29,6 +29,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'foreman-tasks', '>= 0.15.1'
 
   s.add_development_dependency 'factory_bot_rails', '~> 4.8.0'
-  s.add_development_dependency 'rubocop'
+  s.add_development_dependency 'rubocop', '~> 0.85'
```
adamruzicka, line 32: This means `0.85 <= $version < 1.0`. I'd rather stay on the safe side and go with `~> 0.85.0`

mmoll:
With the current master tests again also running on Jenkins, this won't work with the current plugin test setup.

stejskalleos:

adamruzicka:
@mmoll that means we should configure rubocop for rex the same way as in core, correct?

mmoll:
@adamruzicka Yes, also the versions need to be the same. :disappointed:

stejskalleos:
Rubocop updated to Foreman's version.

adamruzicka:
Thank you @stejskalleos !

PR #507, opened 2020-06-09
Fixes #30068 - Allow remote_execution_ssh_keys to be a string

adamruzicka:


stejskalleos:
LGTM

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -49,7 +49,7 @@ def host_params_hash
       keys = remote_execution_ssh_keys
       source = 'global'
       if keys.present?
-        value, safe_value = params.fetch('remote_execution_ssh_keys', {}).values_at(:value, :safe_value).map { |v| v || [] }
+        value, safe_value = params.fetch('remote_execution_ssh_keys', {}).values_at(:value, :safe_value).map { |v| [v].flatten }
```
ezr-ondrej, line 52: Should we ?
```suggestion
        value, safe_value = params.fetch('remote_execution_ssh_keys', {}).values_at(:value, :safe_value).map { |v| [v].flatten }.compact
```

adamruzicka: That should have no effect, it will always be a two element array where the two elements are possibly empty arrays. Calling compact on it wouldn't do anything.

ezr-ondrej: Stupid me... I mean
```suggestion
        value, safe_value = params.fetch('remote_execution_ssh_keys', {}).values_at(:value, :safe_value).map { |v| [v].flatten.compact }
```

adamruzicka: Except we need each of the resulting variables to be arrays. We take both the values we get from here and join both of them with an additional array. Things wouldn't be so nice if we had to check if the value we're adding to is a nil. This way we can always rely on it being an array and safely do the addition

ezr-ondrej: compact will do `[nil] => [] ` in case the v was nil, before we seemed to expect nil value and returning an empty array.

adamruzicka: Apparently I can't brain today, you're right, good point.

PR #508, opened 2020-06-23
Fixes #30188 - Provide the ability to use su with password

patilsuraj767:
This feature adds the ability to use effective user method "su" with password.

### DIFF
--- a/db/migrate/20200623073022_rename_sudo_password_to_effective_user_password.rb
+++ b/db/migrate/20200623073022_rename_sudo_password_to_effective_user_password.rb
@@ -2,6 +2,8 @@ class RenameSudoPasswordToEffectiveUserPassword < ActiveRecord::Migration[6.0]
   def up
     rename_column :job_invocations, :sudo_password, :effective_user_password
 
+    Parameter.where("name = 'remote_execution_sudo_password'").update_all(name: 'remote_execution_effective_user_password')
+
     return unless (password = Setting.find_by(:name => 'remote_execution_sudo_password').try(:value))
 
     Setting.find_by(:name => 'remote_execution_effective_user_password').update(value: password)
@@ -11,6 +13,8 @@ class RenameSudoPasswordToEffectiveUserPassword < ActiveRecord::Migration[6.0]
 
   def down
     rename_column :job_invocations, :effective_user_password, :sudo_password
+
+    Parameter.where("name = 'remote_execution_effective_user_password'").update_all(name: 'remote_execution_sudo_password')
 
     return unless (password = Setting.find_by(:name => 'remote_execution_effective_user_password').try(:value))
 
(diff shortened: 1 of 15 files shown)


adamruzicka:
In general this looks good. One question though, do we want to add a separate field in the form for every single user method we have or do we want the approach ansible took and just have a single password which will be used for whatever method is chosen at the time?

CC @ezr-ondrej @ares

patilsuraj767:
@adamruzicka 
Yes, having the single password for all effective user method makes sense.
Just a small query, if we are going ahead with single password then In the current implementation of `script_runner.rb` class `DzdoUserMethod` inherits from `SudoUserMethod`. Should we do the same for `SuUserMethod`? Or We should make new class `EffectiveUserMethod` and shift logic to it and then inherit `EffectiveUserMethod` to other user methods.

ares:
:+1: for something like `become_password` or `effective_user_method_password`

adamruzicka:
Now I'm just thinking out loud, shouldn't we also ship a migration which would rename all existing parameters (and the setting) from the old name to the new one so if someone has passwords set up as parameters, everything keeps working?

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -178,10 +172,8 @@ def prepare_start
     def initialization_script
       # pipe the output to tee while capturing the exit code in a file
       <<-SCRIPT.gsub(/^\s+\| /, '')
-      | sh <<WRAPPER
-      | (#{@user_method.cli_command_prefix}#{@remote_script} < /dev/null; echo \\$?>#{@exit_code_path}) | /usr/bin/tee #{@output_path}
-      | exit \\$(cat #{@exit_code_path})
-      | WRAPPER
+      | sh -c "(#{@user_method.cli_command_prefix}#{@user_method.instance_of?(ForemanRemoteExecutionCore::SuUserMethod) ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path}) | /usr/bin/tee #{@output_path}
```
adamruzicka, line 175: Could we pull the `@user_method.instance_of? ...` conditional out of the `SCRIPT`, it is getting a bit hard to read

patilsuraj767: Ok. I will do it.

adamruzicka:
Would it make sense to migrate parameters the same way you migrate settings here? Also could you put together a followup PR for foreman_ansible which still uses the old names?

db/migrate/20200623073022_rename_sudo_password_to_effective_user_password.rb
```diff
@@ -0,0 +1,24 @@
+class RenameSudoPasswordToEffectiveUserPassword < ActiveRecord::Migration[6.0]
+  def up
+    rename_column :job_invocations, :sudo_password, :effective_user_password
+
+    Parameter.where("name = 'remote_execution_sudo_password'").update_all(name: 'remote_execution_effective_user_password')
```
adamruzicka, line 5: What would happen if a `remote_execution_effective_user_password` parameter already existed?

patilsuraj767: Yes, db:migrate can fail in that case. I will work on it.

adamruzicka:
Thank you @patilsuraj767 !

PR #511, opened 2020-06-25
Fixes #30213 - long host list breaks page

MariaAga:
in job invocations long host list breaks page, the page is too long to really use or understand when there are too many hosts in the list.
`infoblock` class was removed a long time ago.

before:
Screenshot from 2020-06-25 12-42-22
After
Screenshot from 2020-06-25 12-47-15


adamruzicka:
Thank you @MariaAga !

PR #512, opened 2020-06-29
Update apt cache before installing packages in GH action

adamruzicka:


adamruzicka:
The CI is green, merging

PR #514, opened 2020-07-01
Fixes #30225 - Make ssh debug logging show useful data

adamruzicka:
When debug ssh logging is enabled, it passes a facility as an argument
and a block to the logger. Our implementation of log filter, which
allows only messages of required severity (or higher) to be logger
discarded the block. For this reason, log messages showed only
facility, which wasn't really useful in any way.

### DIFF
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -32,7 +32,9 @@ module ForemanRemoteExecutionCore
       raise "Wrong value '#{@settings[:ssh_log_level]}' for ssh_log_level, must be one of #{SSH_LOG_LEVELS.join(', ')}"
     end
 
-    current = if defined?(SmartProxyDynflowCore)
+    current = if defined?(::Proxy::SETTINGS)
+                ::Proxy::SETTINGS.log_level.to_s.downcase
+              elsif defined?(SmartProxyDynflowCore)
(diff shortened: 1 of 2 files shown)


lib/foreman_remote_execution_core.rb
```diff
@@ -32,7 +32,9 @@ def self.validate_ssh_log_level!
       raise "Wrong value '#{@settings[:ssh_log_level]}' for ssh_log_level, must be one of #{SSH_LOG_LEVELS.join(', ')}"
     end
 
-    current = if defined?(SmartProxyDynflowCore)
+    current = if defined?(::Proxy::SETTINGS)
+                ::Proxy::SETTINGS.log_level.to_s.downcase
```
ezr-ondrej, line 36: shouldn't this be fallback for a `SmartProxyDynflowCore` rather than the other way around? :thinking:

adamruzicka: That's for the case when dynflow core runs inside smart proxy. In that case, settings in `SmartProxyDynflowCore::SETTINGS` are not populated yet

ezr-ondrej: Make sense :+1:

ezr-ondrej:
Thanks @adamruzicka !

PR #515, opened 2020-07-07
Fixes #30330 - Added word-wrap for description field

patilsuraj767:
Before - 
Before-text-wrap

After -
After-text-wrap


adamruzicka:
Thank you @patilsuraj767 !

PR #517, opened 2020-07-14
Fixes #30399 - Fix error when deleting discovered hosts

adamruzicka:
We hook onto Host::Managed and Nic::Managed destroy to perform the ssh known
host key removal. However, we use the host as a base from which we figure out
from which proxies the keys should be removed. Because discovered hosts have
managed nics, but do not implement #remote_execution_proxies, the error
undefined method `remote_execution_proxies` was raised when deleting discovered
hosts.


ezr-ondrej:
Thanks @adamruzicka :+1:

lzap:
Aaaaaaaaaaaaaaaaaaaaah.

adamruzicka:
> Aaaaaaaaaaaaaaaaaaaaah.

Not sure if that's a shout of discovery or pain, but I guess both would be adequate here

PR #518, opened 2020-07-14
Fixes #30403 - Add DSL docs for REX plugin

ofedoren:


app/lib/foreman_remote_execution/renderer/scope/input.rb
```diff
@@ -23,6 +42,14 @@ def cached(key, &block)
           Rails.cache.fetch(cache_key, &block)
         end
 
+        apipie :method, 'Render template by given name' do
+          required :template_name, String, desc: 'name of the template to render'
+          optional :input_values, Hash, desc: 'key:value list of input values for the template'
+          optional :options, Hash, desc: 'TODO: not sure'
```
ofedoren, line 48: Not sure about usage of this one...

adamruzicka: If I'm reading this right so far the only expected key in there is `:with_foreign_input_set` which is expected to be a boolean and controls whether a foreign input set should be considered when rendering the template.

ofedoren: Thought so too, thanks. Updated.

app/lib/foreman_remote_execution/renderer/scope/input.rb
```diff
@@ -23,6 +42,16 @@ def cached(key, &block)
           Rails.cache.fetch(cache_key, &block)
         end
 
+        # rubocop:disable Lint/InterpolationCheck
+        apipie :method, 'Render template by given name' do
+          required :template_name, String, desc: 'name of the template to render'
+          optional :input_values, Hash, desc: 'key:value list of input values for the template'
+          optional :options, Hash, desc: 'Additional options such as :with_foreign_input_set. Set to true if a foreign input set should be considered when rendering the template'
```
adamruzicka, line 49: @ares could you check if this makes sense? You're way more familiar with templates and their rendering than I am

ares: yup, that's correct

PR #519, opened 2020-07-14
Fixes #30406: default_capsule method exists on SmartProxy object

ehelms:


app/services/default_proxy_proxy_selector.rb
```diff
@@ -10,7 +10,7 @@ def initialize
   def available_proxies(host, provider)
     # TODO: Once we have a internal proxy marker/feature on the proxy, we can
     # swap the implementation
-    internal_proxy = ::Katello.default_capsule
+    internal_proxy = ::SmartProxy.default_capsule
```
ezr-ondrej, line 13: It's still comming from a Katello, so this call still implies dependency on `Katello` just less explicitly.
Should we instead implement notion of `internal_proxy` in the Foreman itself?

ehelms:
The method is defined on the smart proxy just within Katellos code. The
dependency is still there on Katello.

ezr-ondrej:
> The method is defined on the smart proxy just within Katellos code. The dependency is still there on Katello.

Yeah, that's what I'm saying :thinking: but with calling it on `Katello` it's obvious it's comming from katello, but like this I'd not be sure. So it doesn't improve the situation much IMHO.

ehelms:
I can add code that will do `fail("default_capsule must be defined") unless  ::SmartProxy.defined?('default_capsule')`

adamruzicka:
It would be better than having it fail with undefined method error in case katello is not available

ehelms:
Note https://github.com/theforeman/foreman_remote_execution/pull/519/files#diff-84297365d8d19c4d3bfa8769908bbb85R5 protects against missing Katello. But I will add another layer.

ezr-ondrej:
Rubocop doesn't like `fail` I'd agree `raise` is what we want.

adamruzicka:
Thank you @ehelms & @ezr-ondrej !

PR #523, opened 2020-07-30
Fixes #30525 - Disable task buttons when job finishes

adamruzicka:


adamruzicka:
> We've just moved this to React, haven't we?

I don't think so, the action buttons were always rendered on the server and refreshed with some jquery. Only the host table at the bottom of the page was rewritten to react recently.

Since this is a thing that used to work and got borked I wanted to fix it and get it out in a release asap. At least for me, rewriting the buttons to react would turn this from a 20 minute fix to something much bigger, which I probably still wouldn't dare to release in a bugfix release

PR #524, opened 2020-08-06
Fixes #30582 - host search filter does not work in job invocation page

stejskalleos:


adamruzicka:
When I type a search query into the search box and press enter, nothing happens, I have to hit the search button with mouse. Is this something that we could easily fix as part of this?

stejskalleos:
> When I type a search query into the search box and press enter, nothing happens, I have to hit the search button with mouse. Is this something that we could easily fix as part of this?

This is a bug what was there before - I tried REX 3.2.2 and it have same behavior. I didn't touched search & pagination at all - see https://github.com/theforeman/foreman_remote_execution/pull/493/files#diff-3aa3b6670498f9ab916dca652df18dbb

In next version it's fixed, search bar and pagination is full react, however after quick look I wasn't able to find solution how to fix it easily.

adamruzicka:
Fair enough

adamruzicka:
Thank you @stejskalleos !

PR #525, opened 2020-08-12
Fixes #30628 - show empty hosts in job invocation api

MariaAga:
based on the solution here: https://github.com/nesquena/rabl/issues/166


adamruzicka:
Could you add a test? There have been too many bugs for such a simple thing

MariaAga:
Changed a test to check that it's an empty array and not just 'empty'

adamruzicka:
Thank you @MariaAga ! Could you please open a cherrypick (with -x) against `3.3-stable`?

MariaAga:
https://github.com/theforeman/foreman_remote_execution/pull/526 @adamruzicka

PR #528, opened 2020-08-18
Fixes #22187 - Show correct job stats

adamruzicka:
When a host gets deleted after a job was run:
- the chart and index shows the full count of hosts on which the job originally run, including the deleted one
- total hosts card splits into two, shows how many hosts are gone
- table shows only present hosts

image

### DIFF
--- a/app/views/job_invocations/_tab_overview.html.erb
+++ b/app/views/job_invocations/_tab_overview.html.erb
@@ -9,7 +9,18 @@
   <% template_invocations.each do |template_invocation| %>
     <%= minicard('user', template_invocation.effective_user || Setting[:remote_execution_effective_user],
                  template_invocation.template.name + ' ' + _('effective user')) %>
-    <%= minicard('cluster', job_invocation.total_hosts_count, _('Total hosts')) %>
+    <div class="row">
+      <% missing = job_invocation.total_hosts_count - job_invocation.targeting.hosts.count %>
+      <% size = missing.zero? ? 12 : 6 %>
+      <div class="col-xs-12 col-sm-<%= size %> col-md-<%= size %>" >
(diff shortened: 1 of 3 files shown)


ares:
does this fully replace https://github.com/theforeman/foreman_remote_execution/pull/309?

adamruzicka:
Yes, however it takes a different approach. If I recall correctly #309 modifies the way how we count the tasks to only include those which have existing hosts. This PR shows the full stats and a warning if some of the hosts disappeared since start of the job

adamruzicka:
image

app/views/job_invocations/_tab_overview.html.erb
```diff
@@ -9,7 +9,19 @@
   <% template_invocations.each do |template_invocation| %>
     <%= minicard('user', template_invocation.effective_user || Setting[:remote_execution_effective_user],
                  template_invocation.template.name + ' ' + _('effective user')) %>
-    <%= minicard('cluster', job_invocation.total_hosts_count, _('Total hosts')) %>
+    <div class="row">
+      <% missing = job_invocation.targeting.resolved? ? job_invocation.total_hosts_count - job_invocation.targeting.hosts.count : 0 %>
```
ares, line 13: should this be a `job_invocation` method instead? we should probably add the same information to the API

ares (changes requested):
one suggestion, otherwise looks good

app/views/job_invocations/_tab_overview.html.erb
adamruzicka: Done, api now shows this, hammer pr will follow shortly

adamruzicka:
Hammer followup https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/34

PR #529, opened 2020-08-18
Fixes #20689 - make REX interface accessible by API

MariaAga:
adds 'execution' attribute to https://foreman/api/hosts/host_id/interfaces

### DIFF
--- a/app/models/concerns/api/v2/interfaces_controller_extensions.rb
+++ b/app/models/concerns/api/v2/interfaces_controller_extensions.rb
@@ -1,0 +1,13 @@
+module Api
+  module V2
+    module InterfacesControllerExtensions
+      extend Apipie::DSL::Concern
+
+      update_api(:create, :update) do
+        param :interface, Hash do
+          param :execution, :bool, :desc => N_('Should this interface be used for remote execution?')
+        end
+      end
+    end
+  end
+end
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -138,6 +138,7 @@ module ForemanRemoteExecution
         end
 
         extend_rabl_template 'api/v2/smart_proxies/main', 'api/v2/smart_proxies/pubkey'
+        extend_rabl_template 'api/v2/interfaces/main', 'api/v2/interfaces/execution_flag'
         describe_host { overview_buttons_provider :host_overview_buttons }
       end
     end
@@ -184,6 +185,7 @@ module ForemanRemoteExecution
       SmartProxy.prepend ForemanRemoteExecution::SmartProxyExtensions
       Subnet.include ForemanRemoteExecution::SubnetExtensions
 
+      ::Api::V2::InterfacesController.include Api::V2::InterfacesControllerExtensions
       # We need to explicitly force to load the Task model due to Rails loader
       # having issues with resolving it to Rake::Task otherwise
       require_dependency 'foreman_tasks/task'
--- a/app/views/api/v2/interfaces/execution_flag.json.rabl
+++ b/app/views/api/v2/interfaces/execution_flag.json.rabl
@@ -1,0 +1,1 @@
+attributes :execution


lib/foreman_remote_execution/engine.rb
```diff
@@ -138,6 +138,7 @@ class Engine < ::Rails::Engine
         end
 
         extend_rabl_template 'api/v2/smart_proxies/main', 'api/v2/smart_proxies/pubkey'
+        extend_rabl_template 'api/v2/interfaces/main', 'api/v2/remote_execution_features/host_extensions'
```
adamruzicka, line 141: It is actually an extension of interface so the name is slightly misleading.

MariaAga:
The user can change the value through api

lib/foreman_remote_execution/engine.rb
MariaAga: renamed to `interface_extensions`

adamruzicka:
Good, could you also add api documentation so hammer can autogenerate options from it? And could you also open a PR against hammer so it can show the execution flag?

MariaAga:
Added api doc and hammer pr (https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/35)

app/models/concerns/api/v2/interfaces_controller_extensions.rb
```diff
@@ -0,0 +1,13 @@
...
+  module V2
+      module InterfacesControllerExtensions
+        extend Apipie::DSL::Concern
+
+        update_api(:create, :update) do
+          param :interface, Hash do
+            param :execution, :bool, :desc => N_('Is the execution interface used for remote execution')
+          end
+        end
+      end
```
adamruzicka, line 11: This is indented one level more than it should be

app/models/concerns/api/v2/interfaces_controller_extensions.rb
```diff
@@ -0,0 +1,13 @@
+module Api
+  module V2
+      module InterfacesControllerExtensions
+        extend Apipie::DSL::Concern
+
+        update_api(:create, :update) do
+          param :interface, Hash do
+            param :execution, :bool, :desc => N_('Is the execution interface used for remote execution')
```
adamruzicka, line 8: The other flags have the description in the format of "Should this interface be used for $SOMETHING?", it would be nice to stick to it

adamruzicka (changes requested):

app/models/concerns/api/v2/interfaces_controller_extensions.rb
MariaAga: right, changed.

app/views/api/v2/remote_execution_features/interface_extensions.rabl
```diff
@@ -0,0 +1 @@
+attributes :execution
```
adamruzicka, line 1: Rabl is just a templating language, the template files usually include the "resulting" format after rendering, so for api the filename should end with `.json.rabl`

lib/foreman_remote_execution/engine.rb
```diff
@@ -138,6 +138,7 @@ class Engine < ::Rails::Engine
         end
 
         extend_rabl_template 'api/v2/smart_proxies/main', 'api/v2/smart_proxies/pubkey'
+        extend_rabl_template 'api/v2/interfaces/main', 'api/v2/remote_execution_features/interface_extensions'
```
adamruzicka, line 141: Usually the templates are named as `$what_is_being_extended/$with_what`. By reading `remote_execution_features/interface_extensions` I'd expect it to add information about interfaces to remote execution features. Probably `interfaces/execution_flag` woudl be a better name

adamruzicka (changes requested):

app/views/api/v2/remote_execution_features/interface_extensions.rabl
MariaAga: Changed, thanks!

lib/foreman_remote_execution/engine.rb
MariaAga: Makes sense, changed, thanks!

adamruzicka (approved):
Looks good to me, will merge when the tests turn green

MariaAga:
all green :tada:

adamruzicka:
Thank you @MariaAga !

PR #530, opened 2020-08-20
Refs #30648 - Allow overriding proxy selector at the feature level

adamruzicka:


adamruzicka:
This is about a RemoteExecutionFeature model in Foreman, not a proxy feature. Or do you mean using the new capabilities api on the smart proxy to determine which one is the "internal" one?

ezr-ondrej:
> This is about a RemoteExecutionFeature model in Foreman, not a proxy feature. Or do you mean using the new capabilities api on the smart proxy to determine which one is the "internal" one?

That is exactly what I mean :)
We are trying to run this exact feature on internal proxy only.
But instead we can ask the proxy what features it supports and only internal would answer it supports this feature :shrug:

adamruzicka:
We were considering going that route but because we started on this pretty late, we decided to go the simpler, but arguably less correct route, see discussion at https://github.com/theforeman/foreman_remote_execution/pull/490

ezr-ondrej:
Oh, so we want to backport this?

db/migrate/20200820122057_add_proxy_selector_override_to_remote_execution_feature.rb
```diff
@@ -0,0 +1,5 @@
+class AddProxySelectorOverrideToRemoteExecutionFeature < ActiveRecord::Migration[4.2]
+  def change
+    add_column :remote_execution_features, :proxy_selector_override, :string
```
ezr-ondrej, line 3: Would `proxy_selector` do?
```suggestion
    add_column :remote_execution_features, :proxy_selector, :string
```
Apart of that, lets get it in! :+1:

ezr-ondrej (approved):

db/migrate/20200820122057_add_proxy_selector_override_to_remote_execution_feature.rb
adamruzicka: With the `_override` suffix it is clear that if set it overrides some other value and if unset, some "normal" value is used. If we change it to `proxy_selector` only I'd say it stops being clear what no value in this column means

ezr-ondrej: Fair enough. I still think blank value = I do not care 'bout my proxy is fair as well, but ok. No strict opinion here.

adamruzicka:
Needs https://github.com/theforeman/foreman_remote_execution/pull/536 to prevent issues where features are registered before db is migrated

PR #531, opened 2020-08-24
Fixes #30685 - Ensure run button appears in job templates page

adamruzicka:
Due to how Rails load code, it could happen that the module with extensions
would get prepended too late and it would be ignored. By prepending the
extensions from an initializer we make it happen earlier in the boot process.

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -32,6 +32,15 @@ module ForemanRemoteExecution
       end
     end
 
+    # A workaround for https://projects.theforeman.org/issues/30685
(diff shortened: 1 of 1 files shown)


pondrejk (approved):

ezr-ondrej (approved):
I do not like this, but it works and there is not enough time to fix properly.
Thanks @adamruzicka ! :+1:

ares:
oh wow, what an uglyness :-) do we have redmine issue for adding this extension point in core?

adamruzicka:
We do now https://projects.theforeman.org/issues/30718

PR #532, opened 2020-09-07
Fixes #30762 - Fix public keys being duplicated

adamruzicka:


app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -44,21 +44,11 @@ def execution_status_label(options = {})
...
-        value, safe_value = params.fetch('remote_execution_ssh_keys', {}).values_at(:value, :safe_value).map { |v| [v].flatten.compact }
-        params['remote_execution_ssh_keys'] = {:value => value + keys, :safe_value => safe_value + keys, :source => source}
-      end
-      [:remote_execution_ssh_user, :remote_execution_effective_user_method,
-       :remote_execution_connect_by_ip].each do |key|
-        value = Setting[key]
-        params[key.to_s] = {:value => value, :safe_value => value, :source => source} unless params.key?(key.to_s)
-      end
-      params
+      @cached_rex_host_params_hash ||= extend_host_params_hash(super)
```
tbrisker, line 49: You should probably add this variable to https://github.com/theforeman/foreman/blob/develop/app/models/concerns/host_params.rb#L14 as well

tbrisker (approved):
Thanks @adamruzicka !

PR #533, opened 2020-09-09
Fixes #30808 - Use IPv6-proxy if no IPv4-proxy

m-bucher:
It would also be possible to add the IPv6-subnet as `:subnet6`, but this would require adding `:subnet6` to the [`strategy`-list in Foreman-Tasks](https://github.com/theforeman/foreman-tasks/blob/904c07ca7f2be446e5453eeb271f9c4bc98628de/app/services/foreman_tasks/proxy_selector.rb#L11).

I would like to put this up for discussion :-)

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -56,7 +56,10 @@ module ForemanRemoteExecution
 
     def remote_execution_proxies(provider, authorized = true)
       proxies = {}
-      proxies[:subnet]   = execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface&.subnet
+      proxies[:subnet] = []
+      proxies[:subnet] += execution_interface.subnet6.remote_execution_proxies.with_features(provider) if execution_interface&.subnet6
+      proxies[:subnet] += execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface&.subnet
+      proxies[:subnet].uniq!
       proxies[:fallback] = smart_proxies.with_features(provider) if Setting[:remote_execution_fallback_proxy]
 
       if Setting[:remote_execution_global_proxy]


app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -56,6 +56,7 @@ def execution_interface
 
     def remote_execution_proxies(provider, authorized = true)
       proxies = {}
+      proxies[:subnet]   = execution_interface.subnet6.remote_execution_proxies.with_features(provider) if execution_interface&.subnet6
       proxies[:subnet]   = execution_interface.subnet.remote_execution_proxies.with_features(provider) if execution_interface&.subnet
```
adamruzicka, line 60: I'd say the proxies from both subnet types should probably be lumped together and `uniq`ed.

adamruzicka (commented):
In general :+1: to the general direction, but maybe it should be a bit smarter. Maybe skip ipv6 subnets if the host has no ipv6 address on the execution interface (and similarly for ipv4)?

app/models/concerns/foreman_remote_execution/host_extensions.rb
m-bucher: > maybe it should be a bit smarter

Full ACK to that. Unfortunately, I do not really know about the possible use-cases and how to determine them.
Could it be necessary to prefer the IPv6-proxy about the IPv4-proxy or vice versa and if so, how can we tell here :thinking:

m-bucher:
@adamruzicka can we solve the issue at hand as proposed here now? I fear a nicer solution will take much more time and consideration.

Maybe we need a more versatile solution in foreman itself to receive subnets (regardless of IPv4, IPv6 or whatever may come next).
I think we also have a similar issue in the UI, when showing the number of hosts for a subnet on the Subnet index-page, where IPv6-only subnets are ignored.

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @m-bucher .

> Maybe we need a more versatile solution in foreman itself to receive subnets 

Could you open a thread on discourse about this so we get more people into the loop?

m-bucher:
Thanks @adamruzicka, I opened the new topic [IPv6 subnet precedence and fallback in host-interfaces](https://community.theforeman.org/t/ipv6-subnet-precedence-and-fallback-in-host-interfaces/20499) for further discussion.

PR #535, opened 2020-09-16
Fixes #30859 - Prevent job from getting stuck after first batch

adamruzicka:
By accounting for missing hosts in a job we broke running jobs on hosts past
first batch. There is a module which periodically polls states of sub-tasks and
overwrites then numbers we were using to decide whether to spawn next batch or
not.

With this change we store the expected total count under a different key in
task's output so it won't get overwritten by the included polling module. For
compatibility with already existing jobs we use the value under the original key
in task's output in case the task doesn't have the value set under the new key.


stejskalleos:
Works as described, tested on job with 800+ hosts and finished without any issues.

adamruzicka:
Thank you @stejskalleos !

PR #536, opened 2020-09-23
Fixes #30901 - Do not register features when there are pending migrations

adamruzicka:


lzap (approved):
Helps to migrate 20200820122057 AddProxySelectorOverrideToRemoteExecutionFeature which was failing for me. I am still having issue with seeding tho.

MariaAga:
Fixed the migrations errors for me and I don't have an issue with db:seed

app/models/remote_execution_feature.rb
```diff
@@ -24,8 +24,10 @@ def self.feature(label)
   end
 
   def self.register(label, name, options = {})
+    pending_migrations = ::Foreman::Plugin.registered_plugins[:foreman_remote_execution]&.pending_migrations
```
ezr-ondrej, line 27: Just a nitpick, it seems better to use `find` as it feels like a real public interface.
```suggestion
    pending_migrations = ::Foreman::Plugin.find(:foreman_remote_execution)&.pending_migrations
```

ezr-ondrej (approved):
:+1:

app/models/remote_execution_feature.rb
adamruzicka: You didn't give me time to actually apply this

ezr-ondrej: I didn't think it's important to change now, I don't see the `registred_plugins` going away anytime soon :)
but sorry, will wait for a bit next time ;)

PR #539, opened 2020-09-30
Fixes #30933, #30935 - Type=search input fixes

adamruzicka:
requires:
- [ ] ~~https://github.com/theforeman/foreman/pull/8035~~
- [x] https://github.com/theforeman/foreman/pull/8060

### DIFF
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -482,7 +482,7 @@ class JobInvocationComposer
 
   def input_value_for(input)
     invocations = pattern_template_invocations
-    default = TemplateInvocationInputValue.new
+    default = TemplateInvocationInputValue.new(:template_input_id => input.id)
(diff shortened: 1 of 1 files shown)


adamruzicka:
Dropped the second commit. It shouldn't be needed after https://github.com/theforeman/foreman/pull/8060 gets in

ezr-ondrej (approved):
Make sense :+1: 

Just a thought for feature improvements:
We have ReportComposer in the core, that tries to acomplish similar task as JobInvocationComposer, would it be worth to make an unified ancestor? :thinking:

adamruzicka:
Not sure if they have all that much in common from the top of my head, but worth investigating in the future

shiramax (approved):
I've tested the PR and it's working great.

ezr-ondrej:
The core PR is in, lets merge this :)
Thanks @adamruzicka and @shiramax for testing! :+1:

PR #540, opened 2020-10-01
Fixes #30947 - replace the notice flash type

amirfefer:
The toast notification component supports in `warning`, `success`, `info` and `error` type
I used the `success` type here though it can also be `info`


adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @amirfefer !

PR #541, opened 2020-10-01
Fix incorrect form field label

indygriffiths:


app/views/job_invocations/_form.html.erb
```diff
@@ -113,7 +113,7 @@
     </div>
 
     <div class="form-group">
-      <%= add_label({ :label => _('Type of query'), :label_help => _("Type has impact on when is the query evaluated to hosts.<br><ul><li><b>Static</b> - evaluates just after you submit this form</li><li><b>Dynamic</b> - evaluates just before the execution is started, so if it's planed in future, targeted hosts set may change before it</li></ul>") }, f, :targetting_type) %>
+      <%= add_label({ :label => _('Type of query'), :label_help => _("Type has impact on when is the query evaluated to hosts.<br><ul><li><b>Static</b> - evaluates just after you submit this form</li><li><b>Dynamic</b> - evaluates just before the execution is started, so if it's planed in future, targeted hosts set may change before it</li></ul>") }, f, :targeting_type) %>
```
adamruzicka, line 116: Doesn't there actually have to be an element with the same id (`targeting_type`) in order for this change to actually do anything? (https://www.w3schools.com/tags/tag_label.asp)

indygriffiths: Ah you're right, for some reason I mis-read the radio buttons and thought they had the same `id` attr as the form label.

image

Will close this off

PR #543, opened 2020-10-07
Fixes #31012 - API for retrieving outputs for multiple hosts

adamruzicka:


ezr-ondrej:
It seems we don't have permission for the new action

adamruzicka:
We do now

ezr-ondrej (commented):
Just two points to discuss, by no means blockers. Apart of those this LGTM :+1: 

- Would `outputs` be descriptive enough given it's quite straight forward what outputs we mean. The `bulk_outputs` invokes for me that it's gonna be outputs of more jobs.
- Should we have `bulk_raw_outputs` / `raw_outputs` as well ?

adamruzicka:
> Would outputs be descriptive enough given it's quite straight forward what outputs we mean. The bulk_outputs invokes for me that it's gonna be outputs of more jobs.

Possibly, `bulk_outputs` was more of a working name that somehow stuck.

> Should we have bulk_raw_outputs / raw_outputs as well ?

I don't really like the split between `output` and `raw_output`. What if for `bulk_outputs`/`outputs` it was just another parameter, defaulting to false?

ezr-ondrej:
> Possibly, `bulk_outputs` was more of a working name that somehow stuck.

So should we rename it before merging? :thinking: 

> I don't really like the split between `output` and `raw_output`. What if for `bulk_outputs`/`outputs` it was just another parameter, defaulting to false?

Yeah, that sounds good to me :+1:

adamruzicka:
Will do both of the changes

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -6,7 +6,7 @@ class JobInvocationsController < ::Api::V2::BaseController
 
       before_action :find_optional_nested_object, :only => %w{output raw_output}
       before_action :find_host, :only => %w{output raw_output}
-      before_action :find_resource, :only => %w{show update destroy clone cancel rerun}
+      before_action :find_resource, :only => %w{show update destroy clone cancel rerun bulk_outputs}
```
ezr-ondrej, line 9:
```suggestion
      before_action :find_resource, :only => %w{show update destroy clone cancel rerun outputs}
```

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -145,7 +163,7 @@ def allowed_nested_id
 
       def action_permission
         case params[:action]
-        when 'output', 'raw_output'
+        when 'output', 'raw_output', 'bulk_outputs'
```
ezr-ondrej, line 166:
```suggestion
        when 'output', 'raw_output', 'outputs'
```

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -160,6 +160,25 @@ class JobInvocationsControllerTest < ActionController::TestCase
         end
       end
 
+      describe '#bulk_outputs' do
```
ezr-ondrej, line 163:
```suggestion
      describe '#outputs' do
```

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -160,6 +160,25 @@ class JobInvocationsControllerTest < ActionController::TestCase
...
+          result = ActiveSupport::JSON.decode(@response.body)
+          host_output = result['outputs'].first
+          assert_equal host_output['id'], @invocation.targeting.host_ids.first
+          assert_equal host_output['refresh'], true
+          assert_equal host_output['output'], []
+          assert_response :success
+        end
+
+        test 'should provide outputs for hosts in the job matching a search query' do
+          get :bulk_outputs, params: { :id => @invocation.id, :search_query => "name = definitely_not_in_the_job" }
```
ezr-ondrej, line 175:
```suggestion
          get :outputs, params: { :id => @invocation.id, :search_query => "name = definitely_not_in_the_job" }
```

test/functional/api/v2/job_invocations_controller_test.rb
```diff
@@ -160,6 +160,25 @@ class JobInvocationsControllerTest < ActionController::TestCase
         end
       end
 
+      describe '#bulk_outputs' do
+        test 'should provide outputs for hosts in the job' do
+          get :bulk_outputs, params: { :id => @invocation.id }
```
ezr-ondrej, line 165:
```suggestion
          get :outputs, params: { :id => @invocation.id }
```

ezr-ondrej (commented):
Just places with forgotten renaming, apart of it, LGTM :+1:

adamruzicka:
I thought I probably missed a place or two, but didn't expect this many

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -137,6 +137,24 @@ def rerun
...
+      param :since, String, :required => false
+      param :raw, String, :required => false
+      def outputs
+        hosts = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
+        hosts = hosts.search_for(params['search_query']) if params['search_query']
+        raw = ActiveRecord::Type::Boolean.new.cast params['raw']
+        default_value = raw ? '' : []
+        outputs = hosts.map do |host|
+          host_output(@job_invocation, host, :default => default_value, :since => params['since'], :raw => raw)
+            .merge(id: host.id)
```
ezr-ondrej, line 152: Sorry for not noticing sooner, but while testing it, I've realized `outputs: [{...., id: 1}]` is quite missleading. First sight it seems like an ID of the output. Should we use `host_id` instead to be more explicit?
```suggestion
            .merge(host_id: host.id)
```

ezr-ondrej (changes requested):
Just one more find from testing.

ezr-ondrej:
And one more in the `raw = true`:
We are using `join` without delimiter, so the output is one line.
Should we use `output.map { |set| set['output'] }.join("\n")` instead?

adamruzicka:
The partial outputs should be ending with newlines so if we join it without any delimiter, it should appear the same way as it was originally produced by the command

ezr-ondrej:
> The partial outputs should be ending with newlines so if we join it without any delimiter, it should appear the same way as it was originally produced by the command

It was not the case for me, but it is quite likely I was just doing something wrong :+1: I'll retest it.
What about the `host_id` in the output? Should we do something about it?

adamruzicka:
> What about the host_id in the output? Should we do something about it?

Applied your suggestion just now

adamruzicka:
And need to fix the tests

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -137,6 +137,24 @@ def rerun
         end
       end
 
+      api :GET, '/job_invocation/:id/outputs', N_('Get outputs of hosts in a job')
```
ezr-ondrej, line 140: The first time around I though "why have I used singular here, we are not github", second time "something is odd".
I've figured that while checking the `/apidocs` Turns out I've been following the documentation without even knowing it xD
```suggestion
      api :GET, '/job_invocations/:id/outputs', N_('Get outputs of hosts in a job')
```

ezr-ondrej (changes requested):
Apart of a documentation typo, works nice :+1:
I was just using bad json parser, that ate the newlines from the raw output :)

ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #545, opened 2020-10-20
Fixes #31109 - use non-interactive mode

sbernhard:


adamruzicka:
What will happen when you try to install a package which requires interactive response? Will it fail? Will it just accept whatever prompt would come up?

sbernhard:
apt-get will  use the default selection.

adamruzicka (approved):
Makes sense, ACK

adamruzicka:
Thank you @sbernhard !

PR #546, opened 2020-10-20
Fixes #31116 - add exprimental page for job wizard

MariaAga:
A starting page for the PF4 job creation wizard will be in the lab features menu.
It doesn't have any content, yet 
Screenshot from 2020-10-20 21-31-49

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -45,6 +45,8 @@ Rails.application.routes.draw do
   get 'cockpit/redirect', to: 'cockpit#redirect'
   get 'ui_job_wizard/categories', to: 'ui_job_wizard#categories'
 
+  match '/experimental/job_wizard', to: 'react#index', :via => [:get]
+
   namespace :api, :defaults => {:format => 'json'} do
     scope '(:apiv)', :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2, :default => true) do
       resources :job_invocations, :except => [:new, :edit, :update, :destroy] do
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -53,6 +53,7 @@ module ForemanRemoteExecution
 
     initializer 'foreman_remote_execution.register_plugin', before: :finisher_hook do |_app|
       Foreman::Plugin.register :foreman_remote_execution do
+        register_global_js_file 'global'
         requires_foreman '>= 2.2'
 
         apipie_documented_controllers ["#{ForemanRemoteExecution::Engine.root}/app/controllers/api/v2/*.rb"]
@@ -137,6 +138,13 @@ module ForemanRemoteExecution
           caption: N_('Jobs'),
           parent: :monitor_menu,
           after: :audits
+
+        menu :labs_menu, :job_wizard,
+          url_hash: { controller: 'job_wizard', action: :index },
+          caption: N_('Job wizard'),
+          parent: :lab_features_menu,
+          url: 'experimental/job_wizard',
+          after: :host_wizard
 
         register_custom_status HostStatus::ExecutionStatus
         # add dashboard widget
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import { Wizard } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+import history from 'foremanReact/history';
+
+export const JobWizard = () => {
+  const steps = [
+    {
+      name: __('Category and template'),
+      component: <p>Category and template</p>,
+    },
+    { name: __('Target hosts'), component: <p>TargetHosts </p> },
+    { name: __('Advanced fields'), component: <p> AdvancedFields </p> },
+    { name: __('Schedule'), component: <p>Schedule</p> },
+    {
+      name: __('Review details'),
+      component: <p>ReviewDetails</p>,
+      nextButtonText: 'Run',
+    },
+  ];
+  const title = __('Run Job');
+  return (
+    <Wizard
+      onClose={() => history.goBack()}
+      navAriaLabel={`${title} steps`}
+      steps={steps}
+      height="70vh"
+    />
+  );
+};
+
+export default JobWizard;
--- a/webpack/JobWizard/index.js
+++ b/webpack/JobWizard/index.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import { Title, Divider } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
+import { JobWizard } from './JobWizard';
+
+const JobWizardPage = () => {
+  const title = __('Run job');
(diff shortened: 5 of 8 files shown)


MariaAga:
@laviro Can you take a look? :)

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
```
amirfefer, line 4: is this needed?

config/routes.rb
```diff
@@ -44,6 +44,8 @@
   end
   get 'cockpit/redirect', to: 'cockpit#redirect'
 
+  match 'job_wizard', to: 'remote_execution_react#index', :via => [:get]
```
amirfefer, line 47: if this is an experimental feature, don't we want to have `experimental` 'prefix'  in the route?
i.e `/experimental/job_wizard`

webpack/react_app/Routes/ForemanREXRouter.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Switch, Route } from 'react-router-dom';
+
+import routes from './ForemanREXRoutes';
+
+const ForemanTasksRouter = () => (
+  <Switch>
```
amirfefer, line 7: instead of using react-router from scratch, you can now register a plugin's route to foreman core router.
related documentation exists in our storybook - `Client Routing` - or in the PR itself https://github.com/theforeman/foreman/pull/6936

webpack/react_app/components/JobWizard/JobWizard.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Wizard } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export const JobWizard = () => {
+  const steps = [
```
amirfefer, line 6: just a thought for the future - this steps array looks like a great candidate for adding a slot  :)

amirfefer (changes requested):
Thanks @MariaAga, looks awesome :+1: 
I left a few inline comments

webpack/react_app/components/JobWizard/JobWizard.js
MariaAga: Added a task for the tracker :) https://projects.theforeman.org/issues/31248

.prettierrc
MariaAga: We have it in foreman and foreman tasks as well, it helps with formatting

amirfefer: I agree, I just wonder if we can have this in our foreman-js linter meta package

config/routes.rb
MariaAga: Added the prefix

webpack/react_app/Routes/ForemanREXRouter.js
MariaAga: Now it's much cleaner, thanks!

.prettierrc
MariaAga: https://github.com/theforeman/foreman-js/blob/914bdbcba2078436efeef8805ca68011ff2c4c05/packages/eslint-plugin-foreman/lib/src/core.js#L20-L24
Looks like it's there but isn't enforced by prettier

lib/foreman_remote_execution/engine.rb
```diff
@@ -53,6 +53,7 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_remote_execution.register_plugin', before: :finisher_hook do |_app|
       Foreman::Plugin.register :foreman_remote_execution do
+        register_global_js_file 'routes'
```
laviro, line 56: can we rename the file to `global` in case we will want to put more things that are not routes in there?

laviro (commented):
works great, nice start!
left just one nitpick.. LGTM beside of that

lib/foreman_remote_execution/engine.rb
MariaAga: Done!

laviro (approved):

webpack/react_app/components/JobWizard/index.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Title, Divider } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
+import { JobWizard } from './JobWizard';
+
+const JobWizardPage = () => {
```
ezr-ondrej, line 7: In other plugins I see folders `Components` and `Pages`.
I'm not advocating either, but I would like to set a standardized folder structure, so my question:
Do we want to keep Pages and components separate or not?

MariaAga: As it's a page that has one component I like to keep them in the same folder. @laviro what do you think?

laviro: I guess it's up to the plugin maintainers,
though I like the convention of `/webpack/${PAGE_NAME}` e.g: `webpack/JobWizard`

And then in routes registry point to this folder's index file

laviro: @MariaAga yes, inside the main page folder I would use the components folder

MariaAga: Moved to `webpack/JobWizard`, thanks @laviro!

amirfefer (approved):
Thanks @MariaAga ! Looks great 👍

lib/foreman_remote_execution/engine.rb
```diff
@@ -138,6 +139,13 @@ class Engine < ::Rails::Engine
           parent: :monitor_menu,
           after: :audits
 
+        menu :labs_menu, :job_wizard,
+          :url_hash => { :controller => 'job_wizard', :action => :index },
+          caption: N_('Job wizard'),
+          parent: :lab_features_menu,
+          :url => 'experimental/job_wizard',
+          after: :host_wizard
```
adamruzicka, line 147: Please don't mix the hash syntax

webpack/JobWizard/JobWizard.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Wizard } from '@patternfly/react-core';
```
adamruzicka, line 2: The wizard shows a "cancel" button. What will it do?

lib/foreman_remote_execution/engine.rb
MariaAga: What should I replace it with? or should I just remove `:url_hash => { :controller => 'job_wizard', :action => :index },`?

adamruzicka: Either have all the key-value pairs as `{caption: 'caption', parent: :lab_features_menu}` or `:caption => 'caption', :parent => :labe_features menu}`. Either one is fine, as long as it is not a mix of both

MariaAga: Oh yeah, that makes sense, changed

webpack/JobWizard/JobWizard.js
MariaAga: I think it should navigate to the previous page, so if the user was in the jobs table it would go there, and if they were in the hosts page it would go there

adamruzicka: Sounds good. Is it doable?

MariaAga: Currently I'm using `history.goBack()` so it should work

MariaAga:
@adamruzicka Can this be merged?

adamruzicka (approved):
Tested once more, works well

adamruzicka:
Thank you @MariaAga, @amirfefer, @laviro & @ezr-ondrej !

PR #547, opened 2020-10-26
Fixes #31180 - create job wizard controller

MariaAga:
instead of adding more functions to the job templates api, create a new controller for the UI needs

### DIFF
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -1,0 +1,18 @@
+class UiJobWizardController < ::Api::V2::BaseController
+  def categories
+    job_categories = resource_scope
+                     .search_for("job_category ~ \"#{params[:search]}\"")
+                     .select(:job_category).distinct
+                     .reorder(:job_category)
+                     .pluck(:job_category)
+    render :json => {:job_categories =>job_categories}
+  end
+
+  def resource_class
+    JobTemplate
+  end
+
+  def action_permission
+    :view_job_templates
+  end
+end
--- a/test/functional/ui_job_wizard_controller_test.rb
+++ b/test/functional/ui_job_wizard_controller_test.rb
@@ -1,0 +1,16 @@
+require 'test_plugin_helper'
+
+class UiJobWizardControllerTest < ActionController::TestCase
+  def setup
+    FactoryBot.create(:job_template, :job_category => 'cat1')
+    FactoryBot.create(:job_template, :job_category => 'cat2')
+    FactoryBot.create(:job_template, :job_category => 'cat2')
+  end
+
+  test 'should respond with categories' do
+    get :categories, :params => {}, :session => set_session_user
+    assert_response :success
+    res = JSON.parse @response.body
+    assert_equal ['cat1','cat2'], res['job_categories']
+  end
+end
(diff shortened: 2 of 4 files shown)


app/controllers/ui_job_wizard_controller.rb
```diff
@@ -0,0 +1,14 @@
+class UiJobWizardController < ::Api::V2::BaseController
+  def categories
+    job_categories = resource_scope.where(['job_category LIKE ?', "%#{params[:search]}%"]).pluck(:job_category).uniq
```
adamruzicka, line 3: `JobTemplate` allows searching by `job_category` using scoped search so you should be able to do `resource_scope.search_for("job_category ~ #{params[:search]}")`

Also since you're after a list of unique job categories, you can let the db calculate it without having to load all the job categories and then removing the duplicates in ruby.
```ruby
resource_scope
  .where(['job_category LIKE ?', "%#{params[:search]}%"])
  .select(:job_category).distinct
  .reorder(:job_category)
  .pluck(:job_category)
```

ezr-ondrej (commented):
Do we even want to load the categories through UI, wouldn't we want to have it in the component's props directly on mount? :thinking:

MariaAga:
@ezr-ondrej we don't want to mount it using rails,but instead using react router :)

ezr-ondrej:
> @ezr-ondrej we don't want to mount it using rails,but instead using react router :)

Right, make sense :+1:
We will want to have all the metadata in one request then though, right?
But it we can add more data later on. Ignore my comment then :upside_down_face:

app/controllers/ui_job_wizard_controller.rb
MariaAga: So should I combine the two 
``` ruby
resource_scope
  .search_for("job_category ~ #{params[:search]}")
  .select(:job_category).distinct
  .reorder(:job_category)
  .pluck(:job_category)
```
or just use the last one? 
``` ruby
resource_scope
  .where(['job_category LIKE ?', "%#{params[:search]}%"])
  .select(:job_category).distinct
  .reorder(:job_category)
  .pluck(:job_category)
```

adamruzicka: Combination would be nice

MariaAga: Done!

adamruzicka:
Could we get the new controller covered with tests?

MariaAga:
@adamruzicka I can try, but this is my first ruby test :)

test/functional/ui_job_wizard_controller_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_plugin_helper'
+
+class UiJobWizardControllerTest < ActionController::TestCase
+  def setup
+    FactoryBot.create(:job_template, :job_category=> 'cat1')
+    FactoryBot.create(:job_template, :job_category=> 'cat2')
+    FactoryBot.create(:job_template, :job_category=> 'cat2')
```
adamruzicka, line 7: Just add spaces before each of the rockets and it will be good to go, not sure why rubocop didn't complain about this

MariaAga: Done!

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga & @ezr-ondrej !

PR #548, opened 2020-11-03
Fixes #31237 - Extend Registration module with :remote_execution_inteface

stejskalleos:
Extending Registration module with option to specify which Host's interface is going to be set as primary for REX.
Follow up to https://github.com/theforeman/foreman/pull/8121, which is also requirement for this PR.

**Features**
* Add `:remote_execution_interface` to allowed arguments & variables
* Extend UI Form with field for `:remote_execution_interface` (_All Hosts > Register Host_)
* Set Host REX interface during host registration

**How to test**
* Register host to the Foreman
```
curl -X GET "https://foreman.example.com/register?location_id=2&organization_id=1" --user admin:changeme | bash
```
* Add multiple interfaces (`dummy1`, `dummy2`, `dummy3`)
* In UI generate command again, but this time with specified `remote_execution_interface`
* Run registration again & check that interface for REX have been set correctly

### DIFF
--- a/app/controllers/foreman_remote_execution/concerns/api/v2/registration_controller_extensions.rb
+++ b/app/controllers/foreman_remote_execution/concerns/api/v2/registration_controller_extensions.rb
@@ -1,12 +1,15 @@
 module ForemanRemoteExecution
   module Concerns
     module Api::V2::RegistrationControllerExtensions
-      extend Apipie::DSL::Concern
-      extend ActiveSupport::Concern
+      module ApipieExtensions
+        extend Apipie::DSL::Concern
 
-      update_api(:global, :host) do
-        param :remote_execution_interface, String, desc: N_("Identifier of the Host interface for Remote execution")
+        update_api(:global, :host) do
+          param :remote_execution_interface, String, desc: N_("Identifier of the Host interface for Remote execution")
+        end
       end
+
+      extend ActiveSupport::Concern
 
       def host_setup_extension
         remote_execution_interface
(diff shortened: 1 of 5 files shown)


app/controllers/foreman_remote_execution/concerns/api/v2/registration_controller_extensions.rb
```diff
@@ -0,0 +1,31 @@
...
+      def host_setup_extension
+        remote_execution_interface
+        super
+      end
+
+      def remote_execution_interface
+        return unless params['remote_execution_interface'].present?
+
+        interfaces = @host.interfaces
+        interfaces.find_by!(identifier: params['remote_execution_interface'])
```
ares, line 23: this results in ugly error message if the interface with such identifier does exist or is not created for any reason (e.g. sub-man does not send facts), proably could be a problem for debian/ubuntu. We should try to find the interface and fail if more descriptive exception message if the interface is nil.

ares (changes requested):
Works well, the user experience for not existing interface should be improved.

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -54,6 +54,17 @@ def execution_interface
       get_interface_by_flag(:execution)
     end
 
+    def set_execution_interface(identifier)
+      if interfaces.find_by(identifier: identifier).nil?
+        msg = (N_("Host's interface '%s' not found.") % identifier)
```
ares, line 59: how about "Interface with the '%s' identifier was specified as a remote execution interface, however the interface was not found on the host. If the interface exists, it needs to be created in Foreman during the registration."

ares (approved):
Works nicely, thanks @stejskalleos, merging!

PR #549, opened 2020-11-05
Fixes #31267 - Add recent jobs card to host details page

amirfefer:
This adds a card to the experimental host details page
recent-joba

- [x] needs https://github.com/theforeman/foreman/pull/8086

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "@theforeman/vendor-dev": "^4.14.0",
     "babel-eslint": "^10.0.0",
     "eslint": "^6.8.0",
-    "prettier": "^1.19.1"
+    "prettier": "^1.19.1",
+    "@patternfly/react-catalog-view-extension": "^4.8.126"
   },
   "peerDependencies": {
     "@theforeman/vendor": "^8.3.0"
--- a/webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
+++ b/webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
@@ -1,0 +1,87 @@
+/* eslint-disable camelcase */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import Skeleton from 'react-loading-skeleton';
+import ElipsisWithTooltip from 'react-ellipsis-with-tooltip';
+
+import { Grid, GridItem } from '@patternfly/react-core';
+import {
+  OkIcon,
+  ErrorCircleOIcon,
+} from '@patternfly/react-icons/dist/js/icons';
+import {
+  PropertiesSidePanel,
+  PropertyItem,
+} from '@patternfly/react-catalog-view-extension';
+import { ArrowIcon } from '@patternfly/react-icons';
+
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import CardItem from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
+import RelativeDateTime from 'foremanReact/components/common/dates/RelativeDateTime';
+import { translate as __ } from 'foremanReact/common/I18n';
+import './styles.css';
+
+const RecentJobsCard = ({ hostDetails: { name } }) => {
+  const jobsUrl =
+    name && `/api/job_invocations?search=host%3D${name}&per_page=3`;
+  const {
+    response: { results: jobs },
+  } = useAPI('get', jobsUrl);
+
+  const iconMarkup = status => {
+    if (status === 1) return <ErrorCircleOIcon color="#C9190B" />;
+    return <OkIcon color="#3E8635" />;
+  };
+
+  return (
+    <CardItem
+      header={
+        <span>
+          {__('Recent Jobs')}{' '}
+          <a href={`/job_invocations?search=host+%3D+${name}`}>
+            <ArrowIcon />
+          </a>
+        </span>
+      }
+    >
+      <PropertiesSidePanel>
+        {jobs?.map(({ status, status_label, id, start_at, description }) => (
+          <PropertyItem
+            key={id}
+            label={
+              description ? (
+                <Grid>
+                  <GridItem span={8}>
+                    <ElipsisWithTooltip>{description}</ElipsisWithTooltip>
+                  </GridItem>
+                  <GridItem span={1}>{iconMarkup(status)}</GridItem>
+                  <GridItem span={3}>{status_label}</GridItem>
+                </Grid>
+              ) : (
+                <Skeleton />
+              )
+            }
+            value={
+              start_at ? (
+                <a href={`/job_invocations/${id}`}>
+                  <RelativeDateTime date={start_at} />
+                </a>
+              ) : (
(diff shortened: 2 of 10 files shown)


webpack/extend/CardItem.js
```diff
@@ -0,0 +1,86 @@
...
+                    </GridItem>
+                    <GridItem span={2}>{iconMarkup(status)}</GridItem>
+                  </Grid>
+                ) : (
+                  <Skeleton />
+                )
+              }
+              value={
+                start_at ? (
+                  <a href={`/foreman_tasks/tasks/${id}`}>
```
ares, line 65: I think we should rather link job invocation detail page than the task itself. Task page is more for debugging purposes. Also I'm thinking the link should probably be rather the job description than the date, but not a strong opinion on this.

webpack/extend/CardItem.js
```diff
@@ -0,0 +1,86 @@
...
+    selectAPIResponse(state, 'HOST_DETAILS_JOBS')
+  );
+
+  const iconMarkup = (status) => {
+    if (status === 1) return <ErrorCircleOIcon style={{ color: 'red' }} />;
+    return <OkIcon style={{ color: 'green' }} />;
+  };
+
+  return (
+    <CardItem header="Recent Jobs">
```
ares, line 46: should this be extracted for the translation?

webpack/extend/CardItem.js
ares: and one more thing, how do we handle permsisions? We should display the card only if user has view_job_invocations permission. Or we can always display it and only authorize the link (does a user have a permission to view the specific job).

webpack/extend/CardItem.js
```diff
@@ -0,0 +1,86 @@
...
+      <PropertiesSidePanel>
+        {jobs?.results?.map(
```
ares, line 48: how does the card look if we didn't find any job for the host? I think we should display some "No jobs for this host found" text :-)

ares (commented):
very nice draft, this card is IMHO pretty useful, I left few comments from when I was going through the code

webpack/extend/CardItem.js
amirfefer: We have a pf4 empty state for this case :)
How about a general empty state for cards on the host details page?
I've opened an [issue](https://projects.theforeman.org/issues/31316) for this.

webpack/extend/CardItem.js
amirfefer: The data is fetched atm via API, I guess if a user doesn't have permission for  `view_job_invocations`,  the given data will be empty, right? If so we can render an empty state, or not render it at all.

webpack/react_app/extend/fills.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { addGlobalFill } from 'foremanReact/components/common/Fill/GlobalFill';
+import RecentJobsCard from '../components/RecentJobsCard';
+
+export default addGlobalFill(
+  'details-cards',
+  '[REX]host-details-latest-jobs',
```
ezr-ondrej, line 7: Can we loose the `[]` it seems weird :thinking: 
```suggestion
  'REX-host-details-latest-jobs',
```

webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
```diff
@@ -0,0 +1,100 @@
...
+} from '@patternfly/react-icons/dist/js/icons';
+import {
+  PropertiesSidePanel,
+  PropertyItem,
+} from '@patternfly/react-catalog-view-extension';
+import ArrowIcon from '@patternfly/react-icons/dist/js/icons/arrow-icon';
+
+import { get } from 'foremanReact/redux/API';
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+import CardItem from 'foremanReact/components/common/CardTemplate/CardItem';
```
ezr-ondrej, line 22: I really feel this import is weird, it would feel more natural to have a something like `HostDetailCardItem` or `HostDetails/cards/AccordeonCardItem`

webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
```diff
@@ -0,0 +1,100 @@
...
+
+import { Grid, GridItem } from '@patternfly/react-core';
+import {
+  OkIcon,
+  ErrorCircleOIcon,
+} from '@patternfly/react-icons/dist/js/icons';
+import {
+  PropertiesSidePanel,
+  PropertyItem,
+} from '@patternfly/react-catalog-view-extension';
```
ezr-ondrej, line 17: This will probably need a foreman-js dependency, release and packaging, right? :thinking:

webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
ezr-ondrej: Thanks for updating it, now we need to update it here :)

webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
amirfefer: yes :)
a PR already exists in foreman-js

amirfefer:
@ezr-ondrej, thanks! updated
also, I added `useAPI` hook.

adamruzicka:
@amirfefer Sorry for the delay. There seems to be a file conflict now, could you rebase please?

adamruzicka:
I fixed the conflict. @ezr-ondrej could I get your blessing on this?

package.json
```diff
@@ -28,7 +28,8 @@
     "@theforeman/vendor-dev": "^4.14.0",
     "babel-eslint": "^10.0.0",
     "eslint": "^6.8.0",
-    "prettier": "^1.19.1"
+    "prettier": "^1.19.1",
+    "@patternfly/react-catalog-view-extension": "^4.8.126"
```
ezr-ondrej, line 32: Do we have this in foreman-js.
We've already talked about it, but there is just not the PR linked and I believe that if we have it in `foreman-js`, we can remove it from here, right?

ezr-ondrej (commented):
I'm fine with the code, but the explicit dependency make me unease :worried:

package.json
adamruzicka: It should be available https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/package.json#L28 and seems to work without it. Dropped

adamruzicka: Although eslint is not happy about it
```
foreman-remote-execution/webpack/react_app/components/RecentJobsCard/RecentJobsCard.js
  16:8  error  '@patternfly/react-catalog-view-extension' should be listed in the project's dependencies. Run 'npm i -S @patternfly/react-catalog-view-extension' to add it  import/no-extraneous-dependencies
  ```

adamruzicka:
Thank you @amirfefer & @ezr-ondrej !

PR #552, opened 2020-12-02
Update GH actions to theforeman SOTA

ezr-ondrej:
Update RuboCop through theforeman-rubocop gem.
Splits the ruby and js tests
Uses `bundler-cache` for gems caching.
Uses `--format github` to annotate code with the offenses.


adamruzicka:
Does it run the same test with different seed twice https://github.com/theforeman/foreman_remote_execution/pull/552/checks?check_run_id=1489453008#step:11:27 ? Once from line 27 to 39, second time from 40 to 50? The numbers of runs/assertions/failures/errors/skip match, which would be an interesting coincidence :thinking:

ezr-ondrej:
Yeah, I can see it, but I can't figure out what's going on. It's happening to me locally as well though, it runs the core tests with `TEST=<path>` multiple times for some reason. I'll try to figure out why.

adamruzicka:
Well, since running tests twice was not introduced by this change, I'd say let's get this in a figure it out later.

.github/workflows/js_ci.yml
```diff
@@ -0,0 +1,29 @@
...
+      fail-fast: false
+      matrix:
+        node-version: [12]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version:  ${{ matrix.node-version }}
+      - name: Nmp install
```
adamruzicka, line 21: `s/mp/pm`

.github/workflows/ruby_ci.yml
```diff
@@ -12,16 +16,17 @@ jobs:
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.6
-      - name: Setup
-        run: |
-          gem install bundler
-          bundle install --jobs=3 --retry=3
+          ruby-version: 2.5
```
adamruzicka, line 19: Why the downgrade from 2.6 to 2.5?

ezr-ondrej: we can go with 2.7, but 2.6 is actually not used in our environments, as I've recently found out, so we should test on 2.5 or 2.7

adamruzicka: I'm fine with it, I was just curious what was the intention behind it

adamruzicka (approved):
ACK

adamruzicka:
Thank you @ezr-ondrej !

PR #553, opened 2020-12-03
Fixes #31441 - Add basic page for advanced in job wizard

MariaAga:
^ This is a fun title 
Depends on https://github.com/theforeman/foreman_remote_execution/pull/551

Only including the basic `Label: value` fields.
Not including Description and template fields

Screenshot from 2020-12-03 15-38-27


webpack/react_app/components/JobWizard/steps/AdvancedFields/Fields.js
```diff
@@ -0,0 +1,215 @@
...
+      labelIcon={helpLabel(
+        __(
+          'Distribute execution over N seconds. If this is set and proxy batch triggering is enabled, then tasks are triggered on the smart proxy in batches of size 1.'
+        ),
+        'time-span'
+      )}
+      fieldId="time-span"
+      validated={validated}
+      helperTextInvalid={__(
+        'Concurrency level has to be a number, and at least 1'
```
MariaAga, line 179: Need to change to Time span :sweat_smile:

webpack/react_app/components/JobWizard/JobWizard.js
```diff
@@ -0,0 +1,82 @@
...
+  const [category, setCategory] = useState('');
+  const [advancedValue, setAdvancedValue] = useState({});
+  const dispatch = useDispatch();
+
+  const setDefaults = response => {
+    const responseJob = response.data;
+    setAdvancedValue({
+      ...advancedValue,
+      effectiveUserValue: responseJob.effective_user?.value || '',
+      timeoutToKill: responseJob.job_template?.executionTimeoutInterval || '',
```
MariaAga, line 21: Need to change to camel case

webpack/react_app/components/JobWizard/steps/form/FormHelpers.js
```diff
@@ -0,0 +1,19 @@
...
+import { Popover } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export const helpLabel = (text, id) => {
+  if (!text) return null;
+  return (
+    <Popover id={`${id}-help`} bodyContent={text} aria-label="help-text">
+      <button
+        aria-label={__('More info for name field')}
```
MariaAga, line 11: This is not the correct label

adamruzicka:
If the page contains too many fields, a scrollbar appears, even if there's more space on the page. Could we make the form grow instead of doing this?

image

@laviro Could you please take a look?

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -8,6 +8,14 @@ def categories
     render :json => {:job_categories =>job_categories}
   end
 
+  def template
+    job_template = JobTemplate.find_by(id: params[:id])
```
adamruzicka, line 12: Probably an edge case, but we should check the user has necessary permissions.

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -8,6 +8,14 @@ def categories
...
+    render :json => {
+      :job_template => job_template,
+      :effective_user => job_template.effective_user,
```
adamruzicka, line 15: Is it necessary to extract the effective user out of a job template? In other words, can't the frontend code just take it from the template?

webpack/JobWizard/JobWizard.js
```diff
@@ -23,23 +50,34 @@ export const JobWizard = () => {
...
       name: __('Advanced fields'),
-      component: <p> AdvancedFields </p>,
-      canJumpTo: !!jobTemplate,
+      component: (
+        <AdvancedFields
+          advancedValue={advancedValue}
+          setAdvancedValue={newValue => {
+            setAdvancedValue(currentAdvancedValue => ({
+              ...currentAdvancedValue,
+              ...newValue,
```
laviro, line 63: maybe it should be `setAdvancedValues`? and `newValues`?

webpack/JobWizard/JobWizardSelectors.js
```diff
@@ -19,3 +23,6 @@ export const selectJobCategories = state =>
 
 export const selectJobCategoriesStatus = state =>
   selectAPIStatus(state, JOB_CATEGORIES);
+
+export const selectJobTemplate = state =>
+  selectAPIResponse(state, JOB_TEMPLATE) || {};
```
laviro, line 28: `selectAPIResponse` already does it for you, so no need for the `|| {}` part

webpack/JobWizard/__tests__/JobWizard.test.js
```diff
@@ -1,20 +1,29 @@
...
-  describe('rendering', () =>
-    testComponentSnapshotsWithFixtures(JobWizard, fixtures));
+describe('JobWizard rendering', () => {
+  describe('render', () => {
+    const component = mount(
+      <Provider store={store}>
+        <JobWizard />
+      </Provider>
+    );
+    expect(component).toMatchSnapshot();
```
laviro, line 26: how about instead of a snapshot, you can test if the tabs exist and try to jump between them?

webpack/JobWizard/__tests__/integration.test.js
```diff
@@ -0,0 +1,138 @@
...
+import { mount } from '@theforeman/test';
+import * as api from 'foremanReact/redux/API';
+import { JobWizard } from '../JobWizard';
+import * as selectors from '../JobWizardSelectors';
+
+jest.spyOn(api, 'get');
+jest.spyOn(selectors, 'selectJobTemplate');
+jest.spyOn(selectors, 'selectJobTemplates');
+jest.spyOn(selectors, 'selectJobCategories');
+jest.spyOn(selectors, 'selectJobCategoriesStatus');
```
laviro, line 13: just wondering if there is a way to pass a mocked store with this values so the selectors would work the same and grab what they need from the predefined store.

webpack/JobWizard/__tests__/integration.test.js
```diff
@@ -0,0 +1,138 @@
...
+    );
+    expect(wrapper.find('.pf-c-wizard__nav-link.pf-m-disabled')).toHaveLength(
+      4
+    );
+    selectors.selectJobCategoriesStatus.mockImplementation(() => 'RESOLVED');
+    expect(store.getActions()).toMatchSnapshot('initial');
+
+    selectors.selectJobTemplate.mockImplementation(() => jobTemplate);
+    wrapper.find('.pf-c-button.pf-c-select__toggle-button').simulate('click');
+    wrapper.find('.pf-c-select__menu-item').simulate('click');
```
laviro, line 91: from what I remember from past implementation, sometimes something breaks when enzyme or jest upgrades and these clicks won't always work.. I would suggest triggering the actual action, e.g: `wrapper.instance().handleClick()`

webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
```diff
@@ -0,0 +1,87 @@
...
+  PasswordField,
+  KeyPassphraseField,
+  EffectiveUserPasswordField,
+  ConcurrencyLevelField,
+  TimeSpanLevelField,
+} from './Fields';
+
+export const AdvancedFields = ({ advancedValue, setAdvancedValue }) => {
+  const jobTemplate = useSelector(selectJobTemplate);
+  const effectiveUser = jobTemplate.effective_user;
```
laviro, line 18: didn't you also pass it from the server https://github.com/theforeman/foreman_remote_execution/pull/553/files#diff-494fb439c4dcbef79b8470f33ba943aa649b2d8e698a9bd26c5353e9e33158f7R15

I guess no need to get it directly again from `jobTemplate`

laviro (commented):
Thanks @MariaAga! didn't test it out yet but added a few comments

app/controllers/ui_job_wizard_controller.rb
laviro: @adamruzicka is there a before_action trick for that?

adamruzicka: Yes and no. Usually a controller is centered around a resource and then you can use such tricks to reduce code duplication. However in our case, the controller is not really centered around a resource (template in this case), so just doing `JobTemplate.authorized.find_by(...)` would be fine.

Also, `find_by` returns `nil` if no resource is found, don't we want `find` which raises an exception if the resource cannot be found?

package.json
```diff
@@ -28,7 +28,8 @@
     "@theforeman/vendor-dev": "^4.14.0",
     "babel-eslint": "^10.0.0",
     "eslint": "^6.8.0",
-    "prettier": "^1.19.1"
+    "prettier": "^1.19.1",
+    "redux-mock-store": "^1.2.2"
```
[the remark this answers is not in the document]
MariaAga: Nope, apparently its not allowed:
`Trailing commas in objects were only introduced in ECMAScript 5. As JSON is based on JavaScript's syntax prior to ES5, trailing commas are not allowed in JSON.`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

webpack/JobWizard/__tests__/integration.test.js
MariaAga: I tried using `wrapper.instance()` but it returns null, because:  `With React 16 and above, instance() returns null for stateless functional components.`
https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/instance.html
What can cause `wrapper.find('.pf-c-select__menu-item').simulate('click');` to break?

laviro: Alright, but you can use `wrapper.props()`? if I recall and then trigger the method you wanted
yes, also if the DOM changes a bit it can break your test while the button is still there...
anyway it's a better practice to trigger the actual method you want

package.json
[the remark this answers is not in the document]
laviro: hmm... I don't like that 🤣 
even in ruby we have the rule to add trailing comma at the end of JSON

app/controllers/ui_job_wizard_controller.rb
MariaAga: It's not in a the template because its a function I think:
https://github.com/theforeman/foreman_remote_execution/blob/f7e711f6aa54300497f967a6d90576133ab87863/app/models/job_template.rb#L131

webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
MariaAga: This is how I pass it from the server, 
```
jobTemplate= {
    job_template : { ... },
    effective_user : { ... }
}
```

webpack/JobWizard/__tests__/integration.test.js
MariaAga: I mocked the selectors twice in the test, once as their initial state, then click something, check that the correct action was called, and mock the selectors a second time with the response

webpack/JobWizard/__tests__/JobWizard.test.js
MariaAga: Added test in the integration test

app/controllers/ui_job_wizard_controller.rb
adamruzicka: Nevermind then

MariaAga:
rebased with the added api error handling

MariaAga:
rubocop failures aren't related

adamruzicka:
@MariaAga could you rebase plase? It should make the rubocop issues go away.

@laviro anything left to do here or are we good to merge?

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -8,6 +8,14 @@ def categories
     render :json => {:job_categories =>job_categories}
   end
 
+  def template
+    job_template = JobTemplate.authorized.find(params[:id])
+    render :json => {
```
laviro, line 13: in case the `job_template` doesn't exist, let's return another JSON with status `:not_found`

webpack/__mocks__/foremanReact/redux/API/APISelectors.js
```diff
@@ -1,3 +1,3 @@
 export const selectAPIStatus = () => 'RESOLVED';
-export const selectAPIResponse = state => state;
+export const selectAPIResponse = (state, key) => state[key] || {};
 export const selectAPIErrorMessage = state => state.error;
```
laviro, line 3: lets add all of the selectors for future needs, copied from Foreman:

```js
import { STATUS } from '../../constants';

export const selectAPI = state => state.API;

export const selectAPIByKey = (state, key) => selectAPI(state)[key] || {};

export const selectAPIStatus = (state, key) =>
  selectAPIByKey(state, key).status;

export const selectAPIPayload = (state, key) =>
  selectAPIByKey(state, key).payload || {};

export const selectAPIResponse = (state, key) =>
  selectAPIByKey(state, key).response || {};

export const selectAPIError = (state, key) =>
  selectAPIStatus(state, key) === STATUS.ERROR
    ? selectAPIResponse(state, key)
    : null;

export const selectAPIErrorMessage = (state, key) => {
  const error = selectAPIError(state, key);
  return error && error.message;
};
```

webpack/JobWizard/__tests__/fixtures.js
```diff
@@ -0,0 +1,33 @@
+export const jobTemplates = [
+  {
+    id: 178,
+    name: 'Run Command - Ansible Default',
+    job_category: 'Ansible Commands',
+    provider_type: 'Ansible',
+    snippet: false,
+    description_format: 'Run %{command}',
+  },
+];
```
laviro, line 10: could become:

```suggestion
export const jobTemplates = [ jobTemplate ];
```

laviro (commented):
Awesome work @MariaAga! thanks for doing the integration test!

app/controllers/ui_job_wizard_controller.rb
adamruzicka: It kinda works even now, but it would be nice if it could say `job_template` instead of `ui_job_wizard`

```
curl -i -u admin:changeme http://localhost:3000/ui_job_wizard/template/91                                                                                  
HTTP/1.1 404 Not Found
Foreman_version: 2.6.0-develop
Foreman_api_version: 2
Foreman_current_organization: ; ANY
Foreman_current_location: ; ANY
Content-Type: application/json; charset=utf-8
-----B<-----MORE HTTP HEADERS-----B<-----

{
  "error": {"message":"Resource ui_job_wizard not found by id '91'"}
}
```

laviro: nice, though if I change it and access from the UI, it kinda crashes the page - at least on my end :)

MariaAga: In the ui is shows as: 
Screenshot from 2021-05-25 12-21-56
@adamruzicka, how can I change the resource name in the error?

adamruzicka: It is taken from `#resource_name` method on the controller. IIRC by default it guesses the name from the controller's name.

MariaAga: The user can't access the next step if there is no template selected. A user can create job categories only in the job template form so there can't be empty categories, and if there are then it's an error.

MariaAga:
`canJumpTo: isTemplate,` @laviro This should handle that

laviro:
> `canJumpTo: isTemplate,` @laviro This should handle that

I meant things like the password fields, usually they are required. but do you want to revisit it later on a follow-up?

MariaAga:
Oh I see, I think we should "block" submit so users can move between input steps until they fill the required fields. I'll open a redmine task to add this in a follow-up

adamruzicka:
Now that the issues for followups are created, can this go in?

laviro (approved):
Thanks @MariaAga !

adamruzicka:
Thank you @MariaAga & @laviro !

PR #554, opened 2020-12-06
Fixes #31459 - Add advanced template input to job wizard

MariaAga:
Depends on https://github.com/theforeman/foreman_remote_execution/pull/553

render the inputs from `template_inputs_with_foreign` into the advanced field step
New:
Screenshot from 2020-12-06 15-24-44

old:
Screenshot from 2020-12-06 15-25-10

### DIFF
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -18,21 +18,19 @@ export const JobWizard = () => {
 
   const setDefaults = useCallback(response => {
     const responseJob = response.data;
-    const templateValues = {};
-    const inputs = responseJob.template_inputs_with_foreign;
-    if (inputs) {
-      inputs
-        .filter(input => input.advanced)
-        .forEach(input => {
-          templateValues[input.name] = input?.default || '';
-        });
+    const advancedTemplateValues = {};
+    const advancedInputs = responseJob.advanced_template_inputs;
+    if (advancedInputs) {
+      advancedInputs.forEach(input => {
+        advancedTemplateValues[input.name] = input?.default || '';
+      });
     }
-    setAdvancedValues({
-      ...advancedValues,
+    setAdvancedValues(currentAdvancedValues => ({
+      ...currentAdvancedValues,
       effectiveUserValue: responseJob.effective_user?.value || '',
       timeoutToKill: responseJob.job_template?.executionTimeoutInterval || '',
-      templateValues,
-    });
+      templateValues: advancedTemplateValues,
+    }));
   }, []);
   useEffect(() => {
     if (jobTemplateID) {
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -18,9 +18,20 @@ export const JobWizard = () => {
 
   const setDefaults = useCallback(response => {
     const responseJob = response.data;
+    const templateValues = {};
+    const inputs = responseJob.template_inputs_with_foreign;
+    if (inputs) {
+      inputs
+        .filter(input => input.advanced)
+        .forEach(input => {
+          templateValues[input.name] = input?.default || '';
+        });
+    }
     setAdvancedValues({
+      ...advancedValues,
       effectiveUserValue: responseJob.effective_user?.value || '',
-      timeoutToKill: responseJob.job_template.execution_timeout_interval || '',
+      timeoutToKill: responseJob.job_template?.executionTimeoutInterval || '',
+      templateValues,
     });
   }, []);
   useEffect(() => {
--- a/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
+++ b/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
@@ -30,9 +30,7 @@ export const AdvancedFields = ({ advancedValues, setAdvancedValues }) => {
         <TemplateInputsFields
           inputs={templateInputs}
           value={advancedValues.templateValues}
-          setValue={newValue =>
-            setAdvancedValues({ ...advancedValues, templateValues: newValue })
-          }
+          setValue={newValue => setAdvancedValues({ templateValues: newValue })}
         />
         {effectiveUser?.overridable && (
           <EffectiveUserField
--- a/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
+++ b/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import { useSelector } from 'react-redux';
 import { Title, Form } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
-import { selectJobTemplate } from '../../JobWizardSelectors';
+import {
+  selectEffectiveUser,
+  selectTemplateInputs,
+} from '../../JobWizardSelectors';
 import {
   EffectiveUserField,
   TimeoutToKillField,
@@ -12,17 +15,25 @@ import {
   EffectiveUserPasswordField,
   ConcurrencyLevelField,
   TimeSpanLevelField,
+  TemplateInputsFields,
 } from './Fields';
 
 export const AdvancedFields = ({ advancedValues, setAdvancedValues }) => {
-  const jobTemplate = useSelector(selectJobTemplate);
-  const effectiveUser = jobTemplate.effective_user;
+  const effectiveUser = useSelector(selectEffectiveUser);
+  const templateInputs = useSelector(selectTemplateInputs);
   return (
     <>
       <Title headingLevel="h2" className="advanced-fields-title">
         {__('Advanced Fields')}
       </Title>
-      <Form>
+      <Form id="advanced-fields-job-template">
+        <TemplateInputsFields
+          inputs={templateInputs}
+          value={advancedValues.templateValues}
+          setValue={newValue =>
+            setAdvancedValues({ ...advancedValues, templateValues: newValue })
+          }
+        />
         {effectiveUser?.overridable && (
           <EffectiveUserField
             value={advancedValues.effectiveUserValue}
--- a/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
+++ b/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
@@ -5,7 +5,7 @@ import { Title, Form } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 import {
   selectEffectiveUser,
-  selectTemplateInputs,
+  selectAdvancedTemplateInputs,
 } from '../../JobWizardSelectors';
 import {
   EffectiveUserField,
@@ -20,7 +20,7 @@ import {
 
 export const AdvancedFields = ({ advancedValues, setAdvancedValues }) => {
   const effectiveUser = useSelector(selectEffectiveUser);
-  const templateInputs = useSelector(selectTemplateInputs);
+  const templateInputs = useSelector(selectAdvancedTemplateInputs);
   return (
     <>
       <Title headingLevel="h2" className="advanced-fields-title">
--- a/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
+++ b/webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
@@ -26,7 +26,7 @@ export const AdvancedFields = ({ advancedValues, setAdvancedValues }) => {
       <Title headingLevel="h2" className="advanced-fields-title">
         {__('Advanced Fields')}
       </Title>
-      <Form id="advanced-fields-job-template">
+      <Form id="advanced-fields-job-template" autoComplete="off">
         <TemplateInputsFields
           inputs={templateInputs}
           value={advancedValues.templateValues}
--- a/webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
+++ b/webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
@@ -4,6 +4,7 @@ import configureMockStore from 'redux-mock-store';
 import * as patternfly from '@patternfly/react-core';
 import { mount } from '@theforeman/test';
 import { AdvancedFields } from '../AdvancedFields';
+import { jobTemplateResponse } from '../../../__tests__/fixtures';
 
 jest.spyOn(patternfly, 'FormGroup');
 patternfly.FormGroup.mockImplementation(props => (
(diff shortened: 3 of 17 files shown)


webpack/react_app/components/JobWizard/JobWizard.js
```diff
@@ -1,36 +1,45 @@
...
+      inputs
+        .filter(input => input.advanced)
+        .forEach(input => {
+          templateValues[input.name] = input?.default;
+        });
+    }
+    setAdvancedValue({
+      ...advancedValue,
+      effectiveUserValue: responseJob.effective_user?.value || '',
+      timeoutToKill: responseJob.job_template?.executionTimeoutInterval || '',
```
MariaAga, line 30: Need to change to camel case

webpack/react_app/components/JobWizard/JobWizard.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const setDefaults = response => {
+    const responseJob = response.data;
+    const templateValues = {};
+    const inputs = responseJob.template_inputs_with_foreign;
+    if (inputs) {
+      inputs
+        .filter(input => input.advanced)
+        .forEach(input => {
+          templateValues[input.name] = input?.default;
```
MariaAga, line 24: change to `input?.default || '';`

adamruzicka:
In the advanced fields step the form gets clipped and shows an internal scrollbar, even though there is still space left on the page. Could we make it use the free space instead of what it does now?

image

MariaAga:
Wizard will now fill the whole page instead of just part of it
Screenshot from 2021-02-09 17-00-02

Screenshot from 2021-02-09 17-00-08

webpack/JobWizard/steps/AdvancedFields/AdvancedFields.js
```diff
@@ -0,0 +1,96 @@
...
+  TimeSpanLevelField,
+  TemplateInputsFields,
+} from './Fields';
+
+export const AdvancedFields = ({ advancedValue, setAdvancedValue }) => {
+  const jobTemplate = useSelector(selectJobTemplate);
+  const effectiveUser = jobTemplate.effective_user;
+  const templateInputs = jobTemplate.template_inputs_with_foreign;
+  return (
+    <Form id="advanced-fields-job-template">
```
MariaAga, line 22: Need to add a title to the step

MariaAga:
Rebasing and will remove draft soon

MariaAga:
The search that is used in the template input doesn't have bookmarks, but it will, in this pr: https://github.com/theforeman/foreman_remote_execution/pull/579

webpack/JobWizard/steps/form/__tests__/Formatter.test.js
```diff
@@ -0,0 +1,77 @@
...
+      advanced: true,
+      value_type: 'plain',
+      resource_type: 'ansible_roles',
+      default: '',
+      hidden_value: false,
+    };
+    expect(shallow(formatter(props, {}, jest.fn()))).toMatchSnapshot();
+  });
+  it('render search input', () => {
+    // patternfly.FormGroup.mockImplementation(props => <div>{props.children}</div>);
```
adamruzicka, line 60: what's this?

adamruzicka (commented):
A couple of nits:
1) Are validations planned? Time span, timeout to kill, concurrency level and time span are expected to be either unset or an integer. Currently we can set anything in there

2) Odd keyboard interaction: Focus on any field (let's say concurrency level), fill a value in there, press enter. Not sure what should happen, but currently help for the first field on the page shows up which is a bit odd

3) When filling in a password and going to the next page, browser prompts if I want to save he password. Not sure how useful this is. I know even the old form does this, but maybe now would be the time to reconsider it

webpack/JobWizard/steps/form/__tests__/Formatter.test.js
MariaAga: leftovers

MariaAga:
1. Validations on submit are planned here: https://projects.theforeman.org/issues/32667
About Time span, timeout to kill, concurrency level and time span, a user can only input number values in there, can you input letters in it?
The number fields should have an up\down arrows like this:
Screenshot from 2021-06-10 11-37-15

adamruzicka:
> Validations on submit are planned here: https://projects.theforeman.org/issues/32667

Ok, let's not worry about it now and deal with that later

> About Time span, timeout to kill, concurrency level and time span, a user can only input number values in there, can you input letters in it?
The number fields should have an up\down arrows like this:

They are number fields, they have the up/down arrows, but nothing is preventing me from typing letter into them.

> Which browser are you using?

Firefox 88.0.1

MariaAga:
I hope this fixed the password and numbers issue. My firefox keeps crashing so I can't check it well there.

MariaAga:
I think the issue was that "enter = submit" so it tried to click the first submit button it saw. I changed the help button type to be specifically button and not submit.

adamruzicka:
> I hope this fixed the password and numbers issue

Works well in chrome, firefox still allows me to input anything

MariaAga:
Screenshot from 2021-06-11 14-11-38
Without an error like this? ):

adamruzicka:
Without anything :/

https://user-images.githubusercontent.com/7326770/121685776-a70b4f80-cac0-11eb-9c6f-f70e3a4dd9b9.mp4

MariaAga:
So firefox won't fire `onchange` for number so I changed the input type to text

adamruzicka:
@laviro could you take a look?

laviro:
@adamruzicka yes, actually I started today, but got distracted :P

webpack/JobWizard/JobWizard.js
```diff
@@ -18,9 +18,20 @@ export const JobWizard = () => {
...
     const responseJob = response.data;
+    const templateValues = {};
+    const inputs = responseJob.template_inputs_with_foreign;
+    if (inputs) {
+      inputs
+        .filter(input => input.advanced)
+        .forEach(input => {
+          templateValues[input.name] = input?.default || '';
+        });
+    }
```
laviro, line 29: I feel like this is something that the server should prepare better for the client,
so we don't need to iterate over and over

webpack/JobWizard/JobWizard.js
```diff
@@ -18,9 +18,20 @@ export const JobWizard = () => {
...
     setAdvancedValues({
+      ...advancedValues,
```
laviro, line 31: can the `setAdvancedValues` state will be managed only inside the component,
Looks like `AdvancedFields` could manage itself, without needing to update the parent, and if I am not wrong,
seems that no one else uses this state in the parent.

MariaAga: No because the parent component will send the advanced value to the server on submit

webpack/JobWizard/JobWizard.js
MariaAga: I'm not sure what should go in the server, @adamruzicka any thoughts?

adamruzicka: The api we're using is used only by the wizard, so we're free to do whatever we please. In general we will have to iterate over the data somewhere to get it into the right shape and it doesn't really matter on which side it happens.

In general I would try to minimize the amount of the data that actually needs to go over the wire.

MariaAga: It would be the same amount of data so I guess I can split it in 2. I will still have to pass both of the inputs to the advanced tab as the description field (https://github.com/theforeman/foreman_remote_execution/pull/555) uses the default inputs as well.

MariaAga: Added a commit to split the advanced

laviro (commented):
as we discussed offline, I think for a complex form with relationships between tabs, we should use `ForemanForm` / `Redux` in order to prevent changes in one tab from clearing other tabs, e.g:

take a look at what happens to the `Effective user` & `Timeout to kill` fields when changing the Job Category from another tab - it gets reset, but the password remains.


https://user-images.githubusercontent.com/26363699/124079871-5ac58680-da52-11eb-974b-8925abf0709d.mp4

I think it would make sense to continue with this PR as is because you did great work here,
but as a follow up I would suggest switching to a different form's state managing.

webpack/JobWizard/steps/AdvancedFields/__tests__/__snapshots__/AdvancedFields.test.js.snap
```diff
@@ -186,61 +281,139 @@ exports[`AdvancedFields rendring 1`] = `
...
+                      noVerticalAlign={true}
+                      size="sm"
+                    />
+                  </button>
+                </Popover>
+              }
+            >
+              <div />
+            </mockConstructor>
+          </NumberInput>
```
laviro, line 416: If you want, you can remove this snapshot file so you don't need to maintain it in other PRs as well, in continuous to the discussion in https://community.theforeman.org/t/rethinking-react-redux-folder-structure/24183/8

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
```diff
@@ -4,14 +4,15 @@ import configureMockStore from 'redux-mock-store';
...
+import { jobTemplateResponse } from '../../../__tests__/fixtures';
 
 jest.spyOn(patternfly, 'FormGroup');
 patternfly.FormGroup.mockImplementation(props => (
   <div>{props.navAriaLabel}</div>
 ));
 const mockStore = configureMockStore([]);
 const store = mockStore({
-  JOB_TEMPLATE: { response: { effective_user: { overridable: true } } },
+  JOB_TEMPLATE: { response: jobTemplateResponse },
```
laviro, line 15: If you want, you can remove this snapshot test file so you don't need to maintain it in other PRs as well, in continuous to the discussion in https://community.theforeman.org/t/rethinking-react-redux-folder-structure/24183/8

you can move the integration test into this file

MariaAga:
Please notice that `Effective User` and `Timeout to kill` are supposed to be reset because different job templates have different defaults. If its confusing for the user I can reset all the values.
https://github.com/theforeman/foreman_remote_execution/blob/7e88ad1e4eaa05ca3dfeb70861606525a3757047/webpack/JobWizard/JobWizard.js#L28-L33

laviro (commented):
Overall looks great, thanks @MariaAga !

webpack/JobWizard/JobWizard.js
laviro: > Please notice that `Effective User` and `Timeout to kill` are supposed to be reset because different job templates have different defaults. If its confusing for the user I can reset all the values.

From debugging it, I see added a few console.logs
and I see that the Advanced step gets remounted, e.g:

- changed `Effective User` and `Timeout to kill`,
`// advancedValues = {"effectiveUserValue":"admin","timeoutToKill":"12","templateValues":{}}`
- changing job category
`// in did mount, advancedValues = {}`
`// after useEffect advancedValues = {"effectiveUserValue":"","timeoutToKill":"","templateValues":{"pre_script":"","post_script":""}}`

so I don't think you can retrieve the old values in `advancedValues`

webpack/JobWizard/JobWizard.js
```diff
@@ -18,10 +18,19 @@ export const JobWizard = () => {
...
     const responseJob = response.data;
-    setAdvancedValues({
+    const advancedTemplateValues = {};
+    const advancedInputs = responseJob.advanced_template_inputs;
+    if (advancedInputs) {
+      advancedInputs.forEach(input => {
+        advancedTemplateValues[input.name] = input?.default || '';
+      });
+    }
+    setAdvancedValues(currentAdvancedValues => ({
```
laviro, line 28: can you use the same debounced setState here and also for the fields?
currently, they re-render the parent state quite a lot:

```
advancedValues {"effectiveUserValue":"a","timeoutToKill":"","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"ad","timeoutToKill":"","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"adm","timeoutToKill":"","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"admi","timeoutToKill":"","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"admin","timeoutToKill":"","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"admin","timeoutToKill":"1","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"admin","timeoutToKill":"12","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"admin","timeoutToKill":"12","templateValues":{"pre_script":"","post_script":""}}
JobWizard.js:18 advancedValues {"effectiveUserValue":"admin","timeoutToKill":"12","templateValues":{"pre_script":"","post_script":""}}
```

webpack/JobWizard/JobWizard.js
MariaAga: I changed the set default function to be 
``` js
    setAdvancedValues(currentAdvancedValues => ({
      ...currentAdvancedValues,
      effectiveUserValue:
        responseJob.effective_user?.value ||
        currentAdvancedValues.effectiveUserValue ||
        '',
 ```
 And after changing job templates or category the old effective user was still there

webpack/JobWizard/JobWizard.js
MariaAga: What debounced setstate do you mean?

laviro: you can use `lodash` for that: https://lodash.com/docs/4.17.15#debounce

MariaAga: I tried to use it in the use state but it looks like it prevents the user from typing too fast.
I used 150 ms and when I tried to type in `1234567890-` it only recorded `-`
https://github.com/theforeman/foreman_remote_execution/blob/7e88ad1e4eaa05ca3dfeb70861606525a3757047/webpack/JobWizard/JobWizard.js#L71

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
MariaAga: Will move that in a separate pr, thanks!

webpack/JobWizard/JobWizard.js
laviro: we discussed it offline and decided to do more research on it as a performance improvement,
note: maybe `throttle` func is better here

laviro (approved):
Thanks @MariaAga!

MariaAga:
@adamruzicka test failures look unrelated

adamruzicka:
Thank you @MariaAga & @laviro !

PR #557, opened 2020-12-10
Fixes #26671 - fix exporting foreign input set

ares:
Export of the foreign input set sets the wrong template name. Therefore
reimporting it fails, because it creates a circular dependency. To
reproduce, just export a template with at least one foreign input set
defined and try to import it back (with overwrite enabled).

This patch correctly looks at the target template instead of exported
template itself.


adamruzicka:
Works nicely, APJ

adamruzicka:
Thank you @ares !

PR #559, opened 2020-12-18
fixes #31428 - making execution status clickable

domitea:
This PR introduces a clickable execution status at host detail page that leads to last job invocation of given host

### DIFF
--- a/app/models/host_status/execution_status.rb
+++ b/app/models/host_status/execution_status.rb
@@ -49,6 +49,13 @@ class HostStatus::ExecutionStatus < HostStatus::Status
     end
   end
 
+  def status_link
+    job_invocation = last_stopped_task.parent_task.job_invocations.first
+    return nil unless User.current.can?(:view_job_invocations, job_invocation)
+
+    Rails.application.routes.url_helpers.job_invocation_path(job_invocation)
+  end
+
   class ExecutionTaskStatusMapper
     attr_accessor :task
 


app/models/host_status/execution_status.rb
```diff
@@ -49,6 +49,10 @@ def to_label(options = {})
     end
   end
 
+  def status_link
+    Rails.application.routes.url_helpers.job_invocation_path(last_stopped_task.parent_task.job_invocations.first)
```
ares, line 53: Works well, however it does not check for permissions. User must have `:view_job_invocations` permissions for this job invocation, otherwise we shouldn't make the link clickable. You can either use `authorized` named scope on JobInvocation model or use `User.current.can?` to check that.

ares (changes requested):
Authorization must be added, otherwise works well.

domitea:
@ares thanks for comment, I totally forgot for permission. PR updated

app/models/host_status/execution_status.rb
```diff
@@ -49,6 +49,13 @@ def to_label(options = {})
     end
   end
 
+  def status_link
+    job_invocation = last_stopped_task.parent_task.job_invocations.first
+    return nil unless User.current.can?(:view_job_invocations, job_invocation)
+
+    Rails.application.routes.url_helpers.job_invocation_path(last_stopped_task.parent_task.job_invocations.first)
```
ares, line 56: please use the `job_invocation` variable here as well to not repeat the code

ares (approved):
Works well, thanks @domitea, please add it to the next demo agenda :-)

PR #560, opened 2021-01-15
Fixes #32326 - Require additional permission to execute jobs against system hosts

adamruzicka:
Part of https://community.theforeman.org/t/infrastructure-roles/22001

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -196,6 +196,10 @@ module Actions
       def verify_permissions(host, template_invocation)
         raise _('User can not execute job on host %s') % host.name unless User.current.can?(:view_hosts, host)
         raise _('User can not execute this job template') unless User.current.can?(:view_job_templates, template_invocation.template)
+        infra_facet = host.infrastructure_facet
+        if (infra_facet&.foreman_instance || infra_facet&.smart_proxy_id) && !User.current.can?(:execute_jobs_on_infrastructure_hosts)
+          raise _('User can not execute job on infrastructure host %s') % host.name
+        end
 
         # we don't want to load all template_invocations to verify so we construct Authorizer object manually and set
         # the base collection to current template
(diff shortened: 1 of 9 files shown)


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -196,6 +196,10 @@ def delegated_output
       def verify_permissions(host, template_invocation)
         raise _('User can not execute job on host %s') % host.name unless User.current.can?(:view_hosts, host)
         raise _('User can not execute this job template') unless User.current.can?(:view_job_templates, template_invocation.template)
+        infra_facet = host.infrastructure_facet
+        if (infra_facet&.foreman_instance || infra_facet&.smart_proxy_id) && !User.current.can?(:execute_jobs_on_infrastructure_hosts)
+          raise _('User can not execute job on an infrastructure host %s') % host.name
```
ehelms, line 201:
```suggestion
          raise _('User can not execute job on infrastructure host %s') % host.name
```

adamruzicka:
This is green-ish

lib/foreman_remote_execution/engine.rb
```diff
@@ -183,6 +183,7 @@ class Engine < ::Rails::Engine
           permission :view_template_invocations, { :template_invocations => [:show],
                                                    'api/v2/template_invocations' => [:template_invocations] }, :resource_type => 'TemplateInvocation'
           permission :create_template_invocations, {}, :resource_type => 'TemplateInvocation'
+          permission :execute_jobs_on_infrastructure_hosts, {}
```
ezr-ondrej, line 186: Shouldn't we have some resource assigned to the permission?
If we won't this will be in miscallanous, which I'm not sure users will look for it there, but no strong opinion here.

Also do we want Execution manager to have it? (also no strong opinion about that)

ezr-ondrej (approved):
Thanks @adamruzicka, looks quite ready, just two minor questions about the permission placing, but apart of that looks good :+1:

lib/foreman_remote_execution/engine.rb
adamruzicka: What would be the right resource for this? Host? JobInvocation?

> Also do we want Execution manager to have it?

No. It was explicitly required that the manager should not have this permission.

PR #561, opened 2021-01-25
Fixes #31711 - Extend API params for generating registration command

stejskalleos:
For: https://github.com/theforeman/foreman/pull/8271


adamruzicka (approved):
Works nicely, but let's hold it for a bit until we get at least one 2.4 release out of the door

adamruzicka:
Thank you @stejskalleos !

PR #562, opened 2021-01-27
Fixes #31737 - Mark RunHostsJob as observable

adamruzicka:
requires:
- [x] https://github.com/theforeman/foreman-tasks/pull/605
- [x] tasks release & dependency bump

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -3,6 +3,7 @@ module Actions
     class RunHostJob < Actions::EntryAction
       include ::Actions::Helpers::WithContinuousOutput
       include ::Actions::Helpers::WithDelegatedAction
+      include ::Actions::ObservableAction
 
       middleware.do_not_use Dynflow::Middleware::Common::Transaction
       middleware.use Actions::Middleware::HideSecrets
@@ -16,7 +17,7 @@ module Actions
       end
 
       def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExecutionProxySelector.new, options = {})
-        action_subject(host, :job_category => job_invocation.job_category, :description => job_invocation.description)
+        action_subject(host, :job_category => job_invocation.job_category, :description => job_invocation.description, :job_invocation_id => job_invocation.id)
 
         template_invocation.host_id = host.id
         template_invocation.run_host_job_task_id = task.id
@@ -121,6 +122,26 @@ module Actions
         delegated_output[:exit_status]
       end
 
+      def host_id
+        input['host']['id']
+      end
+
+      def host_name
+        input['host']['name']
+      end
+
+      def job_invocation_id
+        input['job_invocation_id']
+      end
+
+      def job_invocation
+        @job_invocation ||= ::JobInvocation.authorized.find(job_invocation_id)
+      end
+
+      def host
+        @host ||= ::Host.authorized.find(host_id)
+      end
+
       private
 
       def update_host_status
@@ -172,6 +193,10 @@ module Actions
                   '%{fallback_proxy} in settings') % settings
         end
         proxy
+      end
+
+      class Jail < ::Actions::ObservableAction::Jail
+        allow :host_name, :host_id, :host, :job_invocation_id, :job_invocation
       end
     end
   end
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -195,6 +195,18 @@ module Actions
         proxy
       end
 
+      extend ApipieDSL::Class
+      apipie :class, "An action representing execution of a job against a host" do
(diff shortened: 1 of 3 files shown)


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -173,6 +194,10 @@ def determine_proxy!(proxy_selector, provider, host)
         end
         proxy
       end
+
+      class Jail < ::Actions::ObservableAction::Jail
+        allow :host_name, :host_id, :host, :job_invocation_id, :job_invocation
+      end
```
lzap, line 200: Just a crazy idea: do we also want to expose the script itself? Just in case someone want to archive those? On the other hand, these could contain sensitive data, so it's not bad thing to avoid this.

lzap (approved):
Looks nice and clean, utilizing safemode was a very good idea indeed.

I think it's really good idea NOT to expose script itself in safemode, people can still turn off safe mode and get to it if needed.

Gonna test this today/tomorrow. Have a discovery urgent issue I need to verify.

lzap:
FYI there are some test failures, feel free to fix them before I test this.

app/lib/actions/remote_execution/run_host_job.rb
adamruzicka: > these could contain sensitive data

That's what I thought as well.

adamruzicka:
> FYI there are some test failures, feel free to fix them before I test this.

The errors are coming from this patch depending on https://github.com/theforeman/foreman-tasks/pull/605 . Until that is merged, the tests will be red.

lzap (approved):
It works. Gonna rebase the core PR so we can merge this.

lzap:
Some tests failed, looks relevant.

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -173,6 +194,22 @@ def determine_proxy!(proxy_selector, provider, host)
...
+        property :task, object_of: 'Task', desc: 'Returns the task to which this action belongs'
+        property :host_name, String, desc: "Returns the name of the host"
+        property :host_id, Integer, desc: "Returns the id of the host"
+        property :host, object_of: 'Host', desc: "Returns the host"
+        property :job_invocation_id, Integer, desc: "Returns the id of the job invocation"
+        property :job_invocation, object_of: 'JobInvocation', desc: "Returns the job invocation"
+      end
+      class Jail < ::Actions::ObservableAction::Jail
+        allow :host_name, :host_id, :host, :job_invocation_id, :job_invocation
+      end
```
adamruzicka, line 212: @ofedoren could you please take a look?

ofedoren: Sure, forgot to mention that I generated docs with this PR as well and it seemed OK.

No concerns for now. I finally managed to configure my setup, I'll do some checks with this event occured and I'll try to refer the docs to get some info as a user would. If something will not correspond, I'll create a PR with fix afterwards.

ofedoren (approved):
Works well, the most needed data are available:
ScreenShot-1615149255408

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -173,6 +194,22 @@ def determine_proxy!(proxy_selector, provider, host)
         end
         proxy
       end
+
+      extend ApipieDSL::Class
+      apipie :class, "An action representing execution of a job against a host" do
+        name 'Actions::RemoteExecution::RunHostJob'
+        refs 'Actions::RemoteExecution::RunHostJob'
+        sections only: %w[webhooks]
```
ares, line 202: shouldn't this be `%w[all webhooks]`?

ofedoren: :facepalm: Well, yes. Probably makes sense to show everything related to `ActiveRecord` or `ActiveJob/Action` in `all` section. Although I'm not sure if an action object can be retrieved somewhere else outside of webhook payload.

adamruzicka: We did the same in tasks https://github.com/theforeman/foreman-tasks/pull/611/files#diff-46280f339235610e0e62cc0a085767460b6358550ef06cfd6c8a97f962572669R73 . 

As far as I know webhook payload is currently the only place where an action jail can be accessed by the users. There is a wip pr which give access to them in report templates. Does it make sense to expose those docs under `all` section if they can only be accessed in one (or two) specific context?

ofedoren: Currently the `all` section holds (or at least should) all the macros/properties (except Ruby ones) even if they can be used in e.g. provisioning context only. Then there are sections to filter stuff by context.

adamruzicka: REX: https://github.com/theforeman/foreman_remote_execution/pull/572
tasks: https://github.com/theforeman/foreman-tasks/pull/624

PR #563, opened 2021-01-27
Fixes #31562 - Don't crash when switching context on invocation

tbrisker:
When changing the taxonomy contex on the template invocation page,
the host that was displayed may no longer be viewable in the new
context. Instead of erroring out show an informational message.


ezr-ondrej (approved):
Works nicely, thanks @tbrisker !

PR #565, opened 2021-02-22
Fixes #31933 - Mark output as complete if task is missing

adamruzicka:
and the job is done

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -214,7 +214,7 @@ module Api
       end
 
       def host_output(job_invocation, host, default: nil, since: nil, raw: false)
-        refresh = true
+        refresh = !job_invocation.finished?
 
         if (task = job_invocation.sub_task_for_host(host))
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -214,7 +214,7 @@ def output_lines_since(task, time)
       end
 
       def host_output(job_invocation, host, default: nil, since: nil, raw: false)
-        refresh = true
+        refresh = !job_invocation.finished?
```
ares, line 217: It would be good to be more defensive here, in case the task does not exist yet, `finished?` could fail since it calls `task.pending`. I admit the fix may need to land in `def finished?`.

ares (changes requested):
One thing we discussed (and perhaps you'll find out not necessary at all), otherwise good.

adamruzicka:
The tests are green-ish

adamruzicka:
All :green_circle: now

ares (approved):
Thanks @adamruzicka, squashing and merging!

PR #566, opened 2021-03-08
Fixes #32032 - proxy batch sizing per provider

ezr-ondrej:
Allow number of HostTasks to be send to proxy in one batch to be tweaked
per every provider. If provider won't supply value, we still default to
`Setting['foreman_tasks_proxy_batch_size']`

- [x] https://github.com/theforeman/foreman-tasks/pull/640
- [x] #601

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -31,7 +31,8 @@ module Actions
         set_up_concurrency_control job_invocation
         input.update(:job_category => job_invocation.job_category)
         plan_self(:job_invocation_id => job_invocation.id)
-        input[:proxy_batch_size] ||= Setting['foreman_tasks_proxy_batch_size']
+        provider = job_invocation.pattern_template_invocations.first&.template&.provider
+        input[:proxy_batch_size] ||= provider&.proxy_batch_size || Setting['foreman_tasks_proxy_batch_size']
         trigger_action = plan_action(Actions::TriggerProxyBatch, batch_size: proxy_batch_size, total_count: hosts.count)
         input[:trigger_run_step_id] = trigger_action.run_step_id
       end
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -122,6 +122,10 @@ class RemoteExecutionProvider
       'Proxy::RemoteExecution::Ssh::Actions::RunScript'
     end
 
+    def proxy_batch_size
+      Setting['foreman_tasks_proxy_batch_size']
+    end
+
     # Return a specific proxy selector to use for running a given template
     # Returns either nil to use the default selector or an instance of a (sub)class of ::ForemanTasks::ProxySelector
     def required_proxy_selector_for(template)
--- a/test/unit/actions/run_hosts_job_test.rb
+++ b/test/unit/actions/run_hosts_job_test.rb
@@ -104,6 +104,22 @@ module ForemanRemoteExecution
       planned # To make the expectations happy
     end
 
+    describe '#proxy_batch_size' do
+      it 'defaults to Setting[foreman_tasks_proxy_batch_size]' do
+        Setting.expects(:[]).with('foreman_tasks_proxy_batch_size').returns(14)
+        planned
+        _(planned.proxy_batch_size).must_equal 14
+      end
+
+      it 'gets the provider value' do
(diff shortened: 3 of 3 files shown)


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -65,6 +65,12 @@ def batch(from, size)
         hosts.offset(from).limit(size)
       end
 
+      # Will have default in ProxyBatchTriggering if we cant find provider
+      def proxy_batch_size
```
adamruzicka, line 69: I can't find where this gets called

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -65,6 +65,12 @@ def batch(from, size)
         hosts.offset(from).limit(size)
       end
 
+      # Will have default in ProxyBatchTriggering if we cant find provider
```
adamruzicka, line 68: When will it have a default? There seems to be a fallback to the global setting in base rex provider

app/lib/actions/remote_execution/run_hosts_job.rb
ezr-ondrej: I missed that `ProxyBatchTriggering` is in tasks xD and kept wondering why the tests are failing...
https://github.com/theforeman/foreman-tasks/pull/616

adamruzicka:
Nevermind then, I missed the pr in tasks

adamruzicka:
This wouldn't really work, would it?

There are two batch sizes, one is the planning batch size and the other is the proxy batch size. This PR allows tweaking the latter one. However if the proxy batch size is bigger than the planning batch size there won't be enough "ready" tasks to send over to the proxy. Or am I overlooking something?

ezr-ondrej:
> There are two batch sizes, one is the planning batch size and the other is the proxy batch size. This PR allows tweaking the latter one. However if the proxy batch size is bigger than the planning batch size there won't be enough "ready" tasks to send over to the proxy. Or am I overlooking something?

You are right, I thought we create the sub tasks independently on the proxy triggering. So even the current setting `foreman_tasks_proxy_batch_size` has only range of 1-PLAN_BATCH_SIZE, having it higher has no effect what so ever?

Do you thinkg we should use the same batch_size then?

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -65,6 +65,19 @@ def batch(from, size)
...
+        provider&.proxy_batch_size
+      end
+
+      # Subtasks planning batch_size needs to be the same or higher then proxy_batch_size
+      # if not, the proxy_batch_size has no meaning, as we would not have enough host tasks
+      # to send to the proxy
+      def batch_size
+        proxy_size = proxy_batch_size || Setting['foreman_tasks_proxy_batch_size']
+        [super, proxy_size].compact.max
+      end
```
ezr-ondrej, line 79: @adamruzicka so should we do this to make sure we have enough sub tasks planned, so users can actually set the setting to higher values then the dynflow's default 100 ?

adamruzicka:
I wouldn't tie the batch sizes together. I'd say we should decouple the planning and proxy batch triggering processes. So both can be done independently using their configured batch sizes. It would also make the proxy triggering a bit safer, right now it is rather fragile

xprazak2:
There has not been an activity for a while here, what is the status?

adamruzicka:
This is blocked by https://github.com/theforeman/foreman-tasks/pull/640 which still needs to be polished a bit to be a bit more reliable, @ezr-ondrej knows the specifics

ezr-ondrej:
This is now quite simple change only allowing providers to control the batch size independently

app/models/remote_execution_provider.rb
```diff
@@ -122,6 +122,10 @@ def proxy_action_class
       'Proxy::RemoteExecution::Ssh::Actions::RunScript'
     end
 
+    def proxy_batch_size
+      Setting['foreman_tasks_proxy_batch_size']
```
adamruzicka, line 126: Would it make sense to have yet another setting for this?

ezr-ondrej: I don't believe so, I think this one make sense to share for now :thinking:

adamruzicka (approved):

adamruzicka:
Thank you @ezr-ondrej !

PR #568, opened 2021-03-10
Fixes #32065 - Undeclared :error_message instance field

domq:
Fixes [#32065](https://projects.theforeman.org/issues/32065)


adamruzicka:
Thanks for the patch it seems to work well. Just out of curiosity, how do other template types deal with this?

domq:
I didn't do the required archaeology, but I thought that perhaps job templates used to inherit a `:error_message` field from their superclass but don't anymore? Really I'm not sure.

adamruzicka (approved):
Fixes the issue for me, ACK

adamruzicka:
Thank you @domq !

PR #569, opened 2021-03-12
Fixes #32081 - Add provider specific inputs

xprazak2:
Adds inputs that will be displayed for each template
of a specified provider.

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -64,6 +64,16 @@ module Api
           param :description_format, String, :required => false, :desc => N_('Override the description format from the template for this invocation only')
           param :execution_timeout_interval, Integer, :required => false, :desc => N_('Override the timeout interval from the template for this invocation only')
           param :feature, String, :required => false, :desc => N_('Remote execution feature label that should be triggered, job template assigned to this feature will be used')
+
+          RemoteExecutionProvider.providers.each_value do |provider|
+            next if !provider.respond_to?(:provider_inputs_doc) || provider.provider_inputs_doc.empty?
+            doc = provider.provider_inputs_doc
+            param doc[:namespace], Hash, doc[:opts] do
+              doc[:children].map do |input|
+                param input[:name], input[:type], input[:opts]
+              end
+            end
+          end
         end
       end
 
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -201,7 +201,7 @@ module Api
 
       def permit_provider_inputs(invocation_params)
         providers = RemoteExecutionProvider.providers.values.reject { |provider| !provider.respond_to?(:provider_input_namespace) || provider.provider_input_namespace.empty? }
-        providers.map { |provider| invocation_params.dig(*provider.provider_input_namespace)&.permit! }
+        providers.each { |provider| invocation_params[provider.provider_input_namespace]&.permit! }
       end
 
       def composer_for_feature
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -195,7 +195,13 @@ module Api
         end
         job_invocation_params[:inputs] ||= {}
         job_invocation_params[:inputs].permit!
+        permit_provider_inputs job_invocation_params
         @job_invocation_params = job_invocation_params
+      end
+
+      def permit_provider_inputs(invocation_params)
+        providers = RemoteExecutionProvider.providers.values.reject { |provider| !provider.respond_to?(:provider_input_namespace) || provider.provider_input_namespace.empty? }
+        providers.map { |provider| invocation_params.dig(*provider.provider_input_namespace)&.permit! }
       end
 
       def composer_for_feature
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -20,7 +20,7 @@ class RemoteExecutionProvider
     end
 
     def proxy_command_options(template_invocation, host)
-      {:proxy_operation_name => proxy_operation_name}
+      {:proxy_operation_name => proxy_operation_name}.merge(proxy_command_provider_inputs(template_invocation))
     end
 
     def secrets(_host)
@@ -100,6 +100,14 @@ class RemoteExecutionProvider
 
     def provider_inputs
       []
+    end
+
+    def provider_inputs_doc
+      {}
+    end
+
+    def proxy_command_provider_inputs(template_invocation)
+      {}
     end
 
     def proxy_action_class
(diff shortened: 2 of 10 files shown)


xprazak2:
Not sure if mimicking template input is the best, I am open to suggestions.

ezr-ondrej:
Simpliest IMHO would be just render provider specific template and let providers define the fields. Then we probably could just have an hash of values on provider and no extra model.
Although this is bit more clean...

xprazak2:
Provider inputs can now be submitted via API as regular template inputs.

ares:
> Provider inputs can now be submitted via API as regular template inputs.

Is that a good thing? How do I as a user know which are job inputs and which are execution specific? If I understand this correctly, as a user I'd in hammer do something like inputs="command=mc,tags=a,b,c"?. Which also brings a question of how many levels of commas would actually work.

xprazak2:
I thought we wanted that so that user does not have to distinguish between what is template input and what is provider input. 

I am not sure how hammer handles inputs now, buy I would expect we would be able to take escaped values such as `inputs="command=mc,tags='a,b,c'"`.

Or should I revert to having provider inputs separately?

ares:
> I thought we wanted that so that user does not have to distinguish between what is template input and what is provider input.

Perhaps I wasn't clear much then. I think we should distinguish, since (template) inputs are inputs for the job while provider inputs modify the actual engine execution. They should be on the same level as other execution attributes, e.g. effective user, concurrency level. As discussed online, perhaps the best place would be in provider specific subset (we have ssh and ansible today) where we have the effective user right now.

xprazak2:
All green now.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -195,9 +195,15 @@ def job_invocation_params
         end
         job_invocation_params[:inputs] ||= {}
         job_invocation_params[:inputs].permit!
+        permit_provider_inputs job_invocation_params
         @job_invocation_params = job_invocation_params
       end
 
+      def permit_provider_inputs(invocation_params)
+        providers = RemoteExecutionProvider.providers.values.reject { |provider| !provider.respond_to?(:provider_input_namespace) || provider.provider_input_namespace.empty? }
+        providers.map { |provider| invocation_params.dig(*provider.provider_input_namespace)&.permit! }
```
adamruzicka, line 204: `permit!` modifies the parameters in-place?

app/models/remote_execution_provider.rb
```diff
@@ -35,6 +35,10 @@ def humanized_name
       self.name
     end
 
+    def provider_input_namespace
+      []
```
adamruzicka, line 39: From the name of the method I'd expect a single value, such as a symbol or a string. Why is it an array? Are elements of the array components of the namespace or can a provider have multiple input namespaces?

app/controllers/api/v2/job_invocations_controller.rb
xprazak2: Yes, inspired by what [job_invocation_params](https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb#L197) does. Do you have a different approach in mind?

app/models/remote_execution_provider.rb
xprazak2: Maybe I am making it more complicated than it needs to be, changed.

app/controllers/api/v2/job_invocations_controller.rb
adamruzicka: If we just do this for the side effect then we could `each` instead of `map` since we discard the result anway

xprazak2: Replaced with `each`.

db/migrate/20210312074713_add_provider_inputs.rb
```diff
@@ -0,0 +1,10 @@
+class AddProviderInputs < ActiveRecord::Migration[6.0]
+  def change
+    create_table :template_invocation_provider_input_values do |t|
```
ezr-ondrej, line 3: don't we want to simplify the table name? :thinking:

db/migrate/20210312074713_add_provider_inputs.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.references :template_invocation, :null => false, :index => { :name => 'idx_templ_inv_provider_input_values_on_templ_inv_id' }
```
ezr-ondrej, line 4: why are we naming the index? why not just `index: true`?

app/views/job_invocations/_form.html.erb
```diff
@@ -74,6 +74,14 @@
                     <% end %>
 
                     <div class="advanced hidden">
+                      <%= job_template_fields.fields_for :provider_input_values do |provider_input_fields| %>
```
ezr-ondrej, line 77: Wouldn't it be more straight forward if we would have the namespaces here as well?
Just an idea, I don't insist, but seems like it :shrug:

ezr-ondrej (commented):
Just some general questions.

db/migrate/20210312074713_add_provider_inputs.rb
xprazak2: Any suggestions?

db/migrate/20210312074713_add_provider_inputs.rb
xprazak2: Because autogenerated index name is too long and it crashes the migration.

app/views/job_invocations/_form.html.erb
xprazak2: Well, it would add another level of nesting into the form so I am a bit hesitant to call it straightforward, but I can certainly add it if you prefer.

db/migrate/20210312074713_add_provider_inputs.rb
ezr-ondrej: I think at least `invocation_provider_input_values` :thinking: but no hard feelings aganst this name.

db/migrate/20210312074713_add_provider_inputs.rb
ezr-ondrej: would simplier name of the table help ? :thinking:

app/views/job_invocations/_form.html.erb
ezr-ondrej: no, I'm ok both ways, I just don't like going totally different path API vs UI, but this is minor difference for a good reason, so I can live with that :+1:

xprazak2:
I changed the table name, the generated index is still too long so I need to supply a shorter name.

ezr-ondrej (approved):
I need to test with ansible, but code LGTM!

ezr-ondrej (approved):
I'm unsure what black magic removes the namespace in hammer and makes it directly `--tags` but I guess it is ok and that's onlything I found :shrug:
@adamruzicka anything against merging this?

adamruzicka:
Nope, I'm fine with it, let's get it in.

adamruzicka:
Thank you @xprazak2 & @ezr-ondrej !

PR #570, opened 2021-03-15
Add feature to skip proxy determination

sbernhard:
If the right proxy was already selected on which a job should run,
the REX plugin shouldn't do this again.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -30,10 +30,14 @@ module Actions
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
 
         provider = template_invocation.template.provider
-        proxy_selector = provider.required_proxy_selector_for(template_invocation.template) || proxy_selector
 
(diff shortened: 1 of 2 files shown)


adamruzicka:
> If the right proxy was already selected

selected how?

sbernhard:
Yes. I want to use this for foreman_acd.

The determine_proxy is already done here:
https://github.com/ATIX-AG/foreman_acd/blob/dc49cbad2ddd0af7ebcffbeed94964af73bc34cf/app/services/foreman_acd/app_configurator.rb#L27

Actually, I don't know if the code in this PR works and is doing it the right way. I want to have some feedback first :-)

adamruzicka:
A provider can override `required_proxy_selector_for(template)`. Probably cleanest solution would be to determine a proxy once and then have a custom selector which for a given host returns the host. This way you shouldn't need to touch base rex at all.

sbernhard:
Thx. I will have a look at it.

sbernhard:
Thank you so much. It works!

PR #573, opened 2021-03-25
Fixes #32209 - Try to require task launcher registry in advance

adamruzicka:
There is no guarantee that it was loaded by something else at the point where we try to use it. It is safer to try loading it.


ares (approved):
Makes sense, APJ

PR #575, opened 2021-03-29
Fixes #32219 - Check if SPDC settings are available

adamruzicka:
before trying to use them

### DIFF
--- a/lib/foreman_remote_execution_core.rb
+++ b/lib/foreman_remote_execution_core.rb
@@ -34,7 +34,7 @@ module ForemanRemoteExecutionCore
 
     current = if defined?(::Proxy::SETTINGS)
                 ::Proxy::SETTINGS.log_level.to_s.downcase
-              elsif defined?(SmartProxyDynflowCore)
+              elsif defined?(SmartProxyDynflowCore::SETTINGS)
                 SmartProxyDynflowCore::SETTINGS.log_level.to_s.downcase
               else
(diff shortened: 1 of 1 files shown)


lib/foreman_remote_execution_core.rb
```diff
@@ -34,7 +34,7 @@ def self.validate_ssh_log_level!
 
     current = if defined?(::Proxy::SETTINGS)
                 ::Proxy::SETTINGS.log_level.to_s.downcase
-              elsif defined?(SmartProxyDynflowCore)
+              elsif defined?(SmartProxyDynflowCore::SETTINGS)
```
ares, line 37: would it make sense to do rather ` defined?(SmartProxyDynflowCore) && defined?(SmartProxyDynflowCore::SETTINGS)`

adamruzicka: Is it necessary? What would be the scenario where your suggestion would behave differently?

```
2.7.0 :001 > defined?(::SmartProxyDynflowCore)
 => nil 
2.7.0 :002 > defined?(::SmartProxyDynflowCore::SETTINGS)
 => nil
```

ekohl: @ares ping?

ares: ah true, ignore me

ares (approved):
Thanks @adamruzicka, merging!

PR #577, opened 2021-04-13
Fixes #29384 - Connect by IP: Prefer IPv6 over IPv4

UXabre:


adamruzicka:
Is this safe? Interfaces often have more than 1 ipv6 address, but foreman doesn't model this well (or at all to be precise) so just a single one of them is stored. I don't know from the top of my head how the address-to-be-stored is picked from all the addresses, but it is quite possible it will be a link-local address or something that can't really be used to reach the host.

UXabre:
Hi @adamruzicka 
As far as my mileage goes and afaik, I don't see any subsytem storing link-local IPv6 addresses in the host; only the global ones being configured via the interfaces tab (as this, indeed, only supports a single IPv6 address).
Is there any module you know that would actually store the link-local address or, even worse, overwrite this IPv6 address? My first thought was that the facts scraping module would perhaps overwrite this, but it doesn't seem to be like this.

adamruzicka:
I thought we just use whatever facter (in case of puppet) gives us, which returns link local addresses if there are no other ipv6 addresses on a given interface. But apparently there's some additional filtering going on somewhere because I don't see ll addresses in foreman.

@ares Do you recall if there was something else that made us not implement connect by ip for ipv6?

lzap:
I was actually solving the very same problem in provisioning and my approach is a host parameter:

https://github.com/theforeman/foreman/pull/8447

Just saying.

UXabre:
I've been thinking and was wondering if it would make sense to do some sort of "TCP Ping" to check if the device can be reached on any of the IP addresses instead of just taking IPv6 as the preferred address?
On the other hand, I try to mimic the proposed default of starting with IPv6, as, by now, probably most OS'es, by default, abide to? (I know, this can be circumvented and configured to behave in IPv4 being preferred)
I'm probably just curious to learn who uses IPv6 on a REX-interface without it being actually reachable from foreman? The only thing I can imagine is because somewhere is a router in between which can't IPv6?

lzap:
I would stay away from ping checks, from time to time desperately confused users come to us saying that the ping/echo ICMP/TCP checks in our DHCP workflow made some huge pain to them until they figured it out.

UXabre:
I completely undestand your point and agree that this would indeed be an less then suboptimal solution.
I think it might be a good idea that I draft up an RFC on the subject on the community page first

lzap:
You could turn the setting into IPv4, IPv6 or no (three states). Plus a host parameter could do the trick on a per-host basis.

UXabre:
Oh yes, I actually like that idea a lot! Will implement this accordingly.

adamruzicka:
@UXabre any progress here?

UXabre:
Hi @adamruzicka , no further progress yet, I'm kinda flooded at this point. My next window of opportunity is coming saturday, it's not a lot of work really, so sorry for the delay.

adamruzicka:
Gentle reminder we'd still be interested in this :)

UXabre:
Hi, so sorry for the huge delay, now I'm in my holidays I did find the time!
I've added a setting to configure preference for IPv6 (or not). Not sure if the name isn't too straighforward; any proposal is welcome at this point.

app/models/remote_execution_provider.rb
```diff
@@ -80,7 +80,11 @@ def find_ip_or_hostname(host)
 
     def find_ip(host, interfaces)
       if host_setting(host, :remote_execution_connect_by_ip)
-        interfaces.find { |i| i.ip.present? }.try(:ip)
+        if host_setting(host, :remote_execution_connect_by_ip_prefer_ipv6)
+          interfaces.find { |i| i.ip6.present? }.try(:ip6) || interfaces.find { |i| i.ip.present? }.try(:ip)
+        else
+          interfaces.find { |i| i.ip.present? }.try(:ip)
+        end
```
adamruzicka, line 87:
```suggestion
        ip6 = interfaces.find { |i| i.ip6.present? }.try(:ip6) if host_setting(host, :remote_execution_connect_by_ip_prefer_ipv6)
        ip6 || interfaces.find { |i| i.ip.present? }.try(:ip)
```

adamruzicka (commented):
Could be made a bit DRYer, but in general +1

app/models/remote_execution_provider.rb
UXabre: Thanks for your proposal, I actually did it a bit different as I suddenly realized that I overlooked something: it's a preference, meaning that the fallbacks should simply be switched. What I mean by this is, if IPv6 is preferred, it should fallback to IPv4 if no IPv6 was detected and vice versa.
Nevertheless, interesting and elegant proposal

UXabre:
Allright, I've made a tiny adjustment to actually work as a preference setting, meaning, if IPv6 preference is enabled, it should try an IPv6 address, if none exists, it should fall back to IPv4; and vice versa if the IPv6 preference is disabled.
I've also added a test for this behavior.

adamruzicka (approved):
Works nicely, let's get this in

adamruzicka:
Thank you @UXabre !

PR #581, opened 2021-04-27
Fixes #32107 - job invocation create API should work with the feature param

jturel:
I'll need someone to verify the correctness of the change I made here. The tests I removed assert that the job invocations API should take `host_ids` as an input but that param  is undocumented. I think those tests were meant to describe the behavior of `JobInvocationComposer.for_feature` so I updated the tests accordingly, while also updating the controller to correctly pass the `search_query` param into `for_feature`

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -80,18 +80,15 @@ module Api
       api :POST, '/job_invocations/', N_('Create a job invocation')
       param_group :job_invocation, :as => :create
       def create
-        if job_invocation_params[:feature].present?
-          composer = composer_for_feature
-        else
-          validate_template
-          composer = JobInvocationComposer.from_api_params(
-            job_invocation_params
-          )
-        end
+        composer = JobInvocationComposer.from_api_params(
+          job_invocation_params
+        )
         composer.trigger!
         @job_invocation = composer.job_invocation
         @hosts = @job_invocation.targeting.hosts
         process_response @job_invocation
+      rescue JobInvocationComposer::JobTemplateNotFound, JobInvocationComposer::FeatureNotFound => e
+        not_found(error: { message: e.message })
       end
 
       api :GET, '/job_invocations/:id/hosts/:host_id', N_('Get output for a host')
@@ -190,19 +187,19 @@ module Api
         not_found({ :error => { :message => (_("Host with id '%{id}' was not found") % { :id => params['host_id'] }) } })
       end
 
-      def validate_template
-        JobTemplate.authorized(:view_job_templates).find(job_invocation_params['job_template_id'])
-      rescue ActiveRecord::RecordNotFound
-        not_found({ :error => { :message => (_("Template with id '%{id}' was not found") % { :id => job_invocation_params['job_template_id'] }) } })
-      end
-
       def job_invocation_params
         return @job_invocation_params if @job_invocation_params.present?
 
         job_invocation_params = params.fetch(:job_invocation, {}).dup
+
+        if job_invocation_params[:feature].present? && job_invocation_params[:job_template_id].present?
+          raise _("Only one of feature or job_template_id can be specified")
+        end
+
         if job_invocation_params.key?(:ssh)
           job_invocation_params.merge!(job_invocation_params.delete(:ssh).permit(:effective_user))
         end
+
         job_invocation_params[:inputs] ||= {}
         job_invocation_params[:inputs].permit!
         permit_provider_inputs job_invocation_params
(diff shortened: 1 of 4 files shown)


jturel:
@adamruzicka will please you kick off the tests for me?

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -201,7 +201,7 @@ def job_invocation_params
       def composer_for_feature
         JobInvocationComposer.for_feature(
           job_invocation_params[:feature],
-          job_invocation_params[:host_ids],
+          job_invocation_params[:search_query],
```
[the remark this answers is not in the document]
jturel: No problem at all. Should I add host_ids as a param to the controller method with the description stating that it's deprecated ?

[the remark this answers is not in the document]
adamruzicka: If I remember correctly the original purpose of it was to allow katello's angular ui to trigger rex jobs for package actions. The param wasn't exposed because if was considered internal. I can't really say that I like the idea of documenting a param which shouldn't be used by the wider public just to deprecate it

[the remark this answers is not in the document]
jturel: Ok. Katello doesn't use the REST API but rather we call JobInvocationComposer directly. We do send host ids, though.

jturel:
@adamruzicka updated!

app/controllers/api/v2/job_invocations_controller.rb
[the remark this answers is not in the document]
adamruzicka: I may be misremembering or it might be something that was there some years ago but was eventually removed.

adamruzicka (changes requested):
When triggering a job using a feature through the api, all scheduling params are ignored. I'd expect it to behave the same way as when one explicitly provides a template id instead of a feature

jturel:
@adamruzicka i removed usage of `for_feature` from the API and enhanced `from_api_params` to maintain backward compatibility with the host_ids param. I think I got it right?

app/models/job_invocation_composer.rb
```diff
@@ -96,6 +96,13 @@ class ApiParams
 
     def initialize(api_params)
       @api_params = api_params
+
+      api_params[:targeting_type] ||= Targeting::STATIC_TYPE
```
jturel, line 100: Dunno if this was safe to do

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -70,14 +70,10 @@ def show
       api :POST, '/job_invocations/', N_('Create a job invocation')
       param_group :job_invocation, :as => :create
       def create
-        if job_invocation_params[:feature].present?
-          composer = composer_for_feature
```
adamruzicka, line 74: The composer was, among other things, setting `remote_execution_feature_id` on the `job_invocation` being composed. This no longer happens

app/models/job_invocation_composer.rb
```diff
@@ -227,25 +234,38 @@ def triggering_params
...
+        @bookmark = host_ids
+      elsif host_ids.is_a?(Host::Base)
+        @hosts = [host_ids]
+      elsif host_ids.is_a?(Array)
+        @hosts = host_ids.map do |id|
+          Host::Managed.authorized.friendly.find(id)
+        end
+      elsif host_ids.is_a?(String)
+        @scoped_search = host_ids
+      end
```
adamruzicka, line 251: Could you use a `case` here?

app/controllers/api/v2/job_invocations_controller.rb
jturel: I believe that's fixed now

app/models/job_invocation_composer.rb
jturel: Done!

jturel:
@adamruzicka updated. have some test failures now but will wait for further feedback before cleaning them up

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -80,18 +80,15 @@ def show
...
-        end
+        composer = JobInvocationComposer.from_api_params(
+          job_invocation_params
+        )
         composer.trigger!
         @job_invocation = composer.job_invocation
         @hosts = @job_invocation.targeting.hosts
         process_response @job_invocation
+      rescue JobInvocationComposer::JobTemplateNotFound, JobInvocationComposer::FeatureNotFound => e
+        not_found(error: { message: e.message })
```
jturel, line 91: The previous validate_template method was causing a double render - now it actually returns 404

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -190,19 +187,19 @@ def find_host
...
-        not_found({ :error => { :message => (_("Template with id '%{id}' was not found") % { :id => job_invocation_params['job_template_id'] }) } })
-      end
-
       def job_invocation_params
         return @job_invocation_params if @job_invocation_params.present?
 
         job_invocation_params = params.fetch(:job_invocation, {}).dup
+
+        if job_invocation_params[:feature].present? && job_invocation_params[:job_template_id].present?
+          raise _("Only one of feature or job_template_id can be specified")
```
jturel, line 196: I think this is valid? Maybe the Composer should throw this error

jturel:
@adamruzicka I think I've shored up the differences. Have a look!

parthaa:
I was not able to test the complete workflow, but see that the right call is being made
```
$ hammer job-invocation create --feature=katello_package_install --search-query='centos7.sparta.example.com' --inputs 'package=vim'
Job invocation 8 created
[.....  ] [10%]
```

parthaa (approved):

adamruzicka (approved):
I'm fine with this, let's get it in

adamruzicka:
Thank you @jturel & @parthaa !

PR #582, opened 2021-04-29
Fixes #30277 - Allow spaces in remote_working_dir

pheeeter:


lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -173,8 +173,8 @@ def initialization_script
       su_method = @user_method.instance_of?(ForemanRemoteExecutionCore::SuUserMethod)
       # pipe the output to tee while capturing the exit code in a file
       <<-SCRIPT.gsub(/^\s+\| /, '')
-      | sh -c "(#{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path}) | /usr/bin/tee #{@output_path}
-      | exit \\$(cat #{@exit_code_path})"
+      | sh -c "(#{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script.shellescape.gsub(/"/,'\\\\\\"')} < /dev/null"}; echo \\$?>#{@exit_code_path.shellescape.gsub(/"/,'\\\\\\"')}) | /usr/bin/tee #{@output_path.shellescape.gsub(/"/,'\\\\\\"')}
```
adamruzicka, line 176: This would most likely not work if `su` is used to escalate user privileges

lib/foreman_remote_execution_core/script_runner.rb
```diff
@@ -173,8 +173,8 @@ def initialization_script
...
+      | exit \\$(cat #{@exit_code_path.shellescape.gsub(/"/,'\\\\\\"')})"
```
adamruzicka, line 177: There is a way how we could avoid this substitution and few of the escapes we currently do by hand

```shell
# Instead of current
sh -c "($content) | tee $output"

# we could use
sh <<EOF | tee $output
$content
EOF
```

adamruzicka (changes requested):
We could also escape the directories in `initialize` and not have to do it all over the place.

Is the polling script runner also affected by this or do paths with spaces work just fine out of the box there?

adamruzicka:
Also most of the code from foreman_remote_execution_core got moved to this repo https://github.com/theforeman/smart_proxy_remote_execution_ssh , could you please reopen this PR there?

pheeeter:
PR opened in _smart_proxy_remote_execution_ssh_ repo: [https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/55](url)

PR #583, opened 2021-05-03
Remote execution shutdown command

anandkumaragrawal:


adamruzicka:
Is `/sbin/shutdown` available on *all* platforms? The issue mentions RHEL variants only.

adamruzicka:
Also could you please rebase it so it is just a single commit?

adamruzicka:
How about instead of hardcoding the path to `/usr/sbin/shutdown`, we append `/usr/sbin` to the end of `PATH` env variable and leave the rest untouched?

adamruzicka:
@anandkumaragrawal ping

anandkumaragrawal:
> @anandkumaragrawal ping

Hey @adamruzicka  pong,

Your suggestion looks good. However, I am occupied. Would you like to change it and fix that.

Regards,
Anand

adamruzicka:
Ok, closing in favor of https://github.com/theforeman/foreman_remote_execution/pull/640. Thank you @anandkumaragrawal

PR #584, opened 2021-05-04
Fixes #32357 - integrate rex features with webhooks

lzap:
One nitpick - I am hitting an issue with the new UI, when I try to add a new webhook on the new events, UI tells me Events cannot be blank. I will discuss with Oleh.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -5,6 +5,8 @@ module Actions
       include ::Actions::Helpers::WithDelegatedAction
       include ::Actions::ObservableAction
 
+      execution_plan_hooks.use :emit_feature_event, :on => :success
+
       middleware.do_not_use Dynflow::Middleware::Common::Transaction
       middleware.use Actions::Middleware::HideSecrets
 
@@ -17,7 +19,12 @@ module Actions
       end
 
       def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExecutionProxySelector.new, options = {})
-        action_subject(host, :job_category => job_invocation.job_category, :description => job_invocation.description, :job_invocation_id => job_invocation.id)
+        features = template_invocation.templateremote_execution_features.pluck(:label).uniq
+        action_subject(host,
+          :job_category => job_invocation.job_category,
+          :description => job_invocation.description,
+          :job_invocation_id => job_invocation.id,
+          :job_features => features)
 
         template_invocation.host_id = host.id
         template_invocation.run_host_job_task_id = task.id
@@ -56,6 +63,22 @@ module Actions
       def finalize(*args)
         update_host_status
         check_exit_status
+      end
+
+      def self.feature_job_event_name(label, suffix = :success)
+        ::Foreman::Observable.event_name_for("#{::Actions::RemoteExecution::RunHostJob.event_name_base}_#{label}_#{::Actions::RemoteExecution::RunHostJob.event_name_suffix(suffix)}")
+      end
+
+      def emit_feature_event(execution_plan, hook = :success)
+        return unless root_action?
+
+        payload = event_payload(execution_plan)
+        if input["job_features"]&.any?
+          input['job_features'].each do |feature|
+            name = "#{self.class.event_name_base}_#{feature}_#{self.class.event_name_suffix(hook)}"
+            trigger_hook name, payload: payload
+          end
+        end
       end
 
       def secrets(host, job_invocation, provider)
@@ -186,10 +209,13 @@ module Actions
             'All %{count} applicable proxies are down. Tried %{proxy_names}',
             offline_proxies.count) % settings
         elsif proxy == :not_defined
-          settings = { :global_proxy => 'remote_execution_global_proxy',
-                       :fallback_proxy => 'remote_execution_fallback_proxy' }
+          settings = {
+            global_proxy: 'remote_execution_global_proxy',
+            fallback_proxy: 'remote_execution_fallback_proxy',
+            provider: provider,
+          }
 
-          raise _('Could not use any proxy. Consider configuring %{global_proxy}, ' +
+          raise _('Could not use any proxy for the %{provider} job. Consider configuring %{global_proxy}, ' +
                   '%{fallback_proxy} in settings') % settings
         end
         proxy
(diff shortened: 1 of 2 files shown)


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -17,7 +19,11 @@ def resource_locks
       end
 
       def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExecutionProxySelector.new, options = {})
-        action_subject(host, :job_category => job_invocation.job_category, :description => job_invocation.description, :job_invocation_id => job_invocation.id)
+        action_subject(host,
+          :job_category => job_invocation.job_category,
+          :description => job_invocation.description,
+          :job_invocation_id => job_invocation.id,
+          :job_feature_label => job_invocation&.remote_execution_feature&.label)
```
adamruzicka, line 26: Shouldn't we consider all features tied to the template being run? This way if I go to the UI and instead of selecting "Run all ansible roles" for a host I go into the job invocation form, select the template for running ansible roles and the host and run it, then the event wouldn't be emitted, even though roles were in fact applied to a host.

adamruzicka (commented):
Rubocop is unhappy

app/lib/actions/remote_execution/run_host_job.rb
lzap: Right, rebased.

lzap:
It now works as expected. Make sure to use the updated template if you are testing with it: https://github.com/theforeman/foreman_webhooks/pull/37

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -17,7 +19,12 @@ def resource_locks
       end
 
       def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExecutionProxySelector.new, options = {})
-        action_subject(host, :job_category => job_invocation.job_category, :description => job_invocation.description, :job_invocation_id => job_invocation.id)
+        features = template_invocation&.template&.remote_execution_features&.pluck(:label)&.uniq
```
adamruzicka, line 22: Is the safe navigation everywhere necessary? `template_invocation` should always be non-nil and have an associated template. `remote_execution_features` should always be defined on `template`

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -58,6 +65,24 @@ def finalize(*args)
...
+        return unless root_action?
+
+        if input["job_features"]
+          input['job_features'].each do |feature|
+            name = "#{self.class.event_name_base}_#{feature}_#{self.class.event_name_suffix(hook)}"
+            trigger_hook name, payload: event_payload(execution_plan)
+          end
+        else
+          name = "#{self.class.event_name_base}_#{self.class.event_name_suffix(hook)}"
+          trigger_hook name, payload: event_payload(execution_plan)
```
adamruzicka, line 82: Can the else branch ever happen? And if it does and if I'm reading it right, then it would emit the same generic event that is already being emitted since `RunHostJob` is an observable action. In any case, I'd vote for dropping it so it either emits events for features or nothing at all

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -58,6 +65,24 @@ def finalize(*args)
...
+        ::Foreman::Observable.event_name_for("#{::Actions::RemoteExecution::RunHostJob.event_name_base}_#{label}_#{::Actions::RemoteExecution::RunHostJob.event_name_suffix(suffix)}")
+      end
+
+      def emit_feature_event(execution_plan, hook = :success)
+        return unless root_action?
+
+        if input["job_features"]
+          input['job_features'].each do |feature|
+            name = "#{self.class.event_name_base}_#{feature}_#{self.class.event_name_suffix(hook)}"
+            trigger_hook name, payload: event_payload(execution_plan)
```
adamruzicka, line 78: In case there are multiple features, could we build the event payload just once?

app/lib/actions/remote_execution/run_host_job.rb
lzap: Sure

app/lib/actions/remote_execution/run_host_job.rb
lzap: Absolutely.

lzap:
I amended, so now it test for array emptyness, I am not sure if I understand the design correctly tho - are there jobs (or better - templates) without any features associated? I guess so, or teach me.

adamruzicka:
Yes, I'd even go as far as saying that majority of job templates do not have an associated feature.

app/lib/actions/remote_execution/run_host_job.rb
lzap: Oh yeah I missed your point, I get it now sorry.

lzap:
So I amended, dropped the else statement. Good catch.

adamruzicka:
Could you please rebase so the tests can run now that rubocop issues were resolved in another pr?

ares:
ping @lzap, this is the last story of the list, it would be great to hit the next snap

lzap:
Yeah rebased.

lzap:
All green.

Can we also merge https://github.com/theforeman/foreman_webhooks/pull/37 so I will release 1.1 or 2.0 version of webhooks?

adamruzicka (approved):

adamruzicka:
Thank you @lzap !

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -17,7 +19,12 @@ def resource_locks
       end
 
       def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExecutionProxySelector.new, options = {})
-        action_subject(host, :job_category => job_invocation.job_category, :description => job_invocation.description, :job_invocation_id => job_invocation.id)
+        features = template_invocation.templateremote_execution_features.pluck(:label).uniq
```
ezr-ondrej, line 22: Uh-oh @lzap, @adamruzicka this doesn't seem to be correct :hot_face:

adamruzicka: Yeah, just noticed it as well https://github.com/theforeman/foreman_remote_execution/pull/595

PR #585, opened 2021-05-05
Fixes #32439 - Add error state to job wizard

MariaAga:
When one of the api calls in job wizard fails there is now an error shown.
Currently there are only api calls to get  the categories list and templates list
Screenshot from 2021-05-05 12-40-34

### DIFF
--- a/webpack/JobWizard/JobWizardSelectors.js
+++ b/webpack/JobWizard/JobWizardSelectors.js
@@ -1,6 +1,7 @@
 import {
   selectAPIResponse,
   selectAPIStatus,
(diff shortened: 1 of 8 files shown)


stejskalleos (changes requested):
Hi,
just two nitpicks, otherwise works as described!

MariaAga:
Fixed! thanks!

stejskalleos (approved):

adamruzicka (approved):
Works as expected, ACK

adamruzicka:
Thank you @MariaAga & @stejskalleos !

PR #586, opened 2021-05-06
Fixes #32539 - Stop loading rex core into regular rex

adamruzicka:


ezr-ondrej:
This is really all what it takes to get rid of this? :facepalm:

adamruzicka:
Well, not really. It just prepares the ground for this one https://github.com/theforeman/foreman_remote_execution/pull/587

ezr-ondrej:
I meant to get rid of the dependency of rex to rex_core :) I always thought there is something we really share between proxy and foreman, that I still just don't understand

ezr-ondrej (commented):
Could we drop the dependency from rex gemspec now then, or do I misunderstand? :)

adamruzicka:
I would like to keep that around until we remove rex_core for good so we don't mess up the dependency chain. I'm not sure if anything pulls in rex core explicitly.

PR #587, opened 2021-05-06
Fixes #32542 - Make REX core only a compatibility layer proxy plugin 

adamruzicka:
Follows
- [x] https://github.com/theforeman/foreman_remote_execution/pull/586

Requires
- [ ] https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/52

Notes:
~~This currently breaks foreman_ansible, because it loads foreman_remote_execution_core and accesses its settings, which is probably not what we want to do.~~ Solved by https://github.com/theforeman/foreman_ansible/pull/410


ezr-ondrej (changes requested):
Do we still need to require `foreman_remote_execution_core` gem in `foreman_remote_execution`?
Now it pulls in the smart_proxy plugin as well, what causes trouble as we don't have smart_proxy itself.

adamruzicka:
Where do we do that? After https://github.com/theforeman/foreman_remote_execution/pull/586 went it we shouldn't be doing it anywhere as far as I know

ezr-ondrej:
We require it in gemspec https://github.com/theforeman/foreman_remote_execution/blob/889314eb7d858cd0cda86b34adbcae5a78b313eb/foreman_remote_execution.gemspec#L28 what actually does that require I'm not sure, but even pulling in the smart proxy plugin into gemset is probably not best idea.

adamruzicka:
Before I cut into it: After this gets in, REX core becomes only a compatibility layer, exposing things from `smart_proxy_remote_execution_ssh` (`SP_REX` from now on) under different names. That means REX core should depend on new-enough version of `SP_REX`. If another plugin depends on something from REX core, then it should explicitly depend on it. If this all is true, then we should be able to just drop the dependency on `REX core` from `REX`'s gemspec.

Does this sound correct?

ezr-ondrej (commented):
Works as expected, just one thing

Should we change the `proxy_action_class` in the provider? https://github.com/theforeman/foreman_remote_execution/blob/889314eb7d858cd0cda86b34adbcae5a78b313eb/app/models/remote_execution_provider.rb#L114

It works with the aliasing, but I guess it would be nicer to have the new one explicitly, so we don't confuse people :)

foreman_remote_execution_core.gemspec
```diff
@@ -22,4 +22,5 @@ DESC
   s.add_runtime_dependency('ed25519')
   s.add_runtime_dependency('foreman-tasks-core', '>= 0.3.1')
   s.add_runtime_dependency('net-ssh')
+  s.add_runtime_dependency('smart_proxy_remote_execution_ssh', '>= 0.4.0')
```
ezr-ondrej, line 25: doesn't this make us still release the SPREX before this one anyway?

ezr-ondrej (commented):
As discussed offline, we still want to use the old class, so we have the freedom in release order.

foreman_remote_execution_core.gemspec
adamruzicka: It makes us release sprex before rex_core, but non-core rex can be released independently as it can work with both old and new rex_core

ezr-ondrej (approved):
Thanks @adamruzicka ! I was not insisting, just got held on another meeting, but I believe we need to release REX anyway, so I think this is better :+1:

PR #588, opened 2021-05-10
Fixes #27770 - Add support for ed25519 keys

adamruzicka:
This pr is rather simple, the real work will need to happen in packaging.


lzap:
Big phat warning: blowfish is a blacklisted cipher by NIST and I am actually working on a patch that will add an alternative NIST-approved password hashing into Foreman core. I suggest to ask on the crypto list if you haven't done this already.

If you need assistance with packaging, let me know.

lzap:
So I got the feedback, ed25519 is not allowed in FIPS mode and it looks like ruby-ssh does respect that. I am merging the PR.

PR #591, opened 2021-05-17
Fixes #32606 - Reimplement known host key removal

adamruzicka:
Completely replaces the fix for #21449 and friends.

In the previous implementation, when a host was removed, Foreman tried
to reach out to all the Smart Proxies that could have been used for
remote execution against the host being removed. This made sure all the
traces of a host were removed from everywhere.

The new version performs the cleanup on first execution on a host
through a proxy. It still solves the original issue (not being able to
run jobs after a host is re-provisioned), but leaves bits around when
the host is deleted.

I do not like this and consider it a step back, however it seems to be
what people want.


adamruzicka:
@ezr-ondrej @ares opinions?

ezr-ondrej:
Can't we just not fail on the key removal?
Just silence it out?

adamruzicka:
We can, but what would that solve? I mean, where's the reliability in that?

ezr-ondrej:
Well there is not really one, but this is not great eigher.
I suppose, that we could run cleanup once in a while (or on every successful remove?) by sending list of known hosts in Foreman, that should have their keys in given proxy and proxy could remove all the obsolete ones in case there is more then just the currently deleted one.

adamruzicka:
That would be slightly tricky if a host is reprovisioned with the same name and the cleanup doesn't happen between the teardown of the old and setup of the new host.

adamruzicka:
Let's go with this. It fixes the original issue (run a rex job, reprovision the host, run a job again) and behaves most predictably out of all the options we considered

ares:
Requires a rebase now. :+1: on the direction after chatting. The other option is as suggested, just to fail silently, however this nicer flow is already implemented. If one Foreman manages 100k hosts and has ~10 proxies, we'd see 1M associations but that should be still ok hopefully.

adamruzicka:
Changed merge base to `4.5-stable` (branch before the rex-core removal). Once this gets in, I'll cherry-pick the non-core parts into master and the core bits into smart_proxy_remote_execution_ssh

ezr-ondrej (approved):
I'm fine with this.

adamruzicka:
Merging as discussed on IRC, will open forward-ports shortly

adamruzicka:
Thank you @ezr-ondrej !

PR #593, opened 2021-05-19
Fixes #32628 - Add search definitions for jobs

xprazak2:
Allows searching recurring job invocations by template name
and target host.

`recurring = true && targeted_host_id = 3 && pattern_template_name = "Ansible Roles - Ansible Default" && status = queued`

### DIFF
--- a/app/models/host_status/execution_status.rb
+++ b/app/models/host_status/execution_status.rb
@@ -64,15 +64,15 @@ class HostStatus::ExecutionStatus < HostStatus::Status
 
       case status
         when OK
-          [ "state = 'stopped' AND result = 'success'" ]
+          [ "foreman_tasks_tasks.state = 'stopped' AND result = 'success'" ]
         when CANCELLED
-          [ "state = 'stopped' AND result = 'cancelled'" ]
+          [ "foreman_tasks_tasks.state = 'stopped' AND result = 'cancelled'" ]
         when ERROR
-          [ "state = 'stopped' AND (result = 'error' OR result = 'warning')" ]
+          [ "foreman_tasks_tasks.state = 'stopped' AND (result = 'error' OR result = 'warning')" ]
         when QUEUED
-          [ "state = 'scheduled' OR state IS NULL" ]
+          [ "foreman_tasks_tasks.state = 'scheduled' OR foreman_tasks_tasks.state IS NULL" ]
         when RUNNING
-          [ "state <> 'stopped'" ]
+          [ "foreman_tasks_tasks.state <> 'stopped'" ]
         else
           [ '1 = 0' ]
       end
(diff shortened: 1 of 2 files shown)


app/models/host_status/execution_status.rb
```diff
@@ -64,15 +64,15 @@ def self.sql_conditions_for(status)
 
       case status
         when OK
-          [ "state = 'stopped' AND result = 'success'" ]
+          [ "foreman_tasks_tasks.state = 'stopped' AND result = 'success'" ]
```
adamruzicka, line 67: Why do we need to include the table name?

app/models/job_invocation.rb
```diff
@@ -80,6 +88,14 @@ class Jail < Safemode::Jail
     allow :sub_task_for_host, :template_invocations_hosts
   end
 
+  def self.search_by_targeted_host(key, operator, value)
+    { :conditions => sanitize_sql_for_conditions(["hosts.id = #{value}"]), :joins => :targeted_hosts }
```
adamruzicka, line 92: Could we use a placeholder and let `sanitize_sql_for_conditions` do its thing instead of interpolating strings ourselves?

app/models/job_invocation.rb
```diff
@@ -80,6 +88,14 @@ class Jail < Safemode::Jail
...
+  end
+
+  def self.search_by_pattern_template(key, operator, value)
+    { :conditions => sanitize_sql_for_conditions(["templates.name = '#{value}'"]), :joins => :pattern_templates }
```
adamruzicka, line 96: Ditto

app/models/job_invocation.rb
```diff
@@ -15,8 +15,9 @@ class JobInvocation < ApplicationRecord
 
   belongs_to :targeting, :dependent => :destroy
   has_many :all_template_invocations, :inverse_of => :job_invocation, :dependent => :destroy, :class_name => 'TemplateInvocation'
-  has_many :template_invocations, -> { where('host_id IS NOT NULL') }, :inverse_of => :job_invocation
-  has_many :pattern_template_invocations, -> { where('host_id IS NULL') }, :inverse_of => :job_invocation, :class_name => 'TemplateInvocation'
+  has_many :template_invocations, -> { where('template_invocations.host_id IS NOT NULL') }, :inverse_of => :job_invocation
+  has_many :pattern_template_invocations, -> { where('template_invocations.host_id IS NULL') }, :inverse_of => :job_invocation, :class_name => 'TemplateInvocation'
```
adamruzicka, line 19: Why do we need to explicitly state the table name there?

app/models/job_invocation.rb
xprazak2: Done.

app/models/job_invocation.rb
xprazak2: Done.

app/models/job_invocation.rb
xprazak2: Same as above:
column-err

adamruzicka:
:police_officer: is unhappy about this

xprazak2:
All green now.

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @xprazak2 !

PR #598, opened 2021-06-08
Fixes #32742 - Ensure job rerun through api works with invalid hosts

adamruzicka:
When a job is rerun through the UI, we take out selected fields from the
old job, render the form and then submit it. Essentially it behaves like
running a new job with some fields pre-populated.

When rerunning a job through the api, there is no form middle step. We
just take the old job, run it through job composer, get a new job out of
it and then link the new job's targeting against old job's hosts. This
step was triggering validations on the hosts.


ares (approved):
Thanks @adamruzicka, makes sense and fixes the issue, merging!

PR #599, opened 2021-06-09
Fixes #32797 - Fix issue that monitor > jobs page could not render

sbernhard:
Monitor > Jobs page can not be rendered because total_hosts_count
fails. Unfortunately the total_hosts_count method tries to execute
total_count in ApplicableErrataInstall but this should normally done in
REX > RunHostsJob.

This happens for Jobs created with older REX plugins (< 2.3 of foreman).

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -193,11 +193,16 @@ class JobInvocation < ApplicationRecord
   end
 
   def total_hosts_count
+    count = _('N/A')
+
     if targeting.resolved?
-      task&.main_action&.total_count || targeting.hosts.count
-    else
-      _('N/A')
+      count = if task&.main_action.respond_to?(:total_count)
+                task.main_action.total_count
+              else
+                targeting.hosts.count
+              end
     end
+    count
   end
 
   def pattern_template_invocation_for_host(host)


adamruzicka:
I'm fine with this, but :policeman: is not

sbernhard:
Now Rubocopnis happy

sbernhard:
Is this PR now OK?

app/models/job_invocation.rb
```diff
@@ -193,11 +193,14 @@ def failed_hosts
...
   def total_hosts_count
+    count = _('N/A')
     if targeting.resolved?
-      task&.main_action&.total_count || targeting.hosts.count
-    else
-      _('N/A')
+      count = targeting.hosts.count
+      if task&.main_action.respond_to?(:total_count)
+        count = task.main_action.total_count
+      end
```
adamruzicka, line 201: Could we go with something like this so targeting.hosts.count is evaluated only if action's total_count is not available?

```suggestion
      count = if task&.main_action.respond_to?(:total_count)
        task.main_action.total_count
      else
        targeting.hosts.count
      end
```

adamruzicka (commented):
One more nitpick. Also could we get an issue so it is properly tracked?

adamruzicka (approved):
I'm still not sure how did this happen, but in case it does, this should help. ACK

adamruzicka:
Thank you @sbernhard !

PR #600, opened 2021-06-12
Fixes #32793 - unambigous result field on index query

ezr-ondrej:
This fixes an issue of AmbigousColumn result in JobInvocation query.

It addresses two issues that lead to the error together.

1. We've joined tasks table in two ways, which was not necessary and
   caused double join on foreman_tasks table.
2. We've not referenced table name in the result mapper, which caused
   Ambigous column even when the tasks table was properly aliased.

Also fixing N+1 query introduced by 8f0ba81e3634c2f015814ee77e1b303f77ef3faa. That should probably be a separate PR, but it would have conflicts with this one, so I thought we can do it at once, the resulting history should be fine.

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -67,7 +67,7 @@ class JobInvocationsController < ApplicationController
   end
 
   def index
-    @job_invocations = resource_base_search_and_page.with_task.order('job_invocations.id DESC')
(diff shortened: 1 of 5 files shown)


adamruzicka (commented):
I haven't managed to reproduce the original issue so far, I'll keep this applied for a bit and see if I manage to hit it

adamruzicka (approved):
ACK

adamruzicka:
Thank you @ezr-ondrej !

PR #601, opened 2021-06-23
Fixes #32854 - allow proxy_batch_size to work properly

ezr-ondrej:
The middleware always ran after every plan batch, effectivelly limiting
the tasks sent to proxy to a currectly planned batch.

This moves to a Trigger by action and evoking that action when needed
and than last time at the end.

The trigger action is defined in theforeman/foreman-tasks#640 so that will need to go first and then we need bump dependency.

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -9,7 +9,6 @@ module Actions
       middleware.use Actions::Middleware::BindJobInvocation
       middleware.use Actions::Middleware::RecurringLogic
       middleware.use Actions::Middleware::WatchDelegatedProxySubTasks
-      middleware.use Actions::Middleware::ProxyBatchTriggering
 
       class CheckOnProxyActions; end
 
@@ -32,6 +31,9 @@ module Actions
         set_up_concurrency_control job_invocation
         input.update(:job_category => job_invocation.job_category)
         plan_self(:job_invocation_id => job_invocation.id)
+        input[:proxy_batch_size] ||= Setting['foreman_tasks_proxy_batch_size']
+        trigger_action = plan_action(Actions::TriggerProxyBatch, batch_size: proxy_batch_size, total_count: hosts.count)
+        input[:trigger_run_step_id] = trigger_action.run_step_id
       end
 
       def create_sub_plans
@@ -44,6 +46,25 @@ module Actions
           template_invocation.host_id = host.id
           trigger(RunHostJob, job_invocation, host, template_invocation, proxy_selector, { :use_concurrency_control => uses_concurrency_control })
         end
+      end
+
+      def spawn_plans
+        super
+      ensure
+        trigger_remote_batch
+      end
+
+      def trigger_remote_batch
+        batches_ready = (output[:planned_count] - output[:remote_triggered_count]) / proxy_batch_size
+        return unless batches_ready > 0
+
+        plan_event(Actions::TriggerProxyBatch::TriggerNextBatch[batches_ready], nil, step_id: input[:trigger_run_step_id])
+        output[:remote_triggered_count] += proxy_batch_size * batches_ready
+      end
+
+      def on_planning_finished
+        plan_event(Actions::TriggerProxyBatch::TriggerLastBatch, nil, step_id: input[:trigger_run_step_id])
+        super
       end
 
       def finalize
@@ -67,6 +88,7 @@ module Actions
 
       def initiate
         output[:host_count] = total_count
+        output[:remote_triggered_count] = 0
         super
       end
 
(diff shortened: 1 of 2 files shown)


ares:
could this be related to this? https://community.theforeman.org/t/ansible-default-batch-size-for-remote-execution-job-is-100-hosts-need-to-increase-the-limit-for-remote-execution-in-foreman-ui/24019/4

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -46,6 +48,19 @@ def create_sub_plans
         end
       end
 
+      def spawn_plans
+        super
+        if output[:planned_count] - output[:remote_triggered_count] >= proxy_batch_size
+          plan_event(Actions::TriggerProxyBatch::TriggerLastBatch, nil, step_id: input[:trigger_run_step_id])
```
adamruzicka, line 54: Is this necessary? Won't it lead to TriggerLastBatch being dispatched twice?

ezr-ondrej: That's outside wrong, this should be `TriggerNextBatch`

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -32,6 +32,8 @@ def plan(job_invocation)
         set_up_concurrency_control job_invocation
         input.update(:job_category => job_invocation.job_category)
         plan_self(:job_invocation_id => job_invocation.id)
+        trigger_action = plan_action(Actions::TriggerProxyBatch, batch_size: proxy_batch_size)
```
adamruzicka, line 35:
```suggestion
        trigger_action = plan_action(Actions::TriggerProxyBatch, batch_size: proxy_batch_size, total_count: hosts.count)
```

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -46,6 +48,19 @@ def create_sub_plans
         end
       end
 
+      def spawn_plans
+        super
+        if output[:planned_count] - output[:remote_triggered_count] >= proxy_batch_size
+          plan_event(Actions::TriggerProxyBatch::TriggerNextBatch, nil, step_id: input[:trigger_run_step_id])
+          output[:remote_triggered_count] += proxy_batch_size
+        end
+      end
```
adamruzicka, line 57: `spawn_plans` in one of the ancestors suspends the action, nothing after this `super` will be executed. 

```suggestion
      def spawn_plans
        super
      ensure
        if output[:planned_count] - output[:remote_triggered_count] >= proxy_batch_size
          plan_event(Actions::TriggerProxyBatch::TriggerNextBatch, nil, step_id: input[:trigger_run_step_id])
          output[:remote_triggered_count] += proxy_batch_size
        end
      end
```

adamruzicka (changes requested):
We need to remove use of the middleware from `RunHostsJob`.

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -104,6 +120,10 @@ def humanized_input
       def humanized_name
         '%s:' % _(super)
       end
+
+      def proxy_batch_size
+        Setting['foreman_tasks_proxy_batch_size']
```
adamruzicka, line 125: In theory this could lead to odd behaviours if the setting changed while the action is running. Would you be open to storing the batch size that is set in settings at the start of the action and then using that value till the end?

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -32,6 +31,8 @@ def plan(job_invocation)
         set_up_concurrency_control job_invocation
         input.update(:job_category => job_invocation.job_category)
         plan_self(:job_invocation_id => job_invocation.id)
+        trigger_action = plan_action(Actions::TriggerProxyBatch, batch_size: proxy_batch_size, total_count: hosts.count)
+        input[:trigger_run_step_id] = trigger_action.run_step_id
```
adamruzicka, line 35: What happens if batch triggering is turned off? Will `TriggerProxyBatch` ever finish?

adamruzicka (commented):
The happy path seems to work well.

I had one odd scenario (matchine ran out of memory, had to kill it), where after resume `TriggerProxyBatch` resumed (and suspended), but the main action (`RunHostsJob`) ended up in `skipped`. Could we try reacting to `Skip` event in `RunHostsJob#run` to notify the other action?

app/lib/actions/remote_execution/run_hosts_job.rb
ezr-ondrej: Yeah, it will do nothing and the last `TriggerLastBatch` event will cause it to finish.

app/lib/actions/remote_execution/run_hosts_job.rb
ezr-ondrej: And right away realized it's useless, so I've put it into `input` :)

adamruzicka:
Test failures seem related

adamruzicka (approved):
Seems to work well, ACK

PR #602, opened 2021-06-24
Fixes #32873 - Do not check bookmark permissions

adamruzicka:


tbrisker (approved):
Thanks @adamruzicka !

PR #603, opened 2021-06-24
Fixes #32676 - Add status of REX jobs on dashboard

pkoprda:
Adds widget for latest REX jobs.

Screenshot from 2021-06-24 13-45-30

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -139,6 +139,10 @@ class JobInvocation < ApplicationRecord
     description
   end
 
+  def latest_jobs(limit = 5)
+    JobInvocation.order(id: :desc).limit(limit)
+  end
+
   # returns progress in percents
   def progress(total = nil, done = nil)
     if queued? || !targeting.resolved? || done == 0
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -75,6 +75,8 @@ class JobInvocation < ApplicationRecord
 
   default_scope -> { order('job_invocations.id DESC') }
 
+  scope :latest_jobs, -> { unscoped.joins(:tasks).order('foreman_tasks_tasks.started_at DESC').authorized(:view_job_invocations).limit(5) }
+
   validates_lengths_from_database :only => [:description]
 
   attr_accessor :start_before, :description_format
@@ -137,10 +139,6 @@ class JobInvocation < ApplicationRecord
 
   def to_label
     description
-  end
-
-  def latest_jobs(limit = 5)
-    JobInvocation.order(id: :desc).limit(limit)
   end
 
   # returns progress in percents
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -75,7 +75,7 @@ class JobInvocation < ApplicationRecord
 
   default_scope -> { order('job_invocations.id DESC') }
 
-  scope :latest_jobs, -> { unscoped.joins(:tasks).order('foreman_tasks_tasks.started_at DESC').authorized(:view_job_invocations).limit(5) }
+  scope :latest_jobs, -> { unscoped.joins(:task).order('foreman_tasks_tasks.start_at DESC').authorized(:view_job_invocations).limit(5) }
 
   validates_lengths_from_database :only => [:description]
 
--- a/app/views/dashboard/_latest-jobs.html.erb
+++ b/app/views/dashboard/_latest-jobs.html.erb
@@ -1,0 +1,17 @@
+<% stylesheet 'foreman_remote_execution/foreman_remote_execution' %>
+
+<h4 class="header"><%= link_to _("Latest Jobs"), job_invocations_path(:order=>'started_at DESC')%></h4>
+<table class="<%= table_css_classes('table-fixed') %>">
+  <tr>
+    <th class="col-md-5"><%= _("Name") %></th>
+    <th class="col-md-2"><%= _("State") %></th>
+    <th class="col-md-3"><%= _("Started") %></th>
+  </tr>
+  <% JobInvocation.new.latest_jobs.each do |invocation| %>
+    <tr>
+      <td class="ellipsis"><%= link_to_if_authorized invocation_description(invocation), hash_for_job_invocation_path(invocation).merge(:auth_object => invocation, :permission => :view_job_invocations, :authorizer => authorizer) %></td>
+      <td><%= link_to_invocation_task_if_authorized(invocation) %></td>
+      <td><%= time_in_words_span(invocation.start_at) %></td>
+    </tr>
+  <% end %>
+</table>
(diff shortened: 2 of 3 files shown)


app/views/dashboard/_latest-jobs.html.erb
```diff
@@ -0,0 +1,17 @@
+<% stylesheet 'foreman_remote_execution/foreman_remote_execution' %>
```
ares, line 1: probably not necessary

app/views/dashboard/_latest-jobs.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+<h4 class="header"><%= link_to _("Latest Jobs"), job_invocations_path(:order=>'started_at DESC')%></h4>
+<table class="<%= table_css_classes('table-fixed') %>">
+  <tr>
+    <th class="col-md-5"><%= _("Name") %></th>
+    <th class="col-md-2"><%= _("State") %></th>
+    <th class="col-md-3"><%= _("Started") %></th>
+  </tr>
+  <% JobInvocation.new.latest_jobs.each do |invocation| %>
```
ares, line 10: we shouldn't need to build the instance to load other job invocation instances, either make the `latest_jobs` a class method or convert it to the named scope

app/models/job_invocation.rb
```diff
@@ -139,6 +139,10 @@ def to_label
     description
   end
 
+  def latest_jobs(limit = 5)
+    JobInvocation.order(id: :desc).limit(limit)
```
ares, line 143: This requires permission check, there's `.authorized` named scope that can be used, just pass a permission name to it, the app/* contains many examples of use :-)

ares (changes requested):
Good start, adding notes we discussed on call.

app/models/job_invocation.rb
adamruzicka, line 143: Food for thought: Do we want last N jobs by time when they were submitted (and therefore by id) or by start_at/started_at?

app/models/job_invocation.rb
```diff
@@ -139,6 +139,10 @@ def to_label
...
+  end
```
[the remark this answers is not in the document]
adamruzicka: Dupe of https://github.com/theforeman/foreman_remote_execution/pull/603#discussion_r658642413 , closing.

app/models/job_invocation.rb
ares: Hm, interesting :-) I'd ask the reporter :-)

pheeeter: Good catch :+1: jobs are now sorted by time

app/models/job_invocation.rb
```diff
@@ -75,6 +75,8 @@ class JobInvocation < ApplicationRecord
 
   default_scope -> { order('job_invocations.id DESC') }
 
+  scope :latest_jobs, -> { unscoped.joins(:tasks).order('foreman_tasks_tasks.started_at DESC').authorized(:view_job_invocations).limit(5) }
```
adamruzicka, line 78: Most likely you want `joins(:task)` instead of `joins(:tasks)`. The first looks only at the "main" task of each job, the latter looks at all tasks within each job.

adamruzicka, line 78: This doesn't seem to work 
image

It is possible that NULLs go first and are shown in some implicit order and then the actual shown value comes from `foreman_tasks_task.start_at`. Maybe using that would be a better option?

adamruzicka (changes requested):

app/models/job_invocation.rb
pkoprda: With these suggestions it creates this query:
```sql
SELECT "job_invocations".* FROM "job_invocations" INNER JOIN "foreman_tasks_tasks" ON "foreman_tasks_tasks"."id" = "job_invocations"."task_id" ORDER BY foreman_tasks_task.start_at DESC LIMIT $1  [["LIMIT", 5]]
```

but I think it should join with `foreman_tasks_task`, because this throws error

adamruzicka: Oh, sorry, my bad. That was a typo. The table you're using is correct, but a different field (`start_at` instead of `started_at`) would most likely give better results.

pkoprda: Oh okay, now it should be good :+1:

pkoprda:
Thank you @adamruzicka for your comments :+1:

app/views/dashboard/_latest-jobs.html.erb
```diff
@@ -0,0 +1,15 @@
+<h4 class="header"><%= link_to _("Latest Jobs"), job_invocations_path(:order=>'started_at DESC')%></h4>
```
adamruzicka, line 1: This should be `start_at DESC`, otherwise it is not picked up properly.

app/views/dashboard/_latest-jobs.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <th class="col-md-2"><%= _("State") %></th>
+    <th class="col-md-3"><%= _("Started") %></th>
+  </tr>
+  <% JobInvocation.latest_jobs.each do |invocation| %>
+    <tr>
+      <td class="ellipsis"><%= link_to_if_authorized invocation_description(invocation), hash_for_job_invocation_path(invocation).merge(:auth_object => invocation, :permission => :view_job_invocations, :authorizer => authorizer) %></td>
+      <td><%= link_to_invocation_task_if_authorized(invocation) %></td>
+      <td><%= time_in_words_span(invocation.start_at) %></td>
+    </tr>
+  <% end %>
```
adamruzicka, line 14: If there are no jobs, it just shows an empty table. Could we place a placeholder in there? Something along the lines of what we show in "Host in build mode", "latest events" or "Host configuration chart for ansible"?

adamruzicka (changes requested):

app/views/dashboard/_latest-jobs.html.erb
pkoprda: I think it would be nice to have there some placeholder. But I am not sure if we want to keep link to `/job_invocations` page when no jobs are available.

app/views/dashboard/_latest-jobs.html.erb
```diff
@@ -0,0 +1,21 @@
+<h4 class="header">
+  <%= link_to _("Latest Jobs"), job_invocations_path(:order=>'start_at DESC')%>
+</h4>
+<% if JobInvocation.latest_jobs.limit(1).present? %>
```
adamruzicka, line 4: `.any?` should work instead of `.limit(1).present?`

app/views/dashboard/_latest-jobs.html.erb
```diff
@@ -0,0 +1,21 @@
+<h4 class="header">
+  <%= link_to _("Latest Jobs"), job_invocations_path(:order=>'start_at DESC')%>
```
adamruzicka, line 2: Missing space between `)` and `%>`

adamruzicka (approved):
Looks good

PR #606, opened 2021-06-29
Fixes #32907 - Check permission for job wizard

xprazak2:
cc @MariaAga

Needs https://github.com/theforeman/foreman/pull/8628

A rough sketch that depends on whether we have the permissions in context. This could be even handled in `AppSwitcher` if routes registered with permissions:

```js
const ForemanREXRoutes = [
  {
    path: '/experimental/job_wizard',
    exact: true,
    render: props => <JobWizardPage {...props} />,
    permission: 'create_job_invocations'
  },
]
```

### DIFF
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -82,6 +82,7 @@ export const JobWizard = () => {
       canJumpTo: isTemplate,
     },
   ];
+
   return (
     <Wizard
       onClose={() => history.goBack()}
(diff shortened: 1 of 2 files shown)


xprazak2:
run-job

MariaAga:
Thanks!
Looks like in needs a few linting fixes :) 
I think it should be in `JobWizard/index.js` so that the api calls to view templates won't be made.

xprazak2:
Based on the Discourse thread, it seems like we will not be adding permissions to context, therefore closing. Thanks!

PR #608, opened 2021-07-07
Fixes #32951 - Add rerun button to template invocations

pkoprda:


app/views/template_invocations/show.html.erb
```diff
@@ -19,6 +19,7 @@ end
 <div id="title_action">
   <div class="btn-toolbar pull-right">
     <%= link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation), :class => 'btn btn-default') %>
+    <%= link_to(_('Rerun'), rerun_job_invocation_path(@template_invocation.job_invocation, :host_ids => [ @host.id ]), :class => 'btn btn-default') if @host and @host.id.in? @template_invocation.job_invocation.failed_host_ids %>
```
adamruzicka, line 22: Iirc the convention is to just disable the button when it is not applicable instead of hiding it completely

adamruzicka, line 22: While ruby supports `and` and `or` for flow control, `&&` and `||` are usually used.

adamruzicka, line 22: Is there a possibility `@host` will be `nil`?

pheeeter: Hmm, probably not :+1:

ares: also this should check for create_job_invocations permission I think

adamruzicka: also this :point_up:

pkoprda: Added permission :)

app/views/template_invocations/show.html.erb
```diff
@@ -18,7 +18,8 @@ end
 
 <div id="title_action">
   <div class="btn-toolbar pull-right">
-    <%= link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation), :class => 'btn btn-default') %>
+    <%= button_group(link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation), :class => 'btn btn-default'),
+                     link_to(_('Rerun'), rerun_job_invocation_path(@template_invocation.job_invocation, :host_ids => [ @host.id ]), :class => 'btn btn-default', :disabled => @template_invocation.job_invocation.failed_host_ids.exclude?(@host.id), :permission => :create_job_invocations)) %>
```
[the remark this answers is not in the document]
ares: I'd show it all the time, I find it useful to rerun the job even in case it succeeded.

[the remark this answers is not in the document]
adamruzicka: @pkoprda :point_up:

pkoprda:
@adamruzicka, @ares thank you for your suggestions, please let me know if anything else is needed :).

app/views/template_invocations/show.html.erb
```diff
@@ -18,7 +18,8 @@ end
 
 <div id="title_action">
   <div class="btn-toolbar pull-right">
-    <%= link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation), :class => 'btn btn-default') %>
+    <%= button_group(link_to(_('Back to Job'), job_invocation_path(@template_invocation.job_invocation), :class => 'btn btn-default'),
+                     link_to(_('Rerun'), rerun_job_invocation_path(@template_invocation.job_invocation, :host_ids => [ @host.id ]), :class => 'btn btn-default', :permission => :create_job_invocations)) %>
```
adamruzicka, line 22: Not sure what were the expectations about including `:permission => :create_job_invocations` but it doesn't seem to do anything. From what I'm seeing it just sets `permission="create_job_invocations"` attribute on the link tag.

adamruzicka (changes requested):

app/views/template_invocations/show.html.erb
pkoprda: My bad, now it should be correct :+1:

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @pkoprda !

PR #609, opened 2021-07-07
Fixes #29950 - Map Run Puppet Once template to its feature

adamruzicka:
The code we had was most likely supposed to link the feature against an
already existing template. In case the template didn't exist when the
seeds were run, nothing established the relationship.

### DIFF
--- a/app/views/templates/ssh/module_action.erb
+++ b/app/views/templates/ssh/module_action.erb
@@ -29,6 +29,7 @@ template_inputs:
   input_type: user
   required: false
   advanced: true
+feature: :katello_module_stream_action
 %>
 
 <%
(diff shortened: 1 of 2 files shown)


ares (approved):
Thanks @adamruzicka, makes sense. Merging!

app/views/templates/ssh/module_action.erb
```diff
@@ -29,6 +29,7 @@ template_inputs:
   input_type: user
   required: false
   advanced: true
+feature: :katello_module_stream_action
```
ofedoren, line 32: @adamruzicka, just curious, after I've updated my local checkout and ran db:seed it failed with `... error (Psych::DisallowedClass): Tried to load unspecified class: Symbol`. Is `katello_module_stream_action` should be specified as a symbol?

adamruzicka: Wonder why I didn't see this. https://github.com/theforeman/foreman_remote_execution/pull/634

PR #610, opened 2021-07-07
Try capturing logs from GH actions

adamruzicka:


ezr-ondrej (approved):

ezr-ondrej:
Do you want to get this in, or did you want them just once?

adamruzicka:
There were some test failures that I wanted to look into. Luckily they happen here as well, so I'd keep this one as a playground and then merge it together with the "real" fix

test/unit/job_invocation_composer_test.rb
```diff
@@ -925,12 +925,14 @@ def provider_inputs
...
-        )
+        Foreman.settings._add('remote_execution_form_job_template',
+          context: :test,
+          type: :string,
+          category: 'Setting::RemoteExecution',
+          full_name: '',
+          description: '',
+          default: '')
+        Setting[:remote_execution_form_job_template] = setting_template.name
       end
```
adamruzicka, line 936: @ezr-ondrej is this how settings in tests are supposed to be handled?

test/unit/job_invocation_composer_test.rb
```diff
@@ -925,12 +925,14 @@ def provider_inputs
...
-          :category => 'Setting::RemoteExecution',
-          :value => setting_template.name
-        )
+        Foreman.settings._add('remote_execution_form_job_template',
+          context: :test,
+          type: :string,
+          category: 'Setting::RemoteExecution',
+          full_name: '',
+          description: '',
+          default: '')
```
ezr-ondrej, line 934: this should not be needed, that should be already there, only settings that are added only in tests need to use this.
Simple `Setting[<name>] = value` should do the job for existing settings.

test/unit/job_invocation_composer_test.rb
ezr-ondrej: ^^

adamruzicka:
:green_apple:

ezr-ondrej:
Thanks @adamruzicka ! :)

PR #612, opened 2021-07-12
Fixes #32992 - some rex settings are translated twice

MariaAga:


ezr-ondrej (approved):
Thanks @MariaAga !

ezr-ondrej:
@adamruzicka is it worth cherry-picking or there is not another release planned?

adamruzicka:
I'll cp it when I make the release

PR #613, opened 2021-07-12
Completely drop foreman_remote_execution_core

adamruzicka:


ezr-ondrej:
Isn't there any other plugin depending on foreman_core? or we just simply saying "require the latest alias-only version of rex_core" there is not gonna be another one?

adamruzicka:
According to rubygems we should be safe. The only remaining ones are foreman_ansible_core (discontinued), and smart_proxy_{salt,acd}. Both were patched to not depend on rex core, they just weren't released yet.

ezr-ondrej (approved):
Lets Purge! :)
Thanks @adamruzicka !

PR #615, opened 2021-07-14
Fixes #33007 - Handle taxonomies gracefully on output endpoints

adamruzicka:
When organization_id or location_id parameters were supplied, the nested
resource handling machinery resolved the organization (or location) to
be the parent resource instead of the expected job invocation.

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -162,6 +162,15 @@ module Api
         render :json => { :outputs => outputs }
       end
 
+      def resource_name(resource = controller_name)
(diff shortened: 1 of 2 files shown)


adamruzicka:
@ezr-ondrej could you take a look please?

ezr-ondrej:
The parent_resource mechanism is probably worth revisiting in core, as this will happen for any nested resource.

ezr-ondrej (approved):
Unfortunatelly this is the best fix... There should not be anything relying on getting correct `resource_name('organization')` but if there is this would break it.
Lets merge this, but we shoud really give a look to the `parent_resource_details`

PR #616, opened 2021-07-14
Fixes #32986 - Slow loading of hosts in job invocation page

hao-yu:
When login as non-admin user with viewer permission, the host
list in the job invocation page is taking long time to load.

This is fixed by passing the cached collection when checking
the user permission so the application doesn't need to repeat
the same query for every row.

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -7,6 +7,10 @@ module RemoteExecutionHelper
     @job_hosts_authorizer ||= Authorizer.new(User.current, :collection => @hosts)
   end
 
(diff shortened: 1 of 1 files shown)


adamruzicka (approved):
Very nice :+1: 

@ares you're more versed in everything around authorization than I am, anything to add before merge?

ares:
Permission-wise this is ok. I was looking at where we run the method as caching this for e.g. 10k hosts would cause performance (memory) issues. But if I understand correctly, we use it just for a single page of the table in the job invocation page. Therefore, :+1: and merging. Thanks @hao-yu and @adamruzicka

PR #617, opened 2021-07-14
Fixes #31705 - Handle timezones when scheduling jobs through API

adamruzicka:
We claimed to have support for iso8601 timestamps, but in reality we
discarded timezone information and assumed the timestamp to be in UTC.

### DIFF
--- a/app/models/job_invocation_composer.rb
+++ b/app/models/job_invocation_composer.rb
@@ -209,7 +209,7 @@ class JobInvocationComposer
     def format_datetime(datetime)
       return datetime if datetime.blank?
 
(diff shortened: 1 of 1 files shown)


ezr-ondrej (approved):
:+1:

ezr-ondrej:
This can be quite a change for some users in +6/-6 zones, can we have a upgrade note for this?

adamruzicka:
Upgrade note where?

ezr-ondrej:
> Upgrade note where?

Eh :shrug: we have no such thing for plugins, right? We totally should have, but eh :shrug:

PR #618, opened 2021-07-15
Fixes #32413 - Use setting DSL for REX category

pkoprda:
This uses new DSL to define Remote Execution category.


lib/foreman_remote_execution/engine.rb
```diff
@@ -61,6 +57,100 @@ class Engine < ::Rails::Engine
...
+              description: N_('Choose a job template that is pre-selected in job invocation form'),
+              default: 'Run Command - SSH Default',
+              full_name: N_('Form Job Template'),
+              collection: proc { Hash[JobTemplate.unscoped.map { |template| [template.name, template.name] }] }
+            setting 'remote_execution_job_invocation_report_template',
+              type: :string,
+              description: N_('Select a report template used for generating a report for a particular remote execution job'),
+              default: 'Jobs - Invocation report template',
+              full_name: N_('Job Invocation Report Template'),
+              collection: proc { self.job_invocation_report_templates_select }
```
adamruzicka, line 150: If I'm looking right then `self.job_invocation_report_templates_select` no longer exists.

adamruzicka (changes requested):

lib/foreman_remote_execution/engine.rb
pkoprda: :+1: but also I think that it should be removed from this file: https://github.com/theforeman/foreman_remote_execution/blob/e492aef4b0de384011990b7ddb28447dd3d3c8bf/test/unit/job_invocation_report_template_test.rb

lib/foreman_remote_execution/engine.rb
```diff
@@ -61,6 +57,99 @@ class Engine < ::Rails::Engine
...
+              type: :string,
+              description: N_('Choose a job template that is pre-selected in job invocation form'),
+              default: 'Run Command - SSH Default',
+              full_name: N_('Form Job Template'),
+              collection: proc { Hash[JobTemplate.unscoped.map { |template| [template.name, template.name] }] }
+            setting 'remote_execution_job_invocation_report_template',
+              type: :string,
+              description: N_('Select a report template used for generating a report for a particular remote execution job'),
+              default: 'Jobs - Invocation report template',
+              full_name: N_('Job Invocation Report Template')
```
adamruzicka, line 149: This used to have options to select from

adamruzicka (changes requested):
Needs a rebase

app/models/setting/remote_execution.rb
```diff
@@ -1,88 +1,28 @@
 class Setting::RemoteExecution < Setting
```
ezr-ondrej, line 1: Why are we keeping it? We are referencing the class somewhere?
The best case scenario we get rid of the class :)

pkoprda: I agree, but I ran into some dependency problem. It was trying to load `Setting::RemoteExecution` class which needs to be defined in this file, but I don't know why :/.

pkoprda: Thanks for your help :+1:

ezr-ondrej (approved):
LGTM :+1:

ezr-ondrej:
But you've got an conflict, it needs rebasing, probably some change in the settings in the meantime

ezr-ondrej (approved):
Works well for me :+1:

app/models/setting/remote_execution.rb
```diff
@@ -1,88 +0,0 @@
-class Setting::RemoteExecution < Setting
-
-  ::Setting::BLANK_ATTRS.concat %w{remote_execution_ssh_password remote_execution_ssh_key_passphrase remote_execution_sudo_password remote_execution_effective_user_password remote_execution_cockpit_url remote_execution_form_job_template}
```
ezr-ondrej, line 3: oh, we can't do without this... This can be:

1. hacked into the BLANK_ATTRS
2. wait for the validations to get in and set through the `allow_empty`

pkoprda: I don't know if I understood that correctly. Should I include `BLANK_ATTRS` in the `engine.rb` file?

ezr-ondrej: Yeah, that would be an option, but we are planning to remove that validation completely, so it's better to wait for https://github.com/theforeman/foreman/pull/8470

adamruzicka: Support for validation is now merged in core, would be nice to use it here.

adamruzicka:
Also needs a rebase

ezr-ondrej (approved):
@adamruzicka this is ready IMHO, any more comments?

db/migrate/20210816100932_fix_setting_category.rb
```diff
@@ -0,0 +1,5 @@
+class FixSettingCategory < ActiveRecord::Migration[6.0]
```
ezr-ondrej, line 1: I've just hitted an issue of name clash, because in ansible https://github.com/theforeman/foreman_ansible/pull/448 we took example from this PR and Rails complain about the same name.
It would be probably better to play it safe and name it more specific (both the class and the file needs to be renamed).

lib/foreman_remote_execution/engine.rb
```diff
@@ -61,6 +57,107 @@ class Engine < ::Rails::Engine
...
+              type: :string,
+              description: N_('Where to find the Cockpit instance for the Web Console button.  By default, no button is shown.'),
+              default: nil,
+              full_name: N_('Cockpit URL')
+            setting 'remote_execution_form_job_template',
+              type: :string,
+              description: N_('Choose a job template that is pre-selected in job invocation form'),
+              default: 'Run Command - SSH Default',
+              full_name: N_('Form Job Template'),
+              collection: proc { Hash[JobTemplate.unscoped.map { |template| [template.name, template.name] }] }
```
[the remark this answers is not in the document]
adamruzicka: Looks like I didn't have new-enough Foreman. Nevermind

adamruzicka (approved):
Works like a charm

adamruzicka:
Thank you @pkoprda & @ezr-ondrej !

PR #620, opened 2021-07-16
Refs #32606 - Make method names match

adamruzicka:


adamruzicka:
The tests seem to be running against foreman develop, they should be green if run against 2.5

ares (approved):
Thanks @adamruzicka, merging

PR #624, opened 2021-07-21
Fixes #33087 - Add password options to hammer REX job

pkoprda:
Add options for setting sudo password and ssh password when creating REX job using hammer


adamruzicka (changes requested):
Effective user password is ignored. It gets sent over to the server, but then I gets removed here https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb#L200

pkoprda:
Thanks for your review @adamruzicka :+1: Let me know if anything else is needed

adamruzicka (commented):
While you're at it, could you also add support for providing key passphrase?

adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @pkoprda !

PR #628, opened 2021-07-22
Fixes #33112 - Add parameter for recurring logic purpose

xprazak2:
Needs https://github.com/theforeman/foreman-tasks/pull/644


adamruzicka:
Works nicely. Any plans on exposing this in the current rex form?

adamruzicka:
Works well, my only gripe is that if a RL with given purpose already exists when triggering a job through the API, it just fails with
```
$ hammer job-invocation create --feature ansible_run_playbook --inputs playbook="asd" --search-query='name = test-00.example.com' --cron-line '0 0 * * *'  --purpose testing 
Validation failed:
```

debug logs don't help all that much

```
[ERROR 2021-09-01T13:52:09 API] 500 Internal Server Error
[DEBUG 2021-09-01T13:52:09 API] {
    "error" => {
        "message" => "Validation failed: "
    }
}
```

Could we do something about this?

xprazak2:
Yes, it is an existing problem even for 
```
hammer job-invocation create --feature ansible_run_playbook --inputs playbook="asd" --search-query='name = test-00.example.com' --cron-line foo
```


Fixing separately in https://github.com/theforeman/foreman_remote_execution/pull/648

adamruzicka:
Thank you @xprazak2 !

PR #629, opened 2021-07-23
 Fixes #33125 - autofill wizard with feature and inputs

MariaAga:
depends on: https://github.com/theforeman/foreman_remote_execution/pull/626
example url: `?feature=ansible_run_playbook&inputs[playbook]=hello`
There are some changes in the use effect of selecting category and template.
workflow should be:
No feature in url? select first category -> get categories templates -> select first template
feature in url? select the template -> select the templates category -> get categories templates -> use effect sees that there is a feature in the url and doesnt select the first template. Also deletes the feature so if a user selects a different category the first template of the category will be selected.

PR #632, opened 2021-07-23
Fixes #32990 - job wizard select default template

MariaAga:
The default template is set in the Remote Execution settings
depends on: https://github.com/theforeman/foreman_remote_execution/pull/629

### DIFF
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -6,7 +6,7 @@ class UiJobWizardController < ApplicationController
                      .select(:job_category).distinct
                      .reorder(:job_category)
                      .pluck(:job_category)
-    render :json => {:job_categories =>job_categories, :with_katello => with_katello}
+    render :json => {:job_categories => job_categories, :with_katello => with_katello, default_category: default_category, default_template: default_template&.id}
   end
 
   def template
@@ -22,6 +22,16 @@ class UiJobWizardController < ApplicationController
 
   def map_template_inputs(template_inputs_with_foreign)
     template_inputs_with_foreign.map { |input| input.attributes.merge({:resource_type_tableize => input.resource_type&.tableize }) }
+  end
+
+  def default_category
+    default_template&.job_category
+  end
+
+  def default_template
+    if (setting_value = Setting['remote_execution_form_job_template'])
+      JobTemplate.authorized(:view_job_templates).find_by :name => setting_value
+    end
   end
 
   def resource_name(nested_resource = nil)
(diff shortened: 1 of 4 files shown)


app/controllers/ui_job_wizard_controller.rb
```diff
@@ -24,6 +24,19 @@ def map_template_inputs(template_inputs_with_foreign)
     template_inputs_with_foreign.map { |input| input.attributes.merge({:resource_type_tableize => input.resource_type&.tableize }) }
   end
 
+  def default_category
+    default_template&.job_category
+  end
+
+  def default_template
```
adamruzicka, line 31: This is almost the same as https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_invocation_composer.rb#L643 . For now I wouldn't mind having it duplicated and I'd like to see it eventually removed from the composer once we fully move to the new form.

However, the implementation in composer currently checks permissions, which this one does not. We should check them here as well

adamruzicka (changes requested):
One change needed, apart from it it works nicely

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -24,6 +24,19 @@ def map_template_inputs(template_inputs_with_foreign)
...
+    setting_value = Setting['remote_execution_form_job_template']
+    return nil unless setting_value
```
adamruzicka, line 33: `return nil` is equivalent to just `return`, there's no need to be this explicit

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -24,6 +24,19 @@ def map_template_inputs(template_inputs_with_foreign)
...
+
+    template = JobTemplate.authorized(:view_job_templates).find_by :name => setting_value
+    return nil unless template
+    template
```
adamruzicka, line 37: `find_by` returns either the thing or nil. We don't really have to check if it is nil to return nil or the thing itself

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -24,6 +24,19 @@ def map_template_inputs(template_inputs_with_foreign)
...
+  end
```
adamruzicka, line 38: All in all we could even do something like this, but that boils down to personal preference. Either way is fine as long we don't do unneeded checks

```suggestion
  def default_template
    if setting_value = Setting['remote_execution_form_job_template']
      JobTemplate.authorized(:view_job_templates).find_by :name => setting_value
    end
  end
```

MariaAga: Used this instead, with added parenthesis to make rubocop happy.
Thanks for the review!

adamruzicka (approved):
Works nicely

adamruzicka:
Thank you @MariaAga !

PR #635, opened 2021-07-26
Fixes #33143 - hide rex buttons for unpriviledged

ezr-ondrej:
Users who can't run the Host REX jobs should not see the REX buttons.

### DIFF
--- a/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
@@ -7,7 +7,9 @@ module ForemanRemoteExecution
     end
 
     def multiple_actions
-      super + [ [_('Schedule Remote Job'), new_job_invocation_path, false] ]
+      res = super
+      res += [ [_('Schedule Remote Job'), new_job_invocation_path, false] ] if authorized_for(controller: :job_invocations, action: :new)
+      res
     end
 
     def schedule_job_multi_button(*args)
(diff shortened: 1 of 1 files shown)


app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -38,8 +38,10 @@ def web_console_button(host, *args)
     end
 
     def host_title_actions(*args)
-      title_actions(button_group(schedule_job_multi_button(*args)),
-        button_group(web_console_button(*args)))
+      if authorized_for(permission: 'create_job_invocations', auth_object: args.first)
+        title_actions(button_group(schedule_job_multi_button(*args)),
+          button_group(web_console_button(*args)))
```
adamruzicka, line 43: I don't think this permission is needed to access the web console

ezr-ondrej: Fixed it, it seems both button_group and title_actions are handling nil, quite ok :)

tbrisker (approved):

app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb
```diff
@@ -21,6 +21,7 @@ def schedule_job_multi_button(*args)
     end
 
     def rex_host_features(*args)
+      return unless authorized_for(permission: 'create_job_invocations', auth_object: args.first)
```
tbrisker, line 24: Should the same also be done in `schedule_job_button` below?

ezr-ondrej: Good idea :+1: Done that.

adamruzicka:
There's still a schedule job button on host index. Could you address it as well while you're at it or is it out of scope?

ezr-ondrej:
I've fixed the index page now as well :)

adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @ezr-ondrej !

PR #636, opened 2021-07-27
Refs #33024 - Do not crash during seeding if feature does not exist

adamruzicka:


adamruzicka:
Judging from `feature = RemoteExecutionFeature.feature(feature_name)` here[1] we expect `RemoteExecutionFeature.feature(label)` not to raise. Added a bang variant if this behavior is needed

[1] - https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_template.rb#L171

ezr-ondrej:
I'd expect the template not to be imported at all if the feature is missing.
It'd make sense to me, but I'm not sure if that's something we want :thinking:

adamruzicka:
My assumption is that we have the template here and the feature in katello because the template can be useful even without the feature. But I can't say what the original reasoning was

ezr-ondrej (approved):
Make sense :shrug: :+1:

PR #638, opened 2021-07-27
Fixes #31564 - add email notification

ares:
User can subscribe to either all job notifications, successful jobs or
failed jobs or no notifications. The user who runs the job would then be
notified based on the subscription. It supports both plain text and html
emails.

Since now we have a way to run the notification even on failure, the
notification drawer notification has been also added in case the job
fails.

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -10,6 +10,9 @@ module Actions
       middleware.use Actions::Middleware::RecurringLogic
       middleware.use Actions::Middleware::WatchDelegatedProxySubTasks
 
+      execution_plan_hooks.use :notify_on_success, :on => :success
+      execution_plan_hooks.use :notify_on_failure, :on => :failure
+
       class CheckOnProxyActions; end
 
       def queue
@@ -73,8 +76,22 @@ module Actions
 
         Rails.logger.debug "cleaning cache for keys that begin with 'job_invocation_#{job_invocation.id}'"
         Rails.cache.delete_matched(/\A#{JobInvocation::CACHE_PREFIX}_#{job_invocation.id}/)
-        # creating the success notification should be the very last thing this tasks do
+      end
+
+      def notify_on_success(_plan)
         job_invocation.build_notification.deliver!
+
+        if [RexMailNotification::SUCCEEDED_JOBS, RexMailNotification::ALL_JOBS].include?(mail_notification_preference&.interval)
+          RexJobMailer.job_finished(job_invocation, subject: _("REX job has succeeded - %s") % job_invocation.to_s).deliver_now
+        end
+      end
+
+      def notify_on_failure(_plan)
+        job_invocation.build_notification.deliver!
+
+        if [RexMailNotification::FAILED_JOBS, RexMailNotification::ALL_JOBS].include?(mail_notification_preference&.interval)
+          RexJobMailer.job_finished(job_invocation, subject: _("REX job has failed - %s") % job_invocation.to_s).deliver_now
+        end
       end
 
       def job_invocation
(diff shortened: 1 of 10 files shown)


ares:
When testing, do not forget to seed your DB, then in your user preferences, set the subscription, try running a job that fails and succeeds. Even without the subscription, you should now see a job failure in the notification drawer.

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -52,8 +55,22 @@ def finalize
 
         Rails.logger.debug "cleaning cache for keys that begin with 'job_invocation_#{job_invocation.id}'"
         Rails.cache.delete_matched(/\A#{JobInvocation::CACHE_PREFIX}_#{job_invocation.id}/)
-        # creating the success notification should be the very last thing this tasks do
+      end
+
+      def notify_on_success(plan)
         job_invocation.build_notification.deliver!
+
+        if ['Subscribe to my succeeded jobs', 'Subscribe to all my jobs'].include?(mail_notification_preference&.interval)
```
adamruzicka, line 63: Could we get constants for all of these?

adamruzicka (commented):
In general +1.

Could we maybe instead of setting hooks on success and failure set a single hook on `:stopped` and then inside the hook decide which kind of email to send out to make it a bit DRYer?

app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -52,8 +55,22 @@ def finalize
...
+        end
+      end
+
+      def notify_on_failure(plan)
+        job_invocation.build_notification.deliver!
+
+        if ['Subscribe to my failed jobs', 'Subscribe to all my jobs'].include?(mail_notification_preference&.interval)
+          RexJobMailer.job_finished(job_invocation, subject: _("REX job has failed - %s") % job_invocation.to_s).deliver_now
+        end
       end
```
adamruzicka, line 74: I had in mind something like this, but don't feel too strongly about it

```suggestion
      def notify_on_finish(plan)
        job_invocation.build_notification.deliver!

        event = plan.result == 'success' ? 'succeeded' : 'failed'
        if ["Subscribe to my #{event} jobs", 'Subscribe to all my jobs'].include?(mail_notification_preference&.interval)
          RexJobMailer.job_finished(job_invocation, subject: _("REX job has %{event} - %{job}") % {:job => job_invocation.to_s, :event => event}).deliver_now
        end
    end
```

ares:
I extracted the strings to constants and fixed the rubocop

ezr-ondrej:
The failure is related.

ares:
The problematic test:

```
describe 'notifications' do
  it 'creates notification on sucess run' do
    FactoryBot.create(:notification_blueprint, :name => 'rex_job_succeeded')
    assert_difference 'NotificationRecipient.where(:user_id => targeting.user.id).count' do
      finalize_action planned
    end
  end
end
```

I think that's because I now use a different trigger then a code in finalize. @adamruzicka any thoughts on how to test this? Or should I simply drop the test?

adamruzicka:
IIRC we don't really have a real facility for testing hooks. On the other hand, it is just a regular method. You can just call it in the test to test it behaves how we want it to and then trust dynflow that it will run the hooks. If it helps you sleep any better, hooks are covered with tests in dynflow so there should be no issue there.

adamruzicka:
ping @ares

ares:
Repushed with the tests, sorry for the delay. Can this get merged before the next demo (please ping me on IRC if something else is required). Note the rubocop complains about snake case, but we call all files like that in seeds.d.

adamruzicka:
> rubocop complains about snake case, but we call all files like that in seeds.d.

Could you add it to .rubocop_todo.yml where all the other migration files are ignored?

ares:
And rebased since meanwhile it got a conflict

ares:
The test failure is IMHO unrelated, perhaps caused by https://github.com/theforeman/foreman/pull/8711?

adamruzicka (approved):
Seems to work well, ACK

adamruzicka:
Thank you @ares !

PR #639, opened 2021-07-28
Fixes #33163 - Fix ISE when viewing cancelled scheduled job

adamruzicka:


tbrisker (approved):
Thanks @adamruzicka !

PR #640, opened 2021-07-29
Fixes #32496 - Append usual paths to the end of PATH for power action

adamruzicka:
Prevents errors like shutdown: command not found


tbrisker (approved):
Thanks @adamruzicka !

PR #641, opened 2021-07-29
Fixes #27769 - Catch exception when using ed25519 algorithm

pkoprda:
Catch exception when user generates ssh key with ed25519 algorithm and task will fail with proper error message.


adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @pkoprda !

PR #643, opened 2021-08-11
Fixes #33242 - Allow false boolean parameters to override true settings

adamruzicka:
Previously we treated explicitly set to false and unset as the same thing.


adamruzicka:
Ok, can't set multiple milestones, nevermind then.

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #644, opened 2021-08-11
 Fixes #32890 - job wizard add resource picker

MariaAga:
Follow up for https://github.com/theforeman/foreman/pull/8600

### DIFF
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -10,7 +10,7 @@ class UiJobWizardController < ::Api::V2::BaseController
 
   def template
     job_template = JobTemplate.authorized.find(params[:id])
-    advanced_template_inputs, template_inputs = map_template_inputs(job_template.template_inputs_with_foreign).partition { |x| x["advanced"] }
+    advanced_template_inputs, template_inputs = job_template.template_inputs_with_foreign.partition { |x| x["advanced"] }
     render :json => {
       :job_template => job_template,
       :effective_user => job_template.effective_user,
@@ -23,15 +23,18 @@ class UiJobWizardController < ::Api::V2::BaseController
     nested_resource || 'job_template'
   end
 
-  def map_template_inputs(template_inputs_with_foreign)
-    template_inputs_with_foreign.map { |input| input.attributes.merge({:resource_type => input.resource_type&.tableize }) }
-  end
-
   def resource_class
     JobTemplate
   end
 
   def action_permission
     :view_job_templates
+  end
+
+  def resources
+    resource_type = params[:resource]
+    resource_list = resource_type.constantize.authorized("view_#{resource_type.underscore.pluralize}").all.map { |r| {:name => r.to_s, :id => r.id } }.select { |v| v[:name] =~ /#{params[:name]}/ }
+    render :json => { :results =>
+      resource_list.sort_by { |r| r[:name] }.take(100), :subtotal => resource_list.count}
   end
 end
--- a/webpack/JobWizard/__tests__/fixtures.js
+++ b/webpack/JobWizard/__tests__/fixtures.js
@@ -48,6 +48,16 @@ export const jobTemplateResponse = {
       hidden_value: false,
     },
     {
+      name: 'adv resource select',
+      required: false,
+      input_type: 'user',
+      value_type: 'resource',
+      advanced: true,
+      resource_type: 'ForemanTasks::Task',
+      default: '',
+      hidden_value: false,
+    },
+    {
       name: 'adv search',
       required: false,
       options: '',
@@ -106,7 +116,17 @@ export const testSetup = (selectors, api) => {
     { ...jobTemplate, id: 2, name: 'template2' },
   ]);
   const mockStore = configureMockStore([]);
-  const store = mockStore({});
+  const store = mockStore({
+    ForemanTasksTask: {
+      response: {
+        subtotal: 10,
+        results: [
+          { id: '1', name: 'resource1' },
+          { id: '2', name: 'resource2' },
+        ],
+      },
+    },
+  });
   return store;
 };
 
--- a/webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
+++ b/webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 import React from 'react';
 import { Provider } from 'react-redux';
 import { mount } from '@theforeman/test';
@@ -94,6 +95,9 @@ describe('AdvancedFields', () => {
       selector: 'textarea',
     });
     const selectField = screen.getByText('option 1');
+    const resourceSelectField = screen.getByLabelText(
+      'adv resource select toggle'
+    );
     const searchField = screen.getByPlaceholderText('Filter...');
     const dateField = screen.getByLabelText('adv date', {
       selector: 'input',
@@ -103,6 +107,10 @@ describe('AdvancedFields', () => {
     await act(async () => {
       await fireEvent.click(screen.getByText('option 2'));
       fireEvent.click(screen.getAllByText(WIZARD_TITLES.advanced)[0]); // to remove focus
+
+      fireEvent.click(resourceSelectField);
+      await fireEvent.click(screen.getByText('resource2'));
+
       await fireEvent.change(textField, {
         target: { value: textValue },
       });
@@ -136,6 +144,8 @@ describe('AdvancedFields', () => {
     expect(dateField.value).toBe(dateValue);
     expect(screen.queryAllByText('option 1')).toHaveLength(0);
     expect(screen.queryAllByText('option 2')).toHaveLength(1);
+    expect(screen.queryAllByDisplayValue('resource1')).toHaveLength(0);
+    expect(screen.queryAllByDisplayValue('resource2')).toHaveLength(1);
   });
   it('fill defaults into fields', async () => {
     render(
@@ -163,7 +173,7 @@ describe('AdvancedFields', () => {
         selector: 'input',
       }).value
     ).toBe(
-      'template1 with inputs adv plain hidden="Default val" adv plain select="" adv search="" adv date="" plain hidden="Default val"'
+      'template1 with inputs adv plain hidden="Default val" adv plain select="" adv resource select="" adv search="" adv date="" plain hidden="Default val"'
     );
   });
   it('DescriptionField', async () => {
@@ -296,5 +306,31 @@ describe('AdvancedFields', () => {
         selector: 'input',
       }).value
     ).toBe('Run Default val');
+  });
+
+  it('search resources action', async () => {
+    store.clearActions();
+    jest.useFakeTimers();
+
+    render(
+      <Provider store={store}>
+        <JobWizard />
+      </Provider>
+    );
+    await act(async () => {
+      fireEvent.click(screen.getByText(WIZARD_TITLES.advanced));
+    });
+    const resourceSelectField = screen.getByLabelText(
+      'adv resource select typeahead input'
+    );
+
+    await act(async () => {
+      await fireEvent.change(resourceSelectField, {
+        target: { value: 'some search' },
+      });
+
+      jest.runAllTimers();
+    });
+    expect(store.getActions()).toMatchSnapshot('resource search');
   });
 });
--- a/webpack/JobWizard/steps/form/Formatter.js
+++ b/webpack/JobWizard/steps/form/Formatter.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import SearchBar from 'foremanReact/components/SearchBar';
 import { helpLabel } from './FormHelpers';
 import { SelectField } from './SelectField';
+import { SearchSelect } from './SearchSelect';
 
 const TemplateSearchField = ({
   name,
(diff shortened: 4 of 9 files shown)


MariaAga:
Test failure is due to: https://github.com/patternfly/patternfly-react/issues/6146
Will be fixed when `4.148.0+` will be latest https://www.npmjs.com/package/@patternfly/react-core

webpack/JobWizard/__tests__/fixtures.js
```diff
@@ -48,6 +48,30 @@ export const jobTemplateResponse = {
...
+          name: 'resource2',
+          id: 10,
+        },
+        {
+          name: '',
+          id: 11,
+        },
+      ],
+      advanced: true,
+      value_type: 'plain',
```
laviro, line 70: do we need to also add a fixture of [value_type: 'resource](https://github.com/theforeman/foreman_remote_execution/pull/644/files#diff-494fb439c4dcbef79b8470f33ba943aa649b2d8e698a9bd26c5353e9e33158f7R28)' ?

webpack/JobWizard/steps/form/Formatter.js
```diff
@@ -56,7 +56,11 @@ export const formatter = (input, values, setValue) => {
   const value = values[name];
 
   if (isSelectType) {
-    const options = input.options.split(/\r?\n/).map(option => option.trim());
+    let { options } = input;
+    if (typeof options === 'string' || options instanceof String) {
```
laviro, line 60: can you give an example of how the options look?
when does `options instanceof String` needed?

webpack/JobWizard/steps/form/SelectField.js
```diff
@@ -34,9 +34,16 @@ export const SelectField = ({
         menuAppendTo={() => document.body}
         {...props}
       >
-        {options.map((option, index) => (
-          <SelectOption key={index} value={option} />
-        ))}
+        {options[0]?.id
```
laviro, line 37: if one option has an id, does it mean the rest of the options will also have the same structure?
or should we map them anyway, and if they contain an id, change the value?
e.g:
```js
options.map((option, index) => (
              <SelectOption key={index} value={option.id ? option.id : option} />
           ))}
```

webpack/JobWizard/steps/form/SelectField.js
```diff
@@ -34,9 +34,16 @@ export const SelectField = ({
...
+          ? options.map(({ id, name }) => (
+              <SelectOption key={id} value={id}>
+                {/* Should we show id or blank? */}
```
laviro, line 40: leftover?

webpack/JobWizard/steps/form/SelectField.js
```diff
@@ -34,9 +34,16 @@ export const SelectField = ({
...
+                {name || ' '}
```
laviro, line 41: do we need both a `value` & `children` props to be passed to the `SelectOption`?

webpack/JobWizard/steps/form/Formatter.js
MariaAga: Actually might change it to `if(inputType === 'plain')`
When type is plain the options need to be:
'a b c'
In the template

webpack/JobWizard/steps/form/SelectField.js
MariaAga: A question for the reviewer :)

webpack/JobWizard/steps/form/SelectField.js
MariaAga: Nope, like in the next line `<SelectOption key={index} value={option} />`.
But when an option has an id and a name the value that displays should be the name but the value that is stored should be the id.

MariaAga:
Depends on #646 for pf4

laviro:
@MariaAga need a small rebase

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -34,4 +39,13 @@ def resource_class
   def action_permission
     :view_job_templates
   end
+
+  def resource_value_options(resource_type)
+    return [] unless User.current.allowed_to?("view_#{resource_type.underscore.pluralize}".to_sym)
+
+    resource_type.constantize
+                 .all
+                 .sort_by(&:to_s)
```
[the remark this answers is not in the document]
MariaAga: Removed the sort here, I hope that what you meant :sweat_smile:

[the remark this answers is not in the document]
adamruzicka: Not sure if that's ideal. I mean something like `resource_type_constantize.order(:name).all...`, where the `order` makes the database sort it by the given column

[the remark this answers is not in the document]
MariaAga: Is there a chance that for different `resource_type` the `to_s` will be from a different field? And then I can't know which parameter to give to `order`?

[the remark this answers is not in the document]
adamruzicka: Hmm, that's true.  Nevermind then

laviro (approved):
Thanks @MariaAga !

app/controllers/ui_job_wizard_controller.rb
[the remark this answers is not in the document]
MariaAga: re-added the sort

adamruzicka:
In the old form, the input offers search, the new one does not. Is this something that could be addressed here?

MariaAga:
@adamruzicka I'm not sure, but it looks like resource lists can get very big, should I maybe use graphql for the search or do the filtering in js?

adamruzicka:
I'd prefer if we could let the server do the filtering, if possible. Graphql would probably be a better fit here than using the rest api

MariaAga:
Had to rewrite the pr so now we have live search. graphql is not possible as the filtering relies on `to_s` which is not a real field. 
If the search has more than 100 results this happens:
Screenshot from 2021-09-20 12-48-30

Between steps the resource name is saved so the user can see it when they return to the template inputs step. And when they return to the step the search is set to be their selection.

MariaAga:
Not sure how to fix the ruby tests, @adamruzicka any advice? :)

adamruzicka:
You're adding a new controller endpoint so you need to map a permission to it. Since permissions are explicitly checked by type of the resource, let's just add the endpoint here https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L70

webpack/JobWizard/steps/form/SearchSelect.js
```diff
@@ -0,0 +1,129 @@
...
+export const useNameSearchAPI = (url, apiKey) => {
+  const dispatch = useDispatch();
+  const onSearch = search =>
+    dispatch(
+      get({
+        key: apiKey,
+        url: search
+          ? url.addSearch({
+              name: search,
+              limit: maxResults,
```
MariaAga, line 26: Need to move `addSeach limit` to the formatter component

webpack/JobWizard/steps/form/SearchSelect.js
```diff
@@ -0,0 +1,129 @@
...
+  let selectOptions = [];
+  if (response.subtotal > maxResults) {
+    selectOptions = [
+      <SelectOption
+        isDisabled
+        key={0}
+        description={__('Please refine your search.')}
+      >
+        {sprintf(
+          __('You have %s reults to display. Showing first %s results'),
```
MariaAga, line 69: Typo, results

webpack/JobWizard/steps/form/SearchSelect.js
```diff
@@ -0,0 +1,129 @@
...
+      onSelect={onSelect}
+      onToggle={setIsOpen}
+      isOpen={isOpen}
+      className="without_select2"
+      maxHeight="45vh"
+      onTypeaheadInputChanged={value => {
+        autoSearch(value || '');
+      }}
+      placeholderText={placeholderText}
+      onFilter={() => null}
```
MariaAga, line 108: This is to prevent the default filtering that filters according to the first part of a name so `ello` would not find the `hello` result

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
```diff
@@ -297,4 +307,30 @@ describe('AdvancedFields', () => {
...
+    );
+
+    await act(async () => {
+      await fireEvent.change(resourceSelectField, {
+        target: { value: 'some search' },
+      });
+
+      jest.runAllTimers();
+    });
+    expect(store.getActions()).toMatchSnapshot('resource search');
```
laviro, line 334: instead of taking a snapshot, can you test where the UI changes?

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
```diff
@@ -297,4 +307,30 @@ describe('AdvancedFields', () => {
...
+    store.clearActions();
+    jest.useFakeTimers();
+
+    render(
+      <Provider store={store}>
+        <JobWizard />
+      </Provider>
+    );
+    await act(async () => {
+      fireEvent.click(screen.getByText(WIZARD_TITLES.advanced));
```
laviro, line 321: I see it in `Target hosts and inputs` tab, should it also appear in `Advanced Fields`?

webpack/JobWizard/steps/form/SearchSelect.js
```diff
@@ -0,0 +1,129 @@
...
+  Spinner,
+} from '@patternfly/react-core';
+import Immutable from 'seamless-immutable';
+import URI from 'urijs';
+import { sprintf, translate as __ } from 'foremanReact/common/I18n';
+import { get } from 'foremanReact/redux/API';
+import { selectResponse, selectIsLoading } from '../../JobWizardSelectors';
+
+const maxResults = 100;
+export const useNameSearchAPI = (url, apiKey) => {
```
laviro, line 17: Love the hook!

webpack/JobWizard/steps/form/SearchSelect.js
```diff
@@ -0,0 +1,129 @@
...
+      onToggle={setIsOpen}
+      isOpen={isOpen}
+      className="without_select2"
+      maxHeight="45vh"
+      onTypeaheadInputChanged={value => {
+        autoSearch(value || '');
+      }}
+      placeholderText={placeholderText}
+      onFilter={() => null}
+      typeAheadAriaLabel={`${name} text input`}
```
laviro, line 109: how about:
```suggestion
      typeAheadAriaLabel={`${name} typeahead input`}
```

laviro (commented):
Works great, thanks @MariaAga !

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
MariaAga: I want to see that the correct action is fired as the ui changes are only happening from the api response.

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
MariaAga: It comes from the same function in both: `formatter`

MariaAga:
I'm guessing test failures will be fixed after https://github.com/theforeman/foreman-js/pull/346

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
laviro: the approach of react testing library is to detect the UI changes rather than snapshots or implementation details,
IMHO we can mock the final API call and get results based on our search,
then we can expect to have the results in the DOM.

laviro:
> I'm guessing test failures will be fixed after [theforeman/foreman-js#346](https://github.com/theforeman/foreman-js/pull/346)

Looks like it's now part of `v8.12.2`
let's try to run it again

webpack/JobWizard/steps/AdvancedFields/__tests__/AdvancedFields.test.js
MariaAga: I'm going to split the resource search into having a wrapper, and then test the wrapper it self with a mock responses, is it ok that we leave the test fixes for the next pr? (The host select)

laviro: Sure, np 👍🏼

laviro (approved):

MariaAga:
@adamruzicka can this be merged? :)

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -23,15 +23,18 @@ def resource_name(nested_resource = nil)
...
 
   def action_permission
     :view_job_templates
   end
+
+  def resources
+    resource_type = params[:resource]
+    resource_list = resource_type.constantize.authorized("view_#{resource_type.underscore.pluralize}").all.map { |r| {:name => r.to_s, :id => r.id } }.select { |v| v[:name] =~ /#{params[:name]}/ }
+    render :json => { :results=>
+      resource_list.sort_by { |r| r[:name] }.slice(0..100), :subtotal => resource_list.count}
```
adamruzicka, line 38: Range is inclusive, so `0..100` means first 101 elements. Also `take` might be a better fit here

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -23,15 +23,18 @@ def resource_name(nested_resource = nil)
...
   end
 
   def action_permission
     :view_job_templates
   end
+
+  def resources
+    resource_type = params[:resource]
+    resource_list = resource_type.constantize.authorized("view_#{resource_type.underscore.pluralize}").all.map { |r| {:name => r.to_s, :id => r.id } }.select { |v| v[:name] =~ /#{params[:name]}/ }
+    render :json => { :results=>
```
adamruzicka, line 37: Needs a space before the rocket

adamruzicka (commented):
Last couple of nitpicks, then it is good to go

app/controllers/ui_job_wizard_controller.rb
MariaAga: didn't know about `take` thanks!

app/controllers/ui_job_wizard_controller.rb
MariaAga: added

adamruzicka:
Looks good, waiting for the tests to pass

adamruzicka:
Thank you @MariaAga & @laviro!

PR #645, opened 2021-08-19
Refs #33311 - fix recent audit card alignment

amirfefer:
In a medium-size resolution, the recent audit card becomes too small for rendering all tabs without scrolling:

before:
old

after: 
new


adamruzicka:
While this works, the positioning is sometimes a bit unfortunate

image

I don't want to block this PR on it, but it would be nice if we had at least an issue for it.

adamruzicka (approved):

adamruzicka:
Thank you @amirfefer !

PR #646, opened 2021-08-25
Fixes #33341 - Update foreman js to v8

MariaAga:
Will take it out of draft when test are fixes https://github.com/theforeman/foreman_remote_execution/pull/644#issuecomment-899316408

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,11 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
-    "@theforeman/builder": "^4.14.0",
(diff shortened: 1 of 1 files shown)


laviro (approved):

adamruzicka:
What about the test failures?

laviro:
@adamruzicka looks like it's the same PF4 issues that @MariaAga mentioned on other PRs as well,
also since Foreman dictates the version which is actually used when the app launches, we are using it already.
so this upgrade is basically only for the tests environment.

adamruzicka:
Ok, will merge once it is not a draft

PR #649, opened 2021-09-03
Fixes #33409 - Add facility for injecting alternative names

adamruzicka:


app/models/remote_execution_provider.rb
```diff
@@ -131,5 +131,12 @@ def required_proxy_selector_for(template)
         ::DefaultProxyProxySelector.new
       end
     end
+
+    def alternative_names(host)
+      interfaces = effective_interfaces host
+      fqdn = find_fqdn(interfaces) || raise(_('Could not find any suitable interface for execution'))
```
ezr-ondrej, line 137: do we want to fail the execution on this?

adamruzicka: If it gets to this point then we already know the host has at least one execution interface with ip or fqdn, otherwise an exception is raised even earlier. Not sure if a host can have an interface with an ip but without a fqdn

ezr-ondrej: I'm running the IP only setup in part of my dev env, so it definitelly can, if it is production supported, I guess not, but it can happen and this would break it IIUC :thinking:

adamruzicka: Fair enough, changed.

ezr-ondrej (approved):
Thanks @adamruzicka

PR #650, opened 2021-09-06
Fixes #32677 - Remove refresh tooltip after form refresh

adamruzicka:
When a user clicks the refresh button, the browser does a request in background
and replaces the entire form with the response it receives. However, the tooltip
is rendered directly in body and therefore was left around when the form was
replaced. This change simply removes all currently displayed tooltips once the
refresh is performed.


adamruzicka:
@MariaAga could you please take a look?

MariaAga (approved):
Refresh tooltip is now removed :tada:  (Other tooltips in that page still get stuck but thats for a different pr probably)

ezr-ondrej:
> Refresh tooltip is now removed  (Other tooltips in that page still get stuck but thats for a different pr probably)

Or just wait for proper rewrite into React :)

ezr-ondrej:
Thanks @adamruzicka and @MariaAga !

PR #653, opened 2021-10-05
Improve handling of nil attributes in tests

lzap:
I was getting:

https://gist.github.com/lzap/b30edc8aa25fddeef9b1abe4e75435a8

From time to time, not sure what is going on but apparently nil was passed into create method and that attribute cannot be nil.


ares:
Well the question is, whether we shouldn't (always) rather fix the cause, if I get that correctly, we're registering a feature without a template that implements it, which could cause problems elsewhere. Therefore the failure may be actually desired.

lzap:
Ok incorrect patch then. Thanks.

PR #660, opened 2021-10-29
Fixes #33807 - job wizard add starts before field

MariaAga:
Adds some validation logic, but no messages for the user. Will add them at a later pr, redmine link - https://projects.theforeman.org/issues/33809

"Start before" can be seen in "Schedule future execution" in the original job form.
as it wasnt in "Set up recurring execution" it gets disabled if a user selects a type for repeating.

This pr is removing the advanced schedule button as we don't have a working design for it, and it only adds "start before"
Screenshot from 2021-10-29 17-07-50


laviro (approved):
Thanks @MariaAga !

adamruzicka (approved):
LGTM

adamruzicka:
Thank you @MariaAga & @Ron-Lavi !

PR #661, opened 2021-10-30
Fixes #33813 - Add search on job invocation targeting

xprazak2:


ezr-ondrej:
Any details why would we want this? It will be indead weird looking search query :thinking:

xprazak2:
It will look strange and we need it to find job invocations that target a specific hostgroup to show recurring ansible config jobs for a hostgroup - I'll open a PR for that once I have rebased on latest changes in foreman_ansible.

xprazak2:
It seems like this will not be needed, closing.

PR #662, opened 2021-11-01
Fixes #33823 - job wizard submit data and create a job

MariaAga:
- [x] need to add redirect and toast after submit as user gets no indication currently

### DIFF
--- a/webpack/JobWizard/steps/HostsAndInputs/index.js
+++ b/webpack/JobWizard/steps/HostsAndInputs/index.js
@@ -40,25 +40,7 @@ import {
 import { WizardTitle } from '../form/WizardTitle';
 import { SelectAPI } from './SelectAPI';
 import { SelectGQL } from './SelectGQL';
-
-const buildQuery = (selected, search) => {
-  const { hosts, hostCollections, hostGroups } = selected;
-  const hostsSearch = `(name ^ (${hosts.map(({ id }) => id).join(',')}))`;
-  const hostCollectionsSearch = `(host_collection_id ^ (${hostCollections
-    .map(({ id }) => id)
-    .join(',')}))`;
-  const hostGroupsSearch = `(hostgroup_id ^ (${hostGroups
-    .map(({ id }) => id)
-    .join(',')}))`;
-  return [
-    hosts.length ? hostsSearch : false,
-    hostCollections.length ? hostCollectionsSearch : false,
-    hostGroups.length ? hostGroupsSearch : false,
-    search.length ? `(${search})` : false,
-  ]
-    .filter(Boolean)
-    .join(' or ');
-};
+import { buildHostQuery } from './buildHostQuery';
 
 const HostsAndInputs = ({
   templateValues,
@@ -79,7 +61,7 @@ const HostsAndInputs = ({
           key: HOSTS_API,
           url: '/api/hosts',
           params: {
-            search: buildQuery(selected, hostsSearchQuery),
+            search: buildHostQuery(selected, hostsSearchQuery),
             per_page: HOSTS_TO_PREVIEW_AMOUNT,
           },
         })
@@ -128,7 +110,7 @@ const HostsAndInputs = ({
         <HostPreviewModal
           isOpen={hostPreviewOpen}
           setIsOpen={setHostPreviewOpen}
-          searchQuery={buildQuery(selected, hostsSearchQuery)}
+          searchQuery={buildHostQuery(selected, hostsSearchQuery)}
         />
       )}
       <Form>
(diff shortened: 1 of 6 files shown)


MariaAga:
@laviro can you take a look?
It's missing tests but I'm not sure what to test in this pr

webpack/JobWizard/submit.js
```diff
@@ -0,0 +1,98 @@
...
+      password,
+      key_passphrase: keyPassphrase,
+      recurrence:
+        repeatType === repeatTypes.noRepeat
+          ? null
+          : {
+              cron_line: getCronLine(),
+              max_iteration: repeatAmount,
+              end_time: ends.length ? ends.toISOString() : null,
+              purpose: '', // TODO add value after https://projects.theforeman.org/issues/33814
```
MariaAga, line 72: https://github.com/theforeman/foreman_remote_execution/pull/666

webpack/JobWizard/submit.js
```diff
@@ -0,0 +1,98 @@
...
+            },
+      scheduling: {
+        start_at: starts.length ? starts.toISOString() : null,
```
adamruzicka, line 75: Would it be possible to not include the `scheduling` and `recurrence` keys in the request if the job should execute immediately? The api is confused by this and sets the job to execute in the future at time of the request, which kinda does the same thing but is less efficient than just starting it right away. An alternative would be addressing this in the api

webpack/JobWizard/steps/HostsAndInputs/index.js
```diff
@@ -74,12 +56,13 @@ const HostsAndInputs = ({
   const [hostPreviewOpen, setHostPreviewOpen] = useState(false);
   useEffect(() => {
     debounce(() => {
+      console.log(buildHostQuery(selected, hostsSearchQuery));
```
MariaAga, line 59: leftover to delete

adamruzicka (commented):
If I get to the last page, kill the server and click submit, nothing happens. There are some `Uncaught (in promise) TypeError: _ref5.response is undefined` errors in the browser console, but nothign immediately visible to the user.

MariaAga:
I think this kind of error will usually not be visible to the user, but I added a few fallbacks for the toast message

laviro (commented):
Code wise looks good,
when clicking on the last stage "Run" I get the following error:
```
Uncaught TypeError: starts.toISOString is not a function
    at submit (submit.js:76)
    at onSave (JobWizard.js:194)
    at Wizard.onNext (Wizard.js:34)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:306)
    at executeDispatch (react-dom.development.js:389)
    at executeDispatchesInOrder (react-dom.development.js:414)
    at executeDispatchesAndRelease (react-dom.development.js:3278)
```

MariaAga:
Thanks @laviro!
Added a wrap of `new Date()` around the end and start time to fix that

laviro:
should we do the same in the `starts before field`: https://github.com/theforeman/foreman_remote_execution/pull/660

laviro (approved):

MariaAga:
I added the use of `submitStatus` so if a req has started the user won't be able to click submit again. should a user be able to submit the same job wizard page again if an error accrued? @adamruzicka

adamruzicka:
> should a user be able to submit the same job wizard page again if an error accrued?

In general yes, unless the server indicated there is something wrong with the user's input, which hopefully should not happen.

MariaAga:
I see, so changed it to `isSubmitting` so if the request is running or finished successfully the submit is disabled

adamruzicka:
I'm getting this if I only select a host, fill in the command to run and leave everything else untouched.

image

MariaAga:
Added `start_at` `start_before` and `purpose`

adamruzicka:
This is still valid https://github.com/theforeman/foreman_remote_execution/pull/662#pullrequestreview-797429907 . We can now handle when the server responds nicely (even if with error), but if the server is down completely, it still fails silently

image

adamruzicka:
In chrome 95 I get this 
image

In firefox 92.0 I get:
image

MariaAga:
I couldn't reproduce it, but added a conditional chain `response?.__` so it hopefully won't fail

adamruzicka:
Looks better now, lets' get it in.

adamruzicka (approved):
Works nicely

adamruzicka:
Thank you @MariaAga & @Ron-Lavi !

PR #663, opened 2021-11-01
Fixes #33368 - job wizard fake auto select template values

MariaAga:
When there is a select template input the ui is showing the first result as selected but it's value is not selected in the templateValues state


sharvit (approved):
Thanks @MariaAga LGTM

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga & @sharvit !

PR #664, opened 2021-11-01
Fixes #33662 - Job wizard allow unselecting template inputs

MariaAga:
if the template input is not required there should be an empty option
depends on https://github.com/theforeman/foreman_remote_execution/pull/663

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,11 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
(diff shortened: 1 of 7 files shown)


adamruzicka:
#663 just got in

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @MariaAga !

PR #666, opened 2021-11-02
Fixes #33814 - Job wizard add Purpose field

MariaAga:
Screenshot from 2021-11-02 14-14-33

### DIFF
--- a/webpack/JobWizard/steps/Schedule/index.js
+++ b/webpack/JobWizard/steps/Schedule/index.js
@@ -6,6 +6,7 @@ import { RepeatOn } from './RepeatOn';
 import { QueryType } from './QueryType';
 import { StartEndDates } from './StartEndDates';
 import { WIZARD_TITLES, repeatTypes } from '../../JobWizardConstants';
+import { PurposeField } from './PurposeField';
 import { WizardTitle } from '../form/WizardTitle';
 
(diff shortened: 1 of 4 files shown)


webpack/JobWizard/steps/Schedule/index.js
```diff
@@ -106,6 +108,16 @@ const Schedule = ({ scheduleValue, setScheduleValue, setValid }) => {
             }));
           }}
         />
+        <PurposeField
+          isDisabled={repeatType === repeatTypes.noRepeat}
```
laviro, line 112: should we hide the component when `repeatTypes === noRepeat` ?
it's pretty confusing when the field is disabled initially,
another option is to add more explanation on the tooltip

MariaAga: It only appears in `»Set up recurring execution«` in the original, so I don't think it has a use out of recurrence

adamruzicka: > so I don't think it has a use out of recurrence

Correct. Outside of recurring jobs it has no purpose, pun intended :)

adamruzicka:
Also needs a rebase

MariaAga:
thanks! rebased

adamruzicka (approved):
Looks good

adamruzicka:
Thank you @MariaAga !

PR #669, opened 2021-11-03
Fixes #33844 - wizard multiple select closes after select

MariaAga:
It shouldn't, so a user can select many items

### DIFF
--- a/webpack/JobWizard/steps/form/SearchSelect.js
+++ b/webpack/JobWizard/steps/form/SearchSelect.js
@@ -93,7 +93,6 @@ export const SearchSelect = ({
         autoSearch(value || '');
       }}
       placeholderText={placeholderText}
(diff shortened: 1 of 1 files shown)


adamruzicka:
This doesn't seem to work for the host select in step 2

MariaAga:
moving to draft until I can see if there is a fix or not

adamruzicka:
Any progress here?

MariaAga:
https://github.com/patternfly/patternfly-react/issues/6321

adamruzicka:
Given the issue in PF was closed as stale, should we go with this workaround?

MariaAga:
@Ron-Lavi

adamruzicka (approved):
Seems to do the trick, let's get it in

adamruzicka:
Thank you @MariaAga !

PR #670, opened 2021-11-04
Fixes #33848 - add xl2 support for REX card in host details

amirfefer:
to keep  the card layout consistent across multiple screen sizes


ezr-ondrej (approved):
:shrug: Dunno what's best here, but we already try to be consistent, for all other screens :)
Thanks @amirfefer ! :+1:

PR #672, opened 2021-11-24
Fixes #34001 - remove usage of deprecated auto_complete_search

ezr-ondrej:
This method will go away from foreman core, we need to replace it.

The https://github.com/theforeman/foreman/pull/8951 removes this deprecated method.


adamruzicka (approved):
LGTM

adamruzicka:
Thank you @ezr-ondrej !

PR #673, opened 2021-11-29
Fixes #34032 - Make proxy used for REX more easily discoverable

adamruzicka:
Adds a link to used foreman proxy to template invocation detail page
image

Also links the per-host task against the proxy, making the proxy appear in Locks tab in task's details
image

### DIFF
--- a/app/views/template_invocations/show.html.erb
+++ b/app/views/template_invocations/show.html.erb
@@ -28,7 +28,13 @@ end
   </div>
 </div>
 <% if @host %>
-  <h3><%= _('Target: ') %><%= link_to(@host.name, host_path(@host)) %></h3>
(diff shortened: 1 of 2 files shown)


app/views/template_invocations/show.html.erb
```diff
@@ -28,7 +28,13 @@ end
   </div>
 </div>
 <% if @host %>
-  <h3><%= _('Target: ') %><%= link_to(@host.name, host_path(@host)) %></h3>
+  <% proxy_id = @template_invocation_task.input[:proxy_id] %>
```
ezr-ondrej, line 31: Where is the proxy_id comming from? Is that related to the `input[:proxy_id]`, or are we just exposing it on two independent places?

PR #674, opened 2021-12-02
Add js dependencies needed for packaging

adamruzicka:


evgeni (approved):

ezr-ondrej:
Why aren't those deps in `foreman-js`? It seems like quite common dependencies we will need everywhere.

evgeni:
Right now, only scap and rex use these.

But yeah, long term it should be in fjs.

ezr-ondrej (approved):
Failures unrelated, those are probably related to the new minitest reporters.

PR #675, opened 2021-12-03
stop using removed user_template_input?

evgeni:


ezr-ondrej (approved):
Thanks @evgeni ! :heart:

PR #678, opened 2021-12-22
Fixes #33737 - Add bookmarks to templates in job wizard

MariaAga:
Now when a user has a search input field in the template they can use it's bookmarks.
(Before the foreman fix the new bookmark modal will choose only the last text entered as the value so we can only had one bookmark button perpage)
Screenshot from 2021-12-22 19-01-31

### DIFF
--- a/webpack/JobWizard/JobWizard.scss
+++ b/webpack/JobWizard/JobWizard.scss
@@ -103,4 +103,9 @@
       margin-left: 10px;
     }
   }
+  .foreman-search-field {
+    .autocomplete-search-btn {
+      display: none;
+    }
+  }
 }
--- a/webpack/JobWizard/JobWizardConstants.js
+++ b/webpack/JobWizard/JobWizardConstants.js
@@ -47,7 +47,7 @@ export const hostMethods = {
   searchQuery: __('Search query'),
 };
 
-export const hostQuerySearchID = 'searchBar'; // until https://projects.theforeman.org/issues/33737 is used
+export const hostQuerySearchID = 'mainHostQuery';
 export const hostsController = 'hosts';
 
 export const dataName = {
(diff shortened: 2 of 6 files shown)


adamruzicka (changes requested):
This acts in a slightly weird way. I created a job template with type=search and and resource=host. If I try to fill this input using a bookmarks, it shows me no bookmarks. Looking under the hood it does a request to `/api/bookmarks?search=controller%3DHost` which yields no results. If I go to the hosts index page and one the bookmarks there, it does the request to `http://localhost:3000/api/bookmarks?search=controller%3Dhosts` (`Host` vs `hosts`). We might need to do some kind of mapping there

MariaAga:
Thanks! should work now

adamruzicka (changes requested):
When I select one of the bookmarks, it redirects me to the form again and sets `search` param to the search query of the bookmark. That's probably not what we want

MariaAga:
ok now it should work!
There is a bug in foreman, as it's unusual for the page to not refresh, the dropdown doesn't close on select, will fix in foreman https://projects.theforeman.org/issues/34228

adamruzicka:
If I understand it correctly, the search fields try to do autocompletion, but they do not honor the controller of the resource.

Here we have the form querying `/hosts/auto_complete_search` for job invocations:
image

This leads to errors being shown, even though the search query is in fact correct

image

MariaAga:
Forgot to mention, this pr should be tested with https://github.com/theforeman/foreman/pull/8922

MariaAga:
Thanks @adamruzicka! fixed!

adamruzicka (approved):
Works well

adamruzicka:
Thank you @MariaAga !

PR #679, opened 2022-01-13
Fixes #34260 - add rex features dropdown to host page

amirfefer:
https://user-images.githubusercontent.com/11807069/149392178-43bb0abf-42da-4742-98e9-1a70c385bde3.mov

### DIFF
--- a/webpack/react_app/components/FeaturesDropdown/actions.js
+++ b/webpack/react_app/components/FeaturesDropdown/actions.js
@@ -1,0 +1,13 @@
+import { foremanUrl } from 'foremanReact/common/helpers';
+import { sprintf, translate as __ } from 'foremanReact/common/I18n';
+import { post } from 'foremanReact/redux/API';
+
+export const runFeature = (hostId, feature, label) => dispatch => {
+  const url = foremanUrl(
+    `/job_invocations?feature=${feature}&host_ids%5B%5D=${hostId}`
+  );
+
+  const successToast = () => sprintf(__('%s job has been invoked'), label);
+  const errorToast = ({ message }) => message;
+  dispatch(post({ key: feature.toUpperCase(), url, successToast, errorToast }));
+};
--- a/webpack/react_app/components/FeaturesDropdown/index.js
+++ b/webpack/react_app/components/FeaturesDropdown/index.js
@@ -1,0 +1,73 @@
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import { useDispatch } from 'react-redux';
+import {
+  DropdownItem,
+  Dropdown,
+  DropdownToggle,
+  DropdownToggleAction,
+} from '@patternfly/react-core';
+import { push } from 'connected-react-router';
+
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import { STATUS } from 'foremanReact/constants';
+
+import { REX_FEATURES_API, NEW_JOB_PAGE } from './constant';
+import { runFeature } from './actions';
+
+const FeaturesDropdown = ({ hostId }) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const {
+    response: { results: features },
+    status,
+  } = useAPI('get', foremanUrl(REX_FEATURES_API));
+  const dispatch = useDispatch();
+  const dropdownItems = features
+    ?.filter(feature => feature.host_action_button)
+    ?.map(({ name, label, id, description }) => (
+      <DropdownItem
+        onClick={() => dispatch(runFeature(hostId, label, name))}
+        key={id}
+        description={description}
+      >
+        {name}
+      </DropdownItem>
+    ));
+  const scheduleJob = [
+    <DropdownToggleAction
+      onClick={() => dispatch(push(`${NEW_JOB_PAGE}=${hostId}`))}
+      key="schedule-job-action"
+    >
+      {__('Schedule a job')}
+    </DropdownToggleAction>,
+  ];
+
+  return (
+    <Dropdown
+      alignments={{ default: 'right' }}
+      onSelect={() => setIsOpen(false)}
+      toggle={
+        <DropdownToggle
+          splitButtonItems={scheduleJob}
+          toggleVariant="primary"
+          onToggle={() => setIsOpen(prev => !prev)}
+          isDisabled={status === STATUS.PENDING}
+          splitButtonVariant="action"
+        />
+      }
+      isOpen={isOpen}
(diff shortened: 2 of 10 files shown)


webpack/react_app/components/FeaturesDropdown/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+const FeaturesDropdown = ({ hostId }) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const {
+    response: { results: features },
+    status,
+  } = useAPI('get', foremanUrl(REX_FEATURES_API));
+  const dispatch = useDispatch();
+  const dropdownItems = features
+    ?.filter(feature => feature.host_action_button)
```
amirfefer, line 28: @adamruzicka I tried to filter this by the API call, with no success
I added `scoped_search :on => :host_action_button` 
and created an API call to `/api/remote_execution_features?search=host_action_button=true` 

We can filter it on the client, but if the server can return just the relevant features, it just better :)

webpack/react_app/components/FeaturesDropdown/index.js
```diff
@@ -0,0 +1,73 @@
...
+  ];
+
+  return (
+    <Dropdown
+      alignments={{ default: 'right' }}
+      onSelect={() => setIsOpen(false)}
+      toggle={
+        <DropdownToggle
+          splitButtonItems={scheduleJob}
+          toggleVariant="secondary"
```
amirfefer, line 54: looks like it doesn't work... no `secondary` nor `primary`

jeremylenz: I've noticed this too. Maybe we should file a PF issue..

jeremylenz: I made [this Codepen](https://codesandbox.io/s/holy-breeze-70rwf?file=/index.js) intending to file an issue, but `toggleVariant` seems to work fine there...

webpack/react_app/components/FeaturesDropdown/index.js
adamruzicka: If I'm reading things right the controller currently does not support searching, tracking it as[1]. We could implement that, but for now let's go with frontend only filtering. Standard Satellite deployment has around 20 features, so nothing terrible.

[1] - https://projects.theforeman.org/issues/34284

app/models/remote_execution_feature.rb
```diff
@@ -10,6 +10,7 @@ class RemoteExecutionFeature < ApplicationRecord
   friendly_id :label
 
   scope :with_host_action_button, -> { where(:host_action_button => true) }
+  scoped_search :on => :host_action_button
```
ezr-ondrej, line 13: As of @adamruzicka's answer, I'd keep this out of this PR, as it feels out of scope, especially when not used :)

webpack/react_app/components/FeaturesDropdown/index.js
amirfefer: So...the `toggleVariant` was added in `@patternfly/react-core@4.177.0`
currently we use `"@patternfly/react-core": "~4.162.2"`
:\

anyway, it's not a bug and should be resolved as soon as we will update pf

app/models/remote_execution_feature.rb
amirfefer: removed :)

webpack/react_app/components/FeaturesDropdown/index.js
jeremylenz: But the old prop `isPrimary` doesn't seem to do anything either :thinking:

ezr-ondrej:
So shall we bump patternfly and package before we get this in?

jeremylenz:
> So shall we bump patternfly and package before we get this in?
I'd be +1 to that :+1:

ares:
Any progress on this? Who has the ball for patternfly bump?

amirfefer:
upgrading all PF4 related libraries - https://github.com/theforeman/foreman-js/pull/378

amirfefer:
Worlks well with https://github.com/theforeman/foreman/pull/9110
Screen Shot 2022-02-21 at 15 14 07

amirfefer:
odd, rubocop is passed locally 🤔

adamruzicka:
> odd, rubocop is passed locally thinking

That might have been me with this https://github.com/theforeman/foreman_remote_execution/commit/4d4957a901741e2fe6e7646b0b8027fc8b01006d , https://github.com/theforeman/foreman_remote_execution/pull/695 should fix it.

adamruzicka:
Could you rebase plase? It should make the rubocop issue go away

amirfefer:
@adamruzicka - rebased, but still, https://github.com/theforeman/foreman/pull/9110 hasn't merged yet.

ofedoren:
It is merged now :rocket:

ezr-ondrej (approved):
Thanks @amirfefer for doing this and thaks all the reviewers for the great coordinated effort to get this in properly! :+1:

webpack/react_app/components/FeaturesDropdown/actions.js
```diff
@@ -0,0 +1,13 @@
+import { foremanUrl } from 'foremanReact/common/helpers';
+import { sprintf, translate as __ } from 'foremanReact/common/I18n';
+import { post } from 'foremanReact/redux/API';
+
+export const runFeature = (hostId, feature, label) => dispatch => {
+  const url = foremanUrl(
+    `/job_invocations?feature=${feature}&host_ids%5B%5D=${hostId}`
```
ezr-ondrej, line 7: Unfortunatelly this is wrong as we are passing the Host name no an Id here.
That's something I've complained a lot. In core we are passing the host identifier as hostId prop all around the page, but it is usually a name, not an ID, I believe it should be fixed to properly denote what we are actually passing to this prop. @amirfefer could you fix this instance, so it would prefil the search for the job correctly?
@jeremylenz ^ for awarnes about the mess and what it can cause ;)

jeremylenz: @ezr-ondrej, Yeah, I noticed quite a while back that `host_ids` is named poorly and is actually just a scoped search.  That's the way we treat it in Katello:

```js
`host_ids=name ^ (${hostname})`
```

https://github.com/Katello/katello/blob/2f9e60b313c9cd05f44811db2939ed66fc61e0ac/webpack/components/extensions/HostDetails/Tabs/customizedRexUrlHelpers.js#L13-L14

Let me know if you're going to break this so I can make sure it gets fixed ;)

ezr-ondrej: it is not enrirely true, it works now when we made sure, that hostId is a numerical Id :)

PR #680, opened 2022-01-14
Privileged cockpit

adamruzicka:


adamruzicka:
Apparently github cannot recognize the file as ruby

extra/cockpit/foreman-cockpit-session
```diff
@@ -1,302 +1,377 @@
-#!/usr/bin/env ruby
+#!/usr/bin/tfm-ruby
```
ezr-ondrej, line 1: wont this break on deb? But possibly also EL-8 (unsure about that one).

ezr-ondrej (commented):
I am probably not the best person to review this, @adamruzicka do you have someone who could help here? in general, it would be much easier to review if we would have put the cockpit classes in code files, but I'm not sure if it is possible as I totaly do not get how this all works :(

extra/cockpit/foreman-cockpit-session
adamruzicka: Right, we'll have to mangle it in packaging

adamruzicka:
> do you have someone who could help here?

Not really :/

> in general, it would be much easier to review if we would have put the cockpit classes in code files, but I'm not sure if it is possible

Technically it is possible, although we would have to either ship two mostly unrelated things in a single gem or split this into two separate gems.

> but I'm not sure if it is possible as I totaly do not get how this all works

I'm (again) a bit foggy on how this is supposed to work, but the gist is:
1) User clicks the webconsole button, request gets sent to cockpit controller
2) Cockpit controller does some checks, builds a redirect url and redirects the user to it
3) Client's browser follows the redirect, hits apache, apache proxies the request to `foreman-cockpit` service
4) `foreman-cockpit` forks of `foreman-cockpit-session` and interacts with it through its stdin and stdout
5) `foreman-cockpit-session` calls back to foreman to get details about the host
6) `foreman-cockpit-session` does a request to selected smart proxy
7) smart proxy opens a ssh connection to the remote host and spawns some cockpit thingy there
8) the connection from 6 gets upgraded (both sides agree over http that the connection should be used as a general tcp connection and http doesn't need to be used anymore)

At this point, there is a persistent connection established from the client's browser, through several mitms to the cockpit thingy running on the remote host. From this moment onwards, the only thing `foreman-cockpit-session` does is that it watches inputs from both sides (stdin and the upgraded proxy connection) and each time there is data available on either of them, it proxies the data to the other side (stdin -> upgraded proxy connection, upgraded proxy connection -> stdout).

To make privileged cockpit work, we need to inspect all the messages coming from the remote side and if it is an authorization request, we drop the message and craft a response which has the password and send it back. If the incoming message is not an authorization request, we just pass it.

adamruzicka (commented):
It's been 9 months since I opened this PR and apart from one comment in late February it has been mostly dormant.

At this point I have several options:
1) close it as apparently it has not gained enough traction (not shiny enough feature, touches area noone really knows, or maybe the changes are too big for what it gives in return?)
2) Wait until someone, sometime gets to it
3) Just merge it myself

After such a long time, I don't really remember what I did here so I guess I can do a review myself just as if I were reviewing anyone else's PR.

When I put this PR together it worked against both EL7 and EL8 machines. I took this for a spin now, it still works against an EL7 machine (cockpit-195.12), but stopped working on EL8 machines (cockpit-264.1). I filed https://github.com/cockpit-project/cockpit/issues/17793 to check if there's an easy way how we could still salvage this.

Before my today's testing, while I still assumed this would just work, I was leaning towards option 3. Now, I'm not so sure. If there's an easy way out then yes, I'll stick with it, otherwise number 1 sounds like the way to go.

adamruzicka:
Alright, the EL8 issue was resolved. I'll let this sit around for a bit, test it once more next week or so and unless anyone steps in and objects, I'll merge it.

adamruzicka (commented):
It's been another month, no concerns were voiced, let's get this in.

PR #682, opened 2022-01-18
Fixes #34201 - Use new Foreman <Pagination>

stejskalleos:


ezr-ondrej:
This is breaking change and needs `foreman >= 3.2`, can you please bump the version in `lib/foreman_remote_execution/engine.rb#50`?

amirfefer (approved):
LGTM 👍

ezr-ondrej (changes requested):
Unfortunatelly it won't be this easy :) It shows pagination, but it does not work.
The new pagination takes a bit different `props` then the old one did.
`viewType` is now `variant` and takes different arguments and instead of pagination, it directly takes `perPage` and `page` props see [PF4 guide for props](https://www.patternfly.org/v4/components/pagination#props) or directly at the [Ron's implementation](https://github.com/theforeman/foreman/blob/337ac271d8a87039bfc709bed0749fdbd9651e7e/webpack/assets/javascripts/react_app/components/Pagination/index.js#L19)
Additionally and this is the main change `perPage` is now passed as `per_page` to the change handler, so we will need to process the change little bit diferently now.

I think that to make it work it is quite straight forward, tho I think that we do not need to keep the pagination options in the component state, because the new pagination remembers for us.
Tho, that's bit more difficult to get right with the API requests.

stejskalleos:
@ezr-ondrej Updated, changing `per_page` now works

ezr-ondrej:
I belive you need to refresh the snapshots again :)

webpack/react_app/components/TargetingHosts/index.js
```diff
@@ -30,14 +30,14 @@ const WrappedTargetingHosts = () => {
   const [searchQuery, setSearchQuery] = useState('');
   const [pagination, setPagination] = useState({
     page: 1,
-    perPage,
+    per_page: perPage,
```
Ron-Lavi, line 33: the `perPage` and `perPageOptions` that was provided from core settings are already consumed in the core implementation, so you don't need to pass them twice.

webpack/react_app/components/TargetingHosts/TargetingHostsPage.js
```diff
@@ -39,9 +39,10 @@ const TargetingHostsPage = ({
     <br />
     <TargetingHosts apiStatus={apiStatus} items={items} />
     <Pagination
-      viewType="table"
       itemCount={totalHosts}
-      pagination={pagination}
+      page={pagination.page}
+      perPage={pagination.per_page}
+      perPageOptions={pagination.perPageOptions}
```
Ron-Lavi, line 45: If I am not mistaken, you don't need to pass those 3 props:
- page: is by default 1 in your implementation and is also default 1 in core, and will change based on the URL.
- perPage: I saw that you used `useForemanSettings();` which is also consumed in core Pagination
- perPageOptions: same as the above.

webpack/react_app/components/TargetingHosts/TargetingHostsPage.js
```diff
@@ -39,9 +39,10 @@ const TargetingHostsPage = ({
...
       onChange={args => handlePagination(args)}
       dropdownButtonId="targeting-hosts-pagination-dropdown"
```
Ron-Lavi, line 47: I am not sure PF4 still support this one,
and I guess the className would be enough?

Ron-Lavi (commented):
Thanks @stejskalleos for doing this!

webpack/react_app/components/TargetingHosts/TargetingHostsPage.js
ezr-ondrej: I believe we don't so I'd agree with you here. But to keep the PR small can we keep this as is and do some clean-up in a followup PR? I believe there is a lot of code that is not needed now, but to shift to parameters will be a bit of an effort.

webpack/react_app/components/TargetingHosts/index.js
ezr-ondrej: You're right, but to keep changes to a minimum I believe in "it works" approach for this PR :)

webpack/react_app/components/TargetingHosts/TargetingHostsPage.js
ezr-ondrej: Yeah, we should not use unsupported props and I believe PF4 does not support this, so I'm in favor of dropping this :+1:

stejskalleos:
@Ron-Lavi @ezr-ondrej updated, I removed unused props as Ron suggested plus fixed failing snapshots.

Ron-Lavi (approved):
Code iwise LGTM 👍🏼 
Thanks @stejskalleos

ezr-ondrej (approved):
Thanks @stejskalleos ! :+1:

PR #683, opened 2022-01-20
Fixes #34297 - replace dropped Rails method

ezr-ondrej:
We were using private Rails method `ActionMailer::Base#add_template_helper` that was never ment to be used.
This method is dropped by Rails 6.1 in https://github.com/rails/rails/commit/cb3b37b37975ceb1d38bec9f02305ff5c14ba8e9 and we need to stop using it.
This simply replaces it by supported `helper` is enough.


adamruzicka (approved):
Don't really have anything to add here

adamruzicka:
Thank you @ezr-ondrej !

PR #684, opened 2022-01-24
Fixes #33388 - disable template related fields when loading

MariaAga:
when loading template list template selection is disabled and the selected template is shown as empty.
after selecting a template, until the template is loading all the steps are disabled


adamruzicka (approved):
Seems to work well

adamruzicka:
Thank you @MariaAga !

PR #685, opened 2022-01-25
Fixes #34309 - job wizard Allow repeat amount to be empty

MariaAga:


adamruzicka:
Thank you @MariaAga !

PR #687, opened 2022-01-27
Fixes #34324 - Exclude infra permission from default roles

adamruzicka:
Requires:
- https://github.com/theforeman/foreman/pull/9064
- https://github.com/theforeman/foreman/pull/9073

### DIFF
--- a/db/seeds.d/20-permissions.rb
+++ b/db/seeds.d/20-permissions.rb
@@ -7,15 +7,3 @@ if view_permission && !default_role.permissions.include?(view_permission)
     filter.filterings.build { |f| f.permission = view_permission }
   end
 end
-
-site_manager_role = Role.find_by(name: 'Site manager')
-Role.without_auditing do
-  Role.skip_permission_check do
-    site_manager_role.ignore_locking do |role|
-      names = ForemanRemoteExecution::Engine::USER_PERMISSIONS.map(&:to_s) + %w(execute_jobs_on_infrastructure_hosts)
-      existing = role.permissions.where(:name => names).pluck(:name)
-      to_add = names - existing
-      role.add_permissions! to_add if to_add.any?
-    end
-  end
-end
--- a/db/seeds.d/20-permissions.rb
+++ b/db/seeds.d/20-permissions.rb
@@ -7,3 +7,15 @@ if view_permission && !default_role.permissions.include?(view_permission)
     filter.filterings.build { |f| f.permission = view_permission }
   end
 end
+
+site_manager_role = Role.find_by(name: 'Site manager')
+Role.without_auditing do
+  Role.skip_permission_check do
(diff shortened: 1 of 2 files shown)


adamruzicka:
```
# for role in "Manager" "Site manager" "Organization admin"; do echo "$role"; h role filters --name "$role" | grep JobInvocation; echo; done

Manager
169 | JobInvocation                | none   | yes        | no        | Manager | create_job_invocations, view_job_invocations, execute_jobs_on_infrastructure_...

Site manager
145 | JobInvocation       | none   | yes        | no        | Site manager | create_job_invocations, view_job_invocations, execute_jobs_on_infrastructure_...

Organization admin
173 | JobInvocation                | none   | yes        | no        | Organization admin | create_job_invocations, view_job_invocations, cancel_job_invocations

```

ezr-ondrej:
I've rekicked the job, tho if you want all of this CPed to foreman 3.1, I'll be a sad :panda_face: because we will have to implement the core changes completely anew, as the implementation changed quite significantly in 3.2

db/seeds.d/20-permissions.rb
```diff
@@ -7,3 +7,15 @@
     filter.filterings.build { |f| f.permission = view_permission }
   end
 end
+
+site_manager_role = Role.find_by(name: 'Site manager')
```
ezr-ondrej, line 11: Can't we just add the 'Site manager' to `AUTO_EXTENDED_ROLES` and edit it in the engine file?
 https://github.com/theforeman/foreman/blob/ef2ff06831ae7efcdab247b13c9ff642c6d56688/app/registries/foreman/plugin/rbac_support.rb#L5

adamruzicka: We definitely could, do you know if there's a reason why it currently is excluded from AUTO_EXTENDED_ROLES?

adamruzicka:
> I've rekicked the job, tho if you want all of this CPed to foreman 3.1, I'll be a sad panda_face because we will have to implement the core changes completely anew

Sadly yes. However looking at 3.1, it might be quite a simple change, unless I'm missing something.

db/seeds.d/20-permissions.rb
adamruzicka: Nevermind, did that

ezr-ondrej: Because it was added later and noone actually thought of adding it :shrug:

ezr-ondrej:
> Sadly yes. However looking at 3.1, it might be quite a simple change, unless I'm missing something.

Yeah, it won't be that bad, just that it will be all harder to CP ...

adamruzicka:
Could we get this rolling again?

ezr-ondrej:
Oh, I've missed this did not get in yet. Sorry :facepalm:

PR #690, opened 2022-02-02
Refs #34362 - fix host path to use new host

amirfefer:


ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

PR #691, opened 2022-02-03
Fixes #32676 - Fix long job name on dashboard

pkoprda:
When job name was longer, the tooltip was displayed twice.

tooltip-dual


adamruzicka:
Seems to work fine. @MariaAga could you take a look if this is the way to go?

ezr-ondrej:
I guess we could have a class and extract the styles into a stylesheet, but no hard feelings against inline styles if kept small :)

MariaAga:
^agree 
It's a good solutions, but there are still 2 tooltips, I think one comes from here:
https://github.com/theforeman/foreman_remote_execution/blob/d95b600c75a15d48a45e7d01d36416bba4a9b3a2/app/helpers/remote_execution_helper.rb#L152

pkoprda:
Thanks @MariaAga for advice, I think it's better to keep `ellipsis` class

adamruzicka:
The helper did more than just truncating the text and adding a tooltip. It also capitalized the text and provided a default value, both of which is not missing. Although to be fair the default value doesn't seem to detect empty strings as blank values

image

adamruzicka:
Re-kicked the tests are they turned :green_apple:

adamruzicka:
Thank you @pkoprda !

PR #693, opened 2022-02-15
Fixes #34452 - Allow hosts to have no execution interface

adamruzicka:
Previously we required each host to have exactly one execution interface. There
was a mechanism in place which would ensure host had at least one execution
interface by marking the primary interface as execution if the host had none.
However this wasn't 100% reliable.

This loosens the requirement from exactly one to 0 or 1 and removes the
mechanism enforcing host has at least 1 execution interface.

This is arguably more correct alternative to https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/43

### DIFF
--- a/app/models/concerns/foreman_remote_execution/nic_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/nic_extensions.rb
@@ -3,17 +3,10 @@ module ForemanRemoteExecution
     extend ActiveSupport::Concern
 
     included do
-      before_validation :set_execution_flag
       validate :exclusive_execution_interface
     end
 
     private
-
-    def set_execution_flag
-      return unless primary? && host.present?
-
-      self.execution = true if host.interfaces.detect(&:execution).nil?
-    end
 
     def exclusive_execution_interface
       if host && self.execution?
--- a/app/models/concerns/foreman_remote_execution/nic_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/nic_extensions.rb
@@ -4,9 +4,18 @@ module ForemanRemoteExecution
 
     included do
       validate :exclusive_execution_interface
+      before_validation :move_execution_flag
     end
 
     private
+
+    def move_execution_flag
+      return unless host && self.execution?
+
+      host.interfaces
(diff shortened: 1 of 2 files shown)


ofedoren (approved):
Thanks, @adamruzicka! I agree that it's better than https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/43. Maybe could actually re-use that ticket? :)

Although it's not required, but we could also update https://github.com/theforeman/foreman_remote_execution/blob/f831adf33d49ad28a40e6304fe2a079e31e91b07/app/models/concerns/foreman_remote_execution/nic_extensions.rb#L11 method to ensure that instead of raising an error, we simply flip the `execution` flag for the old one interface to `false` and allow setting `true` to the current one. In this way we still allow 0 to 1, but without errors and without a need to call the API twice to set a different interface for execution. What do you think? :)

adamruzicka:
> flip the execution flag for the old one interface to false and allow setting true to the current one.

What if someone tried to set the execution flag on multiple interfaces at once? My guess would be that then the validations would be processed in undefined order, essentially moving the flag between interfaces multiple times. Are we ok with that?

ofedoren:
> > flip the execution flag for the old one interface to false and allow setting true to the current one.
> 
> What if someone tried to set the execution flag on multiple interfaces at once? My guess would be that then the validations would be processed in undefined order, essentially moving the flag between interfaces multiple times. Are we ok with that?

That's the actually a thing that bothered me and I didn't find a great answer for that :/ The only thing that comes ti my mind: as with options, the later one wins, just make sure that only one.

I'd be okay with that if we won't have any other side effects.

ofedoren (approved):
Thanks, @adamruzicka! Checked with hammer and seems to be working flawlessly :)

The only thing that bothers me, could we use https://projects.theforeman.org/issues/34404 instead since this is the actual fix for that?

ezr-ondrej (approved):
Thanks @adamruzicka and @ofedoren for the review ! :+1: 
I've linked the two issues, so the proper fix for the first can be found and I believe the original issue can be now closed.

PR #694, opened 2022-02-17
Fixes #34471 - Add bookmarks to jobs targeting host

MariaAga:
Uses the same hack to fill the search input text as the job wizard using `getResults`

### DIFF
--- a/webpack/react_app/components/TargetingHosts/TargetingHostsConsts.js
+++ b/webpack/react_app/components/TargetingHosts/TargetingHostsConsts.js
@@ -1,1 +1,2 @@
 export const TARGETING_HOSTS = 'TARGETING_HOSTS';
+export const TARGETING_HOSTS_AUTOCOMPLETE = 'targeting_hosts_search';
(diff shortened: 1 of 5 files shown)


adamruzicka:
Rubocop failures should be unrelated and already fixed in master. Could you rebase so it goes away?

MariaAga:
Working on a fix for this in core instead: https://projects.theforeman.org/issues/34546

MariaAga:
Added also a small fix to pagination so when users move between bookmarks (and searches) the page will reset to 1

adamruzicka:
Unless I'm missing something then when I select a bookmark, it indeed filters the table, but there is no indication anywhere why.  I'm not saying it is wrong, just confusing. The search bar remains as it was and when I open the bookmarks again, it doesn't show that any of the bookmarks is selected.

MariaAga:
If you dont see text in the searchbar then maybe foreman core is out of date, it should have this pr in: https://github.com/theforeman/foreman/pull/9156
or is it a different issue?

> I open the bookmarks again, it doesn't show that any of the bookmarks is selected.

isn't this the normal behavior in all the bookmarks in foreman?

adamruzicka:
> maybe foreman core is out of date, it should have this pr in: https://github.com/theforeman/foreman/pull/9156

Yeah, I didn't have that... Please disregard my previous comment

MariaAga:
@Ron-Lavi can you take a look?

adamruzicka (approved):
Looks good

adamruzicka:
Thank you @MariaAga !

PR #696, opened 2022-02-22
Prevent ssh action for zypper to return 0 on failure

nadjaheitmann:
At the moment, when you want to install a non-existing package on a client with zypper, the installation of the package fails. However, in the return code evaluation of the installation, the return code is overwritten with `0` if the installation failed. This causes the whole job to be shown as successful even if it was not.

This PR removes the line of code where the return code is set to `0`, which causes the return code to be evaluated correctly in line 115/116.

### DIFF
--- a/app/views/templates/ssh/package_action.erb
+++ b/app/views/templates/ssh/package_action.erb
@@ -69,8 +69,11 @@ handle_zypp_res_codes () {
 
   if [ "${ZYPP_RES_CODES[$RETVAL]}" != "" ]; then
     echo ${ZYPP_RES_CODES[$RETVAL]}
+  fi
+  if [[ $RETVAL -ge 100 && $RETVAL -le 103 ]]; then
     RETVAL=0
   fi
+
   return $RETVAL
 }
 <% end -%>


sbernhard (approved):

adamruzicka:
I'm not sure this is exactly what we want. I mean in general yes, but maybe it should be more fine grained. https://github.com/theforeman/community-templates/commit/6f5863f754f0bb1d3244eed2d1bf353b0b41ab8d which introduced this says

> Some zypper results codes are just "information" result codes which don't abort but are shown if the action was still successfull

If I'm reading it right then this would essentially undo it, just an additional line with explanation would appear in the the output. For example if you update packages and zypper determines a reboot is needed, then with this the job would be marked as failed. Or am I missing something?

nadjaheitmann:
I see your point - thanks for the quick feedback. Then I would suggest to remove code 104 and 105 from that list. I think 105 is definitely not an info and in case of 104, the job should fail. 106 and 107 sound like errors as well.

sbernhard:
Maybe we can do it something like this:
Based on https://github.com/openSUSE/zypper/blob/f2bb1f9e082be67d9f88af4855ef7e50007598e1/src/main.h#L20 , errors have return code < 100. info are >= 100 < 200.

nadjaheitmann:
> Maybe we can do it something like this: Based on https://github.com/openSUSE/zypper/blob/f2bb1f9e082be67d9f88af4855ef7e50007598e1/src/main.h#L20 , errors have return code < 100. info are >= 100 < 200.

This does not reflect that a package that is not found should result in an error (code 104).

nadjaheitmann:
@adamruzicka @sbernhard I have modified the PR to have only return codes 100-103 to be ignored, the rest should cause errors. Are you okay with that?

sbernhard:
If you have a look at the errors/info list https://github.com/openSUSE/zypper/blob/6642d8da83bcb473c24a5016b3e950a5239780db/src/main.h#L21 , this PR solves exactly what is currently broken. Thanks!

adamruzicka:
Looks sensible to me

nadjaheitmann:
So is there anybody else who needs to review this PR or is it ready to go, @adamruzicka ?

adamruzicka:
Well, it looks good to me but I'm not really familiar with zypper or the issue you're trying to solve. I was hoping @sbernhard would give this his blessing and then we'd get this in.

nadjaheitmann:
> If you have a look at the errors/info list https://github.com/openSUSE/zypper/blob/6642d8da83bcb473c24a5016b3e950a5239780db/src/main.h#L21 , this PR solves exactly what is currently broken. Thanks!

I think, that he already did :)

adamruzicka:
Right, looks I remembered things happening in different order. Alrighty then, let's get this in

adamruzicka (approved):

adamruzicka:
Thank you @nadjaheitmann & @sbernhard

nadjaheitmann:
Thank you @adamruzicka ! Will the fix be cherry-picked to foreman_remote_execution 5?

PR #699, opened 2022-03-08
Fixes #34574 - Reset known host keys on re-registration

adamruzicka:


ofedoren:
Sorry, I'm new here, but looking at the code just doesn't tell much about how it exactly solves the issue?

adamruzicka:
Known host keys are now handled by foreman tracking if a job was executed on a host using a proxy by linking them together in `host_proxy_invocations` table. If there's a record tying a host and a proxy together, then it is not a first execution and the proxy should trust the hosts' known key. On the other hand, if there's no such record, we consider it being a first run and instruct the proxy to wipe the known key before attempting to run the job.

Previously the contents of `host_proxy_invocations` got cleared only on host (or proxy) deletion, it did not take re-registration into account.

ofedoren (approved):
Thanks for the explanation, makes sense now.

ezr-ondrej (approved):
Thanks @adamruzicka and thanks @ofedoren for the review! :+1:

PR #700, opened 2022-03-13
Fixes #34600 - add web/cockpit console to host's kebab

amirfefer:
This adds the missing `web console` button from the legacy host page.

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -46,6 +46,10 @@ module ForemanRemoteExecution
       end
     end
 
+    def cockpit_url
+      SSHExecutionProvider.cockpit_url_for_host(self.name)
+    end
+
     def execution_status(options = {})
       @execution_status ||= get_status(HostStatus::ExecutionStatus).to_status(options)
     end
--- a/webpack/react_app/components/HostKebab/KebabItems.js
+++ b/webpack/react_app/components/HostKebab/KebabItems.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { DropdownItem } from '@patternfly/react-core';
+import { CodeIcon } from '@patternfly/react-icons';
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { HOST_DETAILS_KEY } from 'foremanReact/components/HostDetails/consts';
+
+const HostKebabItems = () => {
+  const { cockpit_url: consoleUrl } = useSelector(state =>
+    selectAPIResponse(state, HOST_DETAILS_KEY)
+  );
+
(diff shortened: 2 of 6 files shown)


adamruzicka:
Needs a rebase

amirfefer:
@adamruzicka, thanks, rebased.

amirfefer:
@adamruzicka, I think the spinner is a bit misleading here on every click, so I removed it - when the kebab menu opens, it is mounted and rendered, when the menu is hidden the menu is unmounted, however, if the API result keeps the same, it won't trigger any re-render.

ezr-ondrej:
> if the API result keeps the same, it won't trigger any re-render.

Tho API request still happens, right? And that is significantly slower then re-render, so I'd not mind having re-render, but I do mind about too many API requests (if we can help it). Isn't there a way how to store the cockpit url into a host API response that is being already queried and stored?

amirfefer:
> And that is significantly slower then re-render

The API still happens asynchronously to the browser rendering, so it won't cause any UI slowness, though it makes sense to move the `cockpit_url` to the host API 👍

amirfefer:
I've added the `cockpit_url` attribute to the host API, instead of fetching it from a separate API call.
There is no extra API call when a user clicks on the kebab menu.

webpack/react_app/components/HostKebab/KebabItems.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { DropdownItem } from '@patternfly/react-core';
+import { CodeIcon } from '@patternfly/react-icons';
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { HOST_DETAILS_KEY } from 'foremanReact/components/HostDetails/consts';
+
+const HostKebabItems = () => {
+  const { cockpit_url: consoleUrl } = useSelector(state =>
```
amirfefer, line 10: There is no trivial way to propagate the `cockpit_url` from core,  for an easy cherry-pick the attribute is taken directly from redux's store.

ezr-ondrej: In this case it's IMHO fine as it is not one of main attributes :)

ezr-ondrej (approved):

app/models/concerns/foreman_remote_execution/host_extensions.rb
```diff
@@ -46,6 +46,10 @@ def search_by_job_invocation(key, operator, value)
       end
     end
 
+    def cockpit_url
+      SSHExecutionProvider.cockpit_url_for_host(self.name)
```
ezr-ondrej, line 50: This was recently renamed, but we did not update all usages anyway and in order for this to be cherry-pickable, I'm happy we are using the old name here :+1:

ezr-ondrej (approved):
Looks good, works well what's not to like? :)

PR #701, opened 2022-03-14
Fixes #34606 - Job wizard build incorrect hostgroup search

adamruzicka:


webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
```diff
@@ -4,7 +4,7 @@ export const buildHostQuery = (selected, search) => {
   const hostCollectionsSearch = `(host_collection_id ^ (${hostCollections
```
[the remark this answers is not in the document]
adamruzicka: If the [test data](https://github.com/theforeman/foreman_remote_execution/blob/master/webpack/JobWizard/__tests__/fixtures.js#L140) is to be trusted then for host collections searching by id is actually correct. My only machine with katello is currently out of order, so it will take me a while before I can check it

webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
```diff
@@ -4,7 +4,7 @@ export const buildHostQuery = (selected, search) => {
...
     .map(({ id }) => id)
     .join(',')}))`;
-  const hostGroupsSearch = `(hostgroup_id ^ (${hostGroups
+  const hostGroupsSearch = `(hostgroup ^ (${hostGroups
     .map(({ id }) => id)
```
ofedoren, line 8: Actually, the more I look in this file the more I want to change few things :/
 - If `hostGroups`/`hosts`/`hostCollections` contains both `id` and `name`, why we use `id` here although we're searching by name? Moreover, why `id` contains the same value as `name`?
 - If we're going to use names, all the names should be quoted in case they have spaces, e.g. `Katello CentOS 7`, so it should be 
```suggestion
  const hostGroupsSearch = `(hostgroup ^ (${hostGroups
    .map(({ id }) => `"${id}"`)
```

And I guess initially it supposed to use ids instead of names, so here `host*_id` are being used. The better fix for this helper will be to ensure `selected` contains actual `id` value and then use this value to avoid dealing with the strings/spaces/quotes.

UPD: I noticed that for host collections it works as expected, it uses actual ids.

ofedoren (changes requested):
Well, I've played a bit with this and it seems still broken, but only for host groups which have spaces in its names.

webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
adamruzicka: I'm afraid hostgroup names can contain quotes so unless we escaped the name, just wrapping it in quotes still wouldn't be perfect. I'd rather go the id route

webpack/globalIdHelper.js
```diff
@@ -0,0 +1,13 @@
...
+
+export const decodeModelId = model => decodeId(model.id);
+
+export const decodeId = id => {
+  const split = atob(id).split(idSeparator);
+  return parseInt(split[split.length - 1], 10);
+};
+
+export const encodeId = (typename, id) =>
+  btoa([defaultVersion, versionSeparator, typename, idSeparator, id].join(''));
```
adamruzicka, line 13: This is imported from foreman_ansible. Would be nice if some shared package provided this

webpack/JobWizard/steps/HostsAndInputs/SelectGQL.js
```diff
@@ -35,7 +36,7 @@ export const useNameSearchGQL = apiKey => {
       subtotal: data?.[dataName[apiKey]]?.totalCount,
       results:
         data?.[dataName[apiKey]]?.nodes.map(node => ({
-          id: node.name,
+          id: decodeId(node.id),
```
adamruzicka, line 39: @MariaAga does `host.id` being an actual id instead of a name have potential to break something somewhere?

webpack/globalIdHelper.js
MariaAga: This can be added to `foreman/webpack/assets/javascripts/react_app/common/helpers.js` if it's relevant

adamruzicka: https://projects.theforeman.org/issues/34614

webpack/JobWizard/steps/HostsAndInputs/SelectGQL.js
MariaAga: Not from what I see

adamruzicka:
How do js deps work? I hoped having the helpers in foreman@develop would be enough

MariaAga:
It doesn't work in tests ):
Needs a mock in: https://github.com/theforeman/foreman_remote_execution/tree/master/webpack/__mocks__/foremanReact/common
and by mock I mean copy paste the file or a function that always returns the same id like:
` export const decodeId = () => 'test_id' `

adamruzicka:
:green_apple:

adamruzicka:
@MariaAga could I get your blessing on this?

MariaAga (approved):
Tests are happy and the code lgtm

PR #702, opened 2022-03-14
Fixes #34607 - Fix Katello detection

adamruzicka:


app/controllers/ui_job_wizard_controller.rb
```diff
@@ -40,7 +40,7 @@ def resource_name(nested_resource = nil)
   end
 
   def with_katello
-    !!defined?(::Katello)
+    !!defined?(::Katello::Engine)
```
ofedoren, line 43: Just a question, why would we check the constants if we could check it via plugin's API?

adamruzicka: None in particular, changed

ezr-ondrej (approved):
Thanks @adamruzicka and @ofedoren for the review ! :+1:

PR #703, opened 2022-03-16
Fixes #34502 - Add SSH User field in job invocation

pkoprda:


app/models/ssh_execution_provider.rb
```diff
@@ -55,8 +59,12 @@ def cockpit_url_for_host(host)
 
     private
 
-    def ssh_user(host)
-      host.host_param('remote_execution_ssh_user')
+    def ssh_user(template_invocation, host)
```
adamruzicka, line 62: This would break cockpit integration, `ssh_port` is being called from https://github.com/theforeman/foreman_remote_execution/pull/703/files#diff-bac7fac9f5249d58891a94052593b2938f354d1f71469b149d3bf1b774fa333bR49 where no template invocation is available. Maybe reorder the arguments and make template_invocation optional?

adamruzicka (changes requested):
On rerun, the ssh user needs to be retrieved from the previous job and shown. Also it would be nice to somehow indicate in job invocation details that a custom ssh user was provided

adamruzicka (changes requested):
Apart from the things mentioned as separate comments, a bit of food for thought:

This seems to have been implemented closely to how effective user is handled. However, effective user can be in fact set per-job template, where storing the use in template invocations was warranted. For ssh user, where it cannot be set per job template, maybe this is an overkill and storing it in the top level job invocation would be enough? Alternatively we could also allow setting ssh user per job template, although I'd prefer not doing that if we can avoid it

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -60,7 +60,7 @@ def plan(job_invocation, host, template_invocation, proxy_selector = ::RemoteExe
                                :use_concurrency_control => options[:use_concurrency_control],
                                :first_execution => first_execution,
                                :alternative_names => provider.alternative_names(host) }
-        action_options = provider.proxy_command_options(template_invocation, host)
+        action_options = provider.proxy_command_options(template_invocation, host, job_invocation)
```
adamruzicka, line 63: This is slightly problematic. Other plugins implement their own providers (such as [ansible](https://github.com/theforeman/foreman_ansible/blob/3660359a5afb3c9d71b1e52b4c05366881835427/app/models/foreman_ansible/ansible_provider.rb#L25)), but this change changes the interface we expect the providers to implement. While we can technically do that, I'd like to avoid that if possible.

Here we don't really need to pass the job invocation into the function, we can just pull the job invocation out of the template invocation in that function.

pkoprda: Yes, that's true. Thanks for suggestion!

adamruzicka (commented):
Works well, last two nitpicks. One is in inline comment, the other here.

Could we expose the ssh user in api when looking at details of a job? Add it to https://github.com/theforeman/foreman_remote_execution/blob/master/app/views/api/v2/job_invocations/base.json.rabl#L3 should be enough

app/models/remote_execution_provider.rb
```diff
@@ -19,7 +19,7 @@ def provider_names
       providers.keys.map(&:to_s)
     end
 
-    def proxy_command_options(template_invocation, host)
+    def proxy_command_options(template_invocation, host, job_invocation = nil)
```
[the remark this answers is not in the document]
pkoprda: Thanks, I did not notice that

adamruzicka (approved):
Works like a charm

adamruzicka:
Thank you @pkoprda !

PR #704, opened 2022-03-21
Fixes #34653 - Rename SSH provider to script

adamruzicka:
Currently the new provider is just an alias to the SSH provider to make
the transition easier and not require making this change across all
dependenat plugins at once.

### DIFF
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -7,6 +7,15 @@ class RemoteExecutionProvider
       providers[type.to_s] || providers[:script]
     end
 
+    def registered_name
+      klass = self
+      providers.key(klass)
+    end
+
+    def proxy_feature
+      registered_name
+    end
+
     def providers
       @providers ||= { }.with_indifferent_access
     end
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -4,7 +4,7 @@ class RemoteExecutionProvider
 
   class << self
     def provider_for(type)
-      providers[type.to_s] || providers[:SSH]
+      providers[type.to_s] || providers[:script]
     end
 
     def providers
--- a/app/models/ssh_execution_provider.rb
+++ b/app/models/ssh_execution_provider.rb
@@ -1,4 +1,4 @@
-class SSHExecutionProvider < RemoteExecutionProvider
+class ScriptExecutionProvider < RemoteExecutionProvider
   class << self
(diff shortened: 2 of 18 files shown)


app/models/ssh_execution_provider.rb
```diff
@@ -64,3 +64,6 @@ def ssh_port(host)
     end
   end
 end
+
+class ScriptExecutionProvider < SSHExecutionProvider
```
ezr-ondrej, line 68: don't we want to alias that instead and ideally rename to `ScriptExecutionProvider` and make the SSHExecutionProvider an alias?
```suggestion
SSHExecutionProvider = ScriptExecutionProvider
```
if there are concerns for doing it in this PR, then I'd do it the other way arround, but not subclassing, unless we really want to subclass for some reason.

ezr-ondrej (changes requested):
This fails for me with `Failed to initialize: RuntimeError - Could not use any proxy for the script job. Consider configuring remote_execution_global_proxy, remote_execution_fallback_proxy in settings` I believe we need to register `script` as an ability for proxy first?

adamruzicka:
But then even the proxy would have to report the feature as script, which I'm not sure we want. I definitely don't like that idea. How about letting a provider define a list of proxy feature aliases that would be used when picking a proxy?

ezr-ondrej:
can't we have proxy report it as both `ssh` and `script` aliasing the same way we do here?

app/models/remote_execution_provider.rb
```diff
@@ -4,7 +4,15 @@ class RemoteExecutionProvider
 
   class << self
     def provider_for(type)
-      providers[type.to_s] || providers[:SSH]
+      providers[type.to_s] || providers[:script]
+    end
+
+    def registered_name
+      providers.find { |_name, klass| klass == self }.first
```
ezr-ondrej, line 11: This is bit hard to ready, could we use https://ruby-doc.org/core-2.7.0/Hash.html#method-i-key ?
```suggestion
      klass = self
      providers.key(klass)
```

ezr-ondrej (approved):
Works well now :+1:
I'd imagine we want to rename the proxy feature as well, but now we have the option to do it decoupled from this PR :+1:

app/models/remote_execution_provider.rb
adamruzicka: TIL

ezr-ondrej:
I'd like to keep both commits separate as it make sense IMHO, would it be an issue to have both with fixes keyword?

PR #705, opened 2022-03-21
Fixes #34657 - Add aria label to job wizard template select

MariaAga:
Adding the same fields as in `SelectField`
https://github.com/theforeman/foreman_remote_execution/blob/7c73948716a099abefc2248fed3c3a10ed073bdd/webpack/JobWizard/steps/form/SelectField.js#L46-L47


amirfefer (approved):
Thanks @MariaAga !

adamruzicka:
Thank you @MariaAga & @amirfefer !

PR #707, opened 2022-03-23
Refs #34260 - fix the 'schedule a job' button in new host

amirfefer:
The `scedule a job` button redirects to the new job page, but it isn't linked to the current host.

### DIFF
--- a/webpack/react_app/components/FeaturesDropdown/index.js
+++ b/webpack/react_app/components/FeaturesDropdown/index.js
@@ -23,6 +23,7 @@ const FeaturesDropdown = ({ hostId }) => {
     response: { results: features },
     status,
   } = useAPI('get', foremanUrl(REX_FEATURES_API));
+
   const dispatch = useDispatch();
   const dropdownItems = features
     ?.filter(feature => feature.host_action_button)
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
can't we pass the proper id from the core host page? I do not love selecting the id from api response :thinking: But if you think it is the best way :)

amirfefer:
@ezr-ondrej, we can propage the host id from the core's slot, but it force us to create multiple PRs and this just makes the CP process  harder. what concerns do you have for selecting the id from redux's store?

ezr-ondrej:
> what concerns do you have for selecting the id from redux's store?

The place (key) in redux store for this can easily change. I don't like storing the whole API responses there in the first place and using them somewhere else makes it harder to refactor.
Something so vital for the function of the component as hostId should not be diged up from depths of Redux IMHO.

> but it forces us to create multiple PRs and this just makes the CP process harder

There is no plan for cherry-picking this as it would already be quite difficult, that's why I'd rather do it right through multiple PRs if necessary.

amirfefer:
> The place (key) in redux store for this can easily change

the key is taken  directly from core:
```js
import { HOST_DETAILS_KEY } from 'foremanReact/components/HostDetails/consts';
```
The all idea of the API middleware is to store API responses for a later reuse based on constant keys.

> There is no plan for cherry-picking this as it would already be quite difficult

Right, this dropdown depends on a newer version of PF4. So cherry picking it wouldn't be wise.

amirfefer:
Thanks, @ezr-ondrej! it works well with the `hostId`, just needed to change the `proptype` from string to number.

ezr-ondrej (approved):
Thanks @amirfefer !

PR #708, opened 2022-03-30
Refs #34653 - Fix load order

adamruzicka:
Both SSHExecutionProvider and ScriptExecutionProvider live in
ssh_execution_provider.rb. When autoloading kicks in, previously it
tried to load ScriptExecutionProvider from script_execution_provider,
which failed. Switching the order in engine.rb works around this issue.


ezr-ondrej (approved):
Cleanest way technically for autoloading would be to have a file for both, but having file just for an alias makes no sense, so :+1:

PR #709, opened 2022-03-31
Introduce Script proxy feature

adamruzicka:
This should go in before https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/77, once this is in, things should continue working no matter if the proxy has the newer bits or not.


ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #711, opened 2022-04-19
Fixes #34785 - Load helpers in initialize phase

stejskalleos:
Prepending Foreman's helpers with Rails 6.1
must be done in initializer, otherwise it won't work.

Foreman PR: https://github.com/theforeman/foreman/pull/9117

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -26,13 +26,15 @@ module ForemanRemoteExecution
       end
     end
 
-    # A workaround for https://projects.theforeman.org/issues/30685
+    # A workaround for https://projects.theforeman.org/issues/30685 and
+    # https://projects.theforeman.org/issues/34526
     initializer 'foreman_remote_execution.rails_loading_workaround' do
       # Without this, in production environment the module gets prepended too
       # late and the extensions do not get applied
       # TODO: Remove this and from config.to_prepare once there is an extension
       # point in Foreman
       ProvisioningTemplatesHelper.prepend ForemanRemoteExecution::JobTemplatesExtensions
+      HostsHelper.prepend ForemanRemoteExecution::HostsHelperExtensions
     end
 
     initializer 'foreman_remote_execution.apipie' do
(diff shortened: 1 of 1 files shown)


lib/foreman_remote_execution/engine.rb
```diff
@@ -26,13 +26,15 @@ class Engine < ::Rails::Engine
...
-    # A workaround for https://projects.theforeman.org/issues/30685
+    # A workaround for https://projects.theforeman.org/issues/30685 and
+    # https://projects.theforeman.org/issues/34526
     initializer 'foreman_remote_execution.rails_loading_workaround' do
       # Without this, in production environment the module gets prepended too
       # late and the extensions do not get applied
       # TODO: Remove this and from config.to_prepare once there is an extension
       # point in Foreman
       ProvisioningTemplatesHelper.prepend ForemanRemoteExecution::JobTemplatesExtensions
+      HostsHelper.prepend ForemanRemoteExecution::HostsHelperExtensions
```
ezr-ondrej, line 37: This was IMHO wrong as we do have a [way to do this properly](https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#extending-host-ui) (I don't think it is the best method that it could be done, but it is better then this :) )

On top of that we are already almost doing it `host_overview_buttons`:
https://github.com/theforeman/foreman_remote_execution/blob/80d1fbd7334eac5b802049e0a300d6e820776805/lib/foreman_remote_execution/engine.rb#L270

Tho it should be in a separate helper e.g. as `RemoteExecutionHostAdditions` not a patch to existing helper and the methods should have rex specific names e.g. `rex_host_overview_buttons` & `rex_multiple_actions` and then

```ruby
# engine.rb

# Foreman::Plugin.register :rex do
# ...
describe_host { overview_buttons_provider :rex_host_overview_buttons, multiple_actions_provider: :rex_multiple_actions }
```

ezr-ondrej (commented):
We have a way to enhance Host UI, which we should IMHO use, it is bit hard to understand and read, so I get why you might not even know about it, so I'm linking it in the comment with my braindump of what I myself know about it.

It has limited extension points, so we might still need to do this helper patching, but IMO we should avoid Helper patching where we can, because it causes headaches with load order (what is the core issue here IIUC)

lib/foreman_remote_execution/engine.rb
stejskalleos: Yeah you are right, we had discussion yesterday about it and we agreed that this should be the way how to do it,
I'm closing the PR.

stejskalleos:
Follow up PR: [Fix loading of hosts_extensions_helper](https://github.com/theforeman/foreman_remote_execution/pull/714)

PR #714, opened 2022-04-21
Fixes #34785 - Fix loading of hosts_extensions_helper

stejskalleos:
Use plugin provider methods `:hosts_multiple_actions` and
`:host_overview_buttons` for extending the host UI instead of
prepending the module to the Foreman's HostsHelper module.


app/helpers/hosts_extensions_helper.rb
```diff
@@ -0,0 +1,62 @@
...
+    [{ action: [_('Schedule Remote Job'), new_job_invocation_path, false], priority: 1000 }]
+  end
+
+  def host_overview_buttons(host)
+    [
+      { button: link_to_if_authorized(_("Jobs"), hash_for_job_invocations_path(search: "host=#{host.name}"), title: _("Job invocations"), class: 'btn btn-default'), priority: 200 },
+    ]
+  end
+
+  def host_title_actions(*args)
```
ezr-ondrej, line 14: This can also use `describe_host` (method `title_actions_provider`)

app/helpers/hosts_extensions_helper.rb
```diff
@@ -0,0 +1,62 @@
+module HostsExtensionsHelper
+  def hosts_multiple_actions
+    return unless can_schedule_jobs?
```
ezr-ondrej, line 3:
```suggestion
    return [] unless can_schedule_jobs?
```

app/helpers/hosts_extensions_helper.rb
```diff
@@ -0,0 +1,62 @@
+module HostsExtensionsHelper
+  def hosts_multiple_actions
```
ezr-ondrej, line 2: I still think we should use `rex_` prefix.
This could in theory clash with methods from other plugins, couldn't it?

app/helpers/hosts_extensions_helper.rb
stejskalleos: I tried but it didn't work, at least not the way how it is described in dev docs.
This way was the only one for me that worked.

ezr-ondrej: interesting... ok, I did not try myself so ok :+1: this should work pretty safe hopefully (I believe that plugin helpers will always be included after core helpers)

stejskalleos:
Rebased and updated

Just to be sure I also checked extensions of `controllers` and `models` and everything seems to be working well.

ezr-ondrej (approved):
Thanks @stejskalleos ! Awesome work :heart: :)

PR #717, opened 2022-04-28
Fixes #34835 - Open Cockpit console in a new tab

ofedoren:
I could probably use `<Link>` from PF4, but just to be consistent with the core I've used this way.

### DIFF
--- a/webpack/react_app/components/HostKebab/KebabItems.js
+++ b/webpack/react_app/components/HostKebab/KebabItems.js
@@ -13,7 +13,12 @@ const HostKebabItems = () => {
 
   if (!consoleUrl) return null;
   return (
-    <DropdownItem icon={<CodeIcon />} href={consoleUrl}>
+    <DropdownItem
+      icon={<CodeIcon />}
+      href={consoleUrl}
+      target="_blank"
+      rel="noreferrer"
+    >
       {__('Web Console')}
     </DropdownItem>
   );


webpack/react_app/components/HostKebab/KebabItems.js
```diff
@@ -13,7 +13,11 @@ const HostKebabItems = () => {
 
   if (!consoleUrl) return null;
   return (
-    <DropdownItem icon={<CodeIcon />} href={consoleUrl}>
+    <DropdownItem
```
adamruzicka, line 16: Can't we just add `target="_blank"`? It should get passed to whatever the underlying element it

adamruzicka (commented):
This works, but isn't it an overkill?

adamruzicka:
@MariaAga as far as I can tell the js test failures shouldn't be related to the changes proposed here. Any idea what that's about?

webpack/react_app/components/HostKebab/KebabItems.js
MariaAga: I agree with this, its more recommended to use
``` js
 <DropdownItem
      icon={<CodeIcon />}
      href={consoleUrl}
      target="_blank"
      rel="noreferrer"
    >
 ```

MariaAga:
Tests failures are not related and I cant really reproduce it. might be a versions thing and I'll see the tests breaking later

ofedoren:
Thanks, @adamruzicka and @MariaAga, looks clean now :)

adamruzicka (approved):
:shipit:

adamruzicka:
Thank you @ofedoren !

PR #697, opened 2022-02-23
Refs #33010 - fix slot's id

amirfefer:
works with https://github.com/theforeman/foreman/pull/9054

PR #720, opened 2022-05-06
Refs #34876 - Fix slot id for recent jobs card

jeremylenz:
Taking over from https://github.com/theforeman/foreman_remote_execution/pull/697


adamruzicka (approved):
LGTM

PR #721, opened 2022-05-09
Fixes #32104 - Retain tasks used for errata reporting for longer time

adamruzicka:
Companion to https://github.com/Katello/katello/pull/10096


ares (approved):
Makes sense, thanks, merging!

PR #727, opened 2022-05-30
Fixes #24962 - Make graph in job details clickable

adamruzicka:
Known issues:
- the row with chips is most likely wrongly aligned
- the donut chart itself does not indicate the selection, haven't found an easy way how to do that
- the search doesn't really work for hosts which don't have their task yet, but that is nothing new

https://user-images.githubusercontent.com/7326770/171000566-46ac5faf-eef7-49d1-befa-72bd20e0eb98.mp4

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -253,8 +253,9 @@ class JobInvocation < ApplicationRecord
         acc.merge(key => 0)
       end
     else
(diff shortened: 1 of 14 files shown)


stejskalleos (approved):
:green_apple: **LGTM**, works as expected

ares:
Thanks @adamruzicka and @stejskalleos, merging!

PR #729, opened 2022-06-16
Fixes #35071 - Do not configure cleanup for REX jobs

adamruzicka:


adamruzicka:
Setting cleanup interval for those actions won't be necessary once https://github.com/Katello/katello/pull/10096 gets in

adamruzicka:
Can I do [test Redmine issues]?

adamruzicka:
It was decided this should not be hardcoded and should be left to the user to configure. Closing

ofedoren (approved):

PR #730, opened 2022-06-17
Fixes #35077 - Cant run job wizard with weekly

MariaAga:
When selecting weekly recurrence in the job wizard and running the job, this error will show: 
`"cronline":["00 13 * * [object Object] is not valid format of cron line"]`
fixes also an issue where Monday was treated as 0 day when it should be sunday.


adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @MariaAga !

PR #731, opened 2022-06-20
Fixes #35084 - Add rerun to job wizard

MariaAga:
changed the menu url to `experimental/job_wizard/new`
to use this pr go to `experimental/job_wizard/_ID_/rerun` as its experimental.
UI changes are adding the org/location alerts.
Patternfly guidelines say `If an alert message applies globally to the content on a page, place the alert in the page header area just below the title.`
Screenshot from 2022-06-20 11-59-18

### DIFF
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -57,4 +57,19 @@ class UiJobWizardController < ApplicationController
     render :json => { :results =>
       resource_list.sort_by { |r| r[:name] }.take(100), :subtotal => resource_list.count}
   end
+
+  def job_invocation
+    job = JobInvocation.authorized.find(params[:id])
+    composer = JobInvocationComposer.from_job_invocation(job, params).params
+    job_template_inputs = JobTemplate.authorized.find(composer[:template_invocations][0][:template_id]).template_inputs_with_foreign
+    inputs = Hash[job_template_inputs.map{ |input| ["inputs[#{input[:name]}]", (composer[:template_invocations][0][:input_values].select {|value| value[:template_input_id]== input[:id]})[0][:value]] } ]
+    job_organization = Taxonomy.find_by(id: job.task.input[:current_organization_id])
+    job_location = Taxonomy.find_by(id: job.task.input[:current_location_id])
+    render :json => {
+      :job => composer,
+      :job_organization => job_organization,
+      :job_location => job_location,
+      :inputs => inputs,
+    }
+  end
 end
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -62,7 +62,7 @@ class UiJobWizardController < ApplicationController
     job = JobInvocation.authorized.find(params[:id])
     composer = JobInvocationComposer.from_job_invocation(job, params).params
     job_template_inputs = JobTemplate.authorized.find(composer[:template_invocations][0][:template_id]).template_inputs_with_foreign
-    inputs = Hash[job_template_inputs.map{ |input| ["inputs[#{input[:name]}]", (composer[:template_invocations][0][:input_values].select {|value| value[:template_input_id]== input[:id]})[0][:value]] } ]
+    inputs = Hash[job_template_inputs.map { |input| ["inputs[#{input[:name]}]", (composer[:template_invocations][0][:input_values].find { |value| value[:template_input_id] == input[:id] })[:value]] }]
     job_organization = Taxonomy.find_by(id: job.task.input[:current_organization_id])
     job_location = Taxonomy.find_by(id: job.task.input[:current_location_id])
     render :json => {
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -2,6 +2,7 @@
 /* eslint-disable camelcase */
 import React, { useState, useEffect, useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
 import { Wizard } from '@patternfly/react-core';
 import { get } from 'foremanReact/redux/API';
 import history from 'foremanReact/history';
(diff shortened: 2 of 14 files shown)


app/controllers/ui_job_wizard_controller.rb
```diff
@@ -57,4 +57,19 @@ def resources
     render :json => { :results =>
       resource_list.sort_by { |r| r[:name] }.take(100), :subtotal => resource_list.count}
   end
+
+  def job_invocation
+    job = JobInvocation.authorized.find(params[:id])
+    composer = JobInvocationComposer.from_job_invocation(job, params).params
+    job_template_inputs = JobTemplate.authorized.find(composer[:template_invocations][0][:template_id]).template_inputs_with_foreign
+    inputs = Hash[job_template_inputs.map{ |input| ["inputs[#{input[:name]}]", (composer[:template_invocations][0][:input_values].select {|value| value[:template_input_id]== input[:id]})[0][:value]] } ]
```
MariaAga, line 65: Is there a better way to do this? and does this function needs permissions anywhere?

adamruzicka, line 65: I'd do something like this. At least to me it reads a bit better.

```suggestion
    pattern_template_invocation = composer[:template_invocations].first
    job_template_inputs = JobTemplate.authorized.find(pattern_template_invocation[:template_id]).template_inputs_with_foreign
    inputs = job_template_inputs.reduce({}) do |acc, input|
      value = pattern_template_invocation[:input_values].find { |value| value[:template_input_id] == input[:id] }
      acc.merge("inputs[#{input[:name]}]" => value[:value])
    end

```

adamruzicka:
If I have where all the hosts succeeded and I do rerun failed only, in the old form the search would get pre-filled to `name ^ ()` and it would resolve to no hosts. In the wizard the search query becomes empty which then resolves to all hosts known to foreman.

app/controllers/ui_job_wizard_controller.rb
MariaAga: To re-use the autofilling from the url which comes in this format. I can change it but I think since we already have a function that does the same, to re-use it. 
autofill pr for context: https://github.com/theforeman/foreman_remote_execution/pull/629

adamruzicka: Ah, that makes sense. Let's keep it as is then

MariaAga:
I can't find any code results for `rerun_failed`, where should it be passed and what is its value?

adamruzicka:
Ermm, the button is "rerun failed", the actual parameter is `failed_only`, sorry about the confusion. The value is either `true` or `1` or the parameter is not passed.

MariaAga:
Thanks for the review, added the `failed_only`

app/controllers/ui_job_wizard_controller.rb
```diff
@@ -57,4 +57,19 @@ def resources
     render :json => { :results =>
       resource_list.sort_by { |r| r[:name] }.take(100), :subtotal => resource_list.count}
   end
+
+  def job_invocation
+    job = JobInvocation.authorized.find(params[:id])
+    composer = JobInvocationComposer.from_job_invocation(job, params).params
+    job_template_inputs = JobTemplate.authorized.find(composer[:template_invocations][0][:template_id]).template_inputs_with_foreign
+    inputs = Hash[job_template_inputs.map { |input| ["inputs[#{input[:name]}]", (composer[:template_invocations][0][:input_values].select { |value| value[:template_input_id]== input[:id] })[0][:value]] } ]
```
adamruzicka, line 65: Please do use `find` over `select {...}[0]`, there's no need to collect all the elements matching a predicate if we only need the first one

webpack/JobWizard/JobWizard.js
```diff
@@ -48,7 +48,8 @@ export const JobWizard = ({ rerunData }) => {
     hostGroups: [],
   });
   const [hostsSearchQuery, setHostsSearchQuery] = useState('');
-  const [fills, setFills] = useState(useSelector(selectRouterSearch));
+  const routerSearch = useSelector(selectRouterSearch);
```
MariaAga, line 51: Can't put hooks inside conditions

MariaAga:
There was an issue with hooks timing for the filling, should work now

adamruzicka:
Waiting for someone to take a look at the js bits, as far as functionality goes +1 from me

webpack/JobWizard/JobWizard.js
```diff
@@ -102,23 +107,50 @@ export const JobWizard = () => {
...
+          ssh_user: rerunData.ssh_user,
+          concurrency_control: rerunData.concurrency_control,
+        },
+      });
+      setFills({
+        search: rerunData.targeting.search_query,
+        ...rerunData.inputs,
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jeremylenz, line 141: Can this effect be eliminated and instead calculate all the values during render?

https://beta.reactjs.org/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects

jeremylenz: If you feel strongly about keeping this an effect, at least the rule shouldn't be disabled.  But as I was saying before, I think the better solution is to not use an effect at all.

jeremylenz (changes requested):
What are the consequences to me as the user if I run the job from a different org / location?  Does this mean the job will run on different hosts than it did before?

Below, same comment as last time.  (sorry :grimacing:)

adamruzicka:
Needs a rebase now

MariaAga:
Thanks for the reviews!
Rebased, removed some logic from the useeffect, removed `eslint-disable`

> What are the consequences to me as the user if I run the job from a different org / location? Does this mean the job will run on different hosts than it did before?

From what I see the host count is different, @adamruzicka is there something else?

jeremylenz:
I'm getting 
```
No route matches [GET] "/experimental/job_wizard/493/rerun"
```
all of a sudden.  Not sure what I'm doing wrong..

MariaAga:
@jeremylenz I'm not seeing this issue, but when I had it I had to re-run the server

jeremylenz:
@MariaAga got it working now, thanks :)

adamruzicka:
The host resolution is the only thing that should be affected by org/loc selection

MariaAga:
Thanks! new message:
Screenshot from 2022-07-13 13-54-52

jeremylenz (approved):
Message changes look good :+1: 

Works well.

As I said in the previous review I think the useEffect refactor could still go further, but I'll leave that to the future.

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @MariaAga & @jeremylenz !

PR #732, opened 2022-06-20
Fixes #35086 - allow to set advanced install_package options

sbernhard:


app/views/templates/script/package_action.erb
```diff
@@ -113,7 +118,7 @@ handle_zypp_res_codes () {
     end
   -%>
   zypper refresh
-  zypper -n <%= action %> <%= input("package") %>
+  zypper -n <%= action %> <%= input("options") %> <%= input("package") %>
```
adamruzicka, line 121: I don't know zypper by heart, but for yum and apt the options go before the action, but in here the options follow the action. Is that intentional?

sbernhard: For zypper, global options are before the action and action-specific options are after the action. Most commonly you want to set action specific options. Otherwise we would need two additional template parameters which would mood make any sense for apt/yum.

adamruzicka (approved):

adamruzicka:
Thank you @sbernhard !

PR #735, opened 2022-07-01
Fixes #35154 - Job wizard user input for date use datepicker

MariaAga:
When a template has a date field instead of using a text box, the job wizard should use a datetime picker.
To test might need to reinstall npm packages as the `includeSeconds` option is newish


adamruzicka (changes requested):
The date needs to be formatted before sending it over to the backend. The old form sends a date in `YYYY-mm-dd HH:MM:SS` format (`2022-07-27 15:10:00`), with this patch, it gets sent over in some, possibly default format, which is much more verbose `Wed Jul 27 2022 15:10:00 GMT+0200 (Central European Summer Time)`

MariaAga:
Thanks! added the `formatDateTime` function

PR #736, opened 2022-07-07
Fixes #35172 - Do not crash if report template missing

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka, LGTM :)

PR #737, opened 2022-07-08
Fixes #35181 - Honor user's TZ when triggering jobs through API

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka, LGTM and works well!

P.S. :cookie: for branch name :D

ezr-ondrej (approved):
Just clicking buttons (skimmed the code and looks good) :+1:
I just wonder, why @ofedoren's approval is not green yet? can we change that? :thinking:

adamruzicka:
> can we change that?

Good point, yes, done!

PR #739, opened 2022-07-14
Fixes #35218 - Store job outputs outside of tasks

adamruzicka:
Before this change all job outputs were stored deep in actions' outputs.
When we wanted to retrieve them, we had load an instance of the action
and then let it assemble all the outputs into a form that the rest of
the application could consume.

This is a first step away from it. Outputs now live in a separate table,
which gets populated as things happen.

TODO:
- [x] check all possible non-happy paths
- [x] make live output populate the events table and then read the data from there

In the future this should allow us to:
- have a more precise progress reporting (in spirit of https://github.com/theforeman/foreman_remote_execution/pull/712) by inserting "control" events into the log
- retrieve only new outputs from the smart proxy instead of retrieving all of them every single time
- load the outputs in chunks when showing them in the ui

Requires:
- https://github.com/theforeman/foreman-tasks/pull/689

### DIFF
--- a/app/lib/actions/remote_execution/proxy_action.rb
+++ b/app/lib/actions/remote_execution/proxy_action.rb
@@ -1,0 +1,48 @@
+module Actions
+  module RemoteExecution
+    class ProxyAction < ::Actions::ProxyAction
+      include Actions::RemoteExecution::TemplateInvocationProgressLogging
+
+      def on_data(data, meta = {})
+        super
+        process_proxy_data(output[:proxy_output])
+      end
+
+      def run(event = nil)
+        with_template_invocation_error_logging { super }
+      end
+
+      private
+
+      def get_proxy_data(response)
+        data = super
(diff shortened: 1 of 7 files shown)


adamruzicka:
[test Redmine issues]

stejskalleos (approved):
From first look it works fine and I didn't find any issues, tested scenarios:

- Remote execution - happy path, SSH connection errors, bash errors
- Leapp preupgrade
- Ansible - happy path, failed jobs

However I don't feel confident to ACK from the code perspective,
would be lying if I'd say that I fully understand the code changes.

Second thing is that I tried to run the job only at two hosts at one time, maybe would be worth to do some performance testing, or at least notify someone from the performance team to let them know about this changes.

adamruzicka:
Good, thank you for that. I'll merge this in good faith and we'll see where it takes us.

PR #740, opened 2022-07-15
Fixes #35228 - Fix proxy selection with capability

adamruzicka:


ofedoren (approved):
TIL: after some digging through the code, just out of curiosity, it seems [find_by name:](https://github.com/theforeman/foreman/blob/develop/app/models/smart_proxy.rb#L90) can accpet an [array](https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/ssh_execution_provider.rb#L55)...

But regarding this PR: makes sense, let's get this in :)

adamruzicka:
Thank you @ofedoren !

PR #741, opened 2022-07-20
Fixes #35249 - Use only ruby 2.7 in CI

adamruzicka:


ofedoren (approved):
Let's get this in :)

adamruzicka:
Thank you @ofedoren !

PR #742, opened 2022-08-01
Fixes #35307 - Do not rely on tasks always being there

adamruzicka:
when calculating details of execution status.


ofedoren (approved):

PR #743, opened 2022-08-01
Fixes #35308 - Make script provider name match proxy feature

adamruzicka:
When assigning REX proxies to a subnet, proxy features are matched
exactly against registered REX provider names. Since there is a one
letter mismatch, proxies having only Script feature would not show as
available in the list.

### DIFF
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -4,7 +4,7 @@ class RemoteExecutionProvider
 
   class << self
     def provider_for(type)
-      providers[type.to_s] || providers[:script]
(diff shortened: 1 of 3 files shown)


ofedoren (commented):
Thanks, @adamruzicka, but shouldn't we update this line as well https://github.com/adamruzicka/foreman_remote_execution/blob/master/app/models/remote_execution_provider.rb#L7 ?

P.S. It would be better to simply support both cases or even make it not dependent on casing at all :/

ofedoren (approved):
Thanks, @adamruzicka, let's get this in :)

PR #744, opened 2022-08-02
Refs #35308 - Propagate the new registered name to more places

adamruzicka:
Without this, running a job would fail with
`NoMethodError: undefined method 'host_id=' for ["SSH", "Script"]:Array`

### DIFF
--- a/app/models/job_invocation.rb
+++ b/app/models/job_invocation.rb
@@ -213,7 +213,7 @@ class JobInvocation < ApplicationRecord
     providers = available_providers(host)
     providers.each do |provider|
       pattern_template_invocations.each do |template_invocation|
-        if template_invocation.template.provider_type.downcase == provider.downcase
+        if template_invocation.template.provider_type == provider
           return template_invocation
         end
       end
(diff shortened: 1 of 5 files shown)


ofedoren:
Thanks, @adamruzicka, but it seems like a workaround. Although it makes things safer, I'm curious why would we want to do that.  since it looks like it just worked for `SSH`. To elaborate on that:

AFAIU, we previously registred `SSH` only and it worked. That means it compared `SSH` with `SSH`. With this change it would compare `ssh` with `ssh`, which will work as it did before. But since we added `Script` it seems like it tried to compare with `script`, which means we fail to save `Script` instead somewhere. Would be better to find where `script` comes from and fix it there?

But anyway, if this fix is more like for backwards compatibility since we had `script` registred before then simply ACK and I'll merge this :)

adamruzicka:
Yes, things got out of hand a bit.

Problem statement:
- when doing the ssh -> script rename, I didn't consider we might be doing case sensitive comparison somewhere
- provider_type in templates (currently lowercase script) has to match the name under which the provider was registered when the job is being run
- to allow assigning a proxy to a subnet, the name under which the provider is registered has to match the name of the proxy feature (Script)

I took another stab at this and instead of renaming things, I essentially reverted the first part of this fix and removed the requirement that the name under which a provider is registered has to match the name of the proxy feature. Each provider can now specify one or more proxy features. Still not ideal, but hopefully better then the previous attempt

ofedoren (approved):
Thanks, @adamruzicka, seems to be a better approach. As far as I could check it works, so let's get this in.

PR #745, opened 2022-08-03
Fixes #35327 - Add validation for search on job invocation id

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka, this is nice to have, let's get this in.

PR #746, opened 2022-08-08
Fixes #35343 - make sure to install kernel updates

sbernhard:


adamruzicka (approved):
Huh, that's quite surprising behavior on `apt-get`'s side. Let's get this in

adamruzicka:
Thank you @sbernhard !

evgeni:
One shall not use `apt` in scripts (which is what we do here).

It even says so if you do:

```console
# apt install wget -y | cat

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

The correct way here would be to override `action` to be `install`, as `apt-get install <existing package>` does the right thing.

Or you need to call `apt-get dist-upgrade` to allow installing new packages, which `apt-get upgrade` does not.

evgeni:
Alternatively you should be able to set `APT::Get::Upgrade-Allow-New=true` (I think)

PR #749, opened 2022-08-22
Refs #35327 - More search validations

adamruzicka:


ofedoren (approved):

PR #750, opened 2022-08-22
Fixes #35408 - Add time_to_pickup setting

adamruzicka:
overridable by the usual parameter hierarchy and per job.

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -155,6 +155,11 @@ module ForemanRemoteExecution
               default: 'Job - Invocation Report',
               full_name: N_('Job Invocation Report Template'),
               collection: proc { ForemanRemoteExecution.job_invocation_report_templates_select }
+            setting 'remote_execution_time_to_pickup',
+              type: :integer,
+              description: N_('Time in seconds within which the host has to pick up a job. If the job is not picked up within this limit, the job will be cancelled. Defaults to 1 day.'),
+              default: 24 * 60 * 60,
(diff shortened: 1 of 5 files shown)


ofedoren (commented):
Thanks, @adamruzicka, LGTM, but I don't see where this param is used. Is there a corresponding PR for e.g. proxy?

adamruzicka:
> Me, circa yesterday: "I'll open the proxy PR, then the foreman one and link the proxy one there"
> Me, after just opening the PRs: "Another job well done"

Yes, of course, here it is https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/92

lib/foreman_remote_execution/engine.rb
```diff
@@ -155,6 +155,11 @@ class Engine < ::Rails::Engine
               default: 'Job - Invocation Report',
               full_name: N_('Job Invocation Report Template'),
               collection: proc { ForemanRemoteExecution.job_invocation_report_templates_select }
+            setting 'remote_execution_time_to_pickup',
+              type: :integer,
+              description: N_('Time in seconds within which the host has to pick up a job. If the job is not picked up within this limit, the job will be cancelled.'),
```
ofedoren, line 160: Should we mention that by default it's 24 hours?

ofedoren (commented):
Small nit, otherwise LGTM.

lib/foreman_remote_execution/engine.rb
adamruzicka: Done

ofedoren (approved):

PR #752, opened 2022-08-26
Fixes #35437 - host details jobs change from list to table

MariaAga:
Its more correct to use table in this case and its easier for qe to test

### DIFF
--- a/webpack/react_app/components/RecentJobsCard/RecentJobsTable.js
+++ b/webpack/react_app/components/RecentJobsCard/RecentJobsTable.js
@@ -1,15 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import {
-  DataList,
-  DataListItem,
-  DataListItemRow,
-  DataListItemCells,
-  DataListCell,
-  DataListWrapModifier,
-  Text,
-  Bullseye,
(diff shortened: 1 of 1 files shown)


webpack/react_app/components/RecentJobsCard/RecentJobsTable.js
```diff
@@ -33,55 +25,50 @@ const RecentJobsTable = ({ status, hostId }) => {
...
-                    ]}
-                  />
-                </DataListItemRow>
-              </DataListItem>
-            )
-          )}
-      </SkeletonLoader>
-    </DataList>
+        <Tbody>
+          {jobs?.length &&
```
adamruzicka, line 45: If there are no jobs, the card just says `0`
image

Maybe `map`-ing over an empty array should be valid

adamruzicka (changes requested):

adamruzicka:
Also needs a rebase

MariaAga:
rebased and fixed the 0 issue by moving `jobs?.length` up in the tree to trigger the empty state message

adamruzicka (approved):
ACK

adamruzicka:
Thank you @MariaAga !

PR #753, opened 2022-08-30
Fixes #35315 - Properly fail when pattern template cannot be found

adamruzicka:


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -47,7 +47,6 @@ def create_sub_plans
           # composer creates just "pattern" for template_invocations because target is evaluated
           # during actual run (here) so we build template invocations from these patterns
           template_invocation = job_invocation.pattern_template_invocation_for_host(host).deep_clone
-          template_invocation.host_id = host.id
```
ofedoren, line 50: Why does this change?..

adamruzicka: It is being done anyway here https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L29 . It felt cleaner to pass the possible nil value into the child action and handle it properly there rather than handling in two different places

ofedoren (approved):
Thanks, @adamruzicka, LGTM and :green_apple:. Let's get this in.

PR #725, opened 2022-05-26
Fixes #34970 - Allow enabling REX pull in registration form

adamruzicka:
Requires:
- https://github.com/theforeman/foreman/pull/9235

PR #754, opened 2022-09-01
refs #34970 - Host Registration - put interface back

stejskalleos:
Changes from PR#725 accidentaly removes interface
field from registration form.

[0] https://github.com/theforeman/foreman_remote_execution/pull/725


adamruzicka (approved):
ACK as a workaround until https://projects.theforeman.org/issues/35471 gets fixed properly

PR #755, opened 2022-09-01
Fixes #35471 - Slot & Fill renders only one component

MariaAga:
undoing - https://github.com/theforeman/foreman_remote_execution/pull/754
the `name` key was not used and `id` was used instead, `id` was always undefined.


adamruzicka (approved):
Works well

adamruzicka:
Thank you @MariaAga !

PR #757, opened 2022-09-12
Fixes #35515 - Fix proxy selection for subclassed providers

adamruzicka:
::RemoteExecutionProvider does a lookup inside class instance variables set on it. When a custom provider subclasses ::RemoteExecutionProvider, this lookup is done inside the subclass, where the variables are not set so this lookup returns nil.

The other piece is `Host::Managed#remote_execution_proxies(provider)` returning all the proxies, if `provider` is `nil`.

The unfortunate outcome of those two above is that when using subclassed providers (such as in Ansible), proxies were not filtered based on available features.

The fix causes the lookup to be performed on the top-level class, instead of in the subclass.

### DIFF
--- a/app/models/remote_execution_provider.rb
+++ b/app/models/remote_execution_provider.rb
@@ -9,7 +9,7 @@ class RemoteExecutionProvider
 
     def registered_name
       klass = self
-      providers.key(klass)
+      ::RemoteExecutionProvider.providers.key(klass)
     end
 
     def proxy_feature
--- a/test/unit/remote_execution_provider_test.rb
+++ b/test/unit/remote_execution_provider_test.rb
@@ -56,6 +56,28 @@ class RemoteExecutionProviderTest < ActiveSupport::TestCase
     end
   end
 
+  describe '.proxy_feature' do
+    # rubocop:disable Naming/ConstantName
+    it 'handles provider subclasses properly' do
+      old = ::RemoteExecutionProvider
+
+      class P2 < old
+      end
+      ::RemoteExecutionProvider = P2
+
+      class CustomProvider < ::RemoteExecutionProvider
+      end
+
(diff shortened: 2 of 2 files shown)


app/models/remote_execution_provider.rb
```diff
@@ -9,7 +9,7 @@ def provider_for(type)
 
     def registered_name
       klass = self
-      providers.key(klass)
+      ::RemoteExecutionProvider.providers.key(klass)
```
ofedoren, line 12: Seems logical to me if we only allow registering all the providers in the scope of `RemoteExecutionProvider` only. I mean, any subclass of the `RemoteExecutionProvider` can have its own registry and to make `registered_name` work properly for the subclass, it needs to redefine `registered_name` to use `SubclassProvider` scope for the registry.

If it's by design that registration of the providers must be done via `RemoteExecutionProvider.register` only then ACK.

ofedoren: Could this https://github.com/theforeman/foreman_remote_execution/pull/757/files#diff-5c51055144db11e106aa62a1537c9ca8e4cafce1c5503000eb0d8b7ce232b232R31 also fail then? Or any method that uses `providers`?

I mean, I'm not so familiar with the usage of subclasses. If they are more or less `constants only` then it's okay, but if all those methods can be called on a subclass, wouldn't it behave inconsistently?

adamruzicka: In theory it could, but the only place where we call `.provider_for` that I'm aware of is [here](https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_template.rb#L132) and then the rest of the methods called on the returned provider *should* be safe, with the exception of `#proxy_features`.

ofedoren (approved):
Okay then. At least we fix something currently broken, let's get this in.

PR #758, opened 2022-09-30
Fixes #35584 - Make job wizard the default way to create jobs

MariaAga:
* move job wizard from `/experimental/job_wizard/new` to `job_invocations/new`
* move job form from `job_invocations/new` to `old/job_invocations/new`
* remove job wizard from the lab menu tab
* fix bug where hosts list in the url can be a single host (and not an array)
* fix bug where users get to the wizard from the host details page, and then go back (to host details) and again to the job wizard caused some data to be not displayed correctly in the wizard because the `component state` was deleted but the `redux store` stayed the same
* remove unused `job_invocations_buttons` 
* fixed wizards breadcrumbs to go to the correct all jobs url
* if users want to use the old form/new wizard there is a link to the other type in the form/wizard

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,7 +16,11 @@ Rails.application.routes.draw do
     end
   end
 
-  match 'job_invocations/new', to: 'job_invocations#new', via: [:get, :post], as: 'new_job_invocation'
+  match 'job_invocations/new', to: 'react#index', :via => [:get], as: 'new_job_invocation'
+  match 'job_invocations/new', to: 'job_invocations#create', via: [:post], as: 'create_job_invocation'
+  match 'job_invocations/:id/rerun', to: 'react#index', :via => [:get], as: 'rerun_job_invocation'
+  match 'old/job_invocations/new', to: 'job_invocations#new', via: [:get], as: 'form_new_job_invocation'
+  match 'old/job_invocations/:id/rerun', to: 'job_invocations#rerun', via: [:get, :post], as: 'form_rerun_job_invocation'
   resources :job_invocations, :only => [:create, :show, :index] do
     collection do
       post 'refresh'
@@ -25,7 +29,6 @@ Rails.application.routes.draw do
       get 'auto_complete_search'
     end
     member do
-      get 'rerun'
       post 'cancel'
     end
   end
@@ -47,9 +50,6 @@ Rails.application.routes.draw do
   get 'ui_job_wizard/template/:id', to: 'ui_job_wizard#template'
   get 'ui_job_wizard/resources', to: 'ui_job_wizard#resources'
   get 'ui_job_wizard/job_invocation', to: 'ui_job_wizard#job_invocation'
-
-  match '/experimental/job_wizard/new', to: 'react#index', :via => [:get]
-  match '/experimental/job_wizard/:id/rerun', to: 'react#index', :via => [:get]
 
   namespace :api, :defaults => {:format => 'json'} do
     scope '(:apiv)', :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2, :default => true) do
--- a/webpack/JobWizard/index.js
+++ b/webpack/JobWizard/index.js
@@ -1,14 +1,15 @@
 import React from 'react';
-import { Title, Divider } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+import { Title, Divider, Flex, FlexItem, Button } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
 import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
 import { JobWizard } from './JobWizard';
 
-const JobWizardPage = () => {
+const JobWizardPage = ({ location: { search } }) => {
   const title = __('Run job');
   const breadcrumbOptions = {
     breadcrumbItems: [
-      { caption: __('Jobs'), url: `/jobs` },
+      { caption: __('Jobs'), url: `/job_invocations` },
       { caption: title },
     ],
   };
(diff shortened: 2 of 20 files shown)


config/routes.rb
```diff
@@ -16,7 +16,9 @@
     end
   end
 
-  match 'job_invocations/new', to: 'job_invocations#new', via: [:get, :post], as: 'new_job_invocation'
+  match 'job_invocations/new', to: 'react#index', :via => [:get], as: 'new_job_invocation'
+  match 'old/job_invocations/new', to: 'job_invocations#new', via: [:get, :post], as: 'form_new_job_invocation'
```
adamruzicka, line 20: `/old/job_invocations/rerun` seems to be missing

webpack/JobWizard/index.js
```diff
@@ -1,14 +1,15 @@
...
 import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
 import { JobWizard } from './JobWizard';
 
-const JobWizardPage = () => {
+const JobWizardPage = ({ location: { search } }) => {
   const title = __('Run job');
   const breadcrumbOptions = {
     breadcrumbItems: [
-      { caption: __('Jobs'), url: `/jobs` },
+      { caption: __('Jobs'), url: `/job_invocations` },
```
adamruzicka, line 12: On rerun it still links to `/jobs`

webpack/JobWizard/index.js
```diff
@@ -19,14 +20,32 @@ const JobWizardPage = () => {
       searchable={false}
     >
       <React.Fragment>
-        <Title headingLevel="h2" size="2xl">
-          {title}
-        </Title>
+        <Flex>
```
adamruzicka, line 23: This is not shown at all on rerun

adamruzicka (changes requested):
The happy paths seem to work well, but it is still a bit rough around the edges. There are some things that are not directly caused by this PR, but they were not easily testable before this change.

Things that work well:
- `Monitor > Jobs > Run job`
- `Monitor > Jobs > $JOB > Rerun`
- `Monitor > Jobs > $JOB > Rerun failed`
- `Hosts > All Hosts > Schedule remote job`
- `Hosts > All Hosts > $HOST > Schedule a job`
- `Hosts > All Hosts > $HOST > Legacy UI > Schedule a job`

Things that could use some love:
- If `ssh_user` is not filled in, it gets passed on as an empty string (previously it was `nil`). This causes issue later down the pipe when we try to use that empty string as a user
- Description does not seem to get filled in on rerun
- Doing a rerun, changing the inputs and running the job just uses the old, unchanged inputs
- `Hosts > Job Templates > Run` seems to ignore the template (passed to the wizard as `?template_id=$ID`
- `Monitor > Jobs > $JOB > $HOST > Rerun` seems to attempt to rerun the whole `$JOB`, not only `$HOST`

MariaAga:
> If ssh_user is not filled in, it gets passed on as an empty string (previously it was nil). This causes issue later down the pipe when we try to use that empty string as a user

fixed

> Description does not seem to get filled in on rerun

backend issue - doesn't send/save the jobs description format. fixed the issue so when there is no description format it will use the default

> Doing a rerun, changing the inputs and running the job just uses the old, unchanged inputs

fixed

> Hosts > Job Templates > Run seems to ignore the template (passed to the wizard as ?template_id=$ID

added this functionality, and removed some `useeffect` dependencies to avoid double api calls

> Monitor > Jobs > $JOB > $HOST > Rerun seems to attempt to rerun the whole $JOB, not only $HOST

Could not reproduce, when I tested it the job was pre-filled with host search -> `id ^ 140` and then it was run only on on that host

MariaAga:
is a more correct term would be "use legacy form"?

MariaAga:
missing new url for `job_invocations?feature=..`

adamruzicka:
> is a more correct term would be "use legacy form"?

CC @marisvrik 

> Could not reproduce

Odd, on my machine if I do `/job_invocations/$JOB_ID/rerun?host_ids[]=$HOST_ID`, it still gets prepopulated with host ids from the original job.

MariaAga:
bug to fix: `No route matches [POST] "/job_invocations/new"`

MariSvirik:
checking with content team

MariaAga:
> from what I saw it was an issue with rex_host_features using a wrong url so fixed that it will use job_invocation/new?feature= instead of job_invocations?feature=

undid that as 1. it was used in more places 2. before this change post requests to `job_invocations` would go to the `create` function. fixed the create function to redirect to the new url (hopefully)

adamlazik1:
> is a more correct term would be "use legacy form"?

I would avoid using the word "legacy". "Use the old form" is easier to translate for non English speakers.

adamlazik1:
Thanks for the info. I assume that "Use legacy form" refers to the same legacy UI. In that case, it would be indeed better to use "legacy" for consistency, unless the **Legacy UI** button can be renamed, but I assume that is outside of the scope of this PR.

MariaAga:
Thanks for the feedback, changed to `Legacy`. Everything should be working now

adamruzicka (changes requested):
Overall looks good, last two nitpicks

Huh, when I run a job from host index, the search query in the form is correct (`id ^ ($id)`), but when I run the job, it gets transformed into `(id ^ ($id)) or (id = $id)`.

Could we add debounce or something to search query in target hosts and inputs?

MariaAga:
1.Was accidentally added in this pr and removed, thanks
2. the call to `/api/hosts` is denounced with a value of 700, how much should I increase it?

adamruzicka:
>  2. the call to /api/hosts is denounced with a value of 700, how much should I increase it?

Not sure, but I noticed that over the time it took me to type out a reasonably sized search query it tried to updated the count a couple of times.

MariaAga:
changed to 1.5 seconds then

adamruzicka (approved):
Works like a charm now

adamruzicka:
Thank you @MariaAga !

PR #760, opened 2022-10-04
Fixes #31310 - Use redis glob pattern during job finalization

laugmanuel:
Rails cache store does not support regex, but does support glob patterns.

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -75,7 +75,7 @@ module Actions
         job_invocation.save!
 
         Rails.logger.debug "cleaning cache for keys that begin with 'job_invocation_#{job_invocation.id}'"
-        Rails.cache.delete_matched(/\A#{JobInvocation::CACHE_PREFIX}_#{job_invocation.id}/)
+        Rails.cache.delete_matched(cache_deletion_query(job_invocation.id))
       end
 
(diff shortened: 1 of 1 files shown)


app/lib/actions/remote_execution/run_hosts_job.rb
```diff
@@ -75,7 +75,12 @@ def finalize
         job_invocation.save!
 
         Rails.logger.debug "cleaning cache for keys that begin with 'job_invocation_#{job_invocation.id}'"
-        Rails.cache.delete_matched(/\A#{JobInvocation::CACHE_PREFIX}_#{job_invocation.id}/)
+        delete_query = /\A#{JobInvocation::CACHE_PREFIX}_#{job_invocation.id}/
+
+        if redis_cache?
+          delete_query = "#{JobInvocation::CACHE_PREFIX}_#{job_invocation.id}*"
+        end
+        Rails.cache.delete_matched(delete_query)
```
adamruzicka, line 83: Could we do it a little bit differently? Rather than building the query inside `#finalize`, we'd have a method (say `cache_deletion_query`) which would return the right query for the currently configured cache store and in here we'd just do something like `Rails.cache.delete_matched(cache_deletion_query(job_invocation.id))`?

adamruzicka (approved):
Works like a charm

adamruzicka:
Thank you @laugmanuel !

PR #761, opened 2022-10-07
Fixes #35609 - Fix package upgrade on debian/ubuntu

sbernhard:


adamruzicka:
@evgeni could I borrow you and your deb expertise here?

evgeni:
So https://projects.theforeman.org/issues/35609 says "This leads to failed errata application when trying to install packages which are neither installed nor even available from the subscribed repositories."

The "neither installed" part seems irrelevant, we're already using `--only-upgrade`, so no new packages would be installed.
The "nor available" should then be ignored too (or so would the reader assume), but in reality it throws an error: "E: Unable to locate package this-doesnt-exist"

To accommodate that, you're now filtering on whether apt can or can not see that package.

I'll argue this is wrong. For two reasons.

Reason number one:
* Assume you have a source package `example` that builds `example-bin` and `example-data`.
* Both binary packages are installed on `hostA`, but for some reason `example-data` is not visible via `apt-cache show` (maybe because the repos are wrongly configured, the moon is red, or whatever)
* An errata `ERR001` is issued for that source package and the user decides to apply the errata.
* With your change, the user would get `example-bin` updated, but not `example-data`, which by definition doesn't suffice to say "ERR001 was applied", especially as the errata usually doesn't say in which subpackage the real fix is in.
* So if the user said "update packageX", you better update packageX or error out, no weaseling around ;)
* The concern here would be solved by checking if those packages are *installed* vs *visible* (e.g. by using `dpkg-query` and friends), but read on please ;)

Reason two:
* This is called from Katello, right? Katello knows what packages are installed on the system (otherwise it wouldn't have said the errata applies), why the heck isn't it telling us to install the right set of packages, the ones that are installed?

evgeni:
Or maybe to rephrase "reason two": when we apply the same errata via katello-agent or mqtt pull, what happens then?

sbernhard:
It would be nice to only reference packages in an erratum which are really necessary to be upgraded. Unfortunately, a DSA  only provides the affected source package: https://www.debian.org/security/2022/dsa-5271

Therefore, a erratum contains all packages which were build using this source package - and that can be a lot. 
If on the system example-bin is installed, it means, the Erratum is necessary to fix the issue. 

I understand your point that it should also determine if an affected package is already installed - like example-data.
In case the example-data is installed, a "update" should also be possible - and hopefully, the update is available through the provided repostories. 

The question is, what should happen, if example-data is affected and should be updated, but the package is not provided by a repository. Throw an error?

evgeni:
> The question is, what should happen, if example-data is affected and should be updated, but the package is not provided by a repository. Throw an error?

IMHO yes. The user requested "fix ERR01", and if we can't ensure all installed and affected packages were updated, we did not perform the action the user requested.

app/views/templates/script/package_action.erb
```diff
@@ -97,18 +109,27 @@ handle_zypp_res_codes () {
...
   [ -x "$(command -v subscription-manager)" ] && subscription-manager refresh
   export DEBIAN_FRONTEND=noninteractive
   apt-get -y update
-  apt-get -o APT::Get::Upgrade-Allow-New="true" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y <%= input("options") %> <%= action %> <%= input("package") %>
+
+  PACKAGES="<%= input("package") %>"
+  <% if is_pkg_upgrade -%>
+    PACKAGES="$(get_upgradeable_packages "<%= input("package") %>")"
+    echo "Only try to upgrade packages which are already installed on the system: ${PACKAGES}"
+  <% end -%>
```
evgeni, line 130:
```suggestion
  <% if is_pkg_upgrade -%>
    PACKAGES="$(get_upgradeable_packages "<%= input("package") %>")"
    echo "Only try to upgrade packages which are already installed on the system: ${PACKAGES}"
  <% else -%>
    PACKAGES="<%= input("package") %>"
  <% end -%>
```

Or something similar, so that `PACKAGES=…` only shows up once in the rendered version? Otherwise, it seems confusing to the (possible) reader

app/views/templates/script/package_action.erb
```diff
@@ -83,6 +83,18 @@ handle_zypp_res_codes () {
 }
 <% end -%>
 
+<% if package_manager == 'apt' && action == 'update' -%>
+get_upgradeable_packages () {
+  packages=""
+  for i in $1; do
+    if [ $(LANG=C dpkg-query -W -f='${db:Status-Status}' "$i") = "installed" ]; then
```
evgeni, line 90: if the package is not installed, `dpkg-query` will not return anything on stdout resulting in:

```
# dash
dpkg-query: no packages found matching dpkgaaa
[: =: unexpected operator
# bash
dpkg-query: no packages found matching dpkgaaa
[: =: unary operator expected
```

```suggestion
    if [ "$(LANG=C dpkg-query -W -f='${db:Status-Status}' $i)" = "installed" ]; then
```

that would work

evgeni:
Evgeni is being picky on the syntax, but otherwise this now does what I suggested (besides me still hoping Katello would be able to just pass us the right packages, but oh well)

sbernhard:
You are right, the Katello part should be better. Thanks for your review / feedback!

PR #764, opened 2022-11-04
Fixes #35721 - reset to default button to job wizard fields

MariaAga:
Adds a button next to every field that can have a default value to reset the field to that value.
Will show button only for fields that have a default value
example screenshot:
Screenshot from 2022-11-04 20-03-34

### DIFF
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -35,6 +35,7 @@ import ReviewDetails from './steps/ReviewDetails/';
 import { useValidation } from './validation';
 import { useAutoFill } from './autofill';
 import { submit } from './submit';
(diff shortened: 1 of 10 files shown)


adamruzicka (changes requested):
In general it seems to work well.

The only place where it doesn't seem to work if for inputs of type search where clicking the reset button does nothing at all.

Another thing I noticed is that inputs of type resource cannot be used with the wizard at all, the wizard crashes at the review step. That doesn't seem to be caused by this PR so I filed https://projects.theforeman.org/issues/35725 to track it.

MariaAga:
Fixed the search inputs

adamruzicka (approved):
Nothing to add here, well done

adamruzicka:
Thank you @MariaAga !

PR #765, opened 2022-11-09
Fixes #35725 - job wizard resource inputs not working

MariaAga:
Fixed:
* Review step crashing when there is input in a resource input
* rerun not filling resource inputs
* submit sending wrong data for resource inputs

The `ResourceSelect` "new" component is an edited version of `SearchSelect` which was used for it in the previous pr.


adamruzicka (approved):
It works well, I haven't noticed any undesired side effects, let's get it in

adamruzicka:
Thank you @MariaAga !

PR #769, opened 2022-11-30
Fixes #35806 - Introduce sequence_id to new outputs

adamruzicka:
The new storage for job outputs assumed that for each combination of template_invocation_id, event_type, timestamp, there would be up to 1 entry. Considering timestamps are stored with microsecond precision, and sp-rex-ssh collects outputs once per second, this felt somewhat reasonable. However, during transfer the timestamps get serialized (and then deserialized), reducing the precision to seconds in the process. REX then complained about duplicate records when trying to upsert the incoming outputs into the db.

This change introduces sequence_ids to template_invocation_events table. These ids represent the ordering of an output event within the line of events for a given template invocation and as such, every combination of (template_invocation_id, sequence_id) must be unique.

Eventually output providers (pull provider, sp-rex-ssh) should generate the sequence ids themselves, but right now we can generate them on the fly as Foreman is getting full output every time.

### DIFF
--- a/app/lib/actions/remote_execution/proxy_action.rb
+++ b/app/lib/actions/remote_execution/proxy_action.rb
@@ -21,8 +21,11 @@ module Actions
       end
 
       def process_proxy_data(data)
-        events = data['result'].map do |update|
+        events = data['result'].each_with_index.map do |update, seq_id|
           {
+            # For N-1 compatibility, we assume that the output provided here is
+            # complete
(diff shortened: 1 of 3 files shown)


jeremylenz (approved):
Tested and works well

code LGTM

thanks @adamruzicka! ACK :+1:

adamruzicka:
Thank you @jeremylenz !

PR #773, opened 2023-01-19
i18n - pulling from tx

ekohl:


adamruzicka:
@ekohl rubocop is unhappy about a missing trailing newline in the generated action file. Maybe instead of of fixing it, we could ignore the file?

ekohl:
We had a similar conversation before and I wouldn't mind that. Would it make sense to also enhance the task generate a comment `# TRANSLATORS: ...` to describe where it comes from?

For example: https://github.com/theforeman/foreman/blob/22a9408d0168f341713d515be188870ce86a85e4/lib/foreman/gettext/support.rb#L60-L61 ends up as https://github.com/theforeman/foreman/blob/22a9408d0168f341713d515be188870ce86a85e4/locale/foreman.pot#L14717-L14720

adamruzicka:
That would be nice to have

adamruzicka:
Added the trailing newline manually just so we can get this in

adamruzicka:
Thank you @ekohl !

PR #774, opened 2023-01-20
Fixes #35972 - Link to new form from empty state

adamruzicka:
Rails routing gets a bit confused by a lot of things routing to react#index. It was generating a link for job_invocations#new, but that is an endpoint serving the old form.


ofedoren (approved):
Thanks, @adamruzicka, fixes the issue.

PR #775, opened 2023-01-20
Fixes #35975 - validate starts before value constantly in the job wizard

MariaAga:
Users could select a value in the schedule step, go to the review step, wait until the "startsbefore" time is in the past and submit the form.
this solves 2 issues:
1. check every 5 secs that "startsbefore" is still in the future, and disable the submit button if its not, and adds this alert no matter what step the user is in.
Screenshot from 2023-01-20 18-59-28

3. typo in the error toast that prevented users seeing an error if they submitted something invalid to the server 

had to switch the fake timer running in tests because of the added `setInterval` but test times don't look effected in my env.

### DIFF
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -36,6 +36,7 @@ import { useValidation } from './validation';
 import { useAutoFill } from './autofill';
 import { submit } from './submit';
 import { generateDefaultDescription } from './JobWizardHelpers';
+import { StartsBeforeErrorAlert } from './StartsBeforeErrorAlert';
 import './JobWizard.scss';
 
 export const JobWizard = ({ rerunData }) => {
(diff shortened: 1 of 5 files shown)


adamruzicka:
Needs a rebase

MariaAga:
rebased

ofedoren:
Just a question: why would we check the time each 5 seconds instead of checking just once on submit?..

MariaAga:
I think that the submit button should only be active when users can submit, and if its active and gives an error on click it means something is wrong on the foreman side and not on the user side.

ofedoren:
What if the check is being done only at the confirmation (last) step? The button will be disabled (even if other fields are OK), the check is started and if everything OK the button becomes enabled. And just in case a user can leave the form in this state for a while, the check becomes automated (5 second for example).

ofedoren:
Okay then :/ But I won't block on this anyway, that were just some thoughts.

MariaAga:
Thanks for sharing, Im not sure its the best solution but working with times is weird

adamruzicka (approved):

PR #776, opened 2023-01-25
Fixes #35910 - Submit in legacy job form redirects wrong

MariaAga:
Draft until search is fixed in forms, as until then users cant really submit jobs from the legacy form

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -40,6 +40,16 @@ class JobInvocationsController < ApplicationController
     render :action => 'new'
   end
 
+  def legacy_create
+    @composer = prepare_composer
+    if @composer.trigger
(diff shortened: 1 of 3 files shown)


adamruzicka:
> until search is fixed in forms

Where is that happening?

MariaAga:
https://github.com/theforeman/foreman/pull/9593

ianballou:
I tested this and now a job invocation is sent along. I am experiencing the search bug, but perhaps my machine just needs an update. Once I added a bookmark I saw a job invocation made for no hosts, I suppose that at least verifies this particular bug is fixed :)

ianballou (approved):
After an `npm i` this is working great now! Tested with the "Manage Errata" workflow.

MariaAga:
@adamruzicka can we merge this?

PR #777, opened 2023-01-25
Fixes #35995 - Add time to pickup to apidoc and api

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka, LGTM and tested with https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/44.

PR #781, opened 2023-02-03
Fixes #36040 - Mention available providers in explain rake

adamruzicka:
```shell
# rake foreman_remote_execution:explain_proxy_selection HOST=foo
-----B<-----SNIP-----B<----
rake aborted!
Environment variable PROVIDER has to be set to one of SSH, script, Ansible. Note provider aliases SSH = script
/home/aruzicka/vcs/foreman/foreman-remote-execution/lib/foreman_remote_execution/tasks/explain_proxy_selection.rake:27:in `block (2 levels) in <main>'
/home/aruzicka/.distrobox/rh/.asdf/installs/ruby/2.7.6/bin/bundle:23:in `load'
/home/aruzicka/.distrobox/rh/.asdf/installs/ruby/2.7.6/bin/bundle:23:in `<main>'
Tasks: TOP => foreman_remote_execution:explain_proxy_selection
(See full trace by running task with --trace)
```


ofedoren (approved):
Thanks, @adamruzicka, LGTM and works as expected.

The only small question: do we want to support case insensitive (`Script == script`) provider names as a user input?

adamruzicka:
Yeah, that's a nice qol improvement

ofedoren (approved):
Thanks, @adamruzicka, once more.

Although, it now may lead to some funny output, but I don't think this requires a fix:
ScreenShot-1684413670658

One small question again: is this common to use `raise` for halting commands that are meant to be used by not developers? I mean, I don't have anything against that, but the actual help message is bloated with traceback, which is useless for the end user.

adamruzicka:
I could swap it for something like `$stderr.puts(msg); exit(1)` I guess

ofedoren:
Well, I see that we already use here `raise`s on multiple places, so I think we can bear with this until we would need to refactor it a bit by a request :)

PR #782, opened 2023-02-07
Fixes #36045 - Update pf4 to 2022.16+ release

MariaAga:
This + https://github.com/theforeman/foreman/pull/9593 should fix  https://projects.theforeman.org/issues/35911 (and search not being read when creating job templates (needed to select a category))

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,11 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
-    "@theforeman/builder": "^10.1.0",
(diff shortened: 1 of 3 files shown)


ekohl (commented):
Is this now broken with theforeman-vendor 10? If so, does it make sense to raise the minimum Foreman version to 3.6 to be explicit?

ekohl (approved):
Well yes, that's exactly what I meant and I should have checked the source before I made that comment. Thanks!

PR #784, opened 2023-02-13
Fixes #36089 - Job wizard default to no hosts selected

MariaAga:
~~using the magic of `1=0` like ruby does at    `if displayed_search_query.blank?  Host.where('1 = 0')`~~

Checking if any of the host selections are selected after a user clicks the host input, if not it will show an error
Screenshot from 2023-02-20 20-26-14

### DIFF
--- a/webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
+++ b/webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
@@ -13,11 +13,12 @@ export const buildHostQuery = (selected, search) => {
     hostGroups.length ? hostGroupsSearch : false,
     search.length ? search : false,
   ].filter(Boolean);
+
   if (queryParts.length === 0) {
-    return null;
+    return 'name=a AND name=b';
   }
   if (queryParts.length === 1) {
-    return queryParts[0];
+    return queryParts[0] || 'name=a AND name=b';
   }
-  return queryParts.map(p => `(${p})`).join(' or ');
+  return queryParts.map(p => `(${p})`).join(' or ') || 'name=a AND name=b';
 };
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -209,6 +209,11 @@ export const JobWizard = ({ rerunData }) => {
     !templateError &&
     !!jobTemplateID &&
     templateResponse.job_template;
+  const areHostsSelected =
+    selectedTargets.hosts.length > 0 ||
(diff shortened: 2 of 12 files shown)


webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
```diff
@@ -13,11 +13,12 @@ export const buildHostQuery = (selected, search) => {
     hostGroups.length ? hostGroupsSearch : false,
     search.length ? search : false,
   ].filter(Boolean);
+
   if (queryParts.length === 0) {
-    return null;
+    return '1=0';
```
adamruzicka, line 18: The `1=0` idea is a good one, but sadly it will not work in this context. `Host.where('1 = 0')` works, because that gets translated literally to SQL, but the search query goes through scoped search first. Running this should be giving you errors such as this

```
> Host.search_for('1=0')
ScopedSearch::QueryNotSupported: Field '1' not recognized for searching!
```

adamruzicka (changes requested):

webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
MariaAga: `/api/hosts?search="1=0"` gives me zero results. can also change it into `name = a AND name = b` as a hack Marek suggested

adamruzicka: Huh, that's odd
```shell
$ curl -u admin:changeme http://localhost:3000/api/hosts?search="1=0"
{
  "error": {"message":"Field '1' not recognized for searching!","class":"ScopedSearch::QueryNotSupported"}
}
```

MariaAga: Changed so we wont have this mismatch in error/no error

MariaAga:
Need to change the message + logic to allow 0 hosts selected but not allow no host selection

MariaAga:
Redone the pr to treat the hosts field as a required field

adamruzicka:
Once I fill in required inputs, it allows me to go to the review step and kick the job off even though no hosts are selected. Is that intentional?

MariaAga:
missed condition check in the last step, added it. review should be disabled if there are no hosts/search query

adamruzicka (approved):
Works well

adamruzicka:
Thank you @MariaAga !

PR #785, opened 2023-02-15
Update extracted strings

ekohl:
Pulling from Transifex is currently broken in my setup (transifex-client no longer works, I need to switch to transifex-cli) so this updates the extracted strings. The only diff is an added string, so pulling updates would have made little difference.


adamruzicka:
Merged manually as https://github.com/theforeman/foreman_remote_execution/commit/2e7dbd86cef0b07a276c9281668f234e09fd3ba6

adamruzicka:
Thank you @ekohl !

PR #786, opened 2023-02-16
Fixes #36108 - do not load non-existant charts.js

evgeni:
We used charts.js from Foreman, but that was dropped in Foreman 3.0 (https://github.com/theforeman/foreman/commit/69379b51a33344de93836950565ecbfae0c4df23) and results in 404 errors when trying to load the job invocation page.


ekohl (approved):

adamruzicka:
Thank you @evgeni & @ekohl !

PR #787, opened 2023-02-16
Fixes #36114 - Prevent accessing dynflow internals until ready

adamruzicka:
To prevent re-rendering of templates before showing them in the UI, we tried to pull them from the associated tasks. However, we only checked task's presence, not its state. This opened door to us trying to access data that is not there yet. The data should be there once the task goes gets planned completely.


ofedoren (approved):
Thanks, @adamruzicka, LGTM, I've tried to reproduce the issue with and without the fix, and it seems like the patch fixed the issue. The changes make sense, let's get this in.

PR #788, opened 2023-02-16
Fixes #36096 - Add provider inputs to job wizard

MariaAga:
Adding them in the api response as part of the advanced inputs so that the ui will treat them as template inputs, and split them on submit from the advanced inputs.
Tested that they were actually added with ` TemplateInvocation.last.provider_input_values.first.value`


MariaAga:
@adamruzicka are these normal ruby test failures?

MariaAga:
rebased and the test failures are gone

adamruzicka (approved):
Seems to work well, ACK

adamruzicka:
Thank you @MariaAga !

PR #790, opened 2023-03-01
Fixes #36149 - Emit event when job fails

m-bucher:
This is quite a rough draft, I am afraid.

Improvements welcome :wink:

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -7,6 +7,7 @@ module Actions
       include ::Actions::RemoteExecution::TemplateInvocationProgressLogging
 
       execution_plan_hooks.use :emit_feature_event, :on => :success
+      execution_plan_hooks.use :emit_feature_event_failure, :on => :failure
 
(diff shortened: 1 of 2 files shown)


adamruzicka:
When a job succeeds on a host, it emits an event, but only if it is being run for a specific feature. When it fails, it is emitted always. Is this mismatch intentional?

m-bucher:
@adamruzicka I tried it on my test-system and in the failure case the `job_features`-input was empty. Not sure why, but that is the reason for that code.
I am aware that this only creates a 'general' `Actions Remote Execution Run Host Job Failed`-event. Not sure if it is nicer to have one catch-all event for failing jobs or one additional failure-event per feature :thinking:

m-bucher:
It looks ok, but it does not seem to be backportable to foreman-3.1 (which I unfortunately need). In general your PR seems to be exactly what I need :heart_eyes: 
So I am willing to close this one in favour of your refactoring PR :bow:

m-bucher:
Closing in favor of https://github.com/theforeman/foreman_remote_execution/pull/791

PR #793, opened 2023-03-27
Fixes #36231 - Do not raise when checking exit status

adamruzicka:
With the template invocation events based implementation, checking exit status raised an exception if the task was still running (and the event was missing).


ofedoren (approved):
Thanks, @adamruzicka, LGTM.

PR #794, opened 2023-03-28
Fixes #35788 - add run now, skip to review buttons

MariaAga:
community discussion: https://community.theforeman.org/t/choose-a-design-for-new-remote-execution-job-invocation/32193
The poll results say that users want both buttons, so added a button that runs the job immediately with the entered/pre submitted fields, and a button that moves the user to the review step (if all the required fields are filled).
Added a spinner to make it clear that the "run now" button was clicked and is waiting for the server for a redirect.
Originally we said that the button text will be `run on x hosts` but host selection can be dynamic so the number shown wont always be correct.
Screenshot from 2023-03-28 11-49-14

### DIFF
--- a/webpack/JobWizard/Footer.js
+++ b/webpack/JobWizard/Footer.js
@@ -1,0 +1,104 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+import {
+  Button,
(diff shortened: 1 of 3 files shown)


adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #796, opened 2023-03-28
Fixes #36241 - job wizard - datepicker z index fix

MariaAga:
Fixing this:
Screenshot from 2023-03-28 12-57-23
Removed css was not needed as it was trying to solve the issue where users will expend the navigation (on small screen) and could not use the navigation, when the navigation is expanded it has `pf-m-expanded`

### DIFF
--- a/webpack/JobWizard/JobWizard.scss
+++ b/webpack/JobWizard/JobWizard.scss
@@ -6,20 +6,15 @@
 
   .pf-c-wizard__nav.pf-m-expanded {
     z-index: calc(
-      var(--pf-c-wizard__footer--ZIndex) + 2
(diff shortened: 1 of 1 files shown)


adamruzicka:
I've run into two cases where this can misbehave

Sometimes the calendar popup would open to the left and it would get partially obscured by the nav
image

On even smaller screens it would open up and then it would be partially obscured by the wizard step bar
image

MariaAga:
Fixed the second one, the first issue is happening on resizing, and popover act weird with side nav also in pf so I dont have a fix for that, example: https://u1gjys.csb.app/

adamruzicka (approved):
Works well

PR #799, opened 2023-03-29
Fixes #36246 - change org/location alert to be inline

MariaAga:
Requested by Maria from UXD
difference example (the second alert is inline)
Screenshot from 2023-03-29 13-25-51

### DIFF
--- a/webpack/JobWizard/JobWizardPageRerun.js
+++ b/webpack/JobWizard/JobWizardPageRerun.js
@@ -86,6 +86,7 @@ const JobWizardPageRerun = ({
           <React.Fragment>
             {jobOrganization?.id !== currentOrganization?.id && (
               <Alert
+                isInline
                 className="job-wizard-alert"
                 variant="warning"
                 title={sprintf(
@@ -99,6 +100,7 @@ const JobWizardPageRerun = ({
             )}
             {jobLocation?.id !== currentLocation?.id && (
               <Alert
+                isInline
                 className="job-wizard-alert"
                 variant="warning"
                 title={sprintf(


adamruzicka:
Could we change the contents of the banner in case both the organization and location are different so we don't end up with two banners like we currently do? @marisvirik opinions?

image

MariSvirik:
Feel free to change the content of the inline alert and combine those 2.
But PF4 recognises also an inline alert group - which is what we have today.


Job's organization is an organization defined in the job template, correct?
And "current" is the one selected in the masthead, right?
Will this job use org/location and available hosts defined by the job template or the currently selected org/location? That may not be clear here. Should we also add a way to change this?

Screenshot 2023-04-06 at 15 13 44

I would ping Adam Lazik from docs to fix alert phrasing (Do you know his nick?)

adamruzicka:
> But PF4 recognises also an inline alert group

If you don't see anything wrong with it then I'm fine with it. It just felt like we're repeating ourselves quite a bit.

> Job's organization is an organization defined in the job template, correct?

Not really, job templates can have organizations and locations assigned, but that only controls in what contexts the job template is visible.

Job's organization is the organization that is current when the user submits the job. We store this information if the user starts a job, changes org/loc and tries to rerun the original job. Because they are trying to "run the same job", but in a different context, it might run on different hosts which might be unexpected.

> Will this job use org/location and available hosts defined by the job template or the currently selected org/location? That may not be clear here. Should we also add a way to change this?

The currently selected ones. The ones from the template only control if the template is visible and I'd probably leave it at that

>  And "current" is the one selected in the masthead, right?

Yes

> I would ping Adam Lazik from docs to fix alert phrasing (Do you know his nick?)

CC @adamlazik1

adamlazik1:
Adding articles and changing may -> can. Also, it would be good to mention from what are the hosts different. I see **than before** used in the previous iteration, which is honestly better than nothing, but if possible, feel free to provide even more specific comparison.

**Current organization and location differ from the ones defined by the selected job template**
This can result in job running on different hosts than before.

MariaAga:
The "before" refers to the original job. so a user will run a job, and then after it runs they want to run it again (from the job details page), but if the user have a different location/org selected this message will show up.

adamlazik1:
Thanks for the explanation! On a second thought, **than before** is probably sufficient.

MariaAga:
I just noticed that I didnt parse the string correctly, so the message doesnt show the names.
So these are the messages, and I think they should be separate since combining the titles is a bit long
Screenshot from 2023-04-13 14-45-33

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #800, opened 2023-03-29
Fixes #36247 - job wizard host chips improvement

MariaAga:
- visual improvement
- add close all chips per chip group x button
- added translation to the `x more` button here: collapsedText

before:
image
after:
Screenshot from 2023-03-29 15-45-59


adamruzicka (approved):
Works like a charm

PR #801, opened 2023-03-29
Fixes #36248 - Show host preview modal in review step

MariaAga:
No button if number of hosts is less than 3
removed EllipsisWithTooltip as I think it looks better for long host names and it can create issues if it receives a button 
Screenshot from 2023-03-29 17-24-42
Screenshot from 2023-03-29 17-24-30


adamruzicka:
Could we restyle it a little bit so it is a little bit more compact?

Preview in the new form
image


Preview in the old form
image

MariaAga:
yup, added `isinline` to the host buttons 
Screenshot from 2023-04-11 11-27-58

adamruzicka (approved):
Looks good

PR #804, opened 2023-04-04
Fixes #36275 - job wizard Jump to step from review

MariaAga:
In the job wizard review step, users would like to click on the field title to jump to the step that edits it.
so clicking on `job temple` will jump to the template selection step, `Recurrence` will jump to one of the schedule steps...
Screenshot from 2023-04-04 14-07-15


adamruzicka (approved):
Very nice

PR #805, opened 2023-04-04
Fixes #36253 - delete 36096 leftovers in job_invocation

MariaAga:
sorry :upside_down_face:
removes debugging leftovers from https://github.com/theforeman/foreman_remote_execution/pull/788

### DIFF
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -68,9 +68,6 @@ class UiJobWizardController < ApplicationController
     job_organization = Taxonomy.find_by(id: job.task.input[:current_organization_id])
     job_location = Taxonomy.find_by(id: job.task.input[:current_location_id])
     render :json => {
-      :provider_input_values => composer[:template_invocations][0][:provider_input_values],
-      :provider_input_values1 => job[:provider_input_values],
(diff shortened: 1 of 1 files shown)


app/controllers/ui_job_wizard_controller.rb
```diff
@@ -68,9 +68,6 @@ def job_invocation
     job_organization = Taxonomy.find_by(id: job.task.input[:current_organization_id])
     job_location = Taxonomy.find_by(id: job.task.input[:current_location_id])
     render :json => {
-      :provider_input_values => composer[:template_invocations][0][:provider_input_values],
```
adamruzicka, line 71: Don't we need at least one copy of the provider input values? Or are they bundled within the regular inputs?

MariaAga: was added here: https://github.com/theforeman/foreman_remote_execution/pull/788/files#diff-494fb439c4dcbef79b8470f33ba943aa649b2d8e698a9bd26c5353e9e33158f7R22

The keys that were deleted were never used in the js side, and js specified which keys it uses.

adamruzicka (approved):
Let's get rid of it

adamruzicka:
Thank you @MariaAga !

PR #806, opened 2023-04-17
Fixes #36308 - Remove * for non required fields

girijaasoni:
This pr is in reference to foreman pr: https://github.com/theforeman/foreman/pull/9686

### DIFF
--- a/webpack/react_app/components/RegistrationExtension/RexPull.js
+++ b/webpack/react_app/components/RegistrationExtension/RexPull.js
@@ -26,7 +26,6 @@ const options = (value = '') => {
 const RexPull = ({ isLoading, onChange, pluginValues, configParams }) => (
   <FormGroup
     label={__('REX pull mode')}
-    isRequired
     labelIcon={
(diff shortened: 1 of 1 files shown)


girijaasoni:
@MariaAga could you please review it

MariaAga (approved):
Works as expected

Ron-Lavi (approved):

adamruzicka (commented):
Should this also drop isRequired from "REX pull mode", even though the issue doesn't state that explicitly?

stejskalleos (changes requested):
`FormSelect` still has `isRequired`

girijaasoni:
> `FormSelect` still has `isRequired`

updated

Ron-Lavi (approved):
Nice catch @adamruzicka, funny that it worked well even without the other `isRequired`
looks good to me now, thanks @girijaasoni !

adamruzicka:
Thank you @girijaasoni !

PR #807, opened 2023-05-05
Fixes #36363 - Update transifex config

adamruzicka:
and extract strings


.tx/config
```diff
@@ -1,8 +1,10 @@
 [main]
 host = https://www.transifex.com
 
-[foreman.foreman_remote_execution]
-file_filter = locale/<lang>/foreman_remote_execution.edit.po
```
ofedoren, line 5: Does this mean that `.edit` files are no longer needed? Or what's behind this change?

adamruzicka: I just copied what was done (and accepted) in here https://github.com/theforeman/foreman_leapp/pull/124 . What were the .edit files even for?

ofedoren: Have no idea, I think it's somehow used in gettext gem? At least in hammer I didn't dare to ignore those, even though I'm not sure they are really useful...

adamruzicka: You were right, thank you for pointing this out

PR #808, opened 2023-05-10
Fixes #36374 - Wizard should accept search='' in url

MariaAga:
Fixes the bug:
User selects all hosts (not only all in page) in hosts page, and clicks on schedule a job
expected: search to be `name != ''` like in the old form

### DIFF
--- a/webpack/JobWizard/autofill.js
+++ b/webpack/JobWizard/autofill.js
@@ -48,8 +48,11 @@ export const useAutoFill = ({
           })
         );
       }
-      if (search && !hostIds?.length) {
-        setHostsSearchQuery(search);
+      if ((search || search === '') && !hostIds?.length) {
+        // replace an empty string search with a dummy search query to match all hosts
+        // but only if search query was entered (based on presence of :search parameter)
+        const hostSearch = search === '' ? "name != ''" : search;
+        setHostsSearchQuery(hostSearch);
       }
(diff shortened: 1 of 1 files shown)


webpack/JobWizard/autofill.js
```diff
@@ -48,8 +48,11 @@ export const useAutoFill = ({
           })
         );
       }
-      if (search && !hostIds?.length) {
-        setHostsSearchQuery(search);
+      if ((search || search === '') && !hostIds?.length) {
+        // replace an empty string search with a dummy search query to match all hosts
+        // but only if search query was entered (based on presence of :search parameter)
+        const hostSearch = search === '' ? "name != ''" : search;
```
adamruzicka, line 54: Could we somehow not search at all instead of searching using a search query that matches everything?

MariaAga: No because we dont want to go back to on default all hosts selected I assume?

adamruzicka: Right, nevermind then

adamruzicka:
Thank you @MariaAga !

adamruzicka:
Not sure where this went, but it disappeared from master for a while. Put it back as https://github.com/theforeman/foreman_remote_execution/commit/c3ebbfb1749c758b360f126ad3e825a3f70c8437

PR #818, opened 2023-06-26
Fixes #37041 - Change link to docs.theforeman.org

maximiliankolb:
related to https://github.com/theforeman/theforeman.org/pull/2089

### DIFF
--- a/app/views/job_invocations/welcome.html.erb
+++ b/app/views/job_invocations/welcome.html.erb
@@ -7,7 +7,7 @@
         <%= _("Foreman can run arbitrary commands on remote hosts using different providers, such as SSH or Ansible. Communication goes through the Smart Proxy so Foreman does not have to have direct access to the target hosts and can scale to control many hosts.") %></br>
     </p>
     <p><%= link_to _('Learn more about this in the documentation.'),
-           documentation_url('1.ForemanRemoteExecution1.3Manual', :root_url => 'https://www.theforeman.org/plugins/foreman_remote_execution/1.3/index.html#'), :rel => 'external' %></p>
+           external_link_path(type: 'docs', section: 'Managing_Hosts', chapter: 'Configuring_and_Setting_Up_Remote_Jobs_managing-hosts'), :rel => 'external' %></p>
     <div class="blank-slate-pf-main-action">
         <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_("Run Job"), { :action => :create }, { :class => "btn btn-primary btn-lg" }) : nil %>
     </div>


app/views/job_invocations/welcome.html.erb
```diff
@@ -7,7 +7,7 @@
         <%= _("Foreman can run arbitrary commands on remote hosts using different providers, such as SSH or Ansible. Communication goes through the Smart Proxy so Foreman does not have to have direct access to the target hosts and can scale to control many hosts.") %></br>
     </p>
     <p><%= link_to _('Learn more about this in the documentation.'),
-           documentation_url('1.ForemanRemoteExecution1.3Manual', :root_url => 'https://www.theforeman.org/plugins/foreman_remote_execution/1.3/index.html#'), :rel => 'external' %></p>
+           documentation_url('', :root_url => 'https://www.theforeman.org/plugins/foreman_remote_execution/10.0.3/index.html#'), :rel => 'external' %></p>
```
adamruzicka, line 10: https://www.theforeman.org/plugins/foreman_remote_execution/10.0.3/index.html points to nowhere, that page does not exist.

maximiliankolb: Hey Adam, thanks for taking a look at it. My idea was to _not_ point to any subheading if possible, but instead have a small page referring users to docs.theforeman.org. The URL will be live _after_ the PR in theforeman.org has been merged.

This is IMHO just a temporary workaround until _all_ content from Foreman Manual has been migrated to Foreman Documentation _and_ we have a function to add Documentation buttons in Foreman Web UI and all Foreman plug-ins.

Do you think this is a reasonable approach?

ekohl: We have this helper:
https://github.com/theforeman/foreman_remote_execution/blob/2b10ad921c07392004e305d8f7bddc3fa6ea141e/app/helpers/remote_execution_helper.rb#L205-L209

I think it would be preferable to consistently use the same logic. Perhaps we need a separate helper for `root_url`.

It also implies those help buttons are likely to be broken.

ekohl: Oh, and looking further. A more correct usage of the documentation infrastructure is to use it like foreman_discovery:
https://github.com/theforeman/foreman_discovery/blob/417a7550a1b12da84c4787f8bf9ecf7eebc17ea3/app/views/discovered_hosts/welcome.html.erb#L13

So `external_link_path(type: 'plugin_manual', name: 'foreman_discovery', version: discovery_doc_version, section: '#4.3Automaticprovisioning')` (or `external_link_url`).

Or in this case:
```suggestion
           external_link_path(type: 'plugin_manual', name: 'foreman_remote_execution', version: '1.3', section: '1.ForemanRemoteExecution1.3Manual'),  :rel => 'external' %></p>
```

And from there update the rest as well.

I've also opened https://github.com/theforeman/foreman/pull/9756 to be able to link to the new docs.

maximiliankolb:
Updated the PR to point to version "10" as applied in https://github.com/theforeman/theforeman.org/pull/2089. I am also OK if we decide to close this PR because we wait until there's a generic way to link to docs.theforeman.org like https://github.com/theforeman/foreman/pull/9756

adamruzicka:
https://github.com/theforeman/foreman/pull/9756 was merged. @maximiliankolb could you please change this PR to use the new things and link to docs.tfm.org?

maximiliankolb:
build error from yesterday: `Foreman::PluginRequirementError: ERF72-4964 [Foreman::PluginRequirementError]: foreman_remote_execution plugin requires Foreman >= 3.10 but current is 3.9.0`. I assume this is unrelated to my change. :crossed_fingers: 

cc @adamruzicka My change is ready for review.

adamruzicka:
Could you please rebase now that https://github.com/theforeman/foreman_remote_execution/pull/861 got it? Hopefully it will make packit turn green

maximiliankolb:
Thanks Adam; rebased to master.

adamruzicka (approved):

adamruzicka:
Thank you @maximiliankolb !

ekohl:
@adamruzicka will this break branding?

adamruzicka:
I must admit I don't know, but I have it written down in my notes to look into it and fix the possible fallout.

ekohl:
https://github.com/RedHatSatellite/foreman_theme_satellite/pull/32 provides a mechanism to brand these.

PR #819, opened 2023-06-27
Consistent version in documentation URLs

ekohl:
This now uses the plugin manual helper to redirect users. It consistently uses the same version as the REX documentation buttons.

Ideally it would also use the plugin_manual type in documentation_url, but that doesn't work in Foreman itself.

I am a bit worried that it now breaks all links, since https://theforeman.org/plugins/foreman_remote_execution/ only has versions 1.7, 1.3, and some 0.x versions. Given we're already up to 9.1 this isn't going to work for anything. Because of that I think it may be safe to assume nobody clicks the documentation buttons.

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -203,8 +203,7 @@ module RemoteExecutionHelper
   end
 
   def documentation_button_rex(section = '')
-    url = 'http://theforeman.org/plugins/foreman_remote_execution/' +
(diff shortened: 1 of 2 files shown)


adamruzicka:
Does this replace #818 ?

ekohl:
Trivial rebase done.

ekohl:
Closing in favor of https://github.com/theforeman/foreman_remote_execution/pull/820.

PR #821, opened 2023-07-03
Fixes #36564 - Add eslint rule to alert missing ouia-ids

MariaAga:
We keep forgetting adding ouia-ids to react component, this is a rule to alert us in the pr that we forgot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "license": "GPL-3.0",
   "scripts": {
     "lint": "tfm-lint --plugin -d /webpack",
+    "lint:custom": "eslint ./webpack",
     "test": "tfm-test --plugin",
     "test:watch": "tfm-test --plugin --watchAll",
     "test:current": "tfm-test --plugin --watch",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "license": "GPL-3.0",
   "scripts": {
     "lint": "tfm-lint --plugin -d /webpack",
+    "lint:custom": "eslint ./webpack",
```
ofedoren, line 7: Just a nitpick, but why we don't call it `lint:foreman` since it seems we want to use Foreman rules?

ofedoren (approved):
Thanks, @MariaAga, seems reasonable.

package.json
MariaAga: lint:custom will run anything that is in the `plugins` list. which happens to be now the foreman rules.
In foreman core the same command was called spelling, and I had to rename it since now it checked spelling and ouia-ids.
So I want to avoid re-naming things later

ofedoren (approved):
Okay then, I'm fine with that :)

PR #822, opened 2023-07-17
Fixes #36600 - Regex should only match the first ERB comment, not all

jeremylenz:
The `JobTemplate#import_custom_data` method removes metadata from a job template's contents. But the regex actually causes any ERB comment to be removed, not just the metadata at the top of the file:

image

This commit updates the regex so that only the first ERB comment is matched:

image

### DIFF
--- a/app/models/job_template.rb
+++ b/app/models/job_template.rb
@@ -192,7 +192,7 @@ class JobTemplate < ::Template
     self.default = options[:default] unless options[:default].nil?
 
     # job templates have too long metadata, we remove them on parsing until it's stored in separate attribute
-    self.template = self.template.gsub(/<%\#.+?.-?%>\n?/m, '').strip
+    self.template = self.template.sub(/<%\#.+?.-?%>\n?/m, '').strip
   end
 
   def default_input_values(ignore_keys)
--- a/test/unit/job_template_test.rb
+++ b/test/unit/job_template_test.rb
@@ -94,6 +94,8 @@ class JobTemplateTest < ActiveSupport::TestCase
       %>
 
       service <%= input("service_name") %> restart
+
+      <%# test comment %>
       END_TEMPLATE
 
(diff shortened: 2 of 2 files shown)


jeremylenz:
I ran into this before I changed https://github.com/Katello/katello/pull/10652/files to use regular Bash comments instead of ERB comments.

jeremylenz:
> Just a question: why don't we use here the same regexp we use in core (https://github.com/theforeman/foreman/blob/develop/app/models/template.rb#L64)?

I wasn't aware of that one. Maybe I should just change it to
```rb
self.template = self.template_without_metadata
```

@adamruzicka thoughts?

adamruzicka:
Hmm, part of me wants to say it was there for a reason, but I can't think of any. Yet another alternative could be keeping the same regexp and just using `sub` instead of `gsub`.

> Maybe I should just change it to

Let's go with that and we'll see what happens.

jeremylenz:
@adamruzicka updated :+1:

jeremylenz:
Seems that doesn't work:

```
Failure:
JobTemplateTest::importing a new template#test_0002_has a template [/home/runner/work/foreman_remote_execution/foreman_remote_execution/foreman_remote_execution/test/unit/job_template_test.rb:125]
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-"service <%= input(\"service_name\") %> restart"
+"<%# kind: job_template name: Service Restart job_category: Service Restart provider_type: SSH feature: remote_execution_feature_10 template_inputs: - name: service_name input_type: user required: true - name: verbose input_type: user %> service <%= input(\"service_name\") %> restart"
```

jeremylenz:
@adamruzicka Changed as per your `sub` suggestion. Also updated the relevant test to assert that the comment is not removed.

adamruzicka:
Thank you @jeremylenz !

PR #825, opened 2023-07-18
Fixes #36565 - Use correct permissions for job category lookup

adamruzicka:
This previously defaulted to a combination of action_permission and controller name.


ofedoren (approved):
Hmm, I think I've already seen this fix somewhere :D

PR #826, opened 2023-07-19
Fixes #36605 - Job category is not set in job template edit

MariaAga:
Users cant edit and save Job templates without re-setting the job category. Job category should be filled in on edit


adamruzicka (commented):
Works nicely, ACK

adamruzicka:
Thank you @MariaAga !

PR #830, opened 2023-08-01
Fixes #36627 - Discard irrelevant inputs on template change

adamruzicka:
Previously we kept the pre-filled values from the previous template and then sent them to the backend together with the inputs for the new template. This prevented the job from being started as the backend didn't know what to do with the extra inputs.

### DIFF
--- a/webpack/JobWizard/JobWizard.js
+++ b/webpack/JobWizard/JobWizard.js
@@ -169,8 +169,23 @@ export const JobWizard = ({ rerunData }) => {
                   job_template: { name, description_format },
                 },
               }) => {
+                const allowedInputs = template_inputs
+                  .map(({ name: _name }) => _name)
+                  .concat(
+                    advanced_template_inputs.map(({ name: _name }) => _name)
+                  );
+                const prune = inputs =>
+                  allowedInputs.reduce(
+                    (acc, key) =>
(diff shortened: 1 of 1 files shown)


webpack/JobWizard/JobWizard.js
```diff
@@ -169,8 +169,23 @@ export const JobWizard = ({ rerunData }) => {
                   job_template: { name, description_format },
                 },
               }) => {
+                const allowedInputs = template_inputs
+                  .map(({ inputName }) => inputName)
+                  .concat(
+                    advanced_template_inputs.map(({ inputName }) => inputName)
```
MariaAga, line 175: `template_inputs` and `advanced_template_inputs` dont have a key named `inputName`
I assume you meant to do here something like this? (name is already defined so need to use _name)
```suggestion
                  .map(({ name: _name }) => _name)
                  .concat(
                    advanced_template_inputs.map(({ name: _name }) => _name)
```

adamruzicka: Yes, of course. I originally used something like `({ name } => name`, got yelled at by the linter for shadowing an outer variable so I just renamed things without thinking, sorry about that

MariaAga (approved):

adamruzicka:
Thank you @MariaAga !

PR #833, opened 2023-08-09
Pin @adobe/css-tools to ~4.2.0

adamruzicka:


ekohl (approved):
It may make sense to note somewhere (in the commit message) that this is because of NodeJS 12.

adamruzicka:
Reworded

PR #835, opened 2023-08-28
Fixes #36704 - Reject rerun inputs with null values

adamruzicka:


webpack/JobWizard/autofill.js
```diff
@@ -76,7 +76,7 @@ export const useAutoFill = ({
             if (typeof rest[key] === 'string') {
               setTemplateValues(prev => ({ ...prev, [input]: rest[key] }));
               setAdvancedValues(prev => ({ ...prev, [input]: rest[key] }));
-            } else {
+            } else if (rest[key].value !== null) {
```
MariaAga, line 79: is only `null` values are invalid? or should it be `else if (rest[key].value)` ?

adamruzicka: I wanted to be as specific as possible, without catching more general falsey values. Not sure if we can get any apart from `false`

MariaAga:
code LGTM, but I cant test it today

pondrejk:
Verified using stream buid + patch. I applied errata from the hosts page, then re-run the respective rex job. No validation errors occurred.

pondrejk (approved):

adamruzicka:
Thank you @MariaAga & @pondrejk !

PR #836, opened 2023-09-15
Fixes #36751 - check permissions for schedule job button

MariaAga:
Took the permissions from the old host page, schedule job button: https://github.com/theforeman/foreman_remote_execution/blob/0b75d16284b5abc8d121d4b3cf665d0122540f84/app/helpers/hosts_extensions_helper.rb#L56

### DIFF
--- a/app/controllers/api/v2/remote_execution_features_controller.rb
+++ b/app/controllers/api/v2/remote_execution_features_controller.rb
@@ -29,6 +29,14 @@ module Api
         process_response @remote_execution_feature.update(remote_execution_feature_params)
       end
 
+      api :GET, '/api/hosts/:id/available_remote_execution_features', N_('List available remote execution features for a host')
+      param :id, :identifier, :required => true
+      def available_remote_execution_features
+        host = Host.find(params[:id])
+        @remote_execution_features = resource_scope
+        @permissions = {:can_run_job => (authorized_for(controller: :job_invocations, action: :create) && (!host.infrastructure_host? || User.current.can?(:execute_jobs_on_infrastructure_hosts))) }
+      end
+
       private
 
       def parent_scope
--- a/app/views/api/v2/remote_execution_features/available_remote_execution_features.json.rabl
+++ b/app/views/api/v2/remote_execution_features/available_remote_execution_features.json.rabl
@@ -1,0 +1,6 @@
+node :permissions do
+  @permissions
+end
+child @remote_execution_features do
+  extends 'api/v2/remote_execution_features/main'
+end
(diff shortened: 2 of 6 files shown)


app/controllers/api/v2/remote_execution_features_controller.rb
```diff
@@ -29,6 +29,14 @@ def update
         process_response @remote_execution_feature.update(remote_execution_feature_params)
       end
 
+      api :GET, '/api/hosts/:id/available_remote_execution_features', N_('List available remote execution features for a host')
+      param :id, :identifier, :required => true
+      def available_remote_execution_features
+        host = Host.find(params[:id])
+        @remote_execution_features = resource_scope
+        @permissions = {:can_run_job => (authorized_for(controller: :job_invocations, action: :create) && (!host.infrastructure_host? || User.current.can?(:execute_jobs_on_infrastructure_hosts))) }
```
adamruzicka, line 37: I was thinking about not exposing the permission directly and just returning an empty array in case the user doesn't have the necessary permissions, but this works too

MariaAga: It doesn't catch the case when for some reason the user doesnt have any REX features but has permissions to run jobs, if thats possible

adamruzicka: It shouldn't happen, but it is theoretically possible. Nevermind then

adamruzicka:
Does the trick for me. @pnovotny could you please take this for a spin when you have time?

pnovotny:
@adamruzicka Hi, sure, I'll take a look.

pnovotny:
Good news: a user without permissions to schedule jobs, i.e., `View hosts` role, cannot see the `Schedule a job` button in the host detail at all.

Bad news: a user with permissions can see the button, however the dropdown toggle with REX options doesn't work.
This seems to be because of incorrect API call to `.../api/v2/hosts/undefined/available_remote_execution_features`, where the host ID value is `undefined`.
host-detail-rex-features-error

MariaAga:
Fixed, I used `features` instead of `remote_execution_features` accidentally.
the `undefined` host id can happen sometimes when the host id is still not set when the component loaded but another api call will be made once the host id is defined

pnovotny:
LGTM. 
The `Schedule a job` dropdown now works correctly. 
The rest is fine too:


> Good news: a user without permissions to schedule jobs, i.e., `View hosts` role, cannot see the `Schedule a job` button in the host detail at all.

adamruzicka (approved):
Looks good

adamruzicka:
Thank you @MariaAga & @pnovotny !

PR #837, opened 2023-09-26
Fixes #36780 - Hide multiple titles on run job page

kmalyjur:
Run job wizard had multiple titles because of the usage of `beforeToolbarComponent` in the PageLayout. It is dependent on https://github.com/theforeman/foreman/pull/9830.

image

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -47,7 +47,7 @@ module ForemanRemoteExecution
 
     initializer 'foreman_remote_execution.register_plugin', before: :finisher_hook do |_app|
       Foreman::Plugin.register :foreman_remote_execution do
-        requires_foreman '>= 3.8'
+        requires_foreman '>= 3.9'
         register_global_js_file 'global'
         register_gettext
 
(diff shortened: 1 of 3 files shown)


adamruzicka:
Does the trick for me, but the commit message is missing a space between the issue number and the hyphen

adamruzicka:
And one more thing, could you please bump required foreman version to >= 3.9 in engine.rb since we're now relying on a change that will only be available in foreman 3.9 and newer?

pnovotny:
The UI looks good now:
run-job-fixed

The foreman dependency still shows >= 3.8 though:
```
# rpm -qR rubygem-foreman_remote_execution | grep ^foreman
foreman >= 3.8
```
(I did update of the rubygem-foreman_remote_execution RPMs from Packit as well as clean install)

kmalyjur:
I also changed JobWizardPageRerun.js similarly now.

> The foreman dependency still shows >= 3.8 though:

@adamruzicka Don't you know how to fix that?

adamruzicka:
That will need to be fixed in the rpm spec[1]. We can't do anything about it right now, right here. It will have to wait until next release goes out and then adjust it in the packaging repo, but that's part of the release process.

[1] - https://github.com/theforeman/foreman-packaging/blob/rpm/3.8/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec#L4

adamruzicka (approved):
Looks good to me, but let's keep it around for a bit until other things which don't require foreman 3.9 get in

adamruzicka:
Thank you and sorry for the wait, well done @kmalyjur !

PR #838, opened 2023-10-02
Fixes #36785 - Show permission error page for job wizard

kmalyjur:
An error page is shown if a user without necessary permissions (create_job_invocations) goes to /job_invocations/new.

image

I couldn't decide on what's the best way of implementing it, so feel free to tell me if it would be better to create a new API or make the request (which is in /JobWizard/steps/CategoryAndTemplate/index.js) directly in the /JobWizard/index.js file.

The change of error alerts, if a user doesn't have other permissions needed for submitting the job wizard, will be in another PR.


kmalyjur:
This issue will be resolved using an alternative approach.

PR #839, opened 2023-10-11
Fixes #36823 - Create job invocations detail page with main info

kmalyjur:
The commit creates a new job invocation detail page ([see discussion](https://community.theforeman.org/t/rfc-job-invocation-detail-page-redesign/33839)) which is experimental until finished - you can find it in Lab Features or see `/experimental/job_invocations_detail/${JOB_ID}`. It includes the main information.

image

_Update:_ There is a link on the old detail page to the new one if the experimental labs setting is on.

image

### DIFF
--- a/app/views/api/v2/job_invocations/base.json.rabl
+++ b/app/views/api/v2/job_invocations/base.json.rabl
@@ -3,7 +3,11 @@ object @job_invocation
 attributes :id, :description, :job_category, :targeting_id, :status, :start_at, :status_label, :ssh_user, :time_to_pickup
 
 node do |invocation|
+  pattern_template = invocation.pattern_template_invocations.first
   {
+    :template_id => pattern_template&.template_id,
+    :template_name => pattern_template&.template_name,
+    :effective_user => pattern_template&.effective_user,
     :succeeded => invocation_count(invocation, :output_key => :success_count),
     :failed    => invocation_count(invocation, :output_key => :failed_count),
     :pending   => invocation_count(invocation, :output_key => :pending_count),
@@ -14,4 +18,14 @@ end
 
 child :task => :dynflow_task do
   attributes :id, :state
+end
+
+if params.key?(:include_permissions)
+  node :permissions do |invocation|
+    authorizer = Authorizer.new(User.current)
+    edit_job_templates_permission = Permission.where(name: "edit_job_templates", resource_type: "JobTemplate").first
+    {
+      "edit_job_templates" => (edit_job_templates_permission && authorizer.can?("edit_job_templates", invocation, false)),
+    }
+  end
 end
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -244,6 +244,12 @@ module ForemanRemoteExecution
           parent: :monitor_menu,
           after: :audits
 
+        menu :labs_menu, :job_invocations_detail,
+          url_hash: { controller: :job_invocations, action: :show },
+          caption: N_('Job invocations detail'),
+          parent: :lab_features_menu,
+          url: '/experimental/job_invocations_detail/1'
+
         register_custom_status HostStatus::ExecutionStatus
         # add dashboard widget
         # widget 'foreman_remote_execution_widget', name: N_('Foreman plugin template widget'), sizex: 4, sizey: 1
--- a/webpack/JobInvocationDetail/JobInvocationOverview.js
+++ b/webpack/JobInvocationDetail/JobInvocationOverview.js
@@ -1,0 +1,104 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Button,
+  DescriptionList,
+  DescriptionListTerm,
+  DescriptionListGroup,
+  DescriptionListDescription,
+} from '@patternfly/react-core';
+import DefaultLoaderEmptyState from 'foremanReact/components/HostDetails/DetailsCard/DefaultLoaderEmptyState';
+import { translate as __, documentLocale } from 'foremanReact/common/I18n';
+
+const JobInvocationOverview = ({ data }) => {
+  const {
+    start_at: startAt,
+    ssh_user: sshUser,
+    template_id: templateId,
+    template_name: templateName,
+    effective_user: effectiveUser,
+    permissions,
+  } = data;
+  const canEditJobTemplates = permissions
+    ? permissions.edit_job_templates
+    : false;
+  const dateOptions = {
+    day: 'numeric',
+    month: 'short',
+    year: 'numeric',
+    hour: '2-digit',
+    minute: '2-digit',
+    hour12: false,
+    timeZoneName: 'short',
+  };
+  let formattedStartDate = __('Not yet');
+
+  if (startAt) {
+    // Ensures date string compatibility across browsers
+    const convertedDate = new Date(startAt.replace(/[-.]/g, '/'));
+    if (convertedDate.getTime() <= new Date().getTime()) {
+      formattedStartDate = convertedDate.toLocaleString(
+        documentLocale(),
+        dateOptions
+      );
+    }
+  }
+
+  return (
+    <DescriptionList
+      columnModifier={{
+        default: '2Col',
+      }}
+      isHorizontal
+      isCompact
+      isFluid
+      isAutoColumnWidths
+    >
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Effective user:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {effectiveUser || <DefaultLoaderEmptyState />}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Started at:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {formattedStartDate}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('SSH user:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {sshUser || <DefaultLoaderEmptyState />}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Template:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {templateName ? (
+            <Button
+              ouiaId="template-link"
+              variant="link"
+              component="a"
+              isInline
+              isDisabled={!canEditJobTemplates}
+              href={
+                templateId ? `/job_templates/${templateId}/edit` : undefined
+              }
+            >
+              {templateName}
+            </Button>
+          ) : (
+            <DefaultLoaderEmptyState />
+          )}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+    </DescriptionList>
+  );
+};
+
+JobInvocationOverview.propTypes = {
+  data: PropTypes.object.isRequired,
+};
+
+export default JobInvocationOverview;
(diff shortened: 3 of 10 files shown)


webpack/JobInvocationDetail/JobInvocationSelectors.js
```diff
@@ -0,0 +1,4 @@
+import { selectAPIResponse } from 'foremanReact/redux/API/APISelectors';
+
+export const selectItems = state =>
+  selectAPIResponse(state, 'JOB_INVOCATION_KEY') || [];
```
MariaAga, line 4: since items is used as an object `{ description, status_label: statusLabel, task } = items;`, the fallback should be an empty object `{}`

lib/foreman_remote_execution/engine.rb
```diff
@@ -232,6 +232,7 @@ class Engine < ::Rails::Engine
           caption: N_('Job templates'),
           parent: :hosts_menu,
           after: :provisioning_templates
+
```
MariaAga, line 235: extra space :)

webpack/JobInvocationDetail/JobInvocationSelectors.js
MariaAga: Actually, `selectAPIResponse` already gives a fallback, so no fallback needed.
`export const selectAPIResponse = (state, key) =>
  selectAPIByKey(state, key).response || {};`

webpack/JobInvocationDetail/index.js
```diff
@@ -0,0 +1,91 @@
...
+const JobInvocationDetailPage = ({
+  match: {
+    params: { id },
+  },
+}) => {
+  const dispatch = useDispatch();
+  const items = useSelector(selectItems);
+  const { description, status_label: statusLabel, task } = items;
+  const finished = statusLabel === 'failed' || statusLabel === 'succeeded';
+  const autoRefresh = task?.state === 'pending' || false;
```
MariaAga, line 25: please use constants for `failed, succeeded, pending`

webpack/JobInvocationDetail/index.js
```diff
@@ -0,0 +1,91 @@
...
+      get({
+        key: JOB_INVOCATION_KEY,
+        url,
+        handleError: () => {
+          dispatch(stopInterval(JOB_INVOCATION_KEY));
+        },
+      }),
+      1000
+    );
+
```
MariaAga, line 38: Please move to an actions file

webpack/JobInvocationDetail/JobInvocationOverview.js
```diff
@@ -0,0 +1,90 @@
...
+      isAutoColumnWidths
+    >
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Effective user:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {effectiveUser || <DefaultLoaderEmptyState />}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Run on:')}</DescriptionListTerm>
```
MariaAga, line 53: Probably should change the text according to if it already ran or if its going to run in the future, can be done in the next pr but would be nice to have some todo note

MariaAga: image

webpack/JobInvocationDetail/JobInvocationOverview.js
```diff
@@ -0,0 +1,95 @@
...
+    year: 'numeric',
+    hour: '2-digit',
+    minute: '2-digit',
+    hour12: false,
+    timeZoneName: 'short',
+  };
+  const dateConverted = new Date(startAt);
+  const dateLocaleFormatted = dateConverted
+    .toLocaleString(documentLocale(), dateOptions)
+    .replace(/(\d{4}) /, '$1, ');
```
MariaAga, line 34: why have this line? it adds a `,` after the year but in some locales it doesnt make sense to add.

webpack/JobInvocationDetail/JobInvocationOverview.js
```diff
@@ -0,0 +1,95 @@
...
+        <DescriptionListTerm>{__('SSH user:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {sshUser || <DefaultLoaderEmptyState />}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Template:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {templateName ? (
+            <Button
```
MariaAga, line 71: The button should be disabled if the user cant edit the template

MariaAga (changes requested):

webpack/JobInvocationDetail/JobInvocationOverview.js
kmalyjur: I wanted to do exactly what you said, but I didn't realize it may be strange in other locales. I'll just delete it

webpack/JobInvocationDetail/JobInvocationOverview.js
kmalyjur: Thank you, I've just made the updates

MariaAga (approved):
Thanks, LGTM!
@adamruzicka Can you take a look at the ruby parts?

lib/foreman_remote_execution/engine.rb
```diff
@@ -244,6 +244,12 @@ class Engine < ::Rails::Engine
           parent: :monitor_menu,
           after: :audits
 
+        menu :labs_menu, :job_invocations_detail,
+          url_hash: { controller: :job_invocations_detail, action: :index },
+          caption: N_('Job invocations detail'),
+          parent: :lab_features_menu,
+          url: '/experimental/job_invocations_detail/1'
```
adamruzicka, line 251: This is a bit unfortunate as it points at a particular resource which may or may not exist. How about we modify the old details page to include the link to the new one if the experimental page setting is enabled?

webpack/JobInvocationDetail/JobInvocationOverview.js
```diff
@@ -0,0 +1,101 @@
...
+        <DescriptionListDescription>
+          {effectiveUser || <DefaultLoaderEmptyState />}
+        </DescriptionListDescription>
+      </DescriptionListGroup>
+      <DescriptionListGroup>
+        <DescriptionListTerm>{__('Started at:')}</DescriptionListTerm>
+        <DescriptionListDescription>
+          {startAt && dateConverted <= dateCurrent
+            ? dateLocaleFormatted
+            : __('Not yet')}
```
adamruzicka, line 63: Not sure where this goes wrong, but this gives me "Not yet" for all the jobs I have on my instance, even though the time is included in the response and is in the past.

The api response for reference
```json
{
  "id": 1,
  "description": "Install errata whatever",
  "job_category": "Katello",
  "targeting_id": 1,
  "status": 1,
  "start_at": "2023-08-28 10:49:15 +0200",
  "status_label": "failed",
  "ssh_user": null,
  "time_to_pickup": null,
  "dynflow_task": {
    "id": "f0f660b8-2728-4eae-bdeb-3f75800ad647",
    "state": "stopped"
  },
  "template_id": 171,
  "template_name": "Install Errata - Katello Script Default",
  "effective_user": "root",
  "succeeded": 0,
  "failed": 1,
  "pending": 0,
  "total": 1,
  "missing": 0,
  "permissions": {
    "edit_job_templates": true
  },
  "targeting": {
    "bookmark_id": null,
    "search_query": "id ^ (1)",
    "targeting_type": "static_query",
    "user_id": 4,
    "randomized_ordering": null,
    "hosts": [
      {
        "name": "moses-lupton",
        "id": 1,
        "display_name": "moses-lupton"
      }
    ]
  },
  "task": {
    "id": "f0f660b8-2728-4eae-bdeb-3f75800ad647",
    "state": "stopped"
  },
  "template_invocations": [
    {
      "template_id": 171,
      "template_name": "Install Errata - Katello Script Default",
      "host_id": 1,
      "template_invocation_input_values": [
        {
          "template_input_name": "errata",
          "template_input_id": 62,
          "value": "whatever"
        }
      ]
    }
  ]
}
```

lib/foreman_remote_execution/engine.rb
```diff
@@ -244,6 +244,12 @@ class Engine < ::Rails::Engine
           parent: :monitor_menu,
           after: :audits
 
+        menu :labs_menu, :job_invocations_detail,
+          url_hash: { controller: :job_invocations_detail, action: :index },
```
adamruzicka, line 248: If an `url` is provided, then the `url_hash` is used only when checking permissions when deciding whether the menu item should be shown. Since there is no job_invocations_detail controller with index action, this won't show for a non-admin user even with full set of roles. Mapping this to permissions of an existing controller with established permissions should work
```suggestion
          url_hash: { controller: :job_invocations, action: :show },
```

webpack/JobInvocationDetail/JobInvocationOverview.js
adamruzicka: And on a relate note, shouldn't it say when the job is supposed to run instead of just saying "Not yet"?

app/views/api/v2/job_invocations/base.json.rabl
```diff
@@ -15,3 +19,11 @@ end
 child :task => :dynflow_task do
   attributes :id, :state
 end
+
+node :permissions do |invocation|
```
adamruzicka, line 23: Could we hide this behind a parameter? Similarly to what katello does https://github.com/Katello/katello/blob/master/app/views/katello/api/v2/subscriptions/permissions.rabl#L3 ?

lib/foreman_remote_execution/engine.rb
kmalyjur: I was planning to move it from the Lab features after some main functionality is done, but I like your solution.

webpack/JobInvocationDetail/JobInvocationOverview.js
kmalyjur: > shouldn't it say when the job is supposed to run instead of just saying "Not yet"?

As Maria mentioned above, the future "scheduled at" time is going to be there:
277340668-6926ef74-3760-4f3f-a73f-a4dd00b2f0f5

> This gives me "Not yet" for all the jobs I have

This is weird, it works fine for me :thinking: @MariaAga has this worked for you as it's supposed to? I can't figure out why it does that.

MariaAga: It works for me as expected. I tried to take Adams data and it still shows the correct time 
image

adamruzicka: Update, it works for me in chrome, but not in firefox-115.4.0esr. However it could also be a locale thing?

MariaAga: Looks like `new Date('2023-08-28 10:49:15 +0200')` works on chrome but not on firefox, so probably that :/

kmalyjur: @adamruzicka @MariaAga what about now?

adamruzicka: Works for me 👍

app/views/api/v2/job_invocations/base.json.rabl
```diff
@@ -3,7 +3,12 @@ object @job_invocation
 attributes :id, :description, :job_category, :targeting_id, :status, :start_at, :status_label, :ssh_user, :time_to_pickup
 
 node do |invocation|
+  pattern_template = invocation.pattern_template_invocations.first
+  Rails.logger.debug("Pattern Template ID: #{pattern_template&.template_id}")
```
adamruzicka, line 7: Do we need this?

adamruzicka (commented):
One last nitpick, apart from it lgtm

adamruzicka (approved):
Looks good

adamruzicka:

PR #840, opened 2023-10-16
Fixes #36837 - Template for convert2rhel analyze

stejskalleos:


stejskalleos:
@adamruzicka review please

stejskalleos:
Updated
* Check for RH OS family
* root check
* installing tool if not present
* if any errors or warnings are present, exit 1

How to install the tool: https://developers.redhat.com/articles/2023/04/11/how-use-convert2rhel-migrate-centos-rhel?source=sso

app/views/templates/script/convert2rhel_analyze.erb
```diff
@@ -0,0 +1,50 @@
...
+
+<% if input('Data telemetry') != "yes" -%>
+export CONVERT2RHEL_DISABLE_TELEMETRY=1
+<% end -%>
+
+/usr/bin/convert2rhel analyze -y
+
+errors=$(grep -c 'ERROR' '/var/log/convert2rhel/convert2rhel-pre-conversion.json')
+if [ "$errors" -gt 0 ]; then
+    echo "Error: $errors error(s) have been found."
```
stejskalleos, line 36: I just realized that the $errors will be always 1 or 0, as the `json` file is one line only.

app/views/templates/script/convert2rhel_analyze.erb
```diff
@@ -0,0 +1,50 @@
...
+    echo "Exiting ..."
+    exit 1
+fi
+
+skips=$(grep -c 'SKIP' '/var/log/convert2rhel/convert2rhel-pre-conversion.json')
+if [ "$skips" -gt 0 ]; then
+    echo "Warning: $skips check(s) have been skipped."
```
adamruzicka, line 43:
```suggestion
if ! grep -q SKIP /var/log/convert2rhel/convert2rhel-pre-conversion.json; then
    echo "Warning: Some check(s) have been skipped."
```

app/views/templates/script/convert2rhel_analyze.erb
adamruzicka, line 36:
```suggestion
if ! grep -q ERROR /var/log/convert2rhel/convert2rhel-pre-conversion.json; then
    echo "Error: Some error(s) have been found."
```

adamruzicka (commented):
A suggestion to make things a bit simpler if we can't get the counts anyway

app/views/templates/script/convert2rhel_analyze.erb
```diff
@@ -0,0 +1,49 @@
...
+    value_type: plain
+    default: 'yes'
+    hidden_value: false
+model: JobTemplate
+job_category: Convert 2 RHEL
+provider_type: script
+kind: job_template
+%>
+<% if @host.operatingsystem.family == 'Redhat' -%>
+if ! [ $(id -u) = 0 ]; then
```
adamruzicka, line 19: `=` in test does string-based comparison. I'd rather be explicit and use `-eq` to do numeric comparison

app/views/templates/script/convert2rhel_analyze.erb
```diff
@@ -0,0 +1,49 @@
...
+  yum install -y convert2rhel
+fi
+
+<% if input('Data telemetry') != "yes" -%>
+export CONVERT2RHEL_DISABLE_TELEMETRY=1
+<% end -%>
+
+/usr/bin/convert2rhel analyze -y
+
+errors=$(grep -c 'ERROR' '/var/log/convert2rhel/convert2rhel-pre-conversion.json')
```
adamruzicka, line 34: Why is this still here?

adamruzicka (changes requested):

adamruzicka (approved):

PR #841, opened 2023-10-20
Fixes #36851 - Rename migration file so it doesn't break migrations

jeremylenz:
The migration file `2021051713291621250977 AddHostProxyInvocations` always gets treated as the latest migration when you try to do `rails db:rollback`. This is not ideal.

I think it's because the file is oddly named; it has 8 more characters in the timestamp than every other migration file. This causes Rails to think it's a newer timestamp, because I guess the number is the biggest or something?

image

Anyway, I tested it and it seems like renaming it to be in line with the other files works fine.

### DIFF
--- a/db/migrate/20210517132916_add_host_proxy_invocations.rb
+++ b/db/migrate/20210517132916_add_host_proxy_invocations.rb
@@ -1,12 +1,12 @@
 class AddHostProxyInvocations < ActiveRecord::Migration[6.0]
   def change
     # rubocop:disable Rails/CreateTableWithTimestamps
-    create_table :host_proxy_invocations do |t|
+    create_table :host_proxy_invocations, if_not_exists: true do |t|
       t.references :host, :null => false
       t.references :smart_proxy, :null => false
     end
     # rubocop:enable Rails/CreateTableWithTimestamps
 
-    add_index :host_proxy_invocations, [:host_id, :smart_proxy_id], unique: true
+    add_index :host_proxy_invocations, [:host_id, :smart_proxy_id], unique: true, if_not_exists: true
   end
 end


adamruzicka:
If the long name variant is already applied and then the file gets renamed, rails then thinks the migration is not applied and tries to apply it again.

```
## Without this patch
# bundle exec rake db:migrate:status | grep 'Add host'
-----B<-----SNIP-----B<-----
   up     2021051713291621250977  Add host proxy invocations

## With this patch
# bundle exec rake db:migrate:status | grep 'Add host'
-----B<-----SNIP-----B<-----
  down    20210517132916  Add host proxy invocations

# bundle exec rake db:migrate
== 20210517132916 AddHostProxyInvocations: migrating ==========================
-- create_table(:host_proxy_invocations, {})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::DuplicateTable: ERROR:  relation "host_proxy_invocations" already exists
/home/aruzicka/foreman_remote_execution/db/migrate/20210517132916_add_host_proxy_invocations.rb:4:in `change'
/home/aruzicka/.asdf/installs/ruby/2.7.8/bin/bundle:23:in `load'
/home/aruzicka/.asdf/installs/ruby/2.7.8/bin/bundle:23:in `<main>'

Caused by:
ActiveRecord::StatementInvalid: PG::DuplicateTable: ERROR:  relation "host_proxy_invocations" already exists
/home/aruzicka/foreman_remote_execution/db/migrate/20210517132916_add_host_proxy_invocations.rb:4:in `change'
/home/aruzicka/.asdf/installs/ruby/2.7.8/bin/bundle:23:in `load'
/home/aruzicka/.asdf/installs/ruby/2.7.8/bin/bundle:23:in `<main>'

Caused by:
PG::DuplicateTable: ERROR:  relation "host_proxy_invocations" already exists
/home/aruzicka/foreman_remote_execution/db/migrate/20210517132916_add_host_proxy_invocations.rb:4:in `change'
/home/aruzicka/.asdf/installs/ruby/2.7.8/bin/bundle:23:in `load'
/home/aruzicka/.asdf/installs/ruby/2.7.8/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

The migration will need to be changed to be a noop if certain conditions are met

jeremylenz:
@adamruzicka updated. I played around for a bit with detecting the migration status through `ActiveRecord::Base.connection.migration_context.migrations`, but I couldn't find anything useful there.  In the end I realized that the 'dumb' solution of adding `:if_not_exists` should work fine. I tested and confirmed no errors when I start out in master fully migrated, and then switch to this branch and run db:migrate.

adamruzicka (approved):
Works well

adamruzicka:
Thank you @jeremylenz !

PR #842, opened 2023-10-25
Fixes #36785 - Show permission error page for job wizard

kmalyjur:
Depends on https://github.com/theforeman/foreman/pull/9902

It is showing missing permissions needed on the `/job_invocations/new` page. The "Proceed Anyway" button lets you look at the wizard even though it won't be working properly.
image

On the "Category and template" step it shows a warning alert if view_job_templates permission is missing:
image

On the "Target hosts and inputs" step it shows a warning alert if view_hosts permission is missing:
image

### DIFF
--- a/app/controllers/ui_job_wizard_controller.rb
+++ b/app/controllers/ui_job_wizard_controller.rb
@@ -1,4 +1,4 @@
-class UiJobWizardController < ApplicationController
+class UiJobWizardController < Api::V2::BaseController
   include FiltersHelper
   def categories
     job_categories = resource_scope(permission: action_permission)
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -47,7 +47,7 @@ module ForemanRemoteExecution
 
     initializer 'foreman_remote_execution.register_plugin', before: :finisher_hook do |_app|
       Foreman::Plugin.register :foreman_remote_execution do
-        requires_foreman '>= 3.9'
+        requires_foreman '>= 3.10'
         register_global_js_file 'global'
         register_gettext
 
--- a/webpack/JobWizard/JobWizardConstants.js
+++ b/webpack/JobWizard/JobWizardConstants.js
@@ -5,6 +5,10 @@ export const JOB_TEMPLATES = 'JOB_TEMPLATES';
 export const JOB_CATEGORIES = 'JOB_CATEGORIES';
 export const JOB_TEMPLATE = 'JOB_TEMPLATE';
 export const JOB_INVOCATION = 'JOB_INVOCATION';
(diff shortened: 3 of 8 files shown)


ofedoren:
Well, quickly looking at the way we have implemented access control, it seems we only check for one permission per one action. I guess that's why you don't have missing permissions if at least one of them are present. Currently it seems if we add the action into multiple permission blocks, it would treat it as "at least one of those permissions allow this action", which is quite opposite you want to achieve.

As a solution I'd suggest:
 - Either show errors on each requested action (such as loading the page, checking for categories, etc). For loading the page, `create_job_invocations` permission should be enough.
 - Or adjust the whole system to be able to check for multiple permissions for one action.
 
But that's just my observations, @adamruzicka, did I miss anything?

adamruzicka:
Yeah, that's in line with what I've seen.  Out of the two options you suggested I'd vote for the second. There are ~five permissions you really need to be able to trigger a job and I'd prefer not getting to the form at all if I'm missing any rather than seeing the form which bunch of unauthorized errors all over the place. Otoh, and maybe that's just me being pessimistic, I'm pretty sure that there's a part of the wider foreman ecosystem which relies on the current behavior when any permission suffices

ares:
The question is, should we even allow user to get "create_job_invocations" permission without all other 4 required ones? Should it implicitly add these? Should we have any validations? I know this would open a can of worms by introducing dependency between permissions, but from user perspective, I don't want to be able to define meaningless combinations of permissions. With more complex wizards, we'll get to similar situations more often.

ofedoren:
Should we then actually kinda enforce using predefined roles? Which can contain all the needed perms?

kmalyjur:
I've updated the PR to be ready for review. The permissions are taken from a new API (https://github.com/theforeman/foreman/pull/9902). Could anyone please take a look at it?

pondrejk (approved):
ack, checked on stream snap 38 + packit for this and the dependency PR

pondrejk:
I noticed that it takes a bit of time to display the permission denied panel, first the wizard is displayed, then the panel hops in. But that is just a minor annoyance, otherwise works as described

kmalyjur:
> I noticed that it takes a bit of time to display the permission denied panel, first the wizard is displayed, then the panel hops in. But that is just a minor annoyance, otherwise works as described

Thanks for the review. You're right about this - I kinda ignored it, but if it's too annoying, I can for sure add a "loading" page instead of it in the future.

adamruzicka (commented):
Could you please bump the required foreman version in engine.rb to 3.10? Otherwise LGTM

adamruzicka (approved):
Works like a charm

adamruzicka:
Thank you @kmalyjur , @ofedoren, @ares & @pondrejk !

PR #844, opened 2023-11-01
Fixes #36881 - Display full hostgroup name in Job Wizard

MariaAga:
`name` is the short name, while `title` is the full name, including the host group parent


adamruzicka (approved):
Seems to work well

pnovotny:
LGTM
run-job-hg-selection

adamruzicka:
Thank you @MariaAga & @pnovotny !

PR #846, opened 2023-11-07
Use foreman-devel module

adamruzicka:


.packit.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
     trigger: pull_request
     targets:
       centos-stream-8:
-        additional_modules: "foreman:el8,nodejs:12"
+        additional_modules: "foreman:el8,nodejs:14"
```
ekohl, line 32: Similar to https://github.com/theforeman/foreman/commit/8dda6a6c4e7a0931e47ec81b4649215073afd43b we've introduced the `foreman-devel` module that pulls in the correct versions so that when we bump again there's no need to keep it in sync:

```suggestion
        additional_modules: "foreman-devel:el8"
```

ekohl (changes requested):

ekohl (approved):

.packit.yaml
adamruzicka: Done

evgeni (approved):

PR #847, opened 2023-11-09
refs #36837 - Fix condition for errors and remove skips check

stejskalleos:


adamruzicka (approved):
Erm, yes, sorry for the bad suggestion.

PR #849, opened 2023-11-28
Fixes #36952 - update the reported_at field and refresh the host status

sbernhard:


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -195,7 +195,9 @@ def update_host_status
         host = Host.find(input[:host][:id])
         status = host.execution_status_object || host.build_execution_status_object
         status.status = exit_status.to_s == "0" ? HostStatus::ExecutionStatus::OK : HostStatus::ExecutionStatus::ERROR
+        status.reload
```
adamruzicka, line 198: Doesn't this discard the changes done on the previous line?

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -195,7 +195,9 @@ def update_host_status
         host = Host.find(input[:host][:id])
         status = host.execution_status_object || host.build_execution_status_object
         status.status = exit_status.to_s == "0" ? HostStatus::ExecutionStatus::OK : HostStatus::ExecutionStatus::ERROR
+        status.refresh
```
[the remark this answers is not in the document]
ekohl: Probably because it's operating on a non-saved object. Isn't it possible to only save a single attribute, instead of the whole object? I'm guessing `build_execution_status_object` instantiates a new object.

Question is, why does it need to refresh in the first place? Line 195 implies the host object is a fresh object so what are the changes it actually has an outdated cache?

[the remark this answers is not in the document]
sbernhard: I would be happy to only add update_timestamp https://github.com/theforeman/foreman/blob/develop/app/models/host_status/status.rb#L78 but its private.

Should it maybe only call "status.refresh" in case of a already existing execution_status_object?

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -195,7 +195,9 @@ def update_host_status
         host = Host.find(input[:host][:id])
         status = host.execution_status_object || host.build_execution_status_object
         status.status = exit_status.to_s == "0" ? HostStatus::ExecutionStatus::OK : HostStatus::ExecutionStatus::ERROR
+        status.refresh unless status.id.nil?
```
adamruzicka, line 198: This would probably be more rails-y and should do essentially the same thing
```suggestion
        status.refresh unless status.new_record?
```

sbernhard: ok. Make sense. Thanks.

sbernhard:
Do we need to squash the commits now or can this be done when we merge this?

adamruzicka:
No need to, it can be done on merge

adamruzicka:
The only catch I see is that the code that updates the status doesn't get run if the execution fails early (this can happen when the user doesn't have necessary permissions or when no proxy can be used for the execution). Should the host status be updated in such cases? 
Nevertheless, that's not a new thing, it has been around for some time already so we don't really have to solve it here.

sbernhard:
> The only catch I see is that the code that updates the status doesn't get run if the execution fails early (this can happen when the user doesn't have necessary permissions or when no proxy can be used for the execution). Should the host status be updated in such cases? Nevertheless, that's not a new thing, it has been around for some time already so we don't really have to solve it here.

Yes, it should update the status but as you already said, different story. Shouldn't this be possible in https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L74

If not, then hopefully this kind of error handling can be added to another section?

adamruzicka (approved):
Works nicely

adamruzicka:
Thank you @sbernhard !

PR #852, opened 2023-11-29
Fixes #36953 - Fetch configured ansible tags on rerun

girijaasoni:


app/controllers/ui_job_wizard_controller.rb
```diff
@@ -64,7 +64,8 @@ def job_invocation
     job = JobInvocation.authorized.find(params[:id])
     composer = JobInvocationComposer.from_job_invocation(job, params).params
     job_template_inputs = JobTemplate.authorized.find(composer[:template_invocations][0][:template_id]).template_inputs_with_foreign
-    inputs = Hash[job_template_inputs.map { |input| ["inputs[#{input[:name]}]", {:advanced => input[:advanced], :value => (composer[:template_invocations][0][:input_values].find { |value| value[:template_input_id] == input[:id] }).try(:[], :value)}] }]
+    provider_inputs = composer[:template_invocations][0][:provider_input_values]
+    inputs = Hash[job_template_inputs.map { |input| ["inputs[#{input[:name]}]", {:advanced => input[:advanced], :value => (composer[:template_invocations][0][:input_values].find { |value| value[:template_input_id] == input[:id] }).try(:[], :value)}] }].merge(Hash[provider_inputs.map { |input| ["inputs[#{input[:name]}]", {:advanced => true, :value => input[:value]}] }])
```
MariaAga, line 68: provider_inputs might be nil and gives error when trying to map nil

nofaralfasi:
> The bug described here: https://projects.theforeman.org/issues/36953 still happens with this change it looks like `composer[:template_invocations]` doesn't contain the ansible inputs

That's weird, for me it works. When you try to rerun the job, you don't see the tags from the previous run?

girijaasoni:
> Nope and `composer[:template_invocations]` is `{"template_id"=>225, "effective_user"=>"root", "input_values"=>[{"template_input_id"=>125, "value"=>"ls"}]}` My runs are failing/in the future, are your runs successful?

could you specify which job you're running so I could reproduce it here?

app/controllers/ui_job_wizard_controller.rb
nofaralfasi: > provider_inputs might be nil and gives error when trying to map nil

I guess this is also resolved when you apply the changes in the `app/models/job_invocation_composer.rb`. When `provider_inputs` is empty, we get an empty array, which is valid.

girijaasoni: In case of no provider inputs, provider_inputs is passed as an empty array and I think map doesn't throw an error on an empty array as it does on a nil value. I could still add a safety check operator if that's needed

stejskalleos: I'm for refactoring as well, better readability would be nice

MariaAga:
As expected, now that I tested the whole PR everything works as expected :)

nofaralfasi (approved):
LGTM

PR #854, opened 2023-12-07
Fixes #36967 - convert2rhel analyze removes OS from the host

stejskalleos:
Workaround for the issue with 'convert2rhel analyze', where after the job run, OS is unassigned from the host.

See https://issues.redhat.com/browse/RHELC-1280


ares (approved):
Thanks, makes sense. Merging!

PR #856, opened 2023-12-11
10.1.2 - Fetch configured ansible tags on rerun

stejskalleos:


stejskalleos:
Closing the PR, we don't need to CP it to 10.x

PR #858, opened 2023-12-11
Fixes #36974 - show run job button to rex users

MariaAga:
In a separate PR I want to try to deal with the issue that react#index paths dont get the correct auths, but I think we should first return the job buttons to Rex Users. 

Added also `view_remote_execution_features` permission because I think normal users should be able to view the features? just not edit them

test: have a user with only rex user role, open `/job_invocations` before: no button, now: yes button.
Also for job reruns inside `job_invocations/id`, and  in host details page (some hosts are `infrastructure_host` and will need extra permission to run job).

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -33,8 +33,7 @@ module RemoteExecutionHelper
       links << { title: _('Host detail'),
         action: { href: current_host_details_path(host), 'data-method': 'get', id: "#{host.name}-actions-detail" } }
     end
-
-    if authorized_for(hash_for_rerun_job_invocation_path(id: job_invocation, host_ids: [ host.id ], authorizer: job_hosts_authorizer))
+    if authorized_for(controller: :job_invocations, action: :create) && (!host.infrastructure_host? || User.current.can?(:execute_jobs_on_infrastructure_hosts))
       links << { title: (_('Rerun on %s') % host.name),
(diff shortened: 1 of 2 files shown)


ofedoren (commented):
Funny enough, but even without this patch I was able as a user with only `REX user` role to view and run a job by navigating manually to `/job_invocations/new` page...

With this patch I'm able to see the button and run the job. Also can view and run jobs on host details page. So the patch works it seems.

Maybe that's not relevant for this PR, but I've noticed few smaller issues:
  - as a user with viewer only permissions (not a REX user) I'm able to manually navigate to the job invocation page, but at least after I click run a see `Access denied` notification.
  - as a REX user I can't click checkboxes on /tasks page to e.g. cancel few jobs.

Otherwise LGTM and the changes make sense.

@kmalyjur, could you take a look as well?

MariaAga:
Thanks!

> as a user with viewer only permissions (not a REX user) I'm able to manually navigate to the job invocation page, but at least after I click run a see Access denied notification.

@kmalyjur is fixing that.

> as a REX user I can't click checkboxes on /tasks page to e.g. cancel few jobs.

Thats expected since rex users dont have permissions in tasks, thats what the Tasks Manager role is for I assume.

ofedoren (approved):

kmalyjur (approved):
Awesome!

ofedoren:
Okay then, let's get this in. Thanks, @MariaAga and @kmalyjur :rose:

PR #859, opened 2023-12-14
Fixes #36982 - display short hostname on REX job pages

nofaralfasi:
If the `display_fqdn_for_hosts` setting is set to false, show the short hostname on REX job pages.

Requires:
* https://github.com/theforeman/foreman/pull/9968

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -34,7 +34,7 @@ module RemoteExecutionHelper
         action: { href: current_host_details_path(host), 'data-method': 'get', id: "#{host.name}-actions-detail" } }
     end
     if authorized_for(controller: :job_invocations, action: :create) && (!host.infrastructure_host? || User.current.can?(:execute_jobs_on_infrastructure_hosts))
-      links << { title: (_('Rerun on %s') % host.name),
+      links << { title: (_('Rerun on %s') % host),
         action: { href: rerun_job_invocation_path(job_invocation, host_ids: [ host.id ]),
                   'data-method': 'get', id: "#{host.name}-actions-rerun" } }
     end
@@ -274,7 +274,7 @@ module RemoteExecutionHelper
       task = template_invocation.try(:run_host_job_task)
       link_authorized = !task.nil? && authorized_for(hash_for_template_invocation_path(:id => template_invocation).merge(:auth_object => host, :permission => :view_hosts, :authorizer => job_hosts_authorizer))
 
-      { name: host.name,
+      { name: host.to_label,
         link: link_authorized ? template_invocation_path(:id => template_invocation) : '',
         status: template_invocation_status(task, job_invocation.task),
         actions: template_invocation_actions(task, host, job_invocation, template_invocation) }
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -28,6 +28,7 @@ module Actions
         raise _('Could not use any template used in the job invocation') if template_invocation.blank?
         features = template_invocation.template.remote_execution_features.pluck(:label).uniq
         action_subject(host,
+          :host_display_name => host.to_label,
           :job_category => job_invocation.job_category,
           :description => job_invocation.description,
           :job_invocation_id => job_invocation.id,
@@ -101,8 +102,12 @@ module Actions
       end
 
       def humanized_input
-        N_('%{description} on %{host}') % { :host => input[:host].try(:[], :name),
-          :description => input[:description].try(:capitalize) || input[:job_category] }
+        return unless input.present?
+
+        N_('%{description} on %{host}') % {
+          host: input[:host_display_name],
+          description: input[:description].try(:capitalize) || input[:job_category],
+        }
       end
 
       def humanized_name
--- a/webpack/JobWizard/autofill.js
+++ b/webpack/JobWizard/autofill.js
@@ -39,10 +39,14 @@ export const useAutoFill = ({
             handleSuccess: ({ data }) => {
               setSelectedTargets(currentTargets => ({
                 ...currentTargets,
-                hosts: (data.results || []).map(({ id, name }) => ({
-                  id,
-                  name,
-                })),
+                hosts: (data.results || []).map(
+                  // eslint-disable-next-line camelcase
+                  ({ id, name, display_name }) => ({
+                    id,
+                    // eslint-disable-next-line camelcase
+                    name: display_name || name,
+                  })
+                ),
               }));
             },
           })
--- a/webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
+++ b/webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
@@ -22,16 +22,16 @@ export const HostPreviewModal = ({ isOpen, setIsOpen, searchQuery }) => {
     >
       <List isPlain>
         {hosts.map(host => (
-          <ListItem key={host}>
+          <ListItem key={host.name}>
             <Button
               component="a"
-              href={foremanUrl(`/hosts/${host}`)}
+              href={foremanUrl(`/hosts/${host.name}`)}
               variant="link"
               target="_blank"
               rel="noreferrer"
               isInline
             >
-              {host}
+              {host.display_name}
             </Button>
           </ListItem>
         ))}
--- a/webpack/JobWizard/steps/HostsAndInputs/SelectedChips.js
+++ b/webpack/JobWizard/steps/HostsAndInputs/SelectedChips.js
@@ -4,7 +4,7 @@ import { Chip, ChipGroup, Button } from '@patternfly/react-core';
 import { sprintf, translate as __ } from 'foremanReact/common/I18n';
 import { hostMethods } from '../../JobWizardConstants';
 
-const SelectedChip = ({ selected, setSelected, categoryName }) => {
+const SelectedChip = ({ selected, setSelected, categoryName, setLabel }) => {
   const deleteItem = itemToRemove => {
     setSelected(oldSelected =>
       oldSelected.filter(({ id }) => id !== itemToRemove)
@@ -24,14 +24,14 @@ const SelectedChip = ({ selected, setSelected, categoryName }) => {
           setSelected(() => []);
         }}
       >
-        {selected.map(({ name, id }, index) => (
+        {selected.map((result, index) => (
           <Chip
             key={index}
-            id={`${categoryName}-${id}`}
-            onClick={() => deleteItem(id)}
-            closeBtnAriaLabel={`Remove ${name}`}
+            id={`${categoryName}-${result.id}`}
+            onClick={() => deleteItem(result.id)}
+            closeBtnAriaLabel={`Remove ${result.name}`}
           >
-            {name}
+            {setLabel(result)}
           </Chip>
         ))}
       </ChipGroup>
@@ -49,6 +49,7 @@ export const SelectedChips = ({
   setSelectedHostGroups,
   hostsSearchQuery,
   clearSearch,
+  setLabel,
 }) => {
   const clearAll = () => {
     setSelectedHosts(() => []);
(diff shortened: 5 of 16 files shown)


nofaralfasi:
@MariaAga Any idea why the tests are failing?

MariaAga:
`TypeError: (0 , _ForemanContext.useForemanSettings) is not a function` 
It needs to have a mock in the `__mocks__` folder

nofaralfasi:
> `TypeError: (0 , _ForemanContext.useForemanSettings) is not a function` It needs to have a mock in the `__mocks__` folder

Would adding this line [here](https://github.com/theforeman/foreman_remote_execution/blob/master/webpack/__mocks__/foremanReact/Root/Context/ForemanContext/index.js) resolve the issue?
`export const useForemanSettings = () => {};`

Or is there a way I can test it locally? I couldn't find how.

MariaAga:
Yeah it should fix it. You can run `npm test` in the plugin folder to test

webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
```diff
@@ -31,7 +33,7 @@ export const HostPreviewModal = ({ isOpen, setIsOpen, searchQuery }) => {
               rel="noreferrer"
               isInline
             >
-              {host}
+              {displayFqdnForHosts ? host : host.split('.')[0]}
```
ShimShtein, line 36: I would try to avoid this. You can change/add a selector that will add the `display_name`  field for each host. Then you can use it in this modal.

webpack/JobWizard/steps/ReviewDetails/index.js
```diff
@@ -79,12 +80,16 @@ const ReviewDetails = ({
...
   const hostsCount = useSelector(selectHostCount);
   const [hostPreviewOpen, setHostPreviewOpen] = useState(false);
+  const { displayFqdnForHosts } = useForemanSettings();
   const stringHosts = () => {
     if (hosts.length === 0) {
       return __('No Target Hosts');
     }
     if (hosts.length === 1 || hosts.length === 2) {
-      return hosts.join(', ');
+      const displayHosts = displayFqdnForHosts
```
ShimShtein, line 89: Here we can use the same selector as we used for `HostPreviewModal`

webpack/JobWizard/steps/form/SearchSelect.js
```diff
@@ -48,7 +50,7 @@ export const SearchSelect = ({
     ...selectOptions,
     ...Immutable.asMutable(response?.results || [])?.map((result, index) => (
       <SelectOption key={index + 1} value={result.id}>
-        {result.name}
```
ShimShtein, line 51: We should be able to see if `result` has some other properties besides `name`. I would guess it's a host object, and should have a `display_name`

webpack/JobWizard/steps/HostsAndInputs/SelectedChips.js
```diff
@@ -31,7 +33,7 @@ const SelectedChip = ({ selected, setSelected, categoryName }) => {
             onClick={() => deleteItem(id)}
             closeBtnAriaLabel={`Remove ${name}`}
           >
-            {name}
+            {displayFqdnForHosts ? name : name.split('.')[0]}
```
ShimShtein, line 36: `selected` variable is a set of hosts returned by the API, they all have `display_name` property, you can use it here, as long as you change the `map` statement:
instead of:
``` js
selected.map(({ name, id }, index) => (...)
```
You can add the `display_name`:
``` js
selected.map(({ name, display_name, id }, index) => (...)
```

ShimShtein (changes requested):
The changes to the ruby code seem quite OK, but the JS doesn't have to know about this setting. It should just use `display_name` where it is possible.

webpack/JobWizard/steps/form/SearchSelect.js
nofaralfasi: here we are using gql for the hosts query, which means that we don't have the `display_name` property.

webpack/JobWizard/steps/HostsAndInputs/SelectedChips.js
nofaralfasi: I think we are using here gql for the query, which means that we don't have access to the `display_name` property.

ShimShtein: Looks like we are missing `display_name` from GQL definition in foreman. Let's add it first.

nofaralfasi: https://github.com/theforeman/foreman/pull/9968 added to the description

webpack/JobWizard/autofill.js
```diff
@@ -39,10 +39,14 @@ export const useAutoFill = ({
...
-                  name,
-                })),
+                hosts: (data.results || []).map(
+                  // eslint-disable-next-line camelcase
+                  ({ id, name, display_name }) => ({
+                    id,
+                    // eslint-disable-next-line camelcase
+                    name: display_name || name,
+                  })
+                ),
```
nofaralfasi, line 49: @ShimShtein I found that the reason the test is failing is that we were missing the `name` param here (and not the graphql as we thought).
If this fixes the issue, we can discuss whether we want to apply other changes.

webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
```diff
@@ -22,16 +22,16 @@ export const HostPreviewModal = ({ isOpen, setIsOpen, searchQuery }) => {
     >
       <List isPlain>
         {hosts.map(host => (
-          <ListItem key={host}>
+          <ListItem key={host.name}>
             <Button
               component="a"
-              href={foremanUrl(`/hosts/${host}`)}
+              href={foremanUrl(`/hosts/${host.name}`)}
```
ekohl, line 28: I'd be tempted to use host IDs in URLs, not names to avoid breaking when you rename hosts. I know Foreman is inconsistent about that. This is why I think we should avoid constructing URLs all over the place and instead rely on helpers. `foremanUrl` is really an anti-pattern.

If the UI relies on it so much, we should do what GH did and include `html_url` in the API response.

webpack/JobWizard/steps/ReviewDetails/index.js
```diff
@@ -84,7 +84,7 @@ const ReviewDetails = ({
       return __('No Target Hosts');
     }
     if (hosts.length === 1 || hosts.length === 2) {
```
ekohl, line 86: Unrelated, but I'd have written `< 3` and probably made `3` some constant.

ekohl (commented):
If this still depends on https://github.com/theforeman/foreman/pull/9968, can you raise the minimum required Foreman version so we don't accidentally cherry pick it?

app/helpers/remote_execution_helper.rb
```diff
@@ -34,7 +34,7 @@ def template_invocation_actions(task, host, job_invocation, template_invocation)
         action: { href: current_host_details_path(host), 'data-method': 'get', id: "#{host.name}-actions-detail" } }
     end
     if authorized_for(controller: :job_invocations, action: :create) && (!host.infrastructure_host? || User.current.can?(:execute_jobs_on_infrastructure_hosts))
-      links << { title: (_('Rerun on %s') % host.name),
+      links << { title: (_('Rerun on %s') % host),
```
stejskalleos, line 37: Is the result of `host` and `host.to_label` same?
If yes, we should unify the usage, and use it same way across the code.

nofaralfasi: You are right. I'll change it where possible.

stejskalleos (approved):
:green_apple: **LGTM**

Did a quick test with enabled and disabled setting, with the new UI job wizard,
and worked fine.
Setting is applied everywhere and the job execution went well. 

@adamruzicka @MariaAga I'll leave the final ACK on you

adamruzicka:
Could you please rebase? It should make packit turn green

webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
```diff
@@ -22,16 +22,16 @@ export const HostPreviewModal = ({ isOpen, setIsOpen, searchQuery }) => {
     >
       <List isPlain>
         {hosts.map(host => (
-          <ListItem key={host}>
+          <ListItem key={host.name}>
             <Button
               component="a"
-              href={foremanUrl(`/hosts/${host}`)}
+              href={foremanUrl(`/new/hosts/${host.name}`)}
```
ShimShtein, line 28: Shouldn't it be dependent on a switch?
@ron-lavi?

adamruzicka:
Could you also deal with it here https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L104 ?

nofaralfasi:
> Could you also deal with it here https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/actions/remote_execution/run_host_job.rb#L104 ?

Is there a reason for the inconsistency in saving the `action` attribute of the ForemanTasks model? Sometimes it stores the full hostname, and other times it uses the short hostname. Any idea why?

adamruzicka:
Could you give an example?

nofaralfasi:
> ForemanTasks

When I set the `Display FQDN for hosts` setting to true, and I execute any remote job on multiple hosts, some of them have the `action` field containing the short hostname, while others have the full hostname.

image

adamruzicka:
I don't know in what environment you hit that, but it sounds like that either not all processes picked up your code changes or that not all processes picked up the setting change. Either way, I'm failing to reproduce it on a fresh nightly machine

nofaralfasi:
> I don't know in what environment you hit that, but it sounds like that either not all processes picked up your code changes or that not all processes picked up the setting change. Either way, I'm failing to reproduce it on a fresh nightly machine

Thanks. Restarting my local environment seems to have solved the issue for me. Everything works as expected now.

nofaralfasi:
Is there anything else required before we proceed with the merge?

webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
nofaralfasi: You are right, we should check user's settings here. I'll remove it and open a separate PR for that.

Ron-Lavi: I saw the same behaviour across the app so I think it is OK..

Probably we can try to load the correct host path to the React Context Settings and use that helper across the app ( but that's for another PR ). :)

app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -101,8 +101,12 @@ def live_output
       end
 
       def humanized_input
-        N_('%{description} on %{host}') % { :host => input[:host].try(:[], :name),
-          :description => input[:description].try(:capitalize) || input[:job_category] }
+        return unless input.present?
+
+        N_('%{description} on %{host}') % {
+          host: host.to_label,
```
adamruzicka, line 107: Do we _need_ to load the host here? Can't we pre-calculate this somewhere in `#plan` or `#inner_plan` where we already have the host loded?

adamruzicka (commented):
On a relevant side note, what does `RunHostJob#host_name` (a thing usable in webhooks) end up being? Is it the short name or the fqdn?

app/lib/actions/remote_execution/run_host_job.rb
nofaralfasi: I'm not sure where can we add this in `#plan` or `#inner_plan`.
What about adding it here https://github.com/theforeman/foreman-tasks/blob/master/app/models/foreman_tasks/concerns/action_subject.rb#L14:
`hash.update(to_label: to_label) if respond_to? :to_label`

adamruzicka: I can't really tell from the top of my head what side effects adding it to ActionSubject might have. Adding it to the hash on L31-L34 in this file should do the trick

adamruzicka (approved):
Looks good

adamruzicka:
Thank you @nofaralfasi !

PR #860, opened 2023-12-18
Refs #36974 - show run job button to rex users

MariaAga:
Found more places where the button was missing

### DIFF
--- a/app/helpers/concerns/foreman_remote_execution/job_templates_extensions.rb
+++ b/app/helpers/concerns/foreman_remote_execution/job_templates_extensions.rb
@@ -4,7 +4,7 @@ module ForemanRemoteExecution
       original = super(template)
 
       if template.is_a?(JobTemplate)
-        original.unshift(display_link_if_authorized(_('Run'), hash_for_new_job_invocation_path(:template_id => template.id).merge(:authorizer => authorizer))) unless template.snippet
+        original.unshift(authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Run'), hash_for_new_job_invocation_path(:template_id => template.id).merge(:authorizer => authorizer)) : nil) unless template.snippet
       end
 
       original
--- a/app/views/job_invocations/new.html.erb
+++ b/app/views/job_invocations/new.html.erb
@@ -2,8 +2,8 @@
 <% stylesheet 'foreman_remote_execution/foreman_remote_execution' %>
 <% title _('Job invocation') %>
 <% if(request.path_parameters[:action] == "rerun") %>
-  <%= display_link_if_authorized(_('Use new job wizard'), hash_for_rerun_job_invocation_path({:id => request.path_parameters[:id]}).merge(request.query_parameters)) %>
+  <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Use new job wizard'), hash_for_rerun_job_invocation_path({:id => request.path_parameters[:id]}).merge(request.query_parameters)) : nil %>
 <% else %>
-  <%= display_link_if_authorized(_('Use new job wizard'), hash_for_new_job_invocation_path().merge(request.query_parameters)) %>
+  <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Use new job wizard'), hash_for_new_job_invocation_path().merge(request.query_parameters)) : nil %>
 <%end%>
 <%= render :partial => 'form' %>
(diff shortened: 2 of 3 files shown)


MariaAga:
I dont think the tests failure are related as they also happen in a different pr

kmalyjur (approved):
Awesome, I'm sorry for missing those before
@ofedoren could you please take a look and merge?

app/helpers/concerns/foreman_remote_execution/job_templates_extensions.rb
```diff
@@ -4,7 +4,7 @@ def permitted_actions(template)
       original = super(template)
 
       if template.is_a?(JobTemplate)
-        original.unshift(display_link_if_authorized(_('Run'), hash_for_new_job_invocation_path(:template_id => template.id).merge(:authorizer => authorizer))) unless template.snippet
+        original.unshift(authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Run'), hash_for_new_job_invocation_path(:template_id => template.id).merge(:authorizer => authorizer)) : nil) unless template.snippet
       end
```
adamruzicka, line 8: It might be just personal preference, but I'd prefer to avoid `condition ? something : nil` where reasonably possible. Could we move the `authorized_for` outside and either combine it with the unless at the end of the line or with the `template.is_a?` on the line above?

app/views/job_invocations/new.html.erb
```diff
@@ -2,8 +2,8 @@
 <% stylesheet 'foreman_remote_execution/foreman_remote_execution' %>
 <% title _('Job invocation') %>
 <% if(request.path_parameters[:action] == "rerun") %>
-  <%= display_link_if_authorized(_('Use new job wizard'), hash_for_rerun_job_invocation_path({:id => request.path_parameters[:id]}).merge(request.query_parameters)) %>
+  <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Use new job wizard'), hash_for_rerun_job_invocation_path({:id => request.path_parameters[:id]}).merge(request.query_parameters)) : nil %>
 <% else %>
-  <%= display_link_if_authorized(_('Use new job wizard'), hash_for_new_job_invocation_path().merge(request.query_parameters)) %>
+  <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Use new job wizard'), hash_for_new_job_invocation_path().merge(request.query_parameters)) : nil%>
```
adamruzicka, line 7:
```suggestion
  <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Use new job wizard'), hash_for_new_job_invocation_path().merge(request.query_parameters)) : nil %>
```

app/views/job_invocations/welcome.html.erb
```diff
@@ -9,6 +9,6 @@
     <p><%= link_to _('Learn more about this in the documentation.'),
            documentation_url('1.ForemanRemoteExecution1.3Manual', :root_url => 'https://www.theforeman.org/plugins/foreman_remote_execution/1.3/index.html#'), :rel => 'external' %></p>
     <div class="blank-slate-pf-main-action">
-        <%= display_link_if_authorized(_("Run Job"), { :action => :create }, { :class => "btn btn-primary btn-lg" }) %>
+        <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_("Run Job"), { :action => :create }, { :class => "btn btn-primary btn-lg" }) : nil%>
```
adamruzicka, line 12:
```suggestion
        <%= authorized_for(controller: :job_invocations, action: :create) ? link_to(_("Run Job"), { :action => :create }, { :class => "btn btn-primary btn-lg" }) : nil %>
```

adamruzicka (commented):
A couple of really tiny nitpicks, apart from those it works as expected.

app/helpers/concerns/foreman_remote_execution/job_templates_extensions.rb
MariaAga: I'm not sure how to do it while avoiding logic mistakes in this line. I can make a replacement function for the `display_link_if_authorized` that was there before and put the `authorized_for` condition there?

adamruzicka: Nevermind then, let's keep it as is and get this in. We shouldn't go ahead and invent our own flavor of helpers from foreman. Something tells me this is not the last time we're seeing something like this happening, so we may eventually want to fix this properly in foreman itself.

PR #865, opened 2024-01-23
Fixes #37082 - Show 'Awaiting start' as status

adamruzicka:
of template invocations when the backing task is not yet created.

image

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -18,7 +18,7 @@ module RemoteExecutionHelper
   end
 
   def template_invocation_status(task, parent_task)
-    return(parent_task.result == 'cancelled' ? _('cancelled') : 'N/A') if task.nil?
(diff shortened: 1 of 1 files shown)


lhellebr (approved):
I think `pending` and `awaiting start` may be confused for each other but I don't have any better word proposal. LGTM.

app/helpers/remote_execution_helper.rb
```diff
@@ -18,7 +18,7 @@ def host_counter(label, count)
   end
 
   def template_invocation_status(task, parent_task)
-    return(parent_task.result == 'cancelled' ? _('cancelled') : 'N/A') if task.nil?
+    return(parent_task.result == 'cancelled' ? _('cancelled') : _('Awaiting start')) if task.nil?
```
lhellebr, line 21: Should a translation be added in a separate PR?

adamruzicka: Yes, strings are extracted for translation separately.

PR #866, opened 2024-01-25
Fix minitest warnings

adamruzicka:
Like
```
Use _(obj).must_equal instead. This will fail in Minitest 6.
```


test/unit/job_template_test.rb
```diff
@@ -30,7 +30,7 @@ class JobTemplateTest < ActiveSupport::TestCase
       job_template.foreign_input_sets << FactoryBot.build(:foreign_input_set, :target_template => template_with_inputs)
       job_template.foreign_input_sets << FactoryBot.build(:foreign_input_set, :target_template => template_with_inputs)
       assert_not job_template.valid?
-      _(job_template.errors.full_messages.first).must_include 'Duplicated inputs detected: ["command"]'
+      refute_nil job_template.errors.full_messages.first&.include?('Duplicated inputs detected: ["command"]')
```
adamruzicka: This isn't the same, ai--

ekohl (commented):
The `assert` vs `assert_predicate` is probably found by https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestassertpredicate

PR #875, opened 2024-02-20
Fixes #37185 - webpack 5 adjustments

MariaAga:
* to ensure correct loading order we need to wrap js scripts into content for
* '`webpack_asset_paths` is deprecated, use `content_for(:javascripts) { webpacked_plugins_js_for(plugin_name) }` instead.')
* `webpacked_plugins_css_for` is deprecated, plugin css is already loaded

### DIFF
--- a/app/views/job_invocations/show.html.erb
+++ b/app/views/job_invocations/show.html.erb
@@ -1,9 +1,9 @@
 <% title @job_invocation.description %>
-<% stylesheet 'foreman_remote_execution/foreman_remote_execution' %>
-<% javascript 'foreman_remote_execution/template_invocation' %>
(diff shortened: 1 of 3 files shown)


adamruzicka:
Could you please rebase onto latest master? That could make the packit build turn green

MariaAga:
cant use content for because it only runs the scripts on page load, and the script updates only a part of the page

adamruzicka (commented):
The indentation is slightly off, but apart from that it seems to do the trick.

MariaAga:
Fixed indentation

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #876, opened 2024-02-28
Fixes #37213 - Add missing ouia-ids

MariaAga:
It was really optimistic to assume in https://github.com/theforeman/foreman_remote_execution/pull/821 that we didnt need to fix anything...
This pr actually makes it so REX will use the rule.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,5 +6,8 @@
   "extends": [
     "plugin:@theforeman/foreman/core",
     "plugin:@theforeman/foreman/plugins"
-  ]
+  ],
+  "rules":{
+    "@theforeman/rules/require-ouiaid": ["error"]
+  }
 }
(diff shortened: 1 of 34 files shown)


kmalyjur (approved):
Works nicely, all the ouia-ids added, thank you

(It could maybe also fix this warning? However, I get that this PR is about something else.)
image

MariaAga:
> (It could maybe also fix this warning? However, I get that this PR is about something else.)

Lets do that in a seperate one

MariaAga:
@lhellebr does this fix the issue? (https://bugzilla.redhat.com/show_bug.cgi?id=2265127)

lhellebr:
Yes, that BZ is fixed when I apply this PR to Stream. `data-ouia-component-id` is now statically `hosts`.

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga , @kmalyjur & @lhellebr !

PR #878, opened 2024-03-12
Fixes #37242 - Fix live output appearing multiple times

adamruzicka:
The result of calling continuous_output was not persisted, so the original code didn't make much sense. `ContinuousOutput#sort!` returns the same thing as `ContinuousOutput#raw_outputs` so we can skip the second line completely.

`upsert_all` was updating the already existing values, so `exit_status` got updated multiple times with different timestamps. `insert_all` should just skip over the already existing values.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -98,7 +98,6 @@ module Actions
 
       def live_output
         continuous_output.sort!
-        continuous_output.raw_outputs
       end
 
       def humanized_input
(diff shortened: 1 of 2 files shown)


app/lib/actions/remote_execution/run_host_job.rb
```diff
@@ -98,7 +98,6 @@ def check_exit_status
 
       def live_output
         continuous_output.sort!
-        continuous_output.raw_outputs
```
adamruzicka, line 101: Kudos to @avitova for pointing this out when working on her thesis

sbernhard (approved):

ofedoren (approved):
Thanks, @adamruzicka and @sbernhard, let's get this in.

PR #880, opened 2024-03-27
Fixes #37305 - Honor default values

adamruzicka:
for template inputs


shweta83:
Tested the fix and default values for leapp template works fine now.

stejskalleos (approved):
:green_apple: **LGTM**
I haven't tested it, but QAs confirmed functionality, and unit tests cover it.

PR #881, opened 2024-03-27
Fix diff behavior of package removal on debian/rhel based system

sbernhard:


adamruzicka:
I'm afraid I'll need some background for this

sbernhard:
Sorry. 
If you run "yum remove XYZ" and XYZ is not installed, it will have a exit code of 0
If you do the same with apt-get remove XYZ on a debian/ubuntu system, it will fail with exit code 100 as it can not find the package. From a user perspective I would follow the yum behavior because you expect that the package is no longer installed on the system - so if the package was not installed, everything is fine.

app/views/templates/script/package_action.erb
```diff
@@ -115,9 +115,19 @@ handle_zypp_res_codes () {
...
   export DEBIAN_FRONTEND=noninteractive
   apt-get -y update
-  apt-get -o APT::Get::Upgrade-Allow-New="true" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y <%= input("options") %> <%= action %> <%= input("package") %>
+  apt-get -o APT::Get::Upgrade-Allow-New="true" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y <%= input("options") %> <%= action %> <%= input("package") %> 2>&1 | tee $OUTFILE
+  RETVAL=$?
+  if grep -q "Unable to locate" $OUTFILE; then
+    true
+  else
+    setReturnValue() { RETVAL=$1; return $RETVAL; }
+    setReturnValue $RETVAL
```
adamruzicka, line 129: I'm puzzled by this, why create a function that takes an argument, sets RETVAL to the value of the argument and returns the argument, only to call it with the value that is already in RETVAL? What is the trick that I'm missing?

sbernhard: The return value is grabbed here: https://github.com/theforeman/foreman_remote_execution/pull/881/files/2d60047398276ee1e23c7657aba5ec4cc5655ab8#diff-e3f43610dc58afecce93eef51717baa080bcdbda143593286b46922444cfecfaL133

This is used to prevent the original return code from the "apt-get" and set it so that the line L133 uses the return code from the apt-get. 

I tried to find a better solution like a "false" + the $RETVAL from L124 but there is nothing like this.

sbernhard:
Further thoughts @adamruzicka?

app/views/templates/script/package_action.erb
```diff
@@ -115,9 +115,19 @@ handle_zypp_res_codes () {
...
   -%>
+  OUTFILE=$(mktemp)
+  trap "rm -f $OUTFILE" EXIT
+  set -o pipefail
   export DEBIAN_FRONTEND=noninteractive
   apt-get -y update
-  apt-get -o APT::Get::Upgrade-Allow-New="true" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y <%= input("options") %> <%= action %> <%= input("package") %>
+  apt-get -o APT::Get::Upgrade-Allow-New="true" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y <%= input("options") %> <%= action %> <%= input("package") %> 2>&1 | tee $OUTFILE
+  RETVAL=$?
+  if grep -q "Unable to locate" $OUTFILE; then
```
adamruzicka, line 125: This won't work if the machine's language is not english. Should we force it to be english?
```
# LC_ALL=ja_JP.UTF-8 apt-get remove foobar
パッケージリストを読み込んでいます... 完了
依存関係ツリーを作成しています
状態情報を読み取っています... 完了
E: パッケージ foobar が見つかりません
```

sbernhard: Oh. Thats true. Nice catch. 
Run the "apt-get" with LANG=C?

adamruzicka (approved):

adamruzicka:
Thank you @sbernhard !

PR #882, opened 2024-03-28
Fix the tags check for the release gem action

sbernhard:
See https://github.com/theforeman/actions/pull/50

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,10 @@
 name: Release
 
 on:
-  create:
-    ref_type: tag
+  push:
+    # Pattern matched against refs/tags
+    tags:
+      - '**'
 
 jobs:
   release:


adamruzicka:
This never really worked at all, could we drop it instead?

sbernhard:
You mean, to drop the whole github action? 
Actually, for foreman_acd it works pretty good.

adamruzicka:
> You mean, to drop the whole github action?

Yes

>  Actually, for foreman_acd it works pretty good.

Well, this one does not and it doesn't fit into the way how I release things anyway.

sbernhard:
Well, if you have "your way" to release the plugin, feel free to remove it again

adamruzicka (approved):

adamruzicka:
Alright, I'll give it another chance

PR #883, opened 2024-04-04
Fixes #37316 - Drop non-batch triggering

adamruzicka:
Companion to https://github.com/theforeman/foreman-tasks/pull/751


adamruzicka:
Turns out this is still occasionally useful, closing.

PR #203, opened 2016-09-06
Fixes #16438 - Extract common code into foreman_tasks_core

iNecas:

PR #884, opened 2024-04-04
Fixes #37323 - Rely on default implementation of delegated_output

adamruzicka:
The current one was placed in because of changes that happened back in 2016, it is not needed anymore.

### DIFF
--- a/app/lib/actions/remote_execution/run_host_job.rb
+++ b/app/lib/actions/remote_execution/run_host_job.rb
@@ -204,20 +204,6 @@ module Actions
         host.refresh_global_status!
       end
 
-      def delegated_output
-        if input[:delegated_action_id]
-          super
-        elsif phase?(Present)
-          # for compatibility with old actions
-          old_action = all_planned_actions.first
-          if old_action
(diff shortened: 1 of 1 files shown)


ofedoren (commented):
Thanks, @adamruzicka, the test failures are related to Node 14, we have it broken across the org, so not related.

Do we have a ticket which added this method so we could link it for more context?

adamruzicka:
git blame takes me back to https://projects.theforeman.org/issues/16438

ofedoren:
Didn't help much though...

ofedoren (approved):
Looking through code it seems pretty safe. Do you plan to create a RM for this or should we :shipit:?

adamruzicka:
From what I remember, REX had its own way of delegating an action to the smart proxy. This was eventually replaced by the DelegatedAction machinery from foreman-tasks in #203 , but it stored data slightly differently. Rather than migrating the data, we opted to be able to read the data from both places. The old way of storing data should be long gone so it should be safe to not consider it anymore

PR #886, opened 2024-04-05
Fixes #37321 - Rely on output id when inserting events

adamruzicka:
Soft requires:
- https://github.com/theforeman/smart_proxy_dynflow/pull/135
- https://github.com/theforeman/smart_proxy_dynflow/pull/136

### DIFF
--- a/app/lib/actions/remote_execution/proxy_action.rb
+++ b/app/lib/actions/remote_execution/proxy_action.rb
@@ -25,7 +25,7 @@ module Actions
           {
             # For N-1 compatibility, we assume that the output provided here is
             # complete
-            sequence_id: update['sequence_id'] || seq_id,
+            external_id: update['id'] || seq_id,
             template_invocation_id: template_invocation.id,
             event: update['output'],
             timestamp: Time.at(update['timestamp']).getlocal,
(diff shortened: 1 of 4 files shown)


app/lib/actions/remote_execution/proxy_action.rb
```diff
@@ -25,25 +25,25 @@ def process_proxy_data(data)
...
         if data['exit_status']
-          last = events.last || {:sequence_id => -1, :timestamp => Time.zone.now}
+          last = events.last || {:timestamp => Time.zone.now}
           events << {
-            sequence_id: last[:sequence_id] + 1,
+            external_id: 'exit',
             template_invocation_id: template_invocation.id,
             event: data['exit_status'],
-            timestamp: last[:timestamp],
+            timestamp: data['exit_status_timestamp'] || last[:timestamp] + 1,
```
ofedoren, line 41: Do I read this right?

`timestamp` will be either:
  - `String`: `data['exit_status_timestamp']`, which contains a `Float` converted to a `String`, e.g. `Time.now.utc.to_f` (https://github.com/theforeman/smart_proxy_dynflow/pull/136/files#diff-f3c605b1717c7b0b34ac0715dcb3867c32689fad5a09c7f73a9de47708fa326bR62)
  - `Time`: `Time.zone.now` (https://github.com/theforeman/foreman_remote_execution/pull/886/files#diff-60c1608dbb7d26b97a00fa9c6508054d8dfde370bd3c99ed9e8fd4ff5647f7f9R36)
  
 Also, in https://github.com/theforeman/smart_proxy_dynflow/pull/136/files#diff-606cc2cb76c4da10aa50e3b789b003c3452f553e0c425121cacca9a8af776e25R74 it's `Time.now.utc`, but here it's `Time.zone.now`. Shouldn't it be forced to use the same timezone?
 
 Then, in https://github.com/theforeman/foreman_remote_execution/pull/886/files#diff-720316f4bf1726b14fa1324b7fecce4879e350f67177136c17cbf61e26aeba54R162 there is `order`ing by `timestamp`. But what's the value there? Does it compare numbers or strings? How does it work with inconsistencies above?
 
 P.S. I might be totally lost, so please, take this as rather noob questions.

adamruzicka: Fixed loading of the exit status timestamp, hopefully it should make more sense now.

PR #887, opened 2024-04-08
Fix naming of Convert2RHEL job category

stejskalleos:
From `Convert 2 RHEL` to `Convert2RHEL`

See https://github.com/theforeman/foreman_ansible/pull/708

### DIFF
--- a/app/views/templates/script/convert2rhel_analyze.erb
+++ b/app/views/templates/script/convert2rhel_analyze.erb
@@ -11,7 +11,7 @@ template_inputs:
     default: 'yes'
     hidden_value: false
 model: JobTemplate
-job_category: Convert 2 RHEL
+job_category: Convert2RHEL
 provider_type: script
 kind: job_template
 %>


adamruzicka:
This doesn't seem to modify the ones which were already seeded. Will this need a migration?

stejskalleos:
> This doesn't seem to modify the ones which were already seeded. 

Isn't that related more to https://github.com/theforeman/foreman/pull/10038.

adamruzicka:
That would be an option, but then I'd be hesitant to merge this before the pr in foreman is merged.

stejskalleos:
Well, so far it was not much of a problem, there are usually changes in the templates as well so the seeder is _almost_ always executed.

Lennonka (changes requested):
When you use a noun in the name of the other template, i.e. "Convert2RHEL conversion", it would be better to rename this one to "Convert2RHEL analysis" instead of using a verb (analyse).
When we're already renaming the templates, we might as well be consistent in the naming.

PR #893, opened 2024-05-13
Fixes #37441 - new host url in job wizard hosts

MariaAga:


webpack/JobWizard/steps/HostsAndInputs/HostPreviewModal.js
```diff
@@ -13,6 +14,7 @@ export const HostPreviewModal = ({ isOpen, setIsOpen, searchQuery }) => {
   const hostsCount = useSelector(selectHostCount);
   const url = new URI(foremanUrl('/hosts'));
 
+  const hostsIndexUrl = useForemanHostsPageUrl();
```
adamruzicka, line 17: Not sure why but this gives me `/hosts` no matter what the setting is

MariaAga: accidently used the setting for host index and not show, should be good now

PR #895, opened 2024-05-14
Fixes #37453 - Support Zeitwerk loader

ofedoren:
WIP


ekohl:
> requires [theforeman/foreman-tasks#752](https://github.com/theforeman/foreman-tasks/pull/752)

We support `gemfile.d` where you can add the foreman-tasks dependency with your branch:
https://github.com/theforeman/actions/blob/61c576caf6aea7bf18178a2bac7b21178fb44ed9/.github/workflows/foreman_plugin.yml#L121-L123

foreman_remote_execution.gemspec
```diff
@@ -26,7 +26,6 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'deface'
   s.add_dependency 'dynflow', '>= 1.0.2', '< 2.0.0'
-  s.add_dependency 'foreman-tasks', '>= 8.3.0'
```
ekohl, line 29: Technically I think you could even keep this, as long as the `gem` line in `gemfile.d` also satisfies it.

ekohl (commented):
Something is unhappy:
```
NoMethodError: undefined method `config' for nil:NilClass
    test/test_helper.rb:181:in `setup_set_script_name'
    test/test_helper.rb:84:in `before_setup'
```

https://github.com/theforeman/foreman/blob/9fccd97887938c718cda9f638ed8e049091852b3/test/test_helper.rb#L180-L182

Looks like some test doesn't set `@controller`, which is weird because this isn't changing any of that and we haven't changed it in Foreman itself either.

ofedoren:
> Something is unhappy:
> 
> ```
> NoMethodError: undefined method `config' for nil:NilClass
>     test/test_helper.rb:181:in `setup_set_script_name'
>     test/test_helper.rb:84:in `before_setup'
> ```
> 
> https://github.com/theforeman/foreman/blob/9fccd97887938c718cda9f638ed8e049091852b3/test/test_helper.rb#L180-L182
> 
> Looks like some test doesn't set `@controller`, which is weird because this isn't changing any of that and we haven't changed it in Foreman itself either.

It was a silly mistake in test/functional/ui_job_wizard_controller_test.rb. The test case's class name should comply with the actual controller (Ui -> UI).

ekohl:
@evgeni this failed to build and has already been rotated, but I'm assumed it failed to build because it needs a patched Foreman and/or foreman-tasks. I recall you did something similar with a lookaside repo. Do you recall how?

evgeni:
If the only thing you need is a build against a Zeitwerk-enabled Foreman: Add the packit for https://github.com/theforeman/foreman/pull/10131 to `external_repos` of packit and run with it.
If you want it more stable, mirror those builds to an external URL so they remain even if Copr cleans them up.

ekohl:
Ah yes. Looks like I'll manually compose a repository with all packages. Then it's also an easy place to test everything.

evgeni:
That's what I did for webpack5, and it was quite fine to test with

ofedoren:
Ready to merge after https://github.com/theforeman/foreman/pull/10131 is merged, dropped the commit for testing with Foreman PR, the latest run was green:
Screenshot_20240903_143351

adamruzicka:
Thank you @ofedoren !

PR #897, opened 2024-05-21
Use consistent phrasing

quba42:
Both of the following should use the same phrasing:

webpack/JobWizard/JobWizardPageRerun.js
webpack/JobWizard/index.js

Right now the user experience is as follows:

When Creating a new Job through Monitor → Jobs → Run Job, while in the creation wizard, there is an option to “Use Legacy Form”

After Submitting the Job I land on the Job Result Page and click on Rerun which sends me back to the Job wizard, and now the option that was “Use Legacy Form” says " Use old Form"


quba42:
Do I need to get rid of now unneeded translations as well, or will that happen automatically with the next release?

adamruzicka:
Thank you @quba42 !

PR #898, opened 2024-05-22
Fixes #37492 - same response code for pkg removal

sbernhard:
Small addition to https://github.com/theforeman/foreman_remote_execution/pull/881


adamruzicka (approved):
Right, makes sense

adamruzicka:
Thank you @sbernhard !

PR #899, opened 2024-05-22
Fixes #37493 - Store smart proxy id in template invocations

adamruzicka:
Up until now we only stored it somewhere in the depths of tasks and it served us reasonably well, because we only showed the in template invocation's details.


kmalyjur:
Thank you! Could you please also make it work with invocations that are already up?

kmalyjur:
It works for me now, thank you!

ares (approved):
Makes sense, thanks @adamruzicka, merging!

PR #900, opened 2024-05-22
Fixes #37494 - Properly load proxy data

adamruzicka:
when proxy callback does not go through


ofedoren (approved):
Thanks, @adamruzicka !

PR #904, opened 2024-06-12
Fixes #37557 - change job wizard width

MariaAga:
so users will have more space to type in


pondrejk:
Hello, I'm in for the proposed solution but it seems that the search field for hosts is not affected and maintains the narrow width

image

MariaAga:
@pondrejk can you inspect ` #host-selection` to make sure that the pr is applied and its not limited to 500px?
image

MariaAga:
Should be good now

adamruzicka (approved):
LGTM

adamruzicka:
Thank you @MariaAga & @pondrejk !

PR #906, opened 2024-06-20
Fixes #37580 - Auto-import new GPG keys on SLES

tux93:


adamruzicka:
@sbernhard I have no experience with sles, could you please chime in here?

app/views/templates/script/package_action.erb
```diff
@@ -148,7 +148,7 @@ handle_zypp_res_codes () {
     end
   -%>
   zypper refresh
-  zypper -n <%= action %> <%= input("options") %> <%= input("package") %>
+  zypper -n --gpg-auto-import-keys <%= action %> <%= input("options") %> <%= input("package") %>
```
sbernhard, line 151: Should we really do this automatically ? @tux93 
It will simply add the gpg keys. You can use the "option" field if you are sure to add the gpg key

tux93: I'm pretty sure we want this because it will simply make zypper behave the same way as dnf / yum where GPG keys already are automatically added.

Using the "option" field won't work because the flag is position dependant and needs to come before the action.

sbernhard: Can you show me, how this is done in dnf/yum?

tux93: There is no special configuration in yum/dnf as far as I can tell, if gpgkey is set in a repo it will be imported by default, while the default for zypper is to reject it:
```
dnf -y install katello-host-tools-tracer
[...]
warning: /var/cache/dnf/75bba7ec-6d79-4269-841a-15a32b0747d4_Rocky_Client_Rocky_Client_8-1de048887381d109/packages/katello-host-tools-4.4.0-1.el8.noarch.rpm: Header V4 RSA/SHA512 Signature, key ID df50fffb: NOKEY
Rocky Client 8                                                                                                                                                                                       12 kB/s | 1.6 kB     00:00    
Importing GPG key 0xDF50FFFB:
 Userid     : "ATIX AG <info@atix.de>"
 Fingerprint: CA56 F29E 0006 0B63 D089 0D0F 5BD9 6651 DF50 FFFB
 From       : https://or.supp206.mk/katello/api/v2/repositories/15/gpg_key_content
Key imported successfully
[...]
```
```
zypper -n install katello-host-tools-tracer
[...]
katello-host-tools-4.2.3-5suse1500.noarch.rpm:
    Header V4 RSA/SHA512 Signature, key ID 5bd96651df50fffb: NOKEY
    V4 RSA/SHA512 Signature, key ID 5bd96651df50fffb: NOKEY

warning: /var/tmp/AP_0xxZnKxN/Packages/k/katello-host-tools-4.2.3-5suse1500.noarch.rpm: Header V4 RSA/SHA512 Signature, key ID df50fffb: NOKEY
Looking for gpg key ID DF50FFFB in cache /var/cache/zypp/pubkeys.

New repository or package signing key received:

  Repository:       SLES Client 15SP5
  Key Fingerprint:  CA56 F29E 0006 0B63 D089 0D0F 5BD9 6651 DF50 FFFB
  Key Name:         ATIX AG <info@atix.de>
  Key Algorithm:    RSA 4096
  Key Created:      Wed 24 Aug 2022 14:33:33 CEST
  Key Expires:      (does not expire)
  Rpm Name:         gpg-pubkey-df50fffb-63061a9d



    Note: Signing data enables the recipient to verify that no modifications occurred after the data
    were signed. Accepting data with no, wrong or unknown signature can lead to a corrupted system
    and in extreme cases even to a system compromise.

    Note: A GPG pubkey is clearly identified by its fingerprint. Do not rely on the key's name. If
    you are not sure whether the presented key is authentic, ask the repository provider or check
    their web site. Many providers maintain a web page showing the fingerprints of the GPG keys they
    are using.

Do you want to reject the key, or trust always? [r/a/?] (r): r
katello-host-tools-4.2.3-5suse1500.noarch (SLES Client 15SP5): Signature verification failed [4-Signatures public key is not available]
Abort, retry, ignore? [a/r/i] (a): a
Problem occurred during or after installation or removal of packages:
Installation has been aborted as directed.
```

maximiliankolb: I agree with @tux93 here: If you run "dnf install -y vim" and vim comes from a different package, the "-y" option automatically agrees to have dnf import GPG public keys to verify package signatures. So IMHO it makes sense to have the same expectation on OpenSUSE/SLES on the REX job: import GPG public keys if necessary; do not require additional configuration/user input compared to EL8/9.

Minimal reproducer:
````
$ podman run --rm -it centos/centos:stream9

# dnf install -y epel-release && dnf install -y ranger
...

# dnf install ranger
Is this ok [y/N]: y
...
Importing GPG key 0x3228467C:
 Userid     : "Fedora (epel9) <epel@fedoraproject.org>"
 Fingerprint: FF8A D134 4597 106E CE81 3B91 8A38 72BF 3228 467C
 From       : /etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-9
Is this ok [y/N]: y
...
Complete!
````

If you run "dnf install ranger", you will be prompted twice: installing the package and importing the GPG key.

sbernhard (approved):
Thanks @tux93 and @maximiliankolb for your feedback on this.

adamruzicka:
Thank you @tux93 , @maximiliankolb & @sbernhard !

PR #907, opened 2024-06-20
Fixes #37706 - Add a template that enables download and run of a script

ShimShtein:
The main idea is to execute scripts from the Foreman server, similar to https://github.com/theforeman/foreman/pull/10208.

### DIFF
--- a/app/views/templates/script/run_downloaded_script.erb
+++ b/app/views/templates/script/run_downloaded_script.erb
@@ -1,0 +1,15 @@
+<%#
+kind: job_template
+name: Download and run a script
+model: JobTemplate
+job_category: Commands
+description_format: "Download the script from %{url} and run it locally"
+provider_type: script
+template_inputs:
+- name: url
+  description: URL of the script to download and execute
+  input_type: user
+  required: true
+%>
+
+set -o pipefail && curl -sS '<%= input("url") %>' | bash


app/views/templates/script/run_downloaded_script.erb
```diff
@@ -0,0 +1,15 @@
...
+description_format: "Download the script from %{url} and run it locally"
+provider_type: script
+template_inputs:
+- name: url
+  description: URL of the script to download and execute
+  input_type: user
+  required: true
+%>
+
+set -o pipefail && curl -sS <%= input("url") %> | bash
```
ekohl, line 15: Just to be sure, I'd quote the URL because it may contain `&`
```suggestion
set -o pipefail && curl -sS '<%= input("url") %>' | bash
```

.gitignore
```diff
@@ -15,3 +15,4 @@ Gemfile.lock
 node_modules/
 package-lock.json
 coverage/
+public/
```
ekohl, line 18: This looks unrelated

ekohl (changes requested):

.gitignore
ShimShtein: It's still needed, so split it to a separate PR.

ShimShtein:
Added quotes.

ShimShtein:
@ekohl What do you think, would it be OK, or do we want to be more specific and just reuse the snippet from Foreman?

stejskalleos (approved):
:green_apple: **LGTM**, works as expected

ofedoren:
@ShimShtein, could you please create a RM ticket and attach it to the PR?

ShimShtein:
@ofedoren Done.

ofedoren (approved):
Thanks, all, merging.

PR #912, opened 2024-06-27
Fixes #37590 - Trigger ready batches on planning finished

adamruzicka:
Previously we sent the last batch event followed by a regular TriggerNextBatch event with correct number of batches. The proxy batch triggering action received the TriggerLastBatch event, set an internal flag, triggered a single batch and exited, the other event got discarded. This commit changes the ordering so that the TriggerNextBatch event is sent out first when planning is done.

### DIFF
--- a/app/lib/actions/remote_execution/run_hosts_job.rb
+++ b/app/lib/actions/remote_execution/run_hosts_job.rb
@@ -75,6 +75,7 @@ module Actions
       end
 
       def on_planning_finished
+        trigger_remote_batch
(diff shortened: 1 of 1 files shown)


pondrejk:
Hello, I was able to reproduce the issue using an ansible job against a bunch of fake hosts, I confirm that on stream 63 + packit version the issue no longer occurs, the job is no longer stuck on the first host and completes as expected

pondrejk (approved):

PR #917, opened 2024-08-21
Make theforeman-rubocop dev dep

ofedoren:


ekohl (commented):
What is the precise benefit of this? I can understand you want it in a separate group, but right now it's not a runtime dependency anyway.

ofedoren:
Well, theforeman-rubocop caused few issues for me: 1. It through dependencies requires latest concurrent-ruby, which causes problems with dynflow, which requires a specific version, thus making bundle install to fail under certain circumstances (noticed that during dev setup via forklift) 2. It depends on rather later rubocop, which causes issues for Ruby LSP plugin for VS Code :/ Currently all the rubocop stuff makes me hate it rather utilize...

Since rubocop is mostly used only in test environment, I'd love to exclude it from bundle install by default, but rather with some flag, like with groups. Now I think I should update PR to move it under `:test` group? But I'm not sure if it then will be installed in GHA though...

ekohl:
> Well, theforeman-rubocop caused few issues for me: 1. It through dependencies requires latest concurrent-ruby, which causes problems with dynflow, which requires a specific version, thus making bundle install to fail under certain circumstances (noticed that during dev setup via forklift) 2. It depends on rather later rubocop, which causes issues for Ruby LSP plugin for VS Code :/ Currently all the rubocop stuff makes me hate it rather utilize...

I often have this feeling with Ruby gems. Specifically with concurrent-ruby I wonder how much of our usage could be replaced with Ruby 3's fibers but that's beyond the scope of this PR.

> Now I think I should update PR to move it under `:test` group? But I'm not sure if it then will be installed in GHA though...

I often like to do this, except making it a `:rubocop` group. That way you can (but we don't do this today) run RuboCop in one job and the CI suite in another with minimal dependencies.

adamruzicka:
> I often like to do this, except making it a :rubocop group.

@ofedoren how about this ^?

adamlazik1 (approved):
LGTM, is there anything that prevents us from getting this in?

ekohl:
Leaving for @adamruzicka since he had comments before

adamruzicka:
Thank you @ofedoren !

PR #918, opened 2024-08-22
Fixes #37761 - Pass cockpit access token as fragment

adamruzicka:
instead of a query string to avoid request rewrite failures.


pondrejk (approved):
ack from the functional side, I ran robottelo's cockpit tests against stream build with this patch and they passed

adamruzicka:
This got resolved in https://github.com/theforeman/puppet-foreman/pull/1185

PR #919, opened 2024-09-06
Fixes #37750 - Update to Patternfly 5

MariaAga:
Needs https://github.com/theforeman/foreman/pull/10272

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,20 +19,21 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
-    "@theforeman/builder": ">= 12.0.1",
-    "@theforeman/eslint-plugin-foreman": ">= 12.0.1",
-    "@theforeman/eslint-plugin-rules": ">= 12.0.2",
-    "@theforeman/test": ">= 12.0.1",
-    "@theforeman/vendor-dev": ">= 12.0.1",
+    "@theforeman/builder": ">= 15.0.1",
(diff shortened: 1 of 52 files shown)


sjha4 (approved):
As much as I know about foreman remote execution, the pages are all loading correctly with the PF5 and I was able to set up and run a job and look at details etc without any issues. Ack 👍🏼

adamruzicka (commented):
In general this looks good, pages load, I'm able to click around and don't see any errors popping up in the console.

However the lints are not happy and the new job ivnocation page has some spacing issues around the search bar with this.
image

MariaAga:
Test error is from downgrading pf charts

ofedoren:
> Test error is from downgrading pf charts

Can we somehow re-test this with https://github.com/theforeman/foreman-js/pull/483 to see if the later helps?

adamruzicka:
> Can we somehow re-test this with https://github.com/theforeman/foreman-js/pull/483 to see if the later helps?

I'm afraid I don't see how we could do that in a sane way. Would you suggest holding this until the f-js pr is merged and released?

MariaAga:
Updated to use Olehs fix in foreman-js, seems to work as expected

adamruzicka:
Hrm, the nightly repo still has builder and vendor at 15.0.0, nevermind then

PR #921, opened 2024-09-20
Fixes #37847 - Support Rails 7.0

ofedoren:


lib/foreman_remote_execution/engine.rb
```diff
@@ -31,9 +27,11 @@ class Engine < ::Rails::Engine
     end
 
     initializer 'foreman_remote_execution.require_dynflow', :before => 'foreman_tasks.initialize_dynflow' do |app|
-      ForemanTasks.dynflow.require!
-      ForemanTasks.dynflow.config.queues.add(DYNFLOW_QUEUE, :pool_size => Setting['remote_execution_workers_pool_size']) if Setting.table_exists? rescue(false)
-      ForemanTasks.dynflow.config.eager_load_paths << File.join(ForemanRemoteExecution::Engine.root, 'app/lib/actions')
+      app.reloader.to_prepare do
+        ForemanTasks.dynflow.require!
+        ForemanTasks.dynflow.config.queues.add(DYNFLOW_QUEUE, :pool_size => Setting['remote_execution_workers_pool_size']) if Setting.table_exists? rescue(false)
+        ForemanTasks.dynflow.config.eager_load_paths << File.join(ForemanRemoteExecution::Engine.root, 'app/lib/actions')
```
adamruzicka, line 33: What happens if this is not in a `reloader.to_prepare` block? Doing this each time the application is reloaded would most probably lead to memory bloating? Unless the list of paths to eager load get wiped on reload

ofedoren: For a long time, Zeitwerk didn't like if during initialization time an autoloadable constant was addressed. In this case `Setting`, which is a model constant, thus an autoloadable/reloadable constant.

But now it seems redundant to put this into that block...

app/helpers/job_invocations_helper.rb
```diff
@@ -33,7 +33,7 @@ def preview_hosts(template_invocation)
   end
 
   def collapsed_preview(target)
-    title = target || 'N/A'
+    title = target.to_s || 'N/A'
```
ekohl, line 36: `nil.to` returns an empty string so `||` no longer works. Should we use
```suggestion
    title = (target || 'N/A').to_s
```

ofedoren:
As of now, this PR can be merged at any time it's ready (dropped the last commit) since it doesn't depend on Foreman's PR.

adamruzicka:
Thank you @ofedoren !

PR #922, opened 2024-10-17
Fixes #37922 - Handle case where proxy no longer exists

adamruzicka:
Fixes e3578d6


evgeni:
Should this get a new redmine, given #37493 is closed and released? Other than that LGTM

ofedoren (approved):
Thanks, @adamruzicka and @evgeni !

PR #927, opened 2024-11-11
Ship tests in the gem

adamruzicka:
Originally dropped in b8eb5a1


evgeni (approved):
packit failure expected

PR #934, opened 2024-12-12
Fixes #38090 - Add filtering to job invocation hosts table

kmalyjur:
Sorting and filtering were added to the host table of the new job invocation page.
(Ordering by status is not permitted on purpose.)
Needs [Fixes #38134 - Add custom pagination to table](https://github.com/theforeman/foreman/pull/10413)

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationConstants.js
+++ b/webpack/JobInvocationDetail/JobInvocationConstants.js
@@ -42,6 +42,14 @@ export const STATUS_UPPERCASE = {
   PENDING: 'PENDING',
 };
 
+export const STATUS_TITLES = {
+  ALL_STATUSES: { id: 'all_statuses', title: __('All statuses') },
+  SUCCESS: { id: 'success', title: __('Succeeded') },
+  FAILED: { id: 'failed', title: __('Failed') },
+  PENDING: { id: 'pending', title: __('In Progress') },
+  CANCELLED: { id: 'cancelled', title: __('Cancelled') },
+};
+
 export const DATE_OPTIONS = {
   day: 'numeric',
   month: 'short',
@@ -61,7 +69,7 @@ const Columns = () => {
         return { title: __('Failed'), status: 1 };
       case 'planned':
         return { title: __('Scheduled'), status: 2 };
-      case 'running':
+      case 'running' || 'pending':
         return { title: __('Pending'), status: 3 };
       case 'cancelled':
         return { title: __('Cancelled'), status: 4 };
@@ -84,13 +92,15 @@ const Columns = () => {
       wrapper: ({ name }) => (
         <a href={`${hostDetailsPageUrl}${name}`}>{name}</a>
       ),
+      isSorted: true,
       weight: 1,
     },
-    groups: {
+    hostgroup: {
       title: __('Host group'),
       wrapper: ({ hostgroup_id, hostgroup_name }) => (
         <a href={`/hostgroups/${hostgroup_id}/edit`}>{hostgroup_name}</a>
       ),
+      isSorted: true,
       weight: 2,
     },
     os: {
@@ -100,6 +110,7 @@ const Columns = () => {
           {operatingsystem_name}
         </a>
       ),
+      isSorted: true,
       weight: 3,
     },
     smart_proxy: {
@@ -107,6 +118,7 @@ const Columns = () => {
       wrapper: ({ smart_proxy_name, smart_proxy_id }) => (
         <a href={`/smart_proxies/${smart_proxy_id}`}>{smart_proxy_name}</a>
       ),
+      isSorted: true,
       weight: 4,
     },
     status: {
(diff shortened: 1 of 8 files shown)


webpack/JobInvocationDetail/JobInvocationConstants.js
```diff
@@ -33,6 +33,14 @@ export const STATUS_UPPERCASE = {
   PENDING: 'PENDING',
 };
 
+export const STATUS_TITLES = {
+  ALL_STATUSES: { id: 'all_statuses', title: __('All statuses') },
+  SUCCESS: { id: 'success', title: __('Succeeded') },
+  FAILED: { id: 'failed', title: __('Failed') },
+  PENDING: { id: 'pending', title: __('In Progress') },
+  CANCELLED: { id: 'cancelled', title: __('Canceled') },
```
MariaAga, line 41: Canceled should have 2 Ls -> Cancelled (both right, but we use the 2 L one in REX)

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -81,6 +99,34 @@ const JobInvocationHostTable = ({ id, targeting, finished, autoRefresh }) => {
     `/${controller}/auto_complete_search`
   );
 
+  const constructFilter = () => {
+    const baseFilter = `job_invocation.id = ${id}`;
+    const dropdownFilterClause =
+      selectedFilter && selectedFilter !== 'all_statuses'
+        ? `and job_invocation.result = ${selectedFilter}`
+        : '';
```
MariaAga, line 107: @adamruzicka Is it correct to filter `/api/job_invocations/53/hosts?search=` with  `job_invocation.id = 53 and job_invocation.result = success and (name ~  centos9 )` ? It does provide the correct results

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -68,9 +75,20 @@ const JobInvocationHostTable = ({ id, targeting, finished, autoRefresh }) => {
...
     initialSearchQuery: urlSearchQuery,
   });
+  const updateSearchQuery = searchQuery => {
+    setParamsAndAPI({
+      ...apiOptions,
+      search: searchQuery,
+    });
+    setAPIOptions({
+      ...apiOptions,
+      params: { search: searchQuery },
```
MariaAga, line 88: Should include the filter search as well, for example if you select a status and then search the status is not included in the search but still shown as selected

MariaAga (changes requested):
This works almost perfectly, and its nice that the chart also searchs the table, only a few comments?

webpack/JobInvocationDetail/JobInvocationHostTable.js
adamruzicka: It is a bit counter-intuitive, but this indeed seems to be the right way. I was thinking about dropping the `job_invocation.id = 53`, but looking at the sql that would generate we would get hosts which were in this job, but succeeded in another.

webpack/JobInvocationDetail/JobInvocationConstants.js
kmalyjur: Nice catch :smile:

kmalyjur:
I also removed the custom pagination from the `Table` as @MariaAga found out the usage of `setParams` is enough.

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -98,24 +145,28 @@ const JobInvocationHostTable = ({ id, targeting, finished, autoRefresh }) => {
...
+    }
+  }, [initialFilter]);
+
   const onPagination = newPagination => {
     setParamsAndAPI({
       ...params,
       ...newPagination,
       search: urlSearchQuery,
     });
-  };
```
MariaAga, line 160: this creates a second api call, best to use the new one :)

MariaAga: nope my bad, something else is causing the second api call

adamruzicka:
image

image

It is a bit unfortunate to map "running" to "Pending" and "pending" to "pending" with lowercase p and a different icon. Does the design say anything about this?

kmalyjur:
@adamruzicka We don't have anything about it in the design, actually. You're right, it looks bad, I'll change it.

kmalyjur:
- I added the code @MariaAga wrote for fixing the second unneeded API call
- @adamruzicka `running` and `pending` are both mapped to "Pending" now

The PR should be ready to be merged

MariaAga:
Running jobs dont show hosts, so merging should wait until we see if its the pr or just my env

MariaAga:
`job_invocation.id=ID` shows 0 results if the job is running

kmalyjur:
@MariaAga I can see them :thinking: 
image

adamruzicka:
Sadly, yes, the per-host tasks are not created until the job starts running, which makes it a little bit tricky.

adamruzicka:
js tests are red :/

MariaAga:
js test error make no sense, and dont happen locally, maybe re run them?

adamruzicka (approved):

adamruzicka:

PR #935, opened 2024-12-13
Fixes #38101 - Align search bar with hosts list on job details page

nadjaheitmann:
Before:
image

After:
image


nadjaheitmann:
@adamruzicka Do you mind taking a look at this one? It is a really tiny one.

adamruzicka (approved):
Definitely looks better with this, although I wonder why it was there in the first place. @MariaAga any ideas?

MariaAga:
its usually used by erb pages only, probably was copied accidentally, and I changed some css rules for `title_filter` in foreman 3.8 so it probably broke there ( I only have foreman 3.8 to check, and not 3.7)

PR #941, opened 2025-01-28
Fixes #38090 - Add filtering to job invocation hosts table

adamruzicka:
https://github.com/theforeman/foreman_remote_execution/pull/934 with some minor improvements on top. Searching inside scheduled jobs still doesn't really work, but neither does it in the old page so let's leave that one for later.


adamruzicka:
Closing in favor of the original one

PR #942, opened 2025-01-28
Refs #38130 - fix job details layout small screen

MariaAga:
small fix for  https://github.com/theforeman/foreman_remote_execution/pull/939
JS code changes is just moving PageSection under pagelayout and adding a class name
before: 
Screenshot From 2025-01-28 12-54-36
after: 
image

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationDetail.scss
+++ b/webpack/JobInvocationDetail/JobInvocationDetail.scss
@@ -39,9 +39,11 @@
   }
 }
 .job-additional-info {
-  padding-top: 0;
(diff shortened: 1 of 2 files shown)


adamruzicka:
Could we also hide the vertical divider if we don't show the chart and the details side by side?

MariaAga:
I dont think that would be a simple fix since the width of the chart might change with languages, and we dont control when the details are moved down

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #946, opened 2025-01-31
Fixes #38184 - auto scroll to bottom job details

MariaAga:
auto scrolls to the bottom of the host job output once the input changes, only if the user is already at the bottom of the page, to avoid page jumps

### DIFF
--- a/webpack/JobInvocationDetail/TemplateInvocationComponents/OutputCodeBlock.js
+++ b/webpack/JobInvocationDetail/TemplateInvocationComponents/OutputCodeBlock.js
@@ -1,4 +1,9 @@
-import React from 'react';
+import React, {
+  useEffect,
+  useLayoutEffect,
+  useState,
+  useCallback,
+} from 'react';
 import PropTypes from 'prop-types';
 import { Button } from '@patternfly/react-core';
 import { translate as __ } from 'foremanReact/common/I18n';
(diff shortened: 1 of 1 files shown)


webpack/JobInvocationDetail/TemplateInvocationComponents/OutputCodeBlock.js
```diff
@@ -55,6 +55,47 @@ export const OutputCodeBlock = ({ code, showOutputType, scrollElement }) => {
...
+  const onClickScrollToTop = useCallback(() => {
+    scrollElementSeleceted().scrollTo(0, 0);
+  }, [scrollElementSeleceted]);
+
+  const onClickScrollToBottom = useCallback(() => {
+    scrollElementSeleceted().scrollTo(0, scrollElementSeleceted().scrollHeight);
+  }, [scrollElementSeleceted]);
+
+  const [isScrolledBottom, setIsScrolledBottom] = useState();
+  useEffect(() => {
```
MariaAga, line 72: We cant check "isScrolledBottom" after the new output has been received already, so we need to keep checking if the user scrolled to the bottom

webpack/JobInvocationDetail/TemplateInvocationComponents/OutputCodeBlock.js
```diff
@@ -55,6 +55,47 @@ export const OutputCodeBlock = ({ code, showOutputType, scrollElement }) => {
...
+      element.addEventListener('scroll', onScroll);
+    }
+    return () => {
+      if (element) {
+        element.removeEventListener('scroll', onScroll);
+      }
+    };
+  }, [scrollElementSeleceted]);
+
+  useEffect(() => {
```
kmalyjur, line 93: Could the `useLayoutEffect` be used here to prevent flickering when the page scrolls?

kmalyjur (approved):
Works great!

webpack/JobInvocationDetail/TemplateInvocationComponents/OutputCodeBlock.js
MariaAga: Yes, thats much better, thanks for the suggestion!

kmalyjur:
@adamruzicka could you please merge it? Or should the "need to investigate different output flashing when row is expanded" bug be included in this PR? 🤔

adamruzicka:
Could we also handle this case?

https://github.com/user-attachments/assets/0f6d922f-e9ba-48a0-a02d-e113ff21a211

If the container becomes scrollable, I would expect it to scroll to the bottom

kmalyjur:
> Could we also handle this case?

@adamruzicka good idea.
@MariaAga, would you be able / do you have time to include it?

MariaAga:
Thanks! 
set auto scroll to be true on default, rebased, and added ` if (scrollElementSeleceted()) ` to avoid errors in edge cases

MariaAga:
Fixed, thanks for noticing!

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga & @kmalyjur !

PR #947, opened 2025-02-06
Fixes #38187 - Fix kebab items in job invocation

Lukshio:


kmalyjur (changes requested):
The "Rerun" and "Task Details" links should be opened in the same tab and it's nice to have the URL preview (in the bottom-left corner) shown when hovering over it.

You could take inspiration for example from this :slightly_smiling_face: 
https://github.com/theforeman/foreman/blob/b9329fc771be7401324032071aa6a2279439ab5a/webpack/assets/javascripts/react_app/components/HostsIndex/index.js#L316-L326 
(dropdown on the _/new/hosts_ page)

The "Cancel Task" and "Abort task" actions are unfortunately not working in your solution.
Also, not something that was expected from you, but could you please capitalize the 't' in "task"?

Lukshio:
According to the request the dropdown elements was changed to link, also onClick function was restored. New scss file has been added to the folder to change the hover color of links to match other dropdown design.

t is Task is now capitalized.

kmalyjur (approved):
Thank you for the changes, LGTM
The `index.scss` file could be named after the `TemplateActionButtons.js`, but I might be nitpicking.

kmalyjur:
@adamruzicka could you please merge it?

adamruzicka (approved):

adamruzicka:
Thank you @Lukshio & @kmalyjur !

webpack/JobInvocationDetail/TemplateInvocationComponents/TemplateActionButtons.js
```diff
@@ -55,7 +56,7 @@ const actions = ({
   },
   abort: {
     name: 'template-invocation-abort-job',
-    text: __('Abort task'),
+    text: __('Abort Task'),
```
ekohl, line 59: Late to the party, but I don't think we use Title Case anymore. The reason I now see it is because it changed a string in the translations. Should we revert this specific line?

adamruzicka: Do we state somewhere what casing should be used?

MariaAga: https://www.patternfly.org/ux-writing/capitalization

ekohl: https://github.com/theforeman/foreman_remote_execution/pull/959

PR #949, opened 2025-02-11
Fixes #38202 - allow host search in job feature run

MariaAga:
For new hosts index page, use the search from the url when its used like: /job_invocations?feature=ansible_run_host&search=id%20^%20(63,50)
for example for the "run all ansible jobs" button when selecting hosts.
Without this the button doesnt run anything.
I didnt change the ui url calls to api since then we dont have the information on when to redirect to the job form or just run it.

### DIFF
--- a/app/controllers/job_invocations_controller.rb
+++ b/app/controllers/job_invocations_controller.rb
@@ -171,7 +171,7 @@ class JobInvocationsController < ApplicationController
       inputs = params[:inputs].permit!.to_hash if params.include?(:inputs)
       JobInvocationComposer.for_feature(
         params[:feature],
-        params[:host_ids],
+        params[:search].presence || params[:host_ids],
         inputs
       )
     else


adamruzicka (commented):
There is [a bit of a hack in HostIdsTranslator](https://github.com/theforeman/foreman_remote_execution/blob/57854297818e0e3aa314a1e7f84524baa2f2f92b/app/models/job_invocation_composer.rb#L300) which allows you to pass a search query inside `host_ids` and the it will get honored.

Could we pass `params[:search]` instead of `params[:host_ids]` in the controller and leave the composer untouched?

Also, does this mean this never worked?

MariaAga:
the calls in the ui are:
```js 

      ? `/job_invocations?feature=${feature}&host_ids%5B%5D=${hostId}`
      : `/job_invocations?feature=${feature}&search=${hostSearch}`
```
so do you mean to send both in search instead? and then do
```rb
JobInvocationComposer.for_feature(
        params[:feature],
        params[:search]
        inputs
      )
 ```
 ?
 And yes, this has probably never worked in the new hosts page, I only checked that it didnt work before the jquery update as well. the original pr only checked that the url was used but not that the job has started.

**Comment by adamruzicka** [comment]
No, not exactly. The ui calls would stay the same, but in the controller we'd do something like
```ruby
JobInvocationComposer.for_feature(
  params[:feature],
  params[:search].blank? ? params[:host_ids] : params[:search],
  inputs
)
```

MariaAga:
Thanks! updated to only change the controller

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #955, opened 2025-03-03
Fixes #38258 - Fix resolved date of job invocation

kmalyjur:
Fixed reporting of the job status should rely on the local browser time

### DIFF
--- a/app/views/api/v2/job_invocations/main.json.rabl
+++ b/app/views/api/v2/job_invocations/main.json.rabl
@@ -31,7 +31,7 @@ child :targeting do
 end
 
 child :task do
-  attributes :id, :state
+  attributes :id, :state, :started_at
 end
 
 child @template_invocations do
--- a/webpack/JobInvocationDetail/index.js
+++ b/webpack/JobInvocationDetail/index.js
@@ -1,31 +1,33 @@
-import PropTypes from 'prop-types';
-import React, { useEffect, useState } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
 import {
   Divider,
   Flex,
   PageSection,
   PageSectionVariants,
 } from '@patternfly/react-core';
-import { translate as __, documentLocale } from 'foremanReact/common/I18n';
-import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
 import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import { translate as __, documentLocale } from 'foremanReact/common/I18n';
 import { stopInterval } from 'foremanReact/redux/middlewares/IntervalMiddleware';
+import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
+import PropTypes from 'prop-types';
+import React, { useEffect, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { JobAdditionInfo } from './JobAdditionInfo';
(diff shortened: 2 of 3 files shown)


webpack/JobInvocationDetail/index.js
```diff
@@ -56,14 +58,25 @@ const JobInvocationDetailPage = ({
   const handleFilterChange = filter => {
     setSelectedFilter(filter);
   };
+  const getConvertedDate = date =>
+    new Date(
+      date
+        .replace(' ', 'T')
+        .replace(/(-\d{2})(\d{2})$/, '$1:$2')
+        .replace(' ', '')
```
adamruzicka, line 66: Iirc the date string you get from the api should be iso8601-formatted. Is this necessary?

webpack/JobInvocationDetail/index.js
```diff
@@ -56,14 +58,25 @@ const JobInvocationDetailPage = ({
...
   if (startAt) {
-    // Ensures date string compatibility across browsers
-    const convertedDate = new Date(startAt.replace(/[-.]/g, '/'));
-    isAlreadyStarted = convertedDate.getTime() <= new Date().getTime();
-    formattedStartDate = convertedDate.toLocaleString(
+    const convertedStartAt = getConvertedDate(startAt);
+    // eslint-disable-next-line camelcase
+    const convertedResolvedAt = targeting?.resolved_at
+      ? getConvertedDate(targeting.resolved_at)
+      : new Date();
```
adamruzicka, line 76: While this could work, it would be better to rely on the information about the underlying tasks - if you pull out `started_at` in the destructuring on L39, that should be the most accurate.

webpack/JobInvocationDetail/index.js
kmalyjur: I do that because I want to use the `getTime()` function later, which can't be used on the default `startAt` format (`Uncaught TypeError: startAt.getTime is not a function`) and also so that the displayed date is in the format it is (`Started at: 17 Feb 2025, 11:15 CET` using `toLocaleString()`).

webpack/JobInvocationDetail/index.js
kmalyjur: Do you mean using `started_at` from `task`?

webpack/JobInvocationDetail/index.js
kmalyjur: You're right, thank you, I overcomplicated it.

kmalyjur:
@adamruzicka Thank you for your review, is it okay now?
Idk about the failing tests - they're passing locally.

app/views/api/v2/job_invocations/main.json.rabl
```diff
@@ -31,7 +31,7 @@ child :targeting do
 end
 
 child :task do
-  attributes :id, :state
+  attributes :id, :state, :started_at
```
adamruzicka, line 34: Why do we need both this and https://github.com/theforeman/foreman_remote_execution/pull/955/files#diff-727d3d82e0838cb3e4d0e963dd2a03203180ab4eb916a07104e861116becf298R22 ?

kmalyjur: @adamruzicka fixed

adamruzicka (approved):

adamruzicka:
Thank you @kmalyjur !

PR #956, opened 2025-03-04
Fixes #38263 - Assign correct taxonomies on template import

adamlazik1:
Currently, when importing a job template, taxonomies of the request are honored instead of taxonomies of the template itself. For an example see the Redmine issue.

The superclass [Template](https://github.com/theforeman/foreman/blob/develop/app/models/template.rb) of the class `JobTemplate` only extracts taxonomies from metadata on import if the option `:associate` is present, but this option is never accepted by the endpoint or assigned during the processing of the import request. I have added this option as a part of the `template_import_options` param group to stay consistent with imports of other template types.

The new behavior allows user to choose if the taxonomies of the request or the taxonomies in the template metadata should be honored based on the `options[associate]` param

### DIFF
--- a/app/controllers/api/v2/job_templates_controller.rb
+++ b/app/controllers/api/v2/job_templates_controller.rb
@@ -23,10 +23,13 @@ module Api
       end
 
       api :POST, '/job_templates/import', N_('Import a job template from ERB')
+      param_group :template_import_options, ::Api::V2::BaseController
       param :template, String, :required => true, :desc => N_('Template ERB')
       param :overwrite, :bool, :required => false, :desc => N_('Overwrite template if it already exists')
       def import
         options = params[:overwrite] ? { :update => true } : { :build_new => true }
+        # the Template superclass expects hash indexed by symbols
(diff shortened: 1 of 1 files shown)


app/controllers/api/v2/job_templates_controller.rb
```diff
@@ -27,6 +27,7 @@ def index
       param :overwrite, :bool, :required => false, :desc => N_('Overwrite template if it already exists')
       def import
         options = params[:overwrite] ? { :update => true } : { :build_new => true }
+        options[:associate] = 'new'
```
adamruzicka, line 30: Now that I've had time to look at what we do for other template types, I'd say it would be better to stay consistent with what we do there - adding the template_import_options param group here and letting the user control how it behaves rather than going our own way here.

adamlazik1: Applied and updated description.

adamruzicka (approved):

adamruzicka:
Thank you @adamlazik1 !

PR #959, opened 2025-03-18
Use sentence case for abort button task

ekohl:
The patternfly style guide prefers Sentence case instead of Title Case.

Link: https://www.patternfly.org/ux-writing/capitalization
Fixes: 35cd0649f681 ("Fixes #38187 - Fix kebab items in job invocation (#947)")

### DIFF
--- a/webpack/JobInvocationDetail/TemplateInvocationComponents/TemplateActionButtons.js
+++ b/webpack/JobInvocationDetail/TemplateInvocationComponents/TemplateActionButtons.js
@@ -56,7 +56,7 @@ const actions = ({
   },
   abort: {
     name: 'template-invocation-abort-job',
-    text: __('Abort Task'),
+    text: __('Abort task'),
     permission: permissions.cancel_job_invocations,
     onClick: () => {
       dispatch(


adamruzicka (approved):

adamruzicka:
We want to pick this for 3.14, right?

ekohl:
Yes, but for 3.14 we should also figure out the translation situation. 504e9aee49aa334d2cee89aa1e921dd1af28d4a0 changed strings which dropped the translations below 100% (which is also how I found this).

Now that I look closer, I think we should also revert ed4a156ff2cbec50d7222d9e4b75aba9333b5a05.

ekohl:
Ok, fair enough.

adamruzicka:
With that being said, I'd be willing to keep it in master, but pull both of those changes out from the 3.14-compatible stable branch so that the translations are happy. Would that be acceptable?

ekohl:
My suggestion is to update the `.pot` file in master now so Transifex is updated. Then I'll work with the downstream translators to get it fixed, we pull them back in and pull the specific commits into a 3.14-compatible stable branch. https://github.com/theforeman/foreman_remote_execution/commit/504e9aee49aa334d2cee89aa1e921dd1af28d4a0 is already in 3.14 anyway and reverting that feels like the wrong solution because the new texts are clearer IMHO.

ekohl:
Closing the loop: https://github.com/theforeman/foreman_remote_execution/pull/960 did that update. I'll wait a big and do the downstream dance

PR #961, opened 2025-03-25
Fixes #38299 - fix used proxy link in per-host details

Lukshio:


webpack/JobInvocationDetail/TemplateInvocation.js
```diff
@@ -150,10 +156,14 @@ export const TemplateInvocation = ({
...
       {!isInTableView && (
-        <div>
-          {__('Target:')}{' '}
-          <a href={`${hostDetailsPageUrl}${hostName}`}>{hostName}</a>
-        </div>
+        <>
+          <Title ouiaId="template-links-title" headingLevel="h4">
+            {__('Target:')}{' '}
+            <a href={`${hostDetailsPageUrl}${hostName}`}>{hostName}</a>{' '}
+            {__('using Smart Proxy:')}{' '}
```
adamruzicka, line 163: Old:
image

New:
image

Comparing the two side by side, I'd prefer the text to not be bold and I'd probably omit the colon after "Smart Proxy", unless it's dictated by design

Lukshio: Sure, I changed Title to standard div, so the text isn't bold anymore.

adamruzicka (approved):

adamruzicka:
Thank you @Lukshio !

app/controllers/template_invocations_controller.rb
```diff
@@ -38,9 +38,15 @@ def show_template_invocation_by_host
       }
     end
 
+    smart_proxy = @template_invocation.smart_proxy
+    proxy = {
+      name: smart_proxy.name,
+      href: smart_proxy_path(smart_proxy),
+    }
```
kmalyjur, line 45: @adamruzicka, does it also work for you for hosts that don't use the smart proxy? I get an error regarding undefined `smart_proxy.name` here.

adamruzicka: Ah, good catch, I just looked at some static jobs that I had around. Should be addressed in https://github.com/theforeman/foreman_remote_execution/pull/967

Lukshio: Thank you for noticing and fix. @kmalyjur @adamruzicka

PR #963, opened 2025-03-26
Fixes #38309 - Fix kebab items buttons not working on job invocations…

Lukshio:


Lukshio:
It looks like it is PatternFly accessibility feature allowing keyboard control over the options, it overrides the default focus outline. So I think this is the correct working of the component.

adamruzicka (approved):

adamruzicka:
Thank you @Lukshio !

PR #964, opened 2025-03-26
use @theforeman/actions for JS tests too

evgeni:


.github/workflows/js_ci.yml
```diff
@@ -1,32 +1,14 @@
...
-          node-version:  ${{ matrix.node-version }}
-      - name: Npm install
-        run: |
-          npm install
-      - name: Run plugin linter
-        run: |
-          npm run lint
-      - name: Run custom plugin linter
-        run: |
-          npm run lint:custom
```
evgeni, line 29: this is currently not executed by the shared action, but I thought `tfm-lint` would already run eslint?

adamruzicka: @MariaAga ^ opinions?

MariaAga: I was gonna do this pr anyway after I make changes in the shared actions, which will run the lint custom.
I'm making changes in the shared actions after: https://github.com/theforeman/foreman/pull/10239

adamruzicka:
Thank you @evgeni & @MariaAga !

PR #965, opened 2025-03-31
Fixes #38327 - build job names with host names

MariaAga:
So its easier to see from the ui which hosts (/groups/collections) the job runs on

### DIFF
--- a/webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
+++ b/webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
@@ -1,12 +1,31 @@
 export const buildHostQuery = (selected, search) => {
+  const nameEscape = name => `"${name.replaceAll('"', '\\"')}"`;
   const { hosts, hostCollections, hostGroups } = selected;
-  const hostsSearch = `id ^ (${hosts.map(({ id }) => id).join(',')})`;
-  const hostCollectionsSearch = `host_collection_id ^ (${hostCollections
-    .map(({ id }) => id)
-    .join(',')})`;
-  const hostGroupsSearch = `hostgroup_id ^ (${hostGroups
-    .map(({ id }) => id)
-    .join(',')})`;
+  const MAX_NAME_ITEMS = 50;
+  let hostsSearch;
+  if (hosts.length < MAX_NAME_ITEMS)
+    hostsSearch = `name ^ (${hosts.map(({ name }) => name).join(',')})`;
+  else hostsSearch = `id ^ (${hosts.map(({ id }) => id).join(',')})`;
+
+  let hostCollectionsSearch;
+  if (hostCollections.length < MAX_NAME_ITEMS)
+    hostCollectionsSearch = `host_collection ^ (${hostCollections
+      .map(({ name }) => nameEscape(name))
+      .join(',')})`;
+  else
+    hostCollectionsSearch = `host_collection_id ^ (${hostCollections
+      .map(({ id }) => id)
+      .join(',')})`;
+
+  let hostGroupsSearch;
+  if (hostCollections.length < MAX_NAME_ITEMS)
+    hostGroupsSearch = `hostgroup_fullname ^ (${hostGroups
+      .map(({ name }) => nameEscape(name))
+      .join(',')})`;
+  else
+    hostGroupsSearch = `hostgroup_id ^ (${hostGroups
+      .map(({ id }) => id)
+      .join(',')})`;
   const queryParts = [
     hosts.length ? hostsSearch : false,
     hostCollections.length ? hostCollectionsSearch : false,


MariaAga:
@adamruzicka Can we assume that name will always exist for host(/groups/collections)?

MariaAga:
> As a middle-ground, maybe we could use names or ids based on their count? 

so like, 15 hosts - query by name , more than that use ids?

adamruzicka:
Something like that, yes. Looking at the BZs I linked, one of them mentions this started to become a problem above ~240 hosts. The exact number also depends on how long the hostnames are, but we should be able to safely go a bit higher than 15. 50, maybe even 100?

ofedoren:
Just two cents: do we really need to build the query with names? Or should we rather show to the user names, but use the ids?

What prevents us to get a name (or rather names in bulk) from the user, internally translate this to ids, sent the ids and then show names again somewhere in UI when the request is processed? Kinda what hammer does...

MariaAga:
>  The exact number also depends on how long the hostnames are, but we should be able to safely go a bit higher than 15. 50, maybe even 100?

If its mostly for showing in the table there isnt a lot of space anyway in the tooltip? 
so can we say 50, and if  we ever rewrite the table we'll add a better preview? :) 
image

adamruzicka:
> so can we say 50

Yes, let's go with that

nadjaheitmann:
I tested the 'basic' case with 10 hosts which worked fine. Do you need any help/tests with this to finish it up @MariaAga ?

MariaAga:
@nadjaheitmann Thanks! I'm just finishing another pr and then this one is next on my list :) I need to do the limit for 50 hosts

nadjaheitmann:
> @nadjaheitmann Thanks! I'm just finishing another pr and then this one is next on my list :) I need to do the limit for 50 hosts

Cool, feel free to ping me for tests once you are done :)

MariaAga:
This resolves the issue when the job receives hosts from the host_ids param (or by selecting hosts in the wizard), like in the legacy all hosts page, in the new page they send over a search query, I opened https://projects.theforeman.org/issues/38420 (and a Jira for the team that works on the all hosts page)

nadjaheitmann:
HI @MariaAga , I have tested this on my test system. The search query above 50 hosts is shown with IDs, below it shows the query with names. When I hover over the search query, it shows the query with host names. Not sure if this is a flaw in my setup or really a bug, though :thinking:

MariaAga:
> When I hover over the search query 

Can you share a screenshot? @nadjaheitmann

nadjaheitmann:
> > When I hover over the search query
> 
> Can you share a screenshot? @nadjaheitmann

image

As I said, not sure if this is because of my system. But I would expect the tooltip to show the actual query wit ids.

nadjaheitmann:
Hi @MariaAga, I re-did the testing on another system, now it looks fine (I scaled down the number of hosts to 10 for ease of testing):
 
image

adamruzicka (commented):
I'm afraid we'll need to do some escaping. At least host collections can have almost anything in their names (including spaces, commas and quotes)

MariaAga:
escaping by wrapping the name with " or escaping the spaces? cause I think if we escape the spaces the query is wrong

MariaAga:
yup should be back now

webpack/JobWizard/steps/HostsAndInputs/buildHostQuery.js
```diff
@@ -1,12 +1,31 @@
...
-    .map(({ id }) => id)
-    .join(',')})`;
-  const hostGroupsSearch = `hostgroup_id ^ (${hostGroups
-    .map(({ id }) => id)
-    .join(',')})`;
+  const MAX_NAME_ITEMS = 50;
+  let hostsSearch;
+  if (hosts.length < MAX_NAME_ITEMS)
+    hostsSearch = `name ^ (${hosts.map(({ name }) => name).join(',')})`;
+  else hostsSearch = `id ^ (${hosts.map(({ id }) => id).join(',')})`;
```
adamruzicka, line 8: Could we make this whole thing a bit dryer? Something along the lines of this, but applied everywhere?

```suggestion
  const buildSearch = (collection, search_key) => {
    const [real_search_key, attr] = (collection.length < MAX_NAME_ITEMS) ? [search_key, "name"] : ["id", "id"];
    let values = collection.map((item) => item[attr]);
    if (attr === "name") values = values.map((v) => nameEscape(v));
    return `${real_search_key} ^ (${values.join(',')})`;
  }
  
  hostsSearch = buildSearch(hosts, "name");
```

MariaAga: its not really readable, can we not? especially since they live so close together and are only 3?

adamruzicka: I admit I went a bit overboard with code golfing there. 3 repetitions is where I usually draw the line, but I'm willing to let it slide this time

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #967, opened 2025-04-01
Refs #38299 - Address a case where proxy was not determined

adamruzicka:


kmalyjur:
Is this the expected output?
image

adamruzicka:
Expected in which situation?

kmalyjur:
> Expected in which situation?

When there is no smart proxy

adamruzicka:
Yes

kmalyjur:
Nice, then it works, thank you.

kmalyjur (approved):
Nice, then it works, thank you.

PR #968, opened 2025-04-03
Fixes #38306 - dont load all outputs in job details

MariaAga:
Created list_template_invocations_task_by_hosts so we can still see the kebab actions for all hosts without needed to expend anything


app/controllers/template_invocations_controller.rb
```diff
@@ -23,6 +23,30 @@ def show
...
+    hosts.each do |host|
+      template_invocation = @job_invocation.template_invocations.find { |template_inv| template_inv.host_id == host.id }
+      next unless template_invocation
+      template_invocation_task = template_invocation.run_host_job_task
+      template_invocations_task_by_hosts[host.id] =
+        {
+          :task_id => template_invocation_task.id,
+          :task_cancellable => template_invocation_task.cancellable?,
+          :permissions => {
+            :view_foreman_tasks => User.current.allowed_to?(:view_foreman_tasks),
```
adamruzicka, line 38: We could be more specific and instead of asking "Can the user view tasks in general?" we could ask "Can the user view this host's task?". 

```suggestion
            :view_foreman_tasks => User.current.can?(:view_foreman_tasks, template_invocation_task,
```

Alternatively, in the old ui we use this
```
authorized_for(:permission => :view_foreman_tasks, :auth_object => task)
```

app/controllers/template_invocations_controller.rb
```diff
@@ -23,6 +23,30 @@ def show
...
+            :cancel_job_invocations => User.current.allowed_to?(:cancel_job_invocations),
```
adamruzicka, line 39: Same thing as above, in the old ui we use
```
authorized_for(:permission => :cancel_job_invocations, :auth_object => job_invocation)
```

config/routes.rb
```diff
@@ -25,6 +25,7 @@
   match 'experimental/job_invocations_detail/:id', to: 'react#index', :via => [:get], as: 'new_job_invocation_detail'
   match 'job_invocations_detail/:id/host_invocation/:host_id', to: 'react#index', :via => [:get], as: 'new_job_invocation_detail_by_host'
   get 'show_template_invocation_by_host/:host_id/job_invocation/:id', to: 'template_invocations#show_template_invocation_by_host'
+  get 'list_template_invocations_task_by_hosts', to: 'template_invocations#list_template_invocations_task_by_hosts'
```
adamruzicka, line 28: disclaimer: I know there's a precedent for this kind of naming and I've approved the previous instance in the past, but back then we were under time pressure while now we have time.

Our api tries (and usually succeeds) in being REST-ful. Among other things, there are certain expectations about how routes should be formed. The rails routing guide[1] is a nice introduction. 

In this case, template invocations cannot exist on their own so the path to get them shouldn't be top-level. They always have to belong to a job invocation, so endpoints for retrieving information about template invocations should be nested under a path which determines a job invocation, so something like `/job_invocations/<JID>/template_invocations?search=...`. If I'm reading it correctly then in here you're passing the job invocation id as a parameter, which also works, but still the pat-based approach would be preferable.

Additionally, I'm afraid users don't really care about template invocations, they care about "hosts in a job" and maybe it wouldn't hurt to indulge them in that regard. Maybe something like `/job_invocations/<JID>/hosts?search=...` would be a better fit? The answer would include all the task-level details and permissions and so on, without exposing our internals too much. Also, by mentioning "hosts" in the path, you set expectations about the search.

[1] - https://guides.rubyonrails.org/routing.html

adamruzicka (changes requested):
As far as the functionality goes, I have no concerns, although it could use some polishing.

config/routes.rb
adamruzicka: > is that necessary in this case?

Was that meant to react to my whole comment or just parts of it? None of it is strictly necessary and the last paragraph probably falls into the nice-to-have bucket, but we should at least be consistent, even if just for our own sake.

> I guess you also agreed on a similar case before just because it's not an API endpoint?

That's true, but I'd also say we can do better if we have time?

> if we want to explicitly expose this action to the users, I'd vote for a proper API endpoint

I wasn't proposing to make this a "proper, user-facing and stable" api endpoint.  But it is still an api endpoint of sorts, even if it is just frontend-facing and sort of private.

> this seems okay?

Well, it works so there's that, but being a bit more consistent would be nice.

ofedoren: Nevermind me, somehow with all these workarounds for UI I forgot to follow Rails' how-tos

MariaAga: re-remembering this pr:
` /job_invocations/<JID>/hosts?` already exists, and gives info about the host and some of its details, this endpoint only gives a list of hosts ids and their task ids 
> The answer would include all the task-level details and permissions and so on, without exposing our internals too much.

Could you explain what you mean here?

adamruzicka: > is the issue just with the url or also with the method name list_template_invocations_task_by_hosts ?

I would say both. Usually the route and the corresponding method are named in a similar way so if we change one, we should also change the other.

> `/job_invocations/<JID>/hosts`? already exists

Where? I'm aware of `/api(/:apiv)/job_invocations/:id/hosts`, but that's slightly different.

> Could you explain what you mean here?

Something along the lines of this. Just don't explicitly call out template invocations in there

```json
{
  "result": {
    "hosts": [
      { "hostname": "foo.example.com",
        "id": 7,
        "permissions": { "view_task": true, "cancel_job": true, "execute_job": true },
        "task": { "cancellable": true, id: "12345-..." }
      },
      ...
    ]
  }
}
```

MariaAga:
@kmalyjur Can you review the ui part of this?

kmalyjur (approved):
The UI part is great, thank you.
@adamruzicka, I think it can be merged?

config/routes.rb
```diff
@@ -25,6 +25,7 @@
   match 'experimental/job_invocations_detail/:id', to: 'react#index', :via => [:get], as: 'new_job_invocation_detail'
   match 'job_invocations_detail/:id/host_invocation/:host_id', to: 'react#index', :via => [:get], as: 'new_job_invocation_detail_by_host'
   get 'show_template_invocation_by_host/:host_id/job_invocation/:id', to: 'template_invocations#show_template_invocation_by_host'
+  get 'job_invocations/:id/hosts', to: 'template_invocations#list_jobs_hosts'
```
ofedoren, line 28: (not blocking, just two cents)
Is it just me or this is rather confusing? Why do we keep mention `job` but use `template` across at least this PR? I mean, I probably get that we want all the information during 1 call instead of multiple ones, but if we want Rails' way, should we rather be more explicit and place/use things where they are expected to be?..

MariaAga: Thanks, moved

kmalyjur:
Hi @adamruzicka @MariaAga, just checking in on this PR - any updates?

adamruzicka (commented):
If I expand multiple rows, only the last one that I expand keeps refreshing, the others just stop refreshing. I have a feeling that https://github.com/theforeman/foreman_remote_execution/pull/978 might help with that. If that is the case, then I'd rather have the other pr merged first.

adamruzicka:
Needs a rebase

MariaAga:
rebased, added a logic (`(Object.keys(response).length === 0 || autoRefresh)`) so it wont do an api call again if a user expands, un-expands and expand the same row again

webpack/JobInvocationDetail/TemplateInvocation.js
```diff
@@ -54,28 +59,39 @@ export const TemplateInvocation = ({
...
-  const { finished, auto_refresh: autoRefresh } = response;
+
+  const status = useSelector(selectTemplateInvocationStatus(hostID));
+  const response = useSelector(selectTemplateInvocation(hostID));
+  const finished = response.finished || true;
+  const autoRefresh = response.auto_refresh || false;
+  const dispatch = useDispatch();
 
   useEffect(() => {
+    const getData = async () => {
```
ofedoren, line 75: Is it okay to define a function in `useEffect` hook?..

webpack/JobInvocationDetail/TemplateInvocation.js
```diff
@@ -54,25 +59,34 @@ export const TemplateInvocation = ({
...
-    headers: { Accept: 'application/json' },
-    handleError: () => {
-      if (intervalId) clearInterval(intervalId);
-    },
-  });
-  const { finished, auto_refresh: autoRefresh } = response;
-
+  const status = useSelector(selectTemplateInvocationStatus(hostID));
+  const response = useSelector(selectTemplateInvocation(hostID));
+  const finished = response.finished || true;
```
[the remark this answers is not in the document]
ofedoren: I guess this is valid?..

ofedoren (approved):
Thanks, @MariaAga, just two ignorable nits, but otherwise LGTM and seems to work as expected.

@adamruzicka, any concerns?

adamruzicka:
On my end it doesn't seem to be refreshing. When I expand the row, the data gets loaded once, but that's it. To get a fresh view I have to collapse the row and expand it again

webpack/JobInvocationDetail/TemplateInvocation.js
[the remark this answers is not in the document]
adamruzicka: It also seems the refresh issue I'm seeing

webpack/JobInvocationDetail/TemplateInvocation.js
MariaAga: it is recommended

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga !

PR #969, opened 2025-04-08
Use reusable Foreman plugin JS tests action

ekohl:


ekohl:
Oops, consider this a duplicate because I didn't check before opening a PR.

PR #971, opened 2025-05-08
Support Ruby 3.1.0

AltiUP:
Add check and modify processing


ofedoren (commented):
Thanks, @AltiUP !

I think we can drop the version check, since we already call `save_load` with `permitted_classes` in core: https://github.com/theforeman/foreman/blob/develop/app/models/setting.rb#L112

PR #972, opened 2025-05-15
Fixes #38428 - don't add extra column header for table row expansion

jeremylenz:
After https://github.com/theforeman/foreman/pull/10523 we no longer need to pass an extra column for the table row expansion.


kmalyjur:
Sorry for not making a review, thanks for the change.

PR #973, opened 2025-05-16
Fixes #38461 - Replace SSHExecutionProvider with ScriptExecutionProvider

jturel:
I was working in Katello and found that `reload!` from the console is broken (not returning `true`) for a number of reasons. One of them led to this plugin in relation to some aliasing that was introduced in #704 

The error was:

```
TypeError: superclass mismatch for class ScriptExecutionProvider
from /home/vagrant/foreman/.vendor/ruby/3.0.0/gems/foreman_remote_execution-16.0.3/app/models/ssh_execution_provider.rb:1:in `<main>'
```

This PR removes the `SSHExecutionProvider` constant while still providing 'ssh' as a provider by using `ScriptExecutionProvider` directly

### DIFF
--- a/lib/foreman_remote_execution/engine.rb
+++ b/lib/foreman_remote_execution/engine.rb
@@ -71,10 +71,10 @@ module ForemanRemoteExecution
                 full_name: N_('Effective User')
               setting 'remote_execution_effective_user_method',
                 type: :string,
-                description: N_('What command should be used to switch to the effective user. One of %s') % ::SSHExecutionProvider::EFFECTIVE_USER_METHODS.inspect,
+                description: N_('What command should be used to switch to the effective user. One of %s') % ::ScriptExecutionProvider::EFFECTIVE_USER_METHODS.inspect,
                 default: 'sudo',
                 full_name: N_('Effective User Method'),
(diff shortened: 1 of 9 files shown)


lib/foreman_remote_execution/engine.rb
```diff
@@ -328,7 +328,7 @@ class Engine < ::Rails::Engine
       require_dependency 'foreman_tasks/task'
       ForemanTasks::Task.include ForemanRemoteExecution::ForemanTasksTaskExtensions
       ForemanTasks::Cleaner.include ForemanRemoteExecution::ForemanTasksCleanerExtensions
-      RemoteExecutionProvider.register(:SSH, ::SSHExecutionProvider)
+      RemoteExecutionProvider.register(:SSH, ::ScriptExecutionProvider)
```
jturel, line 331: Is this valid?

jturel:
> Thanks, @jturel, I'd love to get rid of old constant :)
> 
> Could you please treat this PR for the constant removal only? It will be easier to review and I'd say this one will be merged sooner than the one with the changes for reloading.
> 
> Otherwise LGTM, but will likely require a major bump.

Done. I'll open another PR with the plugin stuff once this one is merged. Thanks for taking a look

ofedoren (approved):
Thanks, @jturel ! @adamruzicka, any concerns?

adamruzicka (approved):

PR #974, opened 2025-05-20
Add initial tmt test plan

archanaserver:


.packit.yaml
```diff
@@ -35,6 +35,11 @@ jobs:
           - https://yum.theforeman.org/plugins/nightly/el9/x86_64/
     module_hotfixes: true
 
+  - job: tests
+    trigger: pull_request
+    targets:
+      - epel-9
```
adamruzicka, line 41: Can this be `rhel-9`? Alternatively we'll probably need to set up additional repos here

main.fmf
```diff
@@ -0,0 +1 @@
+
```
adamruzicka, line 1: Do we really need to carry an empty file?

tests/rex-uptime/test.sh
```diff
@@ -0,0 +1,33 @@
...
+ORG="Test Org"
+REX_ORG="Remote Execution Org $$"
+
+OLD_ORGS=$(hammer --output csv --no-headers --show-ids proxy info --name "${SMART_PROXY_NAME}" --fields Organizations)
+
+hammer organization create --name "${REX_ORG}"
+hammer host update --name "${SMART_PROXY_NAME}" --new-organization "${REX_ORG}"
+hammer proxy update --name "${SMART_PROXY_NAME}" --organizations "${OLD_ORGS},${REX_ORG}"
+
+hammer job-invocation create \
```
adamruzicka, line 17: This command blocks until the job finishes. If the job failed, the exit code is non-zero so we don't really have to invoke the following two hammer commands and parse them for details.

adamruzicka (commented):
Could we put all those files being added here into a single directory rather than having them scattered over the repo root, .fmf, plans and tests or is the current structure dictated by tmt?

tests/rex-uptime/test.sh
archanaserver: I tried simplify the logic and drop the additional calls.

.packit.yaml
archanaserver: Can you be more descriptive please I might not get it?

archanaserver:
> Could we put all those files being added here into a single directory rather than having them scattered over the repo root, .fmf, plans and tests or is the current structure dictated by tmt?

Thanks for the review @adamruzicka and for pointing this out!
I initially followed the scattered layout based on some earlier references and examples. That said, I’ve tried to re-structure it this way to clearly separate host-side orchestration (via hammer) from guest-side testing, which helps avoid some common pitfalls I believe when integrating TMT with Foreman.

.packit.yaml
adamruzicka: The previous build failed with this - apparently it tried to build the rpm without having access to the other repositories we supply for the `copr_build` job.
```
No matching package to install: 'foreman-assets >= 3.15'
No matching package to install: 'foreman-plugin >= 3.15'
No matching package to install: 'npm(@theforeman/builder) >= 15.0.1'
No matching package to install: 'rubygem(deface)'
No matching package to install: 'rubygem(foreman-tasks) >= 8.3.0'
```

With the `rhel-9` thing, I was hoping that if we used the same target for both of those jobs, things would sort themselves out, but apparently that is not an available target on testing farm, so I'm sorry for that. So back to epel-9 and we'll see where it takes us

PR #975, opened 2025-06-04
Fixes #38446 - Move plugin registration to own file

jturel:
Continuation of work started in #973 to fix code reloading.

The plugin registration calls into a lot of application code. This seems to be the least intrusive way to achieve parity but without registration happening again on each code reload


adamruzicka (approved):

adamruzicka:
Thank you @jturel !

PR #977, opened 2025-06-09
Fixes #38478 - Introduce SSH CA certificate support

adamlazik1:


adamlazik1:
Only drafting stage.

app/lib/proxy_api/remote_execution_ssh.rb
```diff
@@ -11,6 +11,12 @@ def pubkey
       raise ProxyException.new(url, e, N_('Unable to fetch public key'))
     end
 
+    def ca_pubkey
+      get('ca_pubkey')&.strip
+    rescue
+      Rails.logger.info("Unable to fetch CA public key. Using public key authentication instead.")
```
adamruzicka, line 17: So if the proxy doesn't have the ca_pubkey endpoint, the get will raise and we'll get this warning. If the proxy has that endpoint, but doesn't have a ca_pubkey configured, the get will silently return an empty string, right?

app/models/concerns/foreman_remote_execution/smart_proxy_extensions.rb
```diff
@@ -19,13 +23,24 @@ def update_pubkey
       key
     end
 
+    def update_ca_pubkey
+      return unless has_feature?(%w(SSH Script))
+
+      key = ::ProxyAPI::RemoteExecutionSSH.new(:url => url).ca_pubkey
+      # smart proxy is not required to have a CA pubkey, in which case an empty string is returned
+      key = nil if key == ""
```
adamruzicka, line 31: This reads a bit nicer and should do the same thing
```suggestion
      # smart proxy is not required to have a CA pubkey, in which case an empty string is returned
      key = ::ProxyAPI::RemoteExecutionSSH.new(:url => url).ca_pubkey&.presence
```

app/lib/proxy_api/remote_execution_ssh.rb
adamlazik1: Right. Looking at it now I don't know why I made it like this, the message makes little sense. I updated it to reflect the behavior of the pubkey api call.

adamlazik1:
Switching back to draft since the feature got postponed to 3.16

app/lib/proxy_api/remote_execution_ssh.rb
```diff
@@ -11,6 +11,12 @@ def pubkey
       raise ProxyException.new(url, e, N_('Unable to fetch public key'))
     end
 
+    def ca_pubkey
+      get('ca_pubkey')&.strip
+    rescue => e
+      raise ProxyException.new(url, e, N_('Unable to fetch CA public key'))
```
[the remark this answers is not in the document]
adamlazik1: Would returning nil instead of throwing exception suffice? Rest of the application logic treats nil as unset CA pubkey and proceeds with the standard pubkey.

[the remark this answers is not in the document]
adamruzicka: I would prefer not using a catch-all like this to prevent masking real issues.

app/lib/proxy_api/remote_execution_ssh.rb
```diff
@@ -11,6 +11,15 @@ def pubkey
       raise ProxyException.new(url, e, N_('Unable to fetch public key'))
     end
 
+    def ca_pubkey
+      get('ca_pubkey')&.strip
+    rescue RestClient::ResourceNotFound => e
+      Foreman::Logging.exception("Unable to fetch CA public key", e)
```
adamruzicka: _nitpick:_
On my dev setup, this generates roughly 110 lines in the logs which is a bit excessive for a non-error case. I'd be fine with this being silent or just logging a single line

adamruzicka:
Thank you @adamlazik1 !

PR #978, opened 2025-06-13
Fixes #38495 - Fix memory leak bug in TemplateInvocation file

kmalyjur:
image


MariaAga:
Code looks good, Do you remember what actions showed the leak?

kmalyjur:
> Do you remember what actions showed the leak?

@MariaAga, the error always appears once after calling /api/job_invocations/*id*/hosts.

> Probably not related to this pr, but when I tried to changing the dropdown filter and then back to all statuses the table is empty (only when autoreload is true)

I don't know how to reproduce it, but I agree that it's probably a separate issue.
Would it be okay to merge this PR, and we can investigate the filter behavior you're seeing later?

adamruzicka:

PR #979, opened 2025-06-23
Fixes #38527 - Fix job template clone endpoint for locked templates

Gauravtalreja1:
### Description:

- Use dup() instead of clone() to prevent locked status inheritance. 
- Add `deep_clone` method to handle taxonomies properly.

### DIFF
--- a/app/controllers/api/v2/job_templates_controller.rb
+++ b/app/controllers/api/v2/job_templates_controller.rb
@@ -108,9 +108,7 @@ module Api
       param :id, :identifier, :required => true
       param_group :job_template_clone, :as => :create
       def clone
-        @job_template = @job_template.clone
-        load_vars_from_template
-        @job_template.name = job_template_params[:name]
+        @job_template = @job_template.deep_clone(job_template_params[:name])
         process_response @job_template.save
       end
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/v2/job_templates_controller.rb
```diff
@@ -108,9 +108,7 @@ def destroy
       param :id, :identifier, :required => true
       param_group :job_template_clone, :as => :create
       def clone
-        @job_template = @job_template.clone
-        load_vars_from_template
-        @job_template.name = job_template_params[:name]
+        @job_template = @job_template.clone_job_template(job_template_params[:name])
```
adamruzicka, line 111: This all still feels a bit heavy handed, couldn't we just set locked to false before saving it and call it a day?

Gauravtalreja1: @adamruzicka Not exactly, I tried this earlier to add it here only but it doesn't work as expected, and not all associations like taxonomies are copied, which we need handle explicitely in clone_job_template method

app/models/job_template.rb
```diff
@@ -129,6 +129,17 @@ def assign_taxonomies
     end
   end
 
+  # Clone template for user with proper handling of locked status and taxonomies
+  def clone_for_user(new_name)
```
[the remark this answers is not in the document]
Gauravtalreja1: Renamed to deep_clone as discussed to keep it consistent

PR #984, opened 2025-07-21
Fixes #38587 - fix all statuses

Lukshio:


adamruzicka:
If you fill in some search and the clear the search bar, a `search=` query parameter will get added to the url, causing the `Awaiting start` hosts to not show up, no matter what the status search is set to. Could we resolve that one as well?

Lukshio:
> If you fill in some search and the clear the search bar, a `search=` query parameter will get added to the url, causing the `Awaiting start` hosts to not show up, no matter what the status search is set to. Could we resolve that one as well?

Empty url param should be ignored now.

Lukshio:
This PR is required for https://github.com/theforeman/foreman_remote_execution/pull/985

kmalyjur:
I think it makes sense - we also had a call about it, can it get merged then? @adamruzicka

MariaAga:
Fix for rerun: https://github.com/theforeman/foreman_remote_execution/pull/991
This pr is unrelated

adamruzicka:
Test failures are about to be resolved elsewhere, let's get this in

adamruzicka:
Thank you @Lukshio , @MariaAga & @kmalyjur !

PR #985, opened 2025-07-21
Fixes #38592 - Add action to bulk cancel / abort the tasks selected

kmalyjur:


app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -134,16 +134,45 @@ def raw_output
...
-        if @job_invocation.task.cancellable?
-          result = @job_invocation.cancel(params.fetch('force', false))
-          render :json => { :cancelled => result, :id => @job_invocation.id }
+        force = params.fetch('force', false)
+        search = params[:search]
+
+        if search.present?
+          begin
+            hosts_scope = @job_invocation.targeting.hosts.authorized(:view_hosts, Host)
+            matching_hosts = search.present? ? hosts_scope.search_for(search) : hosts_scope
```
adamlazik1, line 148: Here the search is always present due to condition above.

Lukshio:
Not directly affected with this PR. However, opening job invo page will all pending results infinite api requests for host statuses, because `queued` option is missing in statuses that stops requests

https://github.com/theforeman/foreman_remote_execution/blob/febab83a5d2af6c1ab0b3d2c421bf71976957431/webpack/JobInvocationDetail/index.js#L47

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -10,24 +11,38 @@ import {
...
+  bulkParams,
+}) => {
   const [isModalOpen, setIsModalOpen] = useState(false);
   const [showAlert, setShowAlert] = useState(false);
   const [isOpenFailed, setIsOpenFailed] = useState(false);
+  const isTaskCancelable = useSelector(selectTaskCancelable);
+  const dispatch = useDispatch();
   const permissions = useSelector(
     state => selectTemplateInvocation(state)?.permissions
   );
```
Lukshio, line 48: Opening job with all awaiting statuses results undefined permissions. I think it should use this selector `useSelector(state => selectItems(state)?.permissions);`

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -76,10 +91,84 @@ export const CheckboxesActions = ({ selectedIds, failedCount, jobID }) => {
...
     const dropdownItems = [
+      <DropdownItem
+        ouiaId="cancel-host-dropdown-item"
+        onClick={() => handleTaskAction('cancel')}
+        key="cancel"
+        component="button"
+        isDisabled={
+          selectedIds.length === 0 ||
+          !isTaskCancelable ||
+          !hasPermission(permissions, 'cancel_job_invocations')
```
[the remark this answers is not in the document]
MariaAga: When jobs are waiting they cant be canceled (from what I saw in the old ui)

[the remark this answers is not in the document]
Lukshio: Tried in Legacy UI, and successfully canceled/aborted awaiting.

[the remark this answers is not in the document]
kmalyjur: It works for me in the old ui, too

kmalyjur:
> Not directly affected by this PR. However, opening job invo page will all pending results infinite api requests for host statuses, because queued option is missing in statuses that stops requests. Can we add it to this PR? Or maybe create separate one for this.

@MariaAga it makes sense to me, but what's your opinion, please? The issue is that "queued" means it's not finished.

webpack/JobInvocationDetail/CheckboxesActions.js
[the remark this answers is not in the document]
MariaAga: image
awaiting jobs are not clickable
also, awaiting jobs have no task created for the host

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -76,10 +91,82 @@ export const CheckboxesActions = ({ selectedIds, failedCount, jobID }) => {
...
+
+    dispatch(
+      APIActions.post({
+        url: `/api/v2/job_invocations/${jobID}/cancel`,
+        key: 'CANCEL_JOB_TASKS_BY_SEARCH',
+        params: {
+          search: combinedQuery,
+          force: action !== 'cancel',
+        },
+        errorToast: ({ response }) => response?.data?.error,
```
MariaAga, line 117: Needs a fallback to avoid `helpers.js:24 Uncaught TypeError: Cannot read properties of undefined (reading 'length')`

MariaAga: opened https://github.com/theforeman/foreman_remote_execution/pull/986 for the table error toast

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -76,10 +91,82 @@ export const CheckboxesActions = ({ selectedIds, failedCount, jobID }) => {
...
+        successToast: ({ response }) => {
```
MariaAga, line 118:
```suggestion
        successToast: ( response ) => {
```
response is not a key

Lukshio:
> We should fix the dropdown (for bulk actions) not staying open on refresh as its hard to click

I think it is caused by the index page still calling BE each second.

kmalyjur:
@MariaAga @Lukshio @adamlazik1, the dropdown should not be closing constantly now. However, it takes some time for the table to change the status to "cancelled" or the page has to be reloaded - is it normal behaviour? Should I make the page reload automatically after cancelling?

adamlazik1:
I am encountering some unexpected behavior. I run `Run sleep 10; touch /tmp/done` on 4 hosts with concurrency level 1 (before that I make sure that the file is not present on the host). I then proceed to cancel the task on a host which has not yet been scheduled for execution. I get a success message about the canceled task and the UI shows that the task on host has been canceled, but the reality the task is executed and the file is created. Can anyone else replicate this?

Also, when an unscheduled ("Awaiting Start") task is canceled, the status of the host stays in "Awaiting start" instead of changing to "Cancelled" even after page reload.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -134,16 +134,45 @@ def raw_output
...
+          end
         else
-          render :json => { :message => _('The job could not be cancelled.') },
-            :status => :unprocessable_entity
+          if @job_invocation.task.cancellable?
+            result = @job_invocation.cancel(force)
+            render json: { cancelled: result, id: @job_invocation.id }
+          else
+            render json: { message: _('The job could not be cancelled.') }, status: :unprocessable_entity
+          end
```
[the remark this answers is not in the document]
adamlazik1: Although, the bulk_cancel endpoint lives in foreman tasks. Do we want to or are we even allowed to rely on another plugin?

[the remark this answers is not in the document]
adamlazik1: If that is the case then should we just call the tasks endpoint here?

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -10,24 +13,110 @@ import {
...
+      onClick={() => handleTaskAction('cancel')}
+      key="cancel"
+      component="button"
+      isDisabled={
+        selectedIds.length === 0 ||
+        !isTaskCancelable ||
+        !hasPermission(permissions, 'cancel_job_invocations')
+      }
+    >
+      {__('Cancel selected test')}
```
MariaAga, line 56: Probably wrong test

kmalyjur:
@MariaAga thank you, I fixed all the issues you mentioned, however, the `onSuccess` didn't work, so the solution is more complicated. Is this okay?

webpack/JobInvocationDetail/JobInvocationConstants.js
```diff
@@ -34,6 +34,9 @@ export const templateInvocationPageUrl = (hostID, jobID) =>
 export const jobInvocationDetailsUrl = id =>
   `/experimental/job_invocations_detail/${id}`;
 
+export const hasPermission = (permissions, permissionRequired) =>
+  !!permissions?.[permissionRequired];
+
```
MariaAga, line 39: not used

MariaAga:
All bugs are resolved! left some comments for leftovers and not the logic

kmalyjur:
@MariaAga I removed the unused `hasPermission`. Can it be merged now?

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -290,6 +317,7 @@ def set_statuses_and_smart_proxies
         end]
         @smart_proxy_id = Hash[template_invocations.map { |ti| [ti.host_id, ti.smart_proxy_id] }]
         @smart_proxy_name = Hash[template_invocations.map { |ti| [ti.host_id, ti.smart_proxy_name] }]
+        @run_host_job_task_id = Hash[template_invocations.map { |ti| [ti.host_id, ti.run_host_job_task_id] }]
```
adamruzicka, line 320: Where is this used?

kmalyjur: `render json: {
             total: tasks_to_process.size,
              cancelled: cancelled.map(&:run_host_job_task_id),
              skipped: skipped.map(&:run_host_job_task_id),
            }`
In the same file, do you think it should be done differently?

adamruzicka: Are you referring to `cancelled.map(&:run_host_job_task_id)` and `skipped.map(&:run_host_job_task_id)`?

kmalyjur: Yes

adamruzicka (approved):

PR #987, opened 2025-07-22
Sort out the loading of the new job invocation page

kmalyjur:
The skeleton is not included in the description as it doesn't function as intended with the `<Head>`.


MariaAga (approved):

adamruzicka:

PR #988, opened 2025-07-22
Fixes #38591 - Api calls refactor

Lukshio:
Remove useApi duplicates, unify calls

### DIFF
--- a/webpack/JobInvocationDetail/CheckboxesActions.js
+++ b/webpack/JobInvocationDetail/CheckboxesActions.js
@@ -14,23 +14,21 @@ import {
 } from '@patternfly/react-icons';
 import axios from 'axios';
 import { foremanUrl } from 'foremanReact/common/helpers';
-import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
 import { translate as __, sprintf } from 'foremanReact/common/I18n';
 import { addToast } from 'foremanReact/components/ToastsList';
 import PropTypes from 'prop-types';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
 import {
   DIRECT_OPEN_HOST_LIMIT,
-  MAX_HOSTS_API_SIZE,
   templateInvocationPageUrl,
 } from './JobInvocationConstants';
 import {
   selectHasPermission,
   selectTaskCancelable,
 } from './JobInvocationSelectors';
-import OpenAllInvocationsModal, { PopupAlert } from './OpenAllInvocationsModal';
+import OpenAllInvocationsModal from './OpenAllInvocationsModal';
 
 /* eslint-disable camelcase */
 const ActionsKebab = ({
@@ -73,7 +71,7 @@ const ActionsKebab = ({
       onClick={() => handleOpenHosts('failed')}
       isDisabled={failedCount === 0}
     >
-      {sprintf(__('Open all failed runs (%s)'), failedCount)}
+      {sprintf(__('Open all failed runs on this page (%s)'), failedCount)}
     </DropdownItem>,
   ];
 
@@ -104,17 +102,18 @@ const ActionsKebab = ({
 
 export const CheckboxesActions = ({
   selectedIds,
-  failedCount,
+  allJobs,
   jobID,
   filter,
   bulkParams,
+  setShowAlert,
 }) => {
   const [isModalOpen, setIsModalOpen] = useState(false);
-  const [showAlert, setShowAlert] = useState(false);
   const [isOpenFailed, setIsOpenFailed] = useState(false);
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
   const isTaskCancelable = useSelector(selectTaskCancelable);
   const dispatch = useDispatch();
+  const [toBeOpened, setToBeOpened] = useState([]);
 
   const hasCreatePermission = useSelector(
     selectHasPermission('create_job_invocations')
@@ -129,19 +128,14 @@ export const CheckboxesActions = ({
       : '';
   const combinedQuery = `${bulkParams}${filterQuery}`;
 
-  const { response: failedHostsData } = useAPI(
-    'get',
-    foremanUrl(`/api/job_invocations/${jobID}/hosts`),
-    {
-      params: {
-        per_page: MAX_HOSTS_API_SIZE,
-        search: `job_invocation.result = failed`,
-      },
-      skip: failedCount === 0,
-    }
-  );
+  const [failedHosts, setFailedHosts] = useState([]);
 
-  const failedHosts = failedHostsData?.results || [];
+  useEffect(() => {
+    const failed = allJobs.filter(i => i.job_status === 'error');
+    setFailedHosts(failed);
+  }, [allJobs]);
+
+  const failedCount = failedHosts.length;
 
   const openLink = url => {
     const newWin = window.open(url);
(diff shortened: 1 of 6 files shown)


webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -47,25 +42,53 @@ const JobInvocationHostTable = ({
 }) => {
   const columns = Columns();
   const columnNamesKeys = Object.keys(columns);
-  const apiOptions = { key: JOB_INVOCATION_HOSTS };
+
   const history = useHistory();
+  const dispatch = useDispatch();
+
+  const [APIResponse, setAPIResponse] = useState([]);
```
ofedoren, line 49: quick nit: doesn't `APIResponse` means it's a constant now? Although JS is weird and it might work just fine, but I'd rather change it to `apiResponse` or similar.

Lukshio: Yep, it will make more sense with camelCase, thanks.

adamruzicka:
Needs a rebase

Lukshio:
> This makes it so if a users selects all hosts, and opens in a new tab, only the first page is opened, instead of the first 100

Can you please retest it now? Opening in new tabs has its own limit https://github.com/theforeman/foreman_remote_execution/blob/2310a7388226b71584538f58cf414d84532f67c4/webpack/JobInvocationDetail/CheckboxesActions.js#L83-L86

MariaAga:
since nothing calls for the first 100 ( per_page = MAX_HOSTS_API_SIZE) it still only opens the first page instead of the first 100

MariaAga:
also it auto sets per page to 20 instead of the user per_page

MariaAga:
changing per_page adds brackets around the search, and then every page change adds more brackets, searching by clicking on the chart doesnt work anymore

Lukshio:
> changing per_page adds brackets around the search, and then every page change adds more brackets, searching by clicking on the chart doesnt work anymore
Fixed


> also it auto sets per page to 20 instead of the user per_page
Fixed



> since nothing calls for the first 100 ( per_page = MAX_HOSTS_API_SIZE) it still only opens the first page instead of the first 100

I couldn't replicate this issue

MariaAga:
to replicate - have a job run on 21+ hosts, open the job details page, make sure per page is less than the amount of hosts (usually set to 20) click on the open all tab, the number shown in the modal is the per page number, and not all hosts number

Lukshio:
> to replicate - have a job run on 21+ hosts, open the job details page, make sure per page is less than the amount of hosts (usually set to 20) click on the open all tab, the number shown in the modal is the per page number, and not all hosts number

Added old MAX_HOSTS const to load only first 100

adamruzicka:
It seems we're polling periodically, without waiting for the previous poll to finish? Could we do something about that?

image

Lukshio:
After discussion with  @MariaAga new behavior is:
- Open in new tabs in applied to current page if not selected, same for failed open
- All api calls are reduced to 1 which handles ids for selection, failed, etc
- MAX_HOSTS is no longer in use

adamruzicka:
Needs a rebase now

Lukshio:
> Needs a rebase now

Rebased

Lukshio:
> > The open in new tab button - should open the page if nothing is selected, should not have the text "The maximum is 100.«", and should have a matching text (in the modal and tooltip) if its opening the selected in the page or the just the page. open all failed - should say open all failed runs from this page, the modal also shouldnt say "the maximum is 100.«", also ,it doesnt open any tabs when clicked
> 
> Unable to reproduce not opening page if nothing selected, other fixed

Ad. refactored opening new tabs, now it should be working

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -26,52 +27,71 @@ import PropTypes from 'prop-types';
...
-  const constructFilter = (
-    filter = selectedFilter,
-    search = urlSearchQuery
-  ) => {
+  // default
+  const defaultParams = useMemo(
+    () => ({
+      page: urlPage ? Number(urlPage) : 1,
+      ...(urlPerPage && { per_page: Number(urlPerPage) }),
+      order: urlOrder || '',
```
MariaAga, line 81: If I change the order to name desc, and the change the page the order gets ignored. Also it doesnt seem to go to the url which might be the reason

Lukshio:
> clicking the chart seems to add conditions instead of replacing them `(job_invocation.result = failed) AND ((job_invocation.result = success) AND ((job_invocation.result = failed) AND ((job_invocation.result = cancelled) AND ((job_invocation.result = pending) AND ((job_invocation.result = success) AND ((job_invocation.result = failed)))))))`

re fixed it, hopefully this will stop the circle of fixing something, and breaking something else :)

MariaAga (commented):
Seems to work as expected now! thats quite a lot of changes, @kmalyjur can you take a look as well to make sure I'm not missing anything?

webpack/JobInvocationDetail/OpenAllInvocationsModal.js
```diff
@@ -25,21 +21,35 @@ export const PopupAlert = ({ setShowAlert }) => (
...
     }
+    if (selectedIds.length > 0) {
+      return (
+        <>
+          {__('The number of selected invocations is:')}{' '}
+          <b>{selectedIds.length}</b>
+        </>
+      );
+    }
+    return <>{__('Opening current page')} </>;
```
kmalyjur, line 52: I might be nitpicking, but isn't that line unnecessary? It's already mentioned twice in the modal that it's the current page. The number of hosts to be opened, as for the rest of the modal possibilities, might be more useful. 🤔

webpack/JobInvocationDetail/OpenAllInvocationsModal.js
```diff
@@ -25,21 +21,35 @@ export const PopupAlert = ({ setShowAlert }) => (
 
 const OpenAllInvocationsModal = ({
   failedCount,
-  failedHosts,
   isOpen,
   isOpenFailed,
-  jobID,
   onClose,
-  setShowAlert,
```
kmalyjur, line 33: Delete `jobID` and `setShowAlert` from CheckboxesActions.js, too

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -351,21 +341,21 @@ const JobInvocationHostTable = ({
               : null
           }
           params={{
-            page: response?.page || Number(urlPage),
-            per_page: response?.per_page || Number(urlPerPage),
+            page: defaultParams.page || Number(urlPage),
```
kmalyjur, line 344: Missing `?` (although I think it's not needed, it would be inconsistent)

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -26,52 +27,76 @@ import PropTypes from 'prop-types';
...
+  STATUS_UPPERCASE,
   ALL_JOB_HOSTS,
 } from './JobInvocationConstants';
-import { PopupAlert } from './OpenAllInvocationsModal';
 import { TemplateInvocation } from './TemplateInvocation';
 import { RowActions } from './TemplateInvocationComponents/TemplateActionButtons';
+import { PopupAlert } from './OpenAllInvocationsModal';
 
 const JobInvocationHostTable = ({
-  failedCount,
```
kmalyjur, line 44: Needs to be deleted from index.js, too

kmalyjur (changes requested):
Thank you, it works as it should, and the code is great except for these minor things.

webpack/JobInvocationDetail/OpenAllInvocationsModal.js
Lukshio: Maybe I am misunderstanding, but these two are still in use in CheckboxesActions.js

Lukshio: I understand now, thank you, fixed.

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -282,16 +293,14 @@ export const CheckboxesActions = ({
         isDropdownOpen={isDropdownOpen}
         setIsDropdownOpen={setIsDropdownOpen}
       />
-      {showAlert && <PopupAlert setShowAlert={setShowAlert} />}
       <OpenAllInvocationsModal
         isOpen={isModalOpen}
         onClose={() => setIsModalOpen(false)}
         failedCount={failedCount}
-        failedHosts={failedHosts}
         jobID={jobID}
```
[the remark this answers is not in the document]
Lukshio: Sure, Thanks.

PR #989, opened 2025-07-22
Fixes #38597 - Add 'Scheduled' option

Lukshio:
https://github.com/theforeman/foreman_remote_execution/pull/988 Is required for this one

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -118,7 +118,12 @@ module Api
         set_statuses_and_smart_proxies
         @total = @job_invocation.targeting.hosts.size
         @hosts = @hosts.search_for(params[:search], :order => params[:order]).paginate(:page => params[:page], :per_page => params[:per_page])
-        @subtotal = @hosts.respond_to?(:total_entries) ? @hosts.total_entries : @hosts.sizes
+        if params[:awaiting]
+          @hosts = @hosts.select { |host| @host_statuses[host.id] == 'N/A' }
+          @subtotal = @hosts.size
+        else
+          @subtotal = @hosts.respond_to?(:total_entries) ? @hosts.total_entries : @hosts.sizes
+        end
         render :hosts, :layout => 'api/v2/layouts/index_layout'
       end
 
--- a/webpack/JobInvocationDetail/JobInvocationHostTable.js
+++ b/webpack/JobInvocationDetail/JobInvocationHostTable.js
@@ -35,6 +35,7 @@ import Columns, {
   LIST_TEMPLATE_INVOCATIONS,
   STATUS_UPPERCASE,
   ALL_JOB_HOSTS,
+  AWAITING_STATUS_FILTER,
 } from './JobInvocationConstants';
 import { TemplateInvocation } from './TemplateInvocation';
 import { RowActions } from './TemplateInvocationComponents/TemplateActionButtons';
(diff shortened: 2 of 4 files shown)


adamruzicka:
Blocked on https://github.com/theforeman/foreman_remote_execution/pull/988#issuecomment-3103874920 and a rebase

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -119,6 +119,10 @@ def hosts
         @total = @job_invocation.targeting.hosts.size
         @hosts = @hosts.search_for(params[:search], :order => params[:order]).paginate(:page => params[:page], :per_page => params[:per_page])
         @subtotal = @hosts.respond_to?(:total_entries) ? @hosts.total_entries : @hosts.sizes
```
ekohl, line 121: If this needs to be different, can you put this in the `else` branch below to avoid needlessly calculating the subtotal?

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -49,25 +47,53 @@ const JobInvocationHostTable = ({
...
     page: urlPage,
     per_page: urlPerPage,
     order: urlOrder,
   } = useUrlParams();
 
+  // default
+  const defaultParams = useMemo(
+    () => ({
+      page: urlPage ? Number(urlPage) : 1,
+      per_page: urlPerPage ? Number(urlPerPage) : 20,
```
ekohl, line 83: Is there a way to avoid passing default `per_page` so we respect the settings? It's a really bad pattern to copy the hardcoded value everywhere and I don't see why it's needed. Doesn't the API will return the page size it used? That can be used and relied on. If not, something is wrong.

Lukshio: @ekohl already solved in parent PR https://github.com/theforeman/foreman_remote_execution/pull/988

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -147,7 +147,9 @@ const JobInvocationHostTable = ({
       ...newParams,
     };
 
-    if (filterSearch !== '') {
+    if (filterSearch === '(job_invocation.result = N/A)') {
```
MariaAga, line 150: Should use the constant

MariaAga:
Also, not sure if its in the scope of this PR, but searching for `name ~ *` gives empty results for "all results" and "Scheduled" (if there are only scheduled hosts)

Lukshio:
> Also, not sure if its in the scope of this PR, but searching for `name ~ *` gives empty results for "all results" and "Scheduled" (if there are only scheduled hosts)

After discussion with @adamruzicka about this. We decided to do just this "quick fix" for the scheduled option. Because the filter itself is written in some way, that does not allow us to quickly add this option without breaking things or do lot of work to get it done. So the answer is: Yes, it's out of scope of this PR

Lukshio:
> image The chart should reflect this as well, currently clicking on the chart changes the filter to "success"

I took a look at this issue, and it actually looks like, that it is a bug, because I don't think it should filter 'success' while all of the jobs are 'awaiting'. Or does it have some special behavior? That is also question for @kmalyjur 

I think the issue is here: https://github.com/theforeman/foreman_remote_execution/blob/313d8b0b85a9a5df8b49e99d0f2c9cae889046ec/webpack/JobInvocationDetail/JobInvocationSystemStatusChart.js#L69-L77

The index skips "ALL_STATUSES", but it skips it also when there are not any finished jobs.

PR #992, opened 2025-07-22
Fixes #TBFL - Adjust to report macro changes

adamruzicka:
Introduced in
https://github.com/theforeman/foreman/commit/177634991c33f2a143f59dfdb42d6da2eb3775f0

Opening as a draft until https://github.com/theforeman/foreman/pull/10584/files#r2223321135 is answered


ofedoren (approved):
Eh.. What if the original fix had
```ruby
array.map! { |str| str.delete_suffix("\n") }
#...
array.join("\n")
```

But I guess having a trailing new line is common now...

Thanks, @adamruzicka !

adamruzicka:
Thank you @ofedoren !

PR #995, opened 2025-08-05
Fixes #38636 - JobInvocationHostTable doesnt load the template with auto reload

kmalyjur:
The index page now passes `finished` to the `JobInvocationHostTable`, and a `useEffect` is added to refetch the data when `finished` becomes true. I'm not sure if it's the best solution, even though it works as expected, so I'm open to new ideas.

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationHostTable.js
+++ b/webpack/JobInvocationDetail/JobInvocationHostTable.js
@@ -23,7 +23,7 @@ import TableIndexPage from 'foremanReact/components/PF4/TableIndexPage/TableInde
 import { getControllerSearchProps } from 'foremanReact/constants';
 import { Icon } from 'patternfly-react';
 import PropTypes from 'prop-types';
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState, useRef } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { useHistory } from 'react-router-dom';
 import URI from 'urijs';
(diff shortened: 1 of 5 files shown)


MariaAga:
I'm getting `Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. Reac` on host expand

kmalyjur:
@MariaAga Thank you, fixed now

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -138,6 +140,14 @@ const JobInvocationHostTable = ({
     }
   }, [allResponse]);
 
+  useEffect(() => {
+    if (finished && !refreshedOnFinish.current) {
```
MariaAga, line 144: Might be because your job goes immediately to finished and mine goes to running

kmalyjur: You're right! I changed it so it reloads when the status changes. Is that okay?

MariaAga:
Opening a job invocation in a new tab comes back empty `job_invocations_detail/x/host_invocation/y`
image

webpack/JobInvocationDetail/TemplateInvocation.js
```diff
@@ -57,82 +57,86 @@ export const TemplateInvocation = ({
   hostID,
   jobID,
   isInTableView,
+  isVisible,
   hostName,
   hostProxy,
-  isExpanded,
```
MariaAga, line 62: do we need a new prop? can we use just some isExpanded || !isInTableView?

kmalyjur: I changed it back to isExpanded

PR #998, opened 2025-08-28
Fixes #38698 - Missing host jobs in host details page

Lukshio:


ofedoren (commented):
Don't we already have it a bit below? https://github.com/theforeman/foreman_remote_execution/pull/998/files#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5R33

adamruzicka:
Rather than declaring a single route twice in different places, couldn't we move `get 'job_invocations/:id', to: 'react#index', as: 'job_invocation'` _below_ the `resources :job_invocations` block?

Lukshio:
> Rather than declaring a single route twice in different places, couldn't we move `get 'job_invocations/:id', to: 'react#index', as: 'job_invocation'` _below_ the `resources :job_invocations` block?

Sure, thank you! Re-tested it, and it works well

adamruzicka (approved):

PR #999, opened 2025-09-02
Fixes #38709 - Fix hosts job status stuck in "Pending"

pavanshekar:
Fix hosts job status remaining "Pending" after completion

Testing steps:

1. Navigate to Monitor -> Jobs
2. Click on "Run Job" button
3. Job category -> Katello
4. Job template -> Install errata by search query - Katello Script Default
5. Filter the hosts and enter search query
6. Submit
7. Once the job completes successfully, the status should be updated accordingly (e.g., to "Success" or "Completed") for all associated hosts mentioned in the table.

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationHostTable.js
+++ b/webpack/JobInvocationDetail/JobInvocationHostTable.js
@@ -45,8 +45,10 @@ const JobInvocationHostTable = ({
   id,
   initialFilter,
   onFilterUpdate,
-  statusLabel,
   targeting,
+  autoRefresh,
+  statusLabel,
+  hostStatus,
 }) => {
   const columns = Columns();
   const columnNamesKeys = Object.keys(columns);
@@ -55,6 +57,8 @@ const JobInvocationHostTable = ({
   const [selectedFilter, setSelectedFilter] = useState(initialFilter);
   const [expandedHost, setExpandedHost] = useState([]);
   const prevStatusLabel = useRef(statusLabel);
+  const prevHostStatus = useRef({ succeeded: 0, failed: 0, pending: 0 });
+  const lastHostsRefresh = useRef(0);
 
   useEffect(() => {
     if (initialFilter !== selectedFilter) {
@@ -147,6 +151,34 @@ const JobInvocationHostTable = ({
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [statusLabel]);
+
+  const hostSucceeded = hostStatus?.succeeded;
+  const hostFailed = hostStatus?.failed;
+  const hostPending = hostStatus?.pending;
+  const hasResults = !!response?.results;
+
+  useEffect(() => {
+    const statusChanged =
+      prevHostStatus.current.succeeded !== hostSucceeded ||
+      prevHostStatus.current.failed !== hostFailed ||
+      prevHostStatus.current.pending !== hostPending;
+
+    const canRefresh = Date.now() - lastHostsRefresh.current > 3000;
+
+    if (autoRefresh && hasResults && statusChanged && canRefresh) {
+      lastHostsRefresh.current = Date.now();
+      prevHostStatus.current = { hostSucceeded, hostFailed, hostPending };
+      setAPIOptions(prevOptions => ({ ...prevOptions }));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    autoRefresh,
+    hostSucceeded,
+    hostFailed,
+    hostPending,
+    hasResults,
+    setAPIOptions,
+  ]);
 
   const {
     updateSearchQuery: updateSearchQueryBulk,
@@ -441,13 +473,21 @@ JobInvocationHostTable.propTypes = {
   targeting: PropTypes.object.isRequired,
   failedCount: PropTypes.number.isRequired,
   initialFilter: PropTypes.string.isRequired,
-  statusLabel: PropTypes.string,
   onFilterUpdate: PropTypes.func,
+  autoRefresh: PropTypes.bool,
+  statusLabel: PropTypes.string,
+  hostStatus: PropTypes.shape({
+    succeeded: PropTypes.number,
+    failed: PropTypes.number,
+    pending: PropTypes.number,
+  }),
 };
 
 JobInvocationHostTable.defaultProps = {
   onFilterUpdate: () => {},
+  autoRefresh: false,
   statusLabel: undefined,
+  hostStatus: { succeeded: 0, failed: 0, pending: 0 },
 };
 
 export default JobInvocationHostTable;
--- a/webpack/JobInvocationDetail/index.js
+++ b/webpack/JobInvocationDetail/index.js
@@ -50,7 +50,7 @@ const JobInvocationDetailPage = ({
     statusLabel === STATUS.FAILED ||
     statusLabel === STATUS.SUCCEEDED ||
     statusLabel === STATUS.CANCELLED;
-  const autoRefresh = task?.state === STATUS.PENDING || false;
+  const autoRefresh = task?.state === 'running' || false;
   useAPI('get', currentPermissionsUrl, {
     key: CURRENT_PERMISSIONS,
   });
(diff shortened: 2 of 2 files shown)


ekohl (commented):
Is there a way to write a test for this?

app/controllers/job_invocations_controller.rb
```diff
@@ -63,7 +63,7 @@ def show
     @job_invocation = resource_base.includes(:template_invocations => :run_host_job_task).find(params[:id])
     @job_organization = Taxonomy.find_by(id: @job_invocation.task.input[:current_organization_id])
     @job_location = Taxonomy.find_by(id: @job_invocation.task.input[:current_location_id])
-    @auto_refresh = @job_invocation.task.try(:pending?)
+    @auto_refresh = @job_invocation.task.try(:pending?) || has_pending_hosts?
```
adamruzicka, line 66: The `has_pending_host?` only comes into play if `@job_invocation.task.try(:pending?)` returns a false-y value. This can happen if either `@job_invocation.task` does not exist yet or if the task exists but is already done, which would imply all its children have already finished too.

If the `@job_invocation.task` does not exist yet, the child tasks cannot exist so `has_pending_hosts?` will be false as well.

If `@job_invocation.task.pending?` returns false, then all its children should already be done too, so `has_pending_hosts?` _should_ always be false too.

app/controllers/job_invocations_controller.rb
```diff
@@ -222,4 +222,12 @@ def targeting_hosts_resources
     @hosts = resource_base_search_and_page
     @total_hosts = resource_base_with_search.size
   end
+
+  private
+
+  def has_pending_hosts?
+    @job_invocation.template_invocations.joins(:run_host_job_task)
+                   .where(foreman_tasks_tasks: { result: ['pending', 'running'] })
```
adamruzicka, line 230: `running` isn't a result

adamruzicka (commented):
disclaimer:
I don't recall encountering the problem this is meant to fix and I have not tested this, all I wrote here is just me trying to reason about it.

How it is supposed to work:
1) When a job is created it spawns a "parent" task backing the whole job, that's `@job_invocation.task`
2) This parent task spawns child sub-tasks, one for every single host in the job
3) The parent task suspends itself and periodically (every 15 seconds or so) checks it children tasks
4) When all the child tasks are done, the parent marks itself as done

Assuming the code was like this for 5+ years and we're seeing it just now as we switched to the new job invocation page, intuitively I'd look for problems there.

If any of the assumptions I just mentioned don't hold (anymore), then this change feels like putting a band aid on a much deeper problem.

app/controllers/job_invocations_controller.rb
pavanshekar: Ya...you're right about the parent/child task coordination. No backend controller changes needed, the parent/child task coordination works correctly as designed. The issue occurs on the new job invocation page where auto-refresh stops when the parent task completes, but the host table still shows stale "Pending" statuses. Added a refresh trigger that updates the host table data once when the job finishes, ensuring fresh host statuses are displayed without manual refresh.

adamruzicka:
Needs a rebase to make the conflict go away

kmalyjur:
@pavanshekar I couldn't reproduce the exact job from your testing steps, but when I test this on a job that starts in a pending state, I see the update happen in two steps after it finishes:
1) The chart updates correctly
2) The status in the table takes a few extra seconds to catch up
 Is that the same behavior you were seeing?
 I'm looking into the code more.

webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -141,12 +140,17 @@ const JobInvocationHostTable = ({
...
-      prevStatusLabel.current = statusLabel;
+    if (refreshTrigger > 0) {
+      setAPIOptions(prev => ({
+        ...prev,
+        params: { ...prev.params, _refresh: Date.now() },
+      }));
+      setAllAPIOptions(prev => ({
+        ...prev,
+        params: { ...prev.params, _refresh: Date.now() },
+      }));
```
Lukshio, line 151:
```suggestion
    if (refreshTrigger > 0) {
      setAPIOptions(prev => ({
        ...prev,
        params: { ...prev.params, _refresh: Date.now() },
      }));
      setAllAPIOptions(prev => ({
        ...prev,
        params: { ...prev.params, _refresh: Date.now() },
      }));
```

pavanshekar: Thank you for the suggestion! This approach is much better and cleaner. I've implemented the suggested changes by replacing the complex refreshTrigger mechanism with the simple autoRefresh useEffect that responds to the autoRefresh prop already being passed from the parent component.

adamruzicka:
Without this patch (as of v16.2.1), there is a bit of a delay, but then things switch to the proper state.

https://github.com/user-attachments/assets/14c921fa-02e4-47bc-b86b-d0f6944d3f71

With this patch applied, I'm getting this

https://github.com/user-attachments/assets/121ebf0a-5335-4287-8c6f-8b0caf455db8

This doesn't really seem to improve things for me

qcjames53:
@adamruzicka Thank you for the videos. I've been seeing the same thing. Thought I was going crazy! I'm going to set up a meeting with @pavanshekar to see if I'm triggering something incorrectly.

qcjames53 (commented):
@pavanshekar and I just spoke about this issue. Thanks for meeting up! Here are the minimal testing steps to verify the bug:

1. Create several hosts on your system. 20+ is a good number. Pavan shared the following snippet to autogenerate many fake hosts in the rails console:
```
100.times do |i|00" if (i + 1) % 10 == 0
  host = Host.first.dup
  host.name = "fake-host-#{Time.now.to_i}-#{i.to_s.rjust(3, '0')}"
  host.mac = nil
  host.ip = nil
  host.save!
  puts "Created host #{i+1}/100" if (i + 1) % 10 == 0
end
```
2. On the all hosts page, select all hosts and click 'Schedule a job'.
3. Any job will work for this. An easy one is 'Commands' -> 'Run Command - Script Default' and set the command to anything.
4. Click through all pages and run on the selected hosts. The UI will take you to the job's webpage.
5. The table will show all hosts as "In progress". When refreshing the page, some of the hosts will switch to "Failed". Any fix needs to automatically update the table without page refresh.

qcjames53 (changes requested):
Thank you for the help reviewing this, Pavan. The autorefresh works well on my end and seems to be reasonably efficient with the network calls. Nice job! I can confirm the host list refreshes appropriately once the job is "warmed up" but unfortunately there are some issues when the server is slow:

With enough hosts, the calls to refresh table contents start getting to take longer than the job refresh clock (seems to be 1 second). I know this adds extra work but do you think it would be possible to prevent the job invocation element to stop refreshes while waiting for the server to send back the prior request? Implementing this might even give you the next issue for free.

There are also issues with page load. The table begins updating once a second before the initial contents even have the chance to load, causing the element to flip between page load and content load animations. I would disable this autorefresh behavior entirely until the table contents are populated for the first time.

qcjames53 (changes requested):
I feel bad to keep requesting changes but there are still scenarios where this autoupdate slams the server with dozens of requests at once. I'm seeing this when I am waiting on one job to complete and start another, making all of the hosts stay in "pending" for a while. The requests named "7" are the ones driving the autoupdate and they pretty much DOS attack a single-threaded server.
image

qcjames53:
Oh but other that that everything works really well! It's just I can't ACK with this page DOSing slow single threaded servers.

ekohl:
AflFAIK today we don't use it but we may want to consider https://guides.rubyonrails.org/action_cable_overview.html to get updates. Rather than polling you can subscribe to changes. Perhaps an improvement we can make? For example, it could get all job statuses and then subscribe to all jobs for updates. No ddos in terms of requests.

I don't want to push it, but in the past we first didn't use it because it didn't exist. Then we used mod_passenger which didn't support it. Now we can use it but don't. I'm not sure if there's a good reason or not.

qcjames53 (changes requested):
Thanks for these changes Pavan; the page itself runs great now no matter what I do! Huge improvement! It looks like there are some minor changes requested from the JS test cases but after you patch those up I'm okay to ACK.

Ewoud's suggestion is interesting. Let's talk about it in standup.

pavanshekar:
> Hi everyone, meanwhile my PR about API calls refactor got merged. @pavanshekar Can we retest it with latest version? In my case it looks like the issue disappeared. Also if the issue still occurs, it will definitely affect this code, because the API call logic has changed.

Thanks for confirming! I retested with the latest changes and the issue seems to be resolved. Since the original issue was about hosts remaining in “Pending” status even after a job completed successfully, and the expected behavior was that hosts should update to “Success” or “Completed,” it looks like @Lukshios’ refactor addresses this. I’ll go ahead and close this PR.

PR #1000, opened 2025-09-04
Fixes #38714 - use common LabelIcon for popup help

kfamilonidis:
The info button after Hosts -> Create Host -> Advanced -> "Remote Execution Interface" looks ok.


webpack/react_app/components/RegistrationExtension/__tests__/RexInterface.test.js
```diff
@@ -1,9 +1,19 @@
...
 import RexInterface from '../RexInterface';
 
 const fixtures = {
   renders: { isLoading: false, onChange: () => {} },
 };
 
-describe('RexInterface', () =>
-  testComponentSnapshotsWithFixtures(RexInterface, fixtures));
+describe('RexInterface', () => {
+  it('should render label with help icon and popover instructions', () => {
```
MariaAga, line 11: Not sure why the github actions are failing, but I ran it locally and it fails with:
```
 ../foreman_remote_execution/webpack/react_app/components/RegistrationExtension/__tests__/RexInterface.test.js (234 MB heap size)
  RexInterface
    ✕ should render label with help icon and popover instructions (40 ms)

  ● RexInterface › should render label with help icon and popover instructions

    TestingLibraryElementError: Unable to find an element with the text: Identifier of the Host interface for Remote execution. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```

PR #1002, opened 2025-09-09
Fixes #38730 - jobs wizard bottons are same size

kfamilonidis:
In Monitor -> Jobs -> Run Jobs -> all wizard steps footer buttons should be in same size

### DIFF
--- a/webpack/JobWizard/Footer.js
+++ b/webpack/JobWizard/Footer.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


MariaAga (changes requested):
The correct fix for this would be to move the Tooltip to not wrap the buttons, and use triggerRef instead of wrapping them. Read more here: https://v5-archive.patternfly.org/components/tooltip#tooltip

The bug is caused because some of the buttons have a div wrap and some dont, which effects the css 
The buttons can be found here: webpack/JobWizard/Footer.js

Image

MariaAga (approved):
Thank you!

PR #171, opened 2016-04-10
Fixes #14560 - Reuse foreman's core documentation helper.

ShimShtein:
Waits for https://github.com/theforeman/foreman/pull/3407.

PR #1008, opened 2025-09-29
Fixes #38785 - Align docs links with core's documentation versions

kfamilonidis:
**Acceptance Criteria**

Links are adjusted to core docs.

- [ ] Monitor -> Jobs -> link (path: `/job_invocations`)
  - [ ] empty list: Learn more about this in the documentation link
  - [ ] non-empty list: Documentation button
- [ ] Hosts -> Templates -> Job Templates -> link (path: `/job_templates`)
  - [ ] non-empty list: Documentation button

Removes use of :root_url in favor of foreman's core docs params.

Fixes df6ffa2 (Fixes #14560 - Reuse foreman's core documentation helper. (#171))

### DIFF
--- a/app/helpers/remote_execution_helper.rb
+++ b/app/helpers/remote_execution_helper.rb
@@ -54,7 +54,7 @@ module RemoteExecutionHelper
 
   def job_invocations_buttons
     [
-      documentation_button_rex('3.2ExecutingaJob'),
+      documentation_button('Managing_Hosts', type: 'docs', chapter: 'executing-a-remote-job_managing-hosts'),
       authorized_for(controller: :job_invocations, action: :create) ? link_to(_('Run Job'), hash_for_new_job_invocation_path, {:class => "btn btn-primary"}) : '',
     ]
   end
@@ -205,12 +205,6 @@ module RemoteExecutionHelper
       content_tag :span, (time > Time.now.utc ? _('in %s') : _('%s ago')) % time_ago_in_words(time),
         { :'data-original-title' => time.try(:in_time_zone), :rel => 'twipsy' }
     end
-  end
-
-  def documentation_button_rex(section = '')
-    url = 'http://theforeman.org/plugins/foreman_remote_execution/' +
-      "#{ForemanRemoteExecution::VERSION.split('.').take(2).join('.')}/index.html#"
-    documentation_button section, :root_url => url
   end
 
   def description_checkbox_f(f, job_template, disabled)
--- a/app/views/job_templates/index.html.erb
+++ b/app/views/job_templates/index.html.erb
@@ -2,7 +2,7 @@
 <%= javascript 'foreman_remote_execution/template_input' %>
 
 <% title _("Job Templates") %>
-<% title_actions(documentation_button_rex('3.1JobTemplates'),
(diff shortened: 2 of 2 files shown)


kfamilonidis:
this is now waiting for approval [PR](https://github.com/RedHatSatellite/foreman_theme_satellite/pull/271)

app/helpers/remote_execution_helper.rb
```diff
@@ -207,10 +207,8 @@ def time_in_words_span(time)
     end
   end
 
-  def documentation_button_rex(section = '')
-    url = 'http://theforeman.org/plugins/foreman_remote_execution/' +
-      "#{ForemanRemoteExecution::VERSION.split('.').take(2).join('.')}/index.html#"
-    documentation_button section, :root_url => url
+  def documentation_button_rex(chapter = '')
```
ekohl, line 210: This is also used in `app/views/job_templates/index.html.erb` and that needs to be modified too.

Now looking at the whole repo, is there any real point in keeping this helper? It's used twice and can we really assume that all the documentation will be in the `Managing_Hosts` guide? I'd suggest dropping it.

ekohl (changes requested):
Similar to the other PR, please describe what the change does, not what it solves.

app/helpers/remote_execution_helper.rb
kfamilonidis: Indeed, thanks for the suggestion. I also was between keeping this helper and wasn't sure if this helper pattern was used into other repo. It now make more sense since the core `:params` are stable. 

Also, at the `job_templates/index.erb`, the new link is included, but it needs a different `:key` in order to be activated [here](https://github.com/RedHatSatellite/foreman_theme_satellite/pull/271/files#diff-03deddd1ea19da8194b5ae0d27b67e7804f3618c2c8ed058c78a3a8ccbf5d622R81). Thanks for this also!

kfamilonidis:
> Should https://github.com/theforeman/foreman_remote_execution/blob/master/app/views/job_invocations/welcome.html.erb
> reuse the `documentation_button_rex`?

I removed `documentation_button_rex` altogether. empty-state link is a `<anchor>` tag, non-empty state links are `<button>` tags

app/views/job_templates/index.html.erb
```diff
@@ -2,7 +2,7 @@
 <%= javascript 'foreman_remote_execution/template_input' %>
 
 <% title _("Job Templates") %>
-<% title_actions(documentation_button_rex('3.1JobTemplates'),
+<% title_actions(documentation_button('Managing_Hosts', type: 'docs', chapter: 'creating-a-job-template_managing-hosts'),
```
ekohl, line 5: @aneta-petrova This may not be actionable in the moment, but in the old documentation we had a sort of "landing page" for job templates (https://theforeman.org/plugins/foreman_remote_execution/1.7/index.html#3.1JobTemplates). Now we link to https://docs.theforeman.org/3.16/Managing_Hosts/index-katello.html#creating-a-job-template_managing-hosts even though the page also allows importing etc.

Perhaps this is a nice hook into the REX content journey you had in mind.

ekohl (approved):
I'll leave this for others for a final review.

adamruzicka:
Thank you @kfamilonidis !

PR #1011, opened 2025-10-08
Fixes #38811 - Fix position of the overview on the job invocation page

kmalyjur:
Before:
image

After:
image

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationDetail.scss
+++ b/webpack/JobInvocationDetail/JobInvocationDetail.scss
@@ -2,6 +2,8 @@
   $chart_size: 105px;
   padding-top: 0px;
   padding-left: 10px;
+  display: flex;
+  align-items: center;
 
   .chart-donut {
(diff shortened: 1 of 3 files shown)


adamruzicka:
Attaching screenshots before and after of the same size to get a better overview of what the change actually does.

Before
1011-before

After
1011-after

With the patch applied, the first thing that drew my eyes was that there seems to be way more blank space and the table sits lower in the page. The redmine issue doesn't say much, but I assume the problem was that "Succeeded" and "In Progress" were not lined up with "Effective user" and "Started at"? If so, could they be lined up without pushing the table downwards that much?

kmalyjur:
@adamruzicka @Lukshio thank you for noticing and for the suggestion. Adam, does it help? Unfortunately, I don't see the blank space in my environment as you do.

adamruzicka:
Looks good on my end

adamruzicka (approved):

PR #1027, opened 2026-04-08
Fixes #39218 - Update Alert to PF5

Lukshio:


MariaAga (approved):
Code looks good, and the alert looks as it should. Can this be undrafted? 
Tested with packit 

before:
Image

after:
Image

Lukshio:
undrafted, do we need some other ack, or it's ready to go? @MariaAga

PR #1028, opened 2026-04-08
Fixes #39219 - Update Icons to PF5

Lukshio:


webpack/JobInvocationDetail/JobInvocationHostTable.js
```diff
@@ -446,6 +449,7 @@ const JobInvocationHostTable = ({
           isDeleteable={false}
           childrenOutsideTbody
         >
+          {customEmptyState}
```
adamlazik1: I assume this is a leftover from testing or helper for review.
```suggestion
```

adamlazik1 (commented):
Generally LGTM.

One question: Should we replace PlusCircleIcon in empty state for something that makes more sense, for example SearchIcon which is already used in foreman core for the EmptyState component? https://github.com/theforeman/foreman/blob/d296695fc2f032834d9c1f51c0d4165b3f6d4202/webpack/assets/javascripts/react_app/routes/common/EmptyPage/RedirectedEmptyPage.js#L59

webpack/JobInvocationDetail/JobInvocationHostTable.js
Lukshio: Yes, thanks.

Lukshio:
> Generally LGTM.
> 
> One question: Should we replace PlusCircleIcon in empty state for something that makes more sense, for example SearchIcon which is already used in foreman core for the EmptyState component? https://github.com/theforeman/foreman/blob/d296695fc2f032834d9c1f51c0d4165b3f6d4202/webpack/assets/javascripts/react_app/routes/common/EmptyPage/RedirectedEmptyPage.js#L59

I think that the icon is correct, we are using same icon in loading state of TableIndexPage https://github.com/theforeman/foreman/blob/2ff0f022f0dc4c2789fd6c8fb1effe21891dfe7e/webpack/assets/javascripts/react_app/routes/common/EmptyPage/index.js#L17

SearchIcon is used for empty result if I remember correctly.

adamlazik1 (approved):

adamlazik1:
Thanks @Lukshio !

PR #1030, opened 2026-04-24
Fixes #39253 - Implement "Not Found" error handling for invalid Job I…

pondrejk:
…nvocation IDs

SAT-44544

image

### DIFF
--- a/webpack/JobInvocationDetail/CheckboxesActions.js
+++ b/webpack/JobInvocationDetail/CheckboxesActions.js
@@ -14,6 +14,7 @@ import {
 } from '@patternfly/react-icons';
 import axios from 'axios';
 import { foremanUrl } from 'foremanReact/common/helpers';
+import { usePermissions } from 'foremanReact/common/hooks/Permissions/permissionHooks';
 import { translate as __, sprintf } from 'foremanReact/common/I18n';
 import { addToast } from 'foremanReact/components/ToastsList';
 import PropTypes from 'prop-types';
@@ -24,10 +25,7 @@ import {
   DIRECT_OPEN_HOST_LIMIT,
   templateInvocationPageUrl,
 } from './JobInvocationConstants';
-import {
-  selectHasPermission,
-  selectTaskCancelable,
-} from './JobInvocationSelectors';
+import { selectTaskCancelable } from './JobInvocationSelectors';
 import OpenAllInvocationsModal from './OpenAllInvocationsModal';
 
 /* eslint-disable camelcase */
@@ -115,12 +113,8 @@ export const CheckboxesActions = ({
   const dispatch = useDispatch();
   const [toBeOpened, setToBeOpened] = useState([]);
 
-  const hasCreatePermission = useSelector(
-    selectHasPermission('create_job_invocations')
-  );
-  const hasCancelPermission = useSelector(
-    selectHasPermission('cancel_job_invocations')
-  );
+  const hasCreatePermission = usePermissions(['create_job_invocations']);
+  const hasCancelPermission = usePermissions(['cancel_job_invocations']);
   const jobSearchQuery = `job_invocation.id = ${jobID}`;
   const filterQuery =
     filter && filter !== 'all_statuses'
--- a/webpack/JobInvocationDetail/JobInvocationEmptyState.js
+++ b/webpack/JobInvocationDetail/JobInvocationEmptyState.js
@@ -1,0 +1,86 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { useHistory } from 'react-router-dom';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import {
+  Bullseye,
+  Button,
+  EmptyState,
+  EmptyStateBody,
+  EmptyStateIcon,
+  PageSection,
+  PageSectionVariants,
+  EmptyStateVariant,
+  EmptyStateHeader,
+  EmptyStateActions,
+  EmptyStateFooter,
+} from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
+
+const jobInvocationsIndexPath = '/job_invocations';
+
+const JobInvocationEmptyState = ({ jobInvocationId }) => {
+  const history = useHistory();
+  return (
+    <PageSection variant={PageSectionVariants.light}>
+      <Bullseye>
+        <EmptyState
+          ouiaId="job-invocation-empty-state"
+          variant={EmptyStateVariant.lg}
+        >
+          <EmptyStateIcon icon={SearchIcon} />
+          <EmptyStateHeader
+            titleText={<>{sprintf(__('Job invocation not found'))}</>}
+            headingLevel="h5"
+          />
+
+          <EmptyStateBody>
+            {sprintf(
+              __(
+                'There is no job invocation with id %s or there are access permissions needed. Please contact your administrator if this issue continues.'
+              ),
+              jobInvocationId
+            )}
+          </EmptyStateBody>
+          <EmptyStateFooter>
+            <EmptyStateActions>
+              <Button
+                ouiaId="job-invocation-empty-state-go-to-job-invocations-button"
+                variant="primary"
+                component="a"
+                href={foremanUrl(jobInvocationsIndexPath)}
+                isInline
+              >
+                {__('Go to job invocations')}
+              </Button>
+            </EmptyStateActions>
+            <EmptyStateActions>
+              <Button
+                ouiaId="job-invocation-empty-state-create-new-job-invocation-button"
+                variant="link"
+                component="a"
+                href={foremanUrl('/job_invocations/new')}
+              >
+                {__('Create a new job invocation')}
+              </Button>
+              <Button
+                ouiaId="job-invocation-empty-state-return-to-last-page-button"
+                variant="link"
+                onClick={() => history.goBack()}
+              >
+                {__('Return to the last page')}
+              </Button>
+            </EmptyStateActions>
+          </EmptyStateFooter>
+        </EmptyState>
+      </Bullseye>
+    </PageSection>
+  );
+};
+
+JobInvocationEmptyState.propTypes = {
+  jobInvocationId: PropTypes.string.isRequired,
+};
+
+export default JobInvocationEmptyState;
--- a/webpack/JobInvocationDetail/JobInvocationEmptyState.js
+++ b/webpack/JobInvocationDetail/JobInvocationEmptyState.js
@@ -1,86 +1,89 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { useHistory } from 'react-router-dom';
-import { translate as __, sprintf } from 'foremanReact/common/I18n';
-import { foremanUrl } from 'foremanReact/common/helpers';
 import {
-  Bullseye,
   Button,
-  EmptyState,
-  EmptyStateBody,
-  EmptyStateIcon,
+  EmptyStateVariant,
   PageSection,
   PageSectionVariants,
-  EmptyStateVariant,
-  EmptyStateHeader,
-  EmptyStateActions,
-  EmptyStateFooter,
 } from '@patternfly/react-core';
 import { SearchIcon } from '@patternfly/react-icons';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import EmptyStatePattern from 'foremanReact/components/common/EmptyState/EmptyStatePattern';
 
 const jobInvocationsIndexPath = '/job_invocations';
 
-const JobInvocationEmptyState = ({ jobInvocationId }) => {
+const JobInvocationEmptyState = ({ jobInvocationId, errorMessage = null }) => {
   const history = useHistory();
+
+  const descriptionContent = (
+    <>
+      <p>
+        {sprintf(
+          __(
+            'There is no job invocation with id %s or there are access permissions needed. Opening this page requires the view_job_invocations permission. Please contact your administrator if this issue continues.'
+          ),
+          jobInvocationId
+        )}
+      </p>
+      {errorMessage ? (
+        <p className="pf-v5-u-text-break" style={{ whiteSpace: 'pre-wrap' }}>
+          {sprintf(__('The server returned: %s'), errorMessage)}
+        </p>
+      ) : null}
+    </>
+  );
+
   return (
     <PageSection variant={PageSectionVariants.light}>
-      <Bullseye>
-        <EmptyState
-          ouiaId="job-invocation-empty-state"
-          variant={EmptyStateVariant.lg}
-        >
-          <EmptyStateIcon icon={SearchIcon} />
-          <EmptyStateHeader
-            titleText={<>{sprintf(__('Job invocation not found'))}</>}
-            headingLevel="h5"
-          />
-
-          <EmptyStateBody>
-            {sprintf(
-              __(
-                'There is no job invocation with id %s or there are access permissions needed. Please contact your administrator if this issue continues.'
-              ),
-              jobInvocationId
-            )}
-          </EmptyStateBody>
-          <EmptyStateFooter>
-            <EmptyStateActions>
-              <Button
-                ouiaId="job-invocation-empty-state-go-to-job-invocations-button"
-                variant="primary"
-                component="a"
-                href={foremanUrl(jobInvocationsIndexPath)}
-                isInline
-              >
-                {__('Go to job invocations')}
-              </Button>
-            </EmptyStateActions>
-            <EmptyStateActions>
-              <Button
-                ouiaId="job-invocation-empty-state-create-new-job-invocation-button"
-                variant="link"
-                component="a"
-                href={foremanUrl('/job_invocations/new')}
-              >
-                {__('Create a new job invocation')}
-              </Button>
-              <Button
-                ouiaId="job-invocation-empty-state-return-to-last-page-button"
-                variant="link"
-                onClick={() => history.goBack()}
-              >
-                {__('Return to the last page')}
-              </Button>
-            </EmptyStateActions>
-          </EmptyStateFooter>
-        </EmptyState>
-      </Bullseye>
+      <EmptyStatePattern
+        variant={EmptyStateVariant.lg}
+        icon={<SearchIcon />}
+        header={__('Job invocation not found')}
+        description={descriptionContent}
+        action={
+          <Button
+            ouiaId="job-invocation-empty-state-go-to-job-invocations-button"
+            variant="primary"
+            component="a"
+            href={foremanUrl(jobInvocationsIndexPath)}
+            isInline
+          >
+            {__('Go to job invocations')}
+          </Button>
+        }
+        secondaryActions={
+          <>
+            <Button
+              ouiaId="job-invocation-empty-state-create-new-job-invocation-button"
+              variant="link"
+              component="a"
+              href={foremanUrl('/job_invocations/new')}
+            >
+              {__('Create a new job invocation')}
+            </Button>
+            <Button
+              ouiaId="job-invocation-empty-state-return-to-last-page-button"
+              variant="link"
+              onClick={() => history.goBack()}
+            >
+              {__('Return to the last page')}
+            </Button>
+          </>
+        }
+      />
     </PageSection>
   );
 };
 
 JobInvocationEmptyState.propTypes = {
   jobInvocationId: PropTypes.string.isRequired,
+  errorMessage: PropTypes.string,
+};
+
+JobInvocationEmptyState.defaultProps = {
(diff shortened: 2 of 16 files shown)


webpack/JobInvocationDetail/JobInvocationEmptyState.js
```diff
@@ -0,0 +1,86 @@
...
+          <EmptyStateIcon icon={SearchIcon} />
+          <EmptyStateHeader
+            titleText={<>{sprintf(__('Job invocation not found'))}</>}
+            headingLevel="h5"
+          />
+
+          <EmptyStateBody>
+            {sprintf(
+              __(
+                'There is no job invocation with id %s or there are access permissions needed. Please contact your administrator if this issue continues.'
```
adamruzicka: From the top of my head, I could envision four error cases that can happen:

1. The user doesn't have permission to view job invocations at all (backend should give 403 unauthorized)
2. The user doesn't have permission to view this particular job invocation (backend should give 404)
3. The user has permission to view, but the job invocation doesn't exist (backend should give 404)
4. The backend failing to process the request (think 500 ISE)

Should we distinguish 1 vs 2+3 vs 4?

With that being said, the old page showed "Not found" for all cases, except 4 so feel free treat this as optional.

adamruzicka: Since we know what the permissions required are, should we list them as we do on the unauthorized page?

adamruzicka (commented):
Some nitpicks, seems to work well.

webpack/JobInvocationDetail/JobInvocationEmptyState.js
```diff
@@ -0,0 +1,86 @@
...
+  EmptyStateVariant,
+  EmptyStateHeader,
+  EmptyStateActions,
+  EmptyStateFooter,
+} from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
+
+const jobInvocationsIndexPath = '/job_invocations';
+
+const JobInvocationEmptyState = ({ jobInvocationId }) => {
```
ofedoren: AFAIR we have some patterns in core:
 - EmptyState: https://github.com/theforeman/foreman/tree/develop/webpack/assets/javascripts/react_app/components/common/EmptyState
 - PermissionDenied: https://github.com/theforeman/foreman/tree/develop/webpack/assets/javascripts/react_app/components/PermissionDenied

Should we re-use them instead of implementing a brand new specific page for it? I mean if they are not feasible, that's okay, I'm just asking if it was considered? I might lag behind our UI changes, so not sure if those aren't marked for future removal.

webpack/JobInvocationDetail/JobInvocationEmptyState.js
pondrejk: Would it make sense to include the backend message to the page? It could be more informative without us needing to introduce some parsing logic, wdyt?

adamruzicka: Not sure if the backend says anything reasonable, but worth a try.

webpack/JobInvocationDetail/JobInvocationEmptyState.js
pondrejk: I did consider them, but found them a bit too restrictive. Otoh, if its worth it I can also extend them based on what we agree upon here

MariaAga: What about them was too restrictive? might be worth to extend them as you said

webpack/JobInvocationDetail/JobInvocationEmptyState.js
MariaAga: In some cases the backend adds a nice message like for case 3 `»Resource job_invocation not found by id '88'«`

pondrejk:
@ofedoren @MariaAga @adamruzicka since the last time:
- I changed thing so that the EmptyStatePattern from core is used
- I added a note about needed perms
- I display also the message from server

Screenshot 2026-05-12 at 15 41 38

webpack/JobInvocationDetail/JobInvocationEmptyState.js
```diff
@@ -0,0 +1,89 @@
...
+  PageSectionVariants,
+} from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import EmptyStatePattern from 'foremanReact/components/common/EmptyState/EmptyStatePattern';
+
+const jobInvocationsIndexPath = '/job_invocations';
+
+const JobInvocationEmptyState = ({ jobInvocationId, errorMessage = null }) => {
```
ofedoren: We already use defaultProps for that
```suggestion
const JobInvocationEmptyState = ({ jobInvocationId, errorMessage }) => {
```

ofedoren (commented):
It's more a high level review, I'd rather leave the React stuff for someone else :)

1. One of the reasons why I wanted to re-use or enhance and re-use something from Foreman is because we're still trying to unify the UI across the pages at some degree.

Looking at the recent change to models page in Foreman and here I see a bunch of common things, essentially a lot of repetitions, which makes it harder to unify in the future if we'd need some changes. Anyway, at least now these two pages have at least different strings:

Image

Image

It just feels odd :/

2. We'd rather add tests, especially since we can use AI for that, just make sure it actually tests something.

3. There is yet another custom error formatter. And AFAIU there is an idea of creating a general formatter that can be re-used then in other places. Should we wait for that or at least mark somehow to revisit this one and replace after the general/common one is available.

4. Some inline comments.

pondrejk:
> Looking at the recent change to models page in Foreman and here I see a bunch of common things, essentially a lot of repetitions, which makes it harder to unify in the future if we'd need some changes.

Well I thought the point of these patterns was to declare the overall shape of things for the implementations to fill in their own content that could be site specific. To me they look fairly unified, I can extend the foreman pattern, but then we'd need to agree on which parts should be unified and which should stay modifiable

pondrejk:
this is now dependent on https://github.com/theforeman/foreman/pull/10986

adamruzicka (changes requested):
This will also require a required foreman version bump in engine.rb, ideally as a separate commit.

pondrejk:
> This will also require a required foreman version bump in engine.rb, ideally as a separate commit.

Hi, opened https://github.com/theforeman/foreman_remote_execution/pull/1039

webpack/JobInvocationDetail/JobInvocationEmptyState.js
```diff
@@ -0,0 +1,51 @@
...
+      requiredPermissions={['view_job_invocations']}
+      primaryAction={{
+        label: __('Go to job invocations'),
+        onClick: () => visit(foremanUrl(jobInvocationsIndexPath)),
+        ouiaId: 'job-invocation-empty-state-go-to-job-invocations-button',
+      }}
+      secondaryActions={[
+        {
+          label: __('Create a new job invocation'),
+          onClick: () => visit(foremanUrl('/job_invocations/new')),
```
MariaAga: It seems that `ResourceLoadFailedEmptyState` accepts urls, and uses react-router for the actions (which is faster than visit), could the `onClick`s be replaced with a url?

webpack/JobInvocationDetail/index.js
```diff
@@ -51,9 +58,28 @@ const JobInvocationDetailPage = ({
...
-  useAPI('get', currentPermissionsUrl, {
-    key: CURRENT_PERMISSIONS,
-  });
+  const { status: permissionsApiStatus } = useAPI(
+    'get',
+    currentPermissionsUrl,
+    {
+      key: CURRENT_PERMISSIONS,
+    }
+  );
```
MariaAga: I do think its good to check if the permissions api failed, but I think this logic should be moved to core (in a new PR, unrelated to this one)

MariaAga (changes requested):

webpack/JobInvocationDetail/JobInvocationEmptyState.js
pondrejk: I tried it out but didn't work for me, maybe because /job_invocations is still a rails index page?

MariaAga: In what way did it not work? I tried it out locally and using 
```
      primaryAction={{
        label: __('Go to job invocations'),
        url: foremanUrl(jobInvocationsIndexPath),
        ouiaId: 'job-invocation-empty-state-go-to-job-invocations-button',
      }}
      secondaryActions={[
        {
          label: __('Create a new job invocation'),
          url: foremanUrl('/job_invocations/new'),
          ouiaId: 'job-invocation-empty-state-create-new-job-invocation-button',
        },
      ]}
   ```
   Works, and navigates

**Review by MariaAga — CHANGES_REQUESTED** [review]
Since the empty state unloads the job invocation page when the api is done, with an error, it causes some components (JobInvocationToolbarButtons) to unload to fast and cause console warnings. I think this could be fixed by doing something like `toolbarButtons={ jobInvocationApiStatus === API_STATUS.SUCCESS && <JobInvocationToolbarButtons jobId={id} data={items} />}`

File: webpack/JobInvocationDetail/JobInvocationEmptyState.js (hunk 4 above)
Comment (pondrejk) [reply webpack/JobInvocationDetail/JobInvocationEmptyState.js]: right, works for me to :) honestly, I don't know what was the issue last time I tried, sorry for the confusion

**Comment by pondrejk** [comment]
> Since the empty state unloads the job invocation page when the api is done, with an error, it causes some components (JobInvocationToolbarButtons) to unload to fast and cause console warnings. I think this could be fixed by doing something like `toolbarButtons={ jobInvocationApiStatus === API_STATUS.SUCCESS && <JobInvocationToolbarButtons jobId={id} data={items} />}`

thanks for noticing, I added your suggestion, but I had to also add some isMounted conditions to JobInvocationToolbarButtons for the console warning to stop

File: webpack/JobInvocationDetail/index.js (hunk 5 above)
Comment (pondrejk) [reply webpack/JobInvocationDetail/index.js]: I'm not sure how to go about this, since permissionsApiStatus is one of the things we use to find what errorMessage to send to the ResourceLoadFailedEmptyState. Can you suggest where should I put that logic in core and how I'd trigger it?

Comment (MariaAga) [reply webpack/JobInvocationDetail/index.js]: Sorry my bad, I  rechecked, and the data from `currentPermissionsUrl` is already in `useForemanPermissions` in the empty state, so no need to call it at all.

Foreman ui meta data (which goes to `useForemanPermissions`)
` permissions: (User.current.admin? ? Permission.all : User.current.permissions).pluck(:name),`
https://github.com/theforeman/foreman/blob/07ce9438d216cf7d1cef3763de66b83f22669c16/app/helpers/application_helper.rb#L422

api call to `currentPermissionsUrl` 
`@current_permissions = @user.admin? ? Permission.all : @user.permissions`
https://github.com/theforeman/foreman/blob/07ce9438d216cf7d1cef3763de66b83f22669c16/app/controllers/api/v2/permissions_controller.rb#L32

Comment (pondrejk) [reply webpack/JobInvocationDetail/index.js]: makes sense, so I removed those currentPermissionsUrl calls

**Comment by pondrejk** [comment]
@MariaAga Hi, I think I've addressed all the comments, can you plrease take a look again?

File: webpack/JobInvocationDetail/index.js [hunk 6]
```diff
@@ -123,7 +154,11 @@ const JobInvocationDetailPage = ({
       <PageLayout
         header={description}
         breadcrumbOptions={breadcrumbOptions}
-        toolbarButtons={<JobInvocationToolbarButtons jobId={id} data={items} />}
+        toolbarButtons={
+          jobInvocationApiStatus === API_STATUS.SUCCESS && (
```
MariaAga: The STATUS constant from foremanReact/constants has keys PENDING, RESOLVED, and ERROR — there is no SUCCESS. This comparison always evaluates to false, so
  JobInvocationToolbarButtons is never rendered regardless of API state. Should be API_STATUS.RESOLVED.

webpack/JobInvocationDetail/index.js
```diff
@@ -88,6 +98,27 @@ const JobInvocationDetailPage = ({
...
+  const apiFailed = jobInvocationApiStatus === API_STATUS.ERROR;
+
+  const backendErrorMessage = useMemo(() => {
+    if (jobInvocationApiStatus === API_STATUS.ERROR) {
+      return jobInvocationErrorMessage || null;
+    }
+    return null;
+  }, [jobInvocationApiStatus, jobInvocationErrorMessage]);
+
+  const httpStatus = jobInvocationHttpStatus;
```
MariaAga: Used once, no transformation. Pass jobInvocationHttpStatus directly to the prop.

webpack/__mocks__/foremanReact/common/helpers.js
```diff
@@ -1 +1,5 @@
 export const foremanUrl = path => `foreman${path}`;
+
+export const visit = jest.fn(url => {
+  global.window.location.href = url;
+});
```
MariaAga: not used

webpack/JobInvocationDetail/JobInvocationEmptyState.js
```diff
@@ -0,0 +1,51 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { PageSection, PageSectionVariants } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import ResourceLoadFailedEmptyState from 'foremanReact/components/common/EmptyState/ResourceLoadFailedEmptyState';
+
+const jobInvocationsIndexPath = '/job_invocations';
```
MariaAga: should live in JobInvocationConstants.js alongside jobInvocationDetailsUrl.

MariaAga (changes requested):
I ran an AI review, and it caught that with the api permisisons removal, the toolbar and the actions, dont get to see the correct permissions, so it will be good to switch out the permissions checks in `webpack/JobInvocationDetail/CheckboxesActions.js` and `webpack/JobInvocationDetail/JobInvocationToolbarButtons.js` with the `usePermissions` hook. (this causes the buttons next to the search bar to always be disabled)

Also missing a test to make sure the empty state loads.

pondrejk:
> I ran an AI review, and it caught that with the api permisisons removal, the toolbar and the actions, dont get to see the correct permissions, so it will be good to switch out the permissions checks in `webpack/JobInvocationDetail/CheckboxesActions.js` and `webpack/JobInvocationDetail/JobInvocationToolbarButtons.js` with the `usePermissions` hook. (this causes the buttons next to the search bar to always be disabled)
> 
> Also missing a test to make sure the empty state loads.

Hi @MariaAga, I did the requested updates, please take a look when you have time

webpack/JobInvocationDetail/__tests__/MainInformation.test.js
```diff
@@ -50,6 +58,33 @@ jest.mock('foremanReact/common/hooks/API/APIHooks', () => ({
   })),
 }));
 
+jest.mock('foremanReact/common/hooks/Permissions/permissionHooks', () => ({
+  usePermissions: jest.fn(() => true),
+}));
```
MariaAga: If the mock is the same as the mock file, no need to mock it again

webpack/JobInvocationDetail/__tests__/JobInvocationDetailEmptyState.test.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import { createMemoryHistory } from 'history';
+import { Router } from 'react-router-dom';
+import JobInvocationEmptyState from '../JobInvocationEmptyState';
+
+jest.mock('foremanReact/common/hooks/Permissions/permissionHooks', () => ({
+  usePermissions: jest.fn(() => true),
+}));
```
MariaAga: If the mock is the same as the mock file, no need to mock it again

webpack/JobInvocationDetail/__tests__/JobInvocationDetailEmptyState.test.js
```diff
@@ -0,0 +1,82 @@
...
+        </button>
+        {secondaryActions.map(action => (
+          <button key={action.ouiaId} type="button" data-testid={action.ouiaId}>
+            {action.label}
+          </button>
+        ))}
+      </div>
+    ),
+  })
+);
```
MariaAga: Prefer just relying on the original component, rather than mocking it.

webpack/JobInvocationDetail/__tests__/MainInformation.test.js
```diff
@@ -50,6 +58,33 @@ jest.mock('foremanReact/common/hooks/API/APIHooks', () => ({
...
+        <p>{`The ${resourceLabel} with id ${resourceId} could not be found.`}</p>
+        {errorMessage && <p>{`Server returned: ${errorMessage}`}</p>}
+        <button type="button" data-testid={primaryAction.ouiaId}>
+          {primaryAction.label}
+        </button>
+      </div>
+    ),
+  })
+);
+
```
MariaAga: Prefer just relying on the original component, rather than mocking it.

webpack/JobInvocationDetail/__tests__/MainInformation.test.js
```diff
@@ -50,6 +58,33 @@ jest.mock('foremanReact/common/hooks/API/APIHooks', () => ({
   })),
 }));
```
MariaAga: useAPI is no longer used

webpack/JobInvocationDetail/CheckboxesActions.js
```diff
@@ -24,10 +25,7 @@ import {
   DIRECT_OPEN_HOST_LIMIT,
   templateInvocationPageUrl,
 } from './JobInvocationConstants';
-import {
-  selectHasPermission,
```
MariaAga, line 28: Since this is no longer used, it can be removed from `JobInvocationSelectors`

webpack/JobInvocationDetail/index.js
```diff
@@ -5,29 +5,33 @@ import {
...
 import JobInvocationToolbarButtons from './JobInvocationToolbarButtons';
 import { getJobInvocation, getTask } from './JobInvocationActions';
 import './JobInvocationDetail.scss';
 import {
-  CURRENT_PERMISSIONS,
   DATE_OPTIONS,
   JOB_INVOCATION_KEY,
   STATUS,
   STATUS_UPPERCASE,
-  currentPermissionsUrl,
```
MariaAga, line 30: Since this (`currentPermissionsUrl`,`CURRENT_PERMISSIONS`)  is no longer used, it can be removed from `JobInvocationConstants`

webpack/JobInvocationDetail/JobInvocationEmptyState.js
```diff
@@ -0,0 +1,50 @@
...
+      requiredPermissions={['view_job_invocations']}
+      primaryAction={{
+        label: __('Go to job invocations'),
+        url: foremanUrl(jobInvocationsIndexPath),
+        ouiaId: 'job-invocation-empty-state-go-to-job-invocations-button',
+      }}
+      secondaryActions={[
+        {
+          label: __('Create a new job invocation'),
+          url: foremanUrl('/job_invocations/new'),
```
MariaAga: both urls should be consistent, and the  `foremanUrl` can be moved to the consts

MariaAga (changes requested):
Thanks! PR logic works as expected, but needs some code changes

We prefer to test what the user sees, rather than relying on test ids, so `screen.getByTestId(
        'job-invocation-empty-state-go-to-job-invocations-button'
      )` should be changed to things like `screen.getByRole('button', { name: 'Create a new job invocation' })`

You can read more about testing guidelines here: https://github.com/theforeman/foreman/blob/develop/developer_docs/ui-testing-guidelines.asciidoc

webpack/JobInvocationDetail/JobInvocationSelectors.js
```diff
@@ -32,13 +31,3 @@ export const selectTemplateInvocationList = state =>
 
 export const selectCurrentPermisions = state =>
   selectAPIResponse(state, CURRENT_PERMISSIONS);
```
MariaAga: can also be removed since its not used, and CURRENT_PERMISSIONS is undefined.

MariaAga:
I started testing that the page still works, but now when a job is running the buttons (create report, rerun all) are flashing, and it didnt seem to happen without the pr 

https://github.com/user-attachments/assets/18fdbc4c-48f6-4dbc-aaf5-a97262513107

pondrejk:
> I started testing that the page still works, but now when a job is running the buttons (create report, rerun all) are flashing, and it didnt seem to happen without the pr

great chatch! I did some update to fix that with ai assistance, I'm posting it as a separate commit for reviewability

MariaAga (approved):
Thanks! code lgtm, and I tested it

PR #1031, opened 2026-04-24
Fixes #39256 - Fix dependency issues for foreman_remote_execution

ofedoren:


adamruzicka (commented):
I don't think we have any rdoc annotations anywhere. What if we dropped it altogether along with the broken rdoc rake task?

ofedoren:
Thanks, @adamruzicka, gladly did that in a separate commit :)

adamruzicka (approved):
APT

adamruzicka:
Thank you @ofedoren !

PR #1032, opened 2026-04-29
Fixes #39274 - Fix duplicate hosts in job_invocation scoped searches

kmalyjur:
Searching by job_invocation fields returned duplicate hosts due to the use of INNER JOINs. I tried to fix this by `adding :distinct => true`, but it doesn't work because `scoped_search` ignores that parameter.

Test by searching "job_invocation.result = failed" in the all hosts table.

### DIFF
--- a/app/models/concerns/foreman_remote_execution/host_extensions.rb
+++ b/app/models/concerns/foreman_remote_execution/host_extensions.rb
@@ -32,18 +32,21 @@ module ForemanRemoteExecution
 
         def search_by_job_invocation(key, operator, value)
           if key == 'job_invocation.result'
-            operator = operator == '=' ? 'IN' : 'NOT IN'
-            value = TemplateInvocation::TaskResultMap.status_to_task_result(value)
-          end
+            in_operator = operator == '=' ? 'IN' : 'NOT IN'
+            task_results = TemplateInvocation::TaskResultMap.status_to_task_result(value)
+            subquery = TemplateInvocation.joins(:run_host_job_task)
+                                         .where.not(host_id: nil)
+                                         .where("#{ForemanTasks::Task.table_name}.result IN (?)", task_results)
+                                         .select(:host_id)
 
-          mapping = {
-            'job_invocation.id'     => %(#{TemplateInvocation.table_name}.job_invocation_id #{operator} ?),
-            'job_invocation.result' => %(#{ForemanTasks::Task.table_name}.result #{operator} (?)),
-          }
-          {
-            :conditions => sanitize_sql_for_conditions([mapping[key], value_to_sql(operator, value)]),
-            :joins => { :template_invocations => [:run_host_job_task] },
-          }
+            { :conditions => "#{Host.table_name}.id #{in_operator} (#{subquery.to_sql})" }
+          else
+            subquery = TemplateInvocation.where.not(host_id: nil)
+                                         .where("job_invocation_id #{operator} ?", value_to_sql(operator, value))
+                                         .select(:host_id)
+
+            { :conditions => "#{Host.table_name}.id IN (#{subquery.to_sql})" }
+          end
         end
       end
     end


kmalyjur:
Made changes to fix two edge cases with negative searches (NOT IN). This should prevent the query from breaking when hitting NULL host IDs, and ensure hosts with mixed job results are properly excluded when filtering out a specific status.

adamruzicka:
With this fix, `job_invocation.result = failed` translates roughly to "all hosts which failed in _any_ job", correct?

The original intent was to provide a search that could be used to "find all the hosts that failed within the scope of a particular job". Searching for `job_invocation.result = failed` returned duplicate results, but AND-ing it with `job_invocation.id = 7` would filter those duplicates out, giving you a list of hosts that failed in job 7. Now this is one of the cases where two wrongs align to make a right.

To turn this around, the intent never was to use `job_invocation.result = something` on its own as the value of "give me all the hosts that ever had a rex job fail" gets less and less useful the more jobs you run.

With this patch applied, `job_invocation.result = failed AND job_invocation.id = 7` would give you a list of hosts that were involved in job 7 and which failed in _any_ job (ie. also include hosts that succeeded in job 7 but failed in a different one).

It seems we're trading a crude, but working solution for one which is technically correct, but which doesn't address our needs anymore.

kmalyjur:
@adamruzicka Thank you for the explanation, I think I misunderstood it.

What do you think about just handling this at the controller level instead by adding `distinct`?
For example, in `foreman/app/controllers/api/v2/hosts_controller.rb`, we could update it to:
`@hosts = action_scope_for(:index, resource_scope_for_index).distinct`.

adamruzicka:
> What do you think about just handling this at the controller level instead by adding distinct?

That would do the trick, but distinct selects are somewhat expensive operations so I'd be hesitant to run them all the time, especially considering our usecase is just a tiny fraction of the requests that are processed by `Api::V2::HostsController#index`.

How about we mark the current external search method (or maybe just the `job_invocation.result` part) as deprecated and introduce a new one that would encode the job id in the key? Something like `job_invocation.7.result = failed`? The user experience isn't all that great, but it is self-explanatory and we'll be able to avoid "leaking" state between the individual parts of the search. @ofedoren opinions?

ofedoren:
>  @ofedoren opinions?

Ehh, hard to tell. In your https://github.com/theforeman/foreman_remote_execution/pull/1032#issuecomment-4344274246 you explicitly say that `job_invocation.result = something` was never intended to be used alone, which I personally didn't know :/ So looking at it with "newbie eyes" I'd say that if I put `job_invocation.result = failed` in the search for hosts, I'd expect to see all the hosts that have failed invocations regardless of a specific invocation. So this patch would fix that (duplicit entries).

At the same time if I wanted to check for all the hosts of a specific invocation, I'd intuitively tried specifying `and job_invocation.id = 7` to get rid of hosts/invocations I don't care, but this patch would break it :/

`job_invocation.7.result = failed` seems like a nice shortcut for the latter, but if we treat it as a proper fix or replacement for `job_invocation.result = something and job_invocation.id = 7` combo, we then lose the former.

Also, do we have anywhere `entity.id.field = value` syntax support already? Anyway, this would require some visibility at least in release notes.

I've also dug my crazy hat up, so WDYT about this madness:
 - leave `the combo`, but mark as deprecated, so we don't silently change anything
 - add new job_invocation.id.result syntax (as Adam suggested), so we cover the combo usage properly
 - add `job_invocationS.result` to have the UX of fixed `job_invocation.result` syntax, so we cover the first part of `the combo`
 - and as a even crazier alternative we could extend scoped_search with a context-based search. Something like wrapping a search with brackets (or adding a new `with` keyword) would mean that the searches are tight to each other instead of treating them as separate searches and `ANDing` it afterwards to create `the combo`. This might be overkill, but :shrug:

adamruzicka:
> Also, do we have anywhere entity.id.field = value syntax support already?

No, not yet and it wouldn't be exactly hassle-free either to get all the things like autocomplete work right.

> and as a even crazier alternative we could extend scoped_search with a context-based search. Something like wrapping a search with brackets (or adding a new with keyword) would mean that the searches are tight to each other instead of treating them as separate searches and ANDing it afterwards to create the combo. This might be overkill, but 🤷

Well, it is not as crazy as it sounds. It pretty much is the only proper solution we've heard so far, all the others were mostly workarounds.

As a yet-another workaround option, currently we construct the id-and-result form in couple of places ourselves. Maybe if we encoded the id part in a different way (like with implicit scoping through url parameters) and stopped relying on the pure search-based way, then that could be enough?

ofedoren:
> As a yet-another workaround option, currently we construct the id-and-result form in couple of places ourselves. Maybe if we encoded the id part in a different way (like with implicit scoping through url parameters) and stopped relying on the pure search-based way, then that could be enough?

You mean we do that ourselves in the places that are not directly visible to the users (in the code), but relying on something that is broken AND visible to the users (`the combo`)? If the issue invoking this patch was invisible to the users (something that we noticed whilst doing internal changes) then I'd say it could be enough, since for ~9 years we didn't get a filled bug to fix this explicitly. Maybe users don't even need that as much?

It's just if they do and rely on this as well, we need to properly state what's expected, what's not and fix or introduce "workarounds" as part of the fix "for ourselves" :/

In any case, there are 4-5? ways how to proceed, should we vote to pick at least something or it's worth to keep it for a while and play with the options?

PR #1033, opened 2026-05-21
Fixes #39344 - No error toast is shown when API call fails

Lukshio:
When server returns err 500 not toast is shown and skeleton loading continues.

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationActions.js
+++ b/webpack/JobInvocationDetail/JobInvocationActions.js
@@ -24,6 +24,13 @@ export const getJobInvocation = url => dispatch => {
       handleError: () => {
         dispatch(stopInterval(JOB_INVOCATION_KEY));
       },
+      errorToast: ({ response }) =>
+        // eslint-disable-next-line camelcase
+        response?.data?.error?.full_messages?.[0] ||
+        // eslint-disable-next-line camelcase
+        response?.data?.error?.full_messages ||
+        response?.data?.error?.message ||
+        'Error',
     }),
     1000
   );


webpack/JobInvocationDetail/JobInvocationActions.js
```diff
@@ -24,6 +24,7 @@ export const getJobInvocation = url => dispatch => {
       handleError: () => {
         dispatch(stopInterval(JOB_INVOCATION_KEY));
       },
+      errorToast: err => err?.message || 'Error',
```
ofedoren: Without manual testing, it's hard to tell if it actually works :)

Just out of curiosity: why do we check for `message` here only, but in other places (such as below) we either show the whole `err` (which is `response` in those places) or dig through it to find `full_messages || message`

Lukshio: In this case server returns only object below. But yes, there can be some other responses, updated. 

```
{
    "error": {
        "message": "»Internal Server Error: the server was unable to finish the request. This may be caused by unavailability of some required service, incorrect API call or a server-side bug. There may be more information in the server's logs.«"
    }
}
```

webpack/JobInvocationDetail/JobInvocationActions.js
```diff
@@ -24,6 +24,11 @@ export const getJobInvocation = url => dispatch => {
       handleError: () => {
         dispatch(stopInterval(JOB_INVOCATION_KEY));
       },
+      errorToast: ({ response }) =>
+        // eslint-disable-next-line camelcase
+        response?.data?.error?.full_messages ||
+        response?.data?.error?.message ||
```
adamruzicka, line 32: Iirc `full_messages` is an array, while `message` (or the default) are strings. Would the toast be able to deal with that in a sane way?

Lukshio: As I wrote above, I didn't found any error using `full_message` in this place. But we have usages with and without array from `full_messages`. Just to be sure, I've just added the array variant. Hopefully the new helper will come asap :)

adamruzicka (approved):

adamruzicka:
Thank you @Lukshio !

PR #1034, opened 2026-05-26
API performance fixes

adamruzicka:
- **Call set_statuses after pagination**
- **Do not ask for host statuses when reading job invocation details**
- **to_h over Hash[col.map { ... }]**

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -314,8 +314,9 @@ module Api
         template_invocations = @template_invocations.where(host_id: @hosts.select(:id))
                                                     .includes(:run_host_job_task).to_a
         hosts = @hosts.to_a
+        template_invocations_by_host_id = template_invocations.index_by(&:host_id)
         @host_statuses = hosts.to_h do |host|
-          template_invocation = template_invocations.find { |ti| ti.host_id == host.id }
+          template_invocation = template_invocations_by_host_id[host.id]
           task = template_invocation.try(:run_host_job_task)
           [host.id, template_invocation_status(task, @job_invocation.task)]
         end
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -314,13 +314,13 @@ module Api
         template_invocations = @template_invocations.where(host_id: @hosts.select(:id))
                                                     .includes(:run_host_job_task).to_a
         hosts = @hosts.to_a
-        @host_statuses = Hash[hosts.map do |host|
+        @host_statuses = hosts.to_h do |host|
           template_invocation = template_invocations.find { |ti| ti.host_id == host.id }
           task = template_invocation.try(:run_host_job_task)
           [host.id, template_invocation_status(task, @job_invocation.task)]
-        end]
-        @smart_proxy_id = Hash[template_invocations.map { |ti| [ti.host_id, ti.smart_proxy_id] }]
-        @smart_proxy_name = Hash[template_invocations.map { |ti| [ti.host_id, ti.smart_proxy_name] }]
+        end
+        @smart_proxy_id = template_invocations.to_h { |ti| [ti.host_id, ti.smart_proxy_id] }
+        @smart_proxy_name = template_invocations.to_h { |ti| [ti.host_id, ti.smart_proxy_name] }
       end
     end
   end
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -115,10 +115,10 @@ module Api
       param :id, :identifier, :required => true
       def hosts
         set_hosts_and_template_invocations
-        set_statuses_and_smart_proxies
         @total = @hosts.size
         @hosts = @hosts.search_for(params[:search], :order => params[:order]).paginate(:page => params[:page], :per_page => params[:per_page])
         @subtotal = @hosts.total_entries
+        set_statuses_and_smart_proxies
         if params[:awaiting]
           @hosts = @hosts.select { |host| @host_statuses[host.id] == 'N/A' }
         end
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -311,7 +311,8 @@ module Api
       end
 
       def set_statuses_and_smart_proxies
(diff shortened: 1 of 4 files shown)


adamruzicka:
For a job with 10000 hosts moving set_statuses after pagination brings calls to `/api/job_invocations/<job_id>/hosts?include_permissions=true&per_page=1` from 1.5 minutes as admin and 15 minute timeout as a non-admin user to 12 seconds as admin and ~1 minute as a non-admin.

Note, this is an instance with 300 organizations and 1800 locations so a lot of time is burned on permission checks related to them.

ofedoren:
> This review was AI-generated by Claude Code (claude-opus-4-6).

## :bug: What's wrong

### 1. `params[:awaiting]` is broken by commit 1

**Before:**
```ruby
set_statuses_and_smart_proxies   # populates @host_statuses for ALL hosts
@hosts = ...paginate(...)
if params[:awaiting]
  @hosts = @hosts.select { |host| @host_statuses[host.id] == 'N/A' }
end
```

**After:**
```ruby
@hosts = ...paginate(...)
if params[:awaiting]
  @hosts = @hosts.select { |host| @host_statuses[host.id] == 'N/A' }  # @host_statuses NOT SET YET
end
set_statuses_and_smart_proxies   # too late
```

`@host_statuses` is `nil` when the filter runs → `NoMethodError: undefined method '[]' for nil`. The frontend does use this parameter (`JobInvocationHostTable.js:200` sets `finalParams.awaiting = 'true'`).

**Suggestion:** move `set_statuses_and_smart_proxies` between pagination and the awaiting filter, or rework the awaiting logic.

---

## :thinking: What's weird

### 2. The test still expects `?host_status=true`

`MainInformation.test.js:235` asserts:
```js
url: `/api/job_invocations/${jobId}?host_status=true`,
```
But `index.js` now dispatches without it. This test should fail — either it's not being run or the assertion is inside a block that doesn't actually execute. Needs updating either way.

### 3. Inconsistent `Hash[]` vs `to_h` on adjacent lines

Commit 3 converts `@host_statuses` to `to_h` but leaves `@smart_proxy_id` and `@smart_proxy_name` right below in the old style:
```ruby
@smart_proxy_id = Hash[template_invocations.map { |ti| [ti.host_id, ti.smart_proxy_id] }]
@smart_proxy_name = Hash[template_invocations.map { |ti| [ti.host_id, ti.smart_proxy_name] }]
```
Mixing styles in adjacent lines is worse than using either one consistently.

### 4. "Doesn't seem to be used anywhere" is a bit misleading (commit 2)

The `show` endpoint's `host_status=true` path populates `job_status` on each host via RABL (`main.json.rabl:25-28`). Removing it from the React detail page doesn't mean nothing uses it — the API is public and other consumers (Hammer CLI, scripts, plugins) may rely on it. The param still exists and works; it's just no longer requested by the default UI. That's fine, but the commit message implies the data is dead code.

### 5. Pre-existing O(n*m) linear scan (not introduced by this PR)

Since the PR is about performance: `template_invocations.find { |ti| ti.host_id == host.id }` is O(n) per host. For a paginated page of 20 this doesn't matter, but a hash lookup would be trivial:
```ruby
ti_by_host = template_invocations.index_by(&:host_id)
```

---

## :mag: What's missing

### 6. No test for the core pagination fix (commit 1)

This is the most impactful change and there's no test asserting that statuses are correctly computed for the paginated subset. A controller test exercising `GET /api/v2/job_invocations/:id/hosts?page=1&per_page=5` and verifying `job_status` is present would catch regressions.

### 7. No test update for the `host_status=true` removal (commit 2)

The existing test at `MainInformation.test.js:235` asserts the old URL. Needs updating to match the new behavior.

adamruzicka:
Addressed all of the points.

> "Doesn't seem to be used anywhere"

That was meant as "the additional data that was loaded by the frontend was not be used by it anywhere". Not that there is no other consumer

ofedoren:
> This review was AI-generated by Claude Code (claude-opus-4-6).

## Looks good

The core optimization is correct and the numbers are compelling (1.5min → 12s). Moving `set_statuses_and_smart_proxies` after pagination avoids materializing all hosts in the job — `@hosts.to_a` inside that method now only loads the current page. The `index_by` lookup (commit 4) and `to_h` cleanup (commit 3) are good follow-ups. The test in commit 1 covers the important case.

## Partial optimization: `@template_invocations` still loads everything

`@template_invocations` is built in `set_hosts_and_template_invocations` with `.where(host: @hosts)`, where `@hosts` is the **unpaginated** AR relation. ActiveRecord embeds this as a subquery at build time. When `@hosts` is later reassigned to the paginated result, `@template_invocations` still references the original subquery. So this line in `set_statuses_and_smart_proxies`:

```ruby
template_invocations = @template_invocations.includes(:run_host_job_task).to_a
```

still loads template invocations for **all** hosts in the job, not just the page. For a 10k-host job, that's 10k `TemplateInvocation` + `ForemanTasks::Task` records loaded into memory to look up 20.

Same applies to `@smart_proxy_id` and `@smart_proxy_name` — they're built from the full `template_invocations` array.

Not a bug (extra hash keys are harmless, and `hosts.to_a` is correctly scoped to the page), but it leaves performance on the table. Could be addressed by re-scoping:

```ruby
template_invocations = @template_invocations.where(host_id: @hosts.select(:id))
                                            .includes(:run_host_job_task).to_a
```

## Nit: unrelated reformatting in commit 2

`MainInformation.test.js` has a whitespace-only change to the `Date.prototype.toLocaleString` override that isn't related to the `host_status` removal. Probably an auto-formatter — harmless but adds noise to the diff.

ofedoren:
Thanks, @adamruzicka, I'd like to get this in, but I'm not sure what's your plan exactly given the two commits have different issues attached and the other two are issue-less... Otherwise, "@template_invocations still loads everything" concern from Claude seems to be worth considering.

adamruzicka:
> what's your plan exactly given the two commits have different issues attached and the other two are issue-less

A plan is a strong word :) The two issue-less commits were more of a "while I'm at it" sort of thing. They're not necessarily related to either of the issue-linked ones, but at the same time they're rather small so filing a redmine for those felt like an unnecessary overhead. I'll look into the template invocation thing, maybe I'll find a way how to include those two in it.

adamruzicka:
Added redmine for the hash-related commits, resolved the template_invocation thing as suggested and dropped the unrelated reformatting.

ofedoren (approved):
Thanks, @adamruzicka !

PR #1036, opened 2026-05-27
Stop serializing all hosts and template invocations in job invocation

adamruzicka:
> The show endpoint serialized every host (via hosts/base) and every template invocation on each request, including the 1-second polling requests from the detail page. None of this data is used by the frontend — hosts are fetched separately by the host table, and per-host template invocations are fetched on row expand.

The biggest problem with this is that removing it changes a public api. To offer some alternatives, we could:
- add a parameter that would just disable the parts that I'm proposing to remove
- add a new endpoint, deprecate the old one and eventually remove it, but then we'd eventually end up with an oddly shaped api

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -19,14 +19,10 @@ module Api
 
       api :GET, '/job_invocations/:id', N_('Show job invocation')
       param :id, :identifier, :required => true
-      param :host_status, :bool, required: false, desc: N_('Show Job status for the hosts')
       def show
-        set_hosts_and_template_invocations
+        @pattern_template_invocations = @job_invocation.pattern_template_invocations.includes(:input_values)
         @job_organization = Taxonomy.find_by(id: @job_invocation.task.input[:current_organization_id])
         @job_location = Taxonomy.find_by(id: @job_invocation.task.input[:current_location_id])
-        if params[:host_status] == 'true'
-          set_statuses_and_smart_proxies
-        end
       end
 
       def_param_group :job_invocation do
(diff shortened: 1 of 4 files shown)


adamlazik1:
Avoiding changing public API could save us from potential issues in the future. I would like to go with your suggested way to add a parameter. e.g. serialize_all_hosts (default: true).

adamruzicka:
> Avoiding changing public API could save us from potential issues in the future

I know, I know, it's just that it is something that should never have been added. Keeping compatibility is nice, having to opt-out from footguns not so much. Anyway, done, but if we go this way this will also need a corresponding change to h-rex to opt out from getting data it doesn't need.

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -19,13 +19,16 @@ def index
 
       api :GET, '/job_invocations/:id', N_('Show job invocation')
       param :id, :identifier, :required => true
-      param :host_status, :bool, required: false, desc: N_('Show Job status for the hosts')
+      param :include_hosts, :bool, :required => false, :default_value => true, :desc => N_('Include hosts and template invocations in the response. Defaults to true for backwards compatibility, pass false to skip serializing all hosts.')
```
adamlazik1:
```suggestion
      param :include_hosts, :bool, :required => false, :default_value => true, :desc => N_('Include hosts and template invocations in the response. Defaults to true for backwards compatibility. Pass false to skip serializing all hosts.')
```

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -84,13 +87,14 @@ def show
 
       api :POST, '/job_invocations/', N_('Create a job invocation')
       param_group :job_invocation, :as => :create
+      param :include_hosts, :bool, :required => false, :default_value => true, :desc => N_('Include hosts and template invocations in the response. Defaults to true for backwards compatibility, pass false to skip serializing all hosts.')
```
adamlazik1:
```suggestion
      param :include_hosts, :bool, :required => false, :default_value => true, :desc => N_('Include hosts and template invocations in the response. Defaults to true for backwards compatibility. Pass false to skip serializing all hosts.')
```

adamlazik1 (approved):
LGTM, works as expected. Minor wording suggestion.

adamlazik1 (approved):

adamlazik1:
Thanks @adamruzicka !

PR #1039, opened 2026-06-09
Refs #39253 - Bump requires_foreman to 5.0

pondrejk:
needed for https://github.com/theforeman/foreman_remote_execution/pull/1030

### DIFF
--- a/lib/foreman_remote_execution/plugin.rb
+++ b/lib/foreman_remote_execution/plugin.rb
@@ -1,5 +1,5 @@
 Foreman::Plugin.register :foreman_remote_execution do
-  requires_foreman '>= 3.17'
+  requires_foreman '>= 5.0'
   register_global_js_file 'global'
   register_gettext
 


ofedoren:
If the PR is still open, it should be a part of it.

pondrejk:
> If the PR is still open, it should be a part of it.

ah ok, I'll shove it there once it gets approved

ofedoren:
> ah ok, I'll shove it there once it gets approved

One of the reasons to not approve -- is to have this dependency not a part of the original :)

MariaAga:
oh no, I forgot about this.
Should this be merged now that https://github.com/theforeman/foreman_remote_execution/pull/1030 is in?

pondrejk:
> oh no, I forgot about this. Should this be merged now that #1030 is in?
I'd say it has to go in as it is a dependency, its still a separate commit anyway

MariaAga:
Can you add the redmine from the other PR to this commit?
Also is anything needed more from this pr?

ofedoren:
> oh no, I forgot about this.

That's why we usually ask the person who `require changes` on the PR if they are OK with merging/proceeding :)

In the original PR I see that some of Adam's `required changes` are not addressed, but the PR was merged: https://github.com/theforeman/foreman_remote_execution/pull/1030#pullrequestreview-4403622327.

Anyway, let's get this in, I don't think we even need a ticket for this, just use `Refs #39253 - Bump requires_foreman to 5.0` as the commit name.

pondrejk:
@ofedoren I updated the commit message. I meant this PR to address that Adam's comment, anyway I'll know better next time

ofedoren:
> @ofedoren I updated the commit message. I meant this PR to address that Adam's comment, anyway I'll know better next time

Ehh, it depends on the context. Sometimes it's better to address something in a separate PR, sometimes in a separate commit. In this case, a separate commit within that PR would be nicer.

Anyway, happens :)

ofedoren (approved):
Thanks, @pondrejk ! The JS failures are not related.

PR #1041, opened 2026-07-03
Fixes #39494 - Fix create report button is enabled when state is 'running'

Lukshio:
Bug where UI did not check for state 'running' -> button enabled instead of disabled

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationConstants.js
+++ b/webpack/JobInvocationDetail/JobInvocationConstants.js
@@ -38,6 +38,7 @@ export const STATUS = {
   SUCCEEDED: 'succeeded',
   FAILED: 'failed',
   CANCELLED: 'cancelled',
+  RUNNING: 'running',
 };
 
 export const STATUS_UPPERCASE = {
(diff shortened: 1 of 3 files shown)


MariaAga:
it seems like it will clash with https://github.com/theforeman/foreman_remote_execution/pull/1030/changes#diff-753af1c76d3524442c12c00b42c011b9170e139aadccb65a894cdd54904a0795 
@pondrejk

Lukshio:
> it seems like it will clash with https://github.com/theforeman/foreman_remote_execution/pull/1030/changes#diff-753af1c76d3524442c12c00b42c011b9170e139aadccb65a894cdd54904a0795 @pondrejk

It should be ok, the issue is the same, so this will fix it, and the #1030 PR should be fine

MariaAga:
Could you update your pr? the other one was merged

Lukshio:
@MariaAga I also replaced test with RTL store helpers in new commit

Lukshio:
Linter issues unrelated. Redmine dont like second commit

MariaAga (approved):
code looks good, ran the JS tests locally and they pass, and tested with packit that the button is disabled and enabled when needed.

PR #1042, opened 2026-07-08
Fixes #39506 - remove invalid apt-get update option

sbernhard:


app/views/templates/script/package_action.erb
```diff
@@ -127,7 +127,7 @@ handle_zypp_res_codes () {
     LANG=C apt-get -o APT::Get::Upgrade-Allow-New="true" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y <%= input("options") %> <%= action %> <%= input("package") %> 2>&1
   }
   export DEBIAN_FRONTEND=noninteractive
-  apt-get -y update
+  apt-get update
```
dosas, line 130: I disagree:

from the man page

```
      -y, --yes, --assume-yes
           Automatic yes to prompts; assume "yes" as answer to all prompts and run non-interactively. If an undesirable situation, such as changing a held package, trying to install an
           unauthenticated package or removing an essential package occurs then apt-get will abort. Configuration Item: APT::Get::Assume-Yes.
```

dosas (changes requested):

app/views/templates/script/package_action.erb
sbernhard: Isn't this for PKG install/ u install etc.?

sbernhard: Actually, I had the same impression than you. I had a look at the apt-get source and asked AI - the opton “-y” is not yet used in apt-get update. BUT, its is not YET used. Maybe somewhere in the future. It does not hurt right now - so, let is as it yes.

I will close this PR - Dont fix whats not broken!

PR #1044, opened 2026-07-16
Fixes #39533 - "View task" disabled when job is running

MariaAga:
`include_permissions` wasnt present in every api call when job is running, causing the view task button to be disabled as it didnt have permissions data.
Also just removed the usage of `include_permissions` api requests for permissions, as all permissions are loaded to the foreman metadata.

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationActions.js
+++ b/webpack/JobInvocationDetail/JobInvocationActions.js
@@ -19,7 +19,7 @@ export const getJobInvocation = url => dispatch => {
   const fetchData = withInterval(
     get({
       key: JOB_INVOCATION_KEY,
-      params: { include_permissions: true, include_hosts: false },
+      params: { include_hosts: false },
       url,
       handleError: () => {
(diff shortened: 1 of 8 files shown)


kmalyjur (changes requested):
Works as it should.

In `webpack/JobInvocationDetail/__tests__/TemplateInvocation.test.js`, lines 198-202 (permissions) are now unused and could be deleted.

In `webpack/JobInvocationDetail/__tests__/fixtures.js`, the `export const mockPermissionsData` seems to be unused, too.

MariaAga:
Thanks! removed it, js tests are passing locally

kmalyjur:
@MariaAga Thank you. Additionally, I see that when I have a user with only one permission `view_job_invocations`, on master the "View task" button is disabled, but on this branch it's enabled. I think that it's not intended, right?

MariaAga:
I cant reproduce this issue anymore so closing. Couldnt see anything that would have fixed it recently, so not sure if my env just had issues.

PR #1043, opened 2026-07-13
Fixes #39522 - Fix failing new lint rules

MariaAga:
Fix for https://github.com/theforeman/foreman/pull/11008 https://github.com/theforeman/foreman/pull/11098 https://github.com/theforeman/foreman/pull/10968

PR #1045, opened 2026-07-16
Fixes #39352 - Job inv page ignores remote_execution_job_invocation_report_template setting

kmalyjur:
New job invocation page hardcoded `'Job - Invocation Report'` instead of reading the `remote_execution_job_invocation_report_template` setting. This fixes it by fetching the setting value via `/api/settings/remote_execution_job_invocation_report_template` and using it to search for the template. Falls back to the default if the setting fetch fails.


adamruzicka:
Could we leave the lint fixing to https://github.com/theforeman/foreman_remote_execution/pull/1043 ? Right now both PRs try to fix that, but with subtle differences.

webpack/JobInvocationDetail/JobInvocationToolbarButtons.js
```diff
@@ -80,22 +81,37 @@ const JobInvocationToolbarButtons = ({ jobId, data }) => {
...
-        url: '/api/report_templates',
-        handleSuccess: ({ data: { results } }) => {
-          if (isMounted) {
-            setReportTemplateJobId(
-              results.find(result => result.name === 'Job - Invocation Report')
-                ?.id
-            );
-          }
+        key: GET_REPORT_TEMPLATE_SETTING,
+        url: '/api/settings/remote_execution_job_invocation_report_template',
```
Lukshio: I am not sure, but hopefully @adamruzicka will check that.

I think that user needs `view_settings` permission to access this endpoint, so basically all "normal" users won't be able to get the response. So we should change backend to return correct template so `generate_report_templates` permission will be sufficient

adamruzicka: > I think that user needs view_settings permission to access this endpoint, so basically all "normal" users won't be able to get the response

That's right, good catch. 

> So we should change backend to return correct template

Return correct template _from where_?

Lukshio: I meant adding something like this to `job_invocations/base.json.rabl`

```
if params.key?(:include_permissions)
  node :report_template do |invocation|
    next unless User.current&.can?(:generate_report_templates)
    template = job_report_template
    next unless template
    job_id_input = template.template_inputs.find_by(name: 'job_id')
    next unless job_id_input
    {
      id: template.id,
      job_id_input_id: job_id_input.id,
    }
  end
end
```

adamruzicka: Adding what we need to a place where it would be convenient is tempting, but it doesn't really belong there.

What if we introduced a new api endpoint like `/api/v2/job_invocations/:id/report` which would under the hood render the right template from settings, without ever leaking the setting value to the frontend? Frontend would just check if the user has the generate_report_templates permission, clicking the button would call that endpoint. There would still be some slightly rougher edges, but probably still better than bypassing permissions.

Lukshio: Yup, I think we can do it like that, do you agree @kmalyjur?

kmalyjur:
@Lukshio @adamruzicka Thank you, I changed it as you suggested

test/functional/api/v2/job_invocations_controller_test.rb
Lukshio, line 1: Could you please add some tests for `generate_report_templates` permissions?

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -215,6 +215,21 @@ def outputs
...
+        template = job_report_template
+        unless template
+          return not_found(_('Report template not found or not configured properly'))
+        end
+
+        unless ReportTemplate.authorized(:generate_report_templates).find_by(:id => template.id)
+          return deny_access(_('Missing permissions to generate report templates'))
+        end
+
+        redirect_to generate_report_template_path(template, job_report_template_parameters(@job_invocation, template))
```
Lukshio: should this be in API controller? I think that it should be handled by frontend. @adamruzicka

kmalyjur: You're right, I changed it.

adamruzicka: If I'm reading it correctly, then:
1) When the button is clicked an API call is made to the API controller a returns a different url
2) Once the request is processed, the frontend navigates to the returned url, effectively performing a client-assisted redirect

That's not exactly what I had in mind. My original idea was that you click the button and you get a report back, no questions asked. However, looking now at how it worked before (it takes you through the report generation form with some values pre-filled), that would be a change in behaviour. Sorry for not realizing this before

So either we can:
1. Do what I originally suggested, changing behaviour.
2. Accept that a template ID will be leaked even if the user doesn't have permission to view settings and:
    1. Implement some sort of redirect mechanism (like something that is currently there)
    2. Start introducing something like [HAL](https://en.wikipedia.org/wiki/Hypertext_Application_Language) to our api responses
3. Be very strict about permissions and just don't show the button for people who don't have the necessary permission. This is probably unnecessarily strict, but an option nevertheless.

Personally I'm rather fond of 2.2, but that is a new pattern that should be ideally rolled out in a more systematic way, leaving us with 1, 2.1 and 3. Out of those, I'd be fine with either, but since 2.1 is mostly there, I'd lean towards it.

> should this be in API controller?

I would say a regular, non api-controller would be a better place. You navigate to a specific url, it fills in the gaps and redirects you to the right place. That is arguably cleaner than calling an api, getting an url back and navigating to it yourself.

Lukshio: I did not realize that on a first look, and I also meant the option n. 1. Option 3 would be some regression, option 2 could lead to some sec issues :/

kmalyjur: @adamruzicka Thank you for laying out all the options. I'm going with 2.1, so I'll move the redirect to a controller and keep the button as a link.

test/functional/job_invocations_controller_test.rb
```diff
@@ -60,6 +60,51 @@ class JobInvocationsControllerTest < ActionController::TestCase
     assert_response :success
   end
 
+  context '#report' do
+    setup do
+      @invocation = FactoryBot.create(:job_invocation, :with_template, :with_task)
+      @report_template = FactoryBot.create(:report_template, :name => 'Job - Invocation Report', :template => '<%= "report output" %>')
```
Lukshio, line 66: should we also test some non-default job name?

test/functional/job_invocations_controller_test.rb
```diff
@@ -60,6 +60,51 @@ class JobInvocationsControllerTest < ActionController::TestCase
...
+      @report_template.template_inputs.create!(:name => 'job_id', :input_type => 'user')
+      Setting['remote_execution_job_invocation_report_template'] = @report_template.name
+    end
+
+    test 'should redirect to report generation page' do
+      get :report, params: { :id => @invocation.id }, session: set_session_user
+      assert_response :redirect
```
Lukshio: Does this mean every redirect, so it don't test where the user is redirected?

Lukshio (commented):
Rebase should resolve failing CI

test/functional/job_invocations_controller_test.rb
kmalyjur: You're right, I added `assert_redirected_to`

adamruzicka (commented):
I won't have the capacity to take this for a spin, but from reading the code it looks good. If it works, ship it

Lukshio:
ack, manually tested and verified

PR #1046, opened 2026-07-23
Refs #39253 - Add explicit polling to host table

adamlazik1:
The host table previously relied on an accidental refresh cascade: parent polling updated Redux state, which changed callback identities, which triggered useEffect, which re-fetched hosts. This adds an explicit setTimeout-based polling chain (5s interval) that stops when the job finishes or on API error, and restarts when filters change.


adamlazik1:
Accidentally opened.

PR #1048, opened 2026-07-27
Refs #39253 - perf 2/4 - Replace interval polling with setTimeout chains 

adamruzicka:
Follows https://github.com/theforeman/foreman_remote_execution/pull/1047

TODOs:
- [x] pull in missing tests from the "big PR"

PR #1049, opened 2026-07-27
Refs #39253 - perf 3/4 - Consolidate host data into the API endpoint

adamruzicka:
Follows https://github.com/theforeman/foreman_remote_execution/pull/1048

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -117,12 +117,14 @@ module Api
       param_group :search_and_pagination, ::Api::V2::BaseController
       add_scoped_search_description_for(JobInvocation)
       param :id, :identifier, :required => true
+      param :include_permissions, :bool, :required => false, :desc => N_('Include per-host task and permission data in the response')
       def hosts
         set_hosts_and_template_invocations
         @total = @hosts.size
         @hosts = @hosts.search_for(params[:search], :order => params[:order]).paginate(:page => params[:page], :per_page => params[:per_page])
         @subtotal = @hosts.total_entries
         set_statuses_and_smart_proxies
+        set_tasks_and_permissions if Foreman::Cast.to_bool(params[:include_permissions])
         if params[:awaiting]
           @hosts = @hosts.select { |host| @host_statuses[host.id] == 'N/A' }
         end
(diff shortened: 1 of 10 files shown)


adamlazik1:
LGTM

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -321,15 +323,35 @@ def set_statuses_and_smart_proxies
...
+          template_invocation = @template_invocations_by_host_id[host.id]
           task = template_invocation.try(:run_host_job_task)
           [host.id, template_invocation_status(task, @job_invocation.task)]
         end
         @smart_proxy_id = template_invocations.to_h { |ti| [ti.host_id, ti.smart_proxy_id] }
         @smart_proxy_name = template_invocations.to_h { |ti| [ti.host_id, ti.smart_proxy_name] }
       end
+
+      def set_tasks_and_permissions
+        can_view_tasks = User.current.can?(:view_foreman_tasks)
```
adamlazik1: I missed this previously but this was scoped by object in the old version, it should probably be scoped here too.

adamruzicka: Somehow it ended up in part 4[1]. I'll pull it back.

[1] - https://github.com/theforeman/foreman_remote_execution/pull/1050/changes/dc32c1da27ac1b55c1e5cf9519324435b259b7f9#diff-9aace261f6f5fbcd07b27afe2dda1549112ae9b0b98997d4ec9b0b0ef7b842c1R349

adamruzicka: pulled it back

app/controllers/api/v2/job_invocations_controller.rb
```diff
@@ -321,15 +323,35 @@ def set_statuses_and_smart_proxies
...
+
+        hosts = @hosts.to_a
+        @task_by_host = hosts.to_h do |host|
+          task = @template_invocations_by_host_id[host.id].try(:run_host_job_task)
+          [host.id, task ? { :id => task.id, :cancellable => task.try(:cancellable?) || false } : nil]
+        end
+        @permissions_by_host = hosts.to_h do |host|
+          task = @template_invocations_by_host_id[host.id].try(:run_host_job_task)
+          [host.id, {
+            :view_foreman_tasks => task.present? && host_tasks_authorizer.can?(:view_foreman_tasks, task),
```
adamruzicka: wait, what

app/controllers/api/v2/job_invocations_controller.rb
adamlazik1: task needs to be a task record instead of a plain hash, otherwise it crashes silently and :view_foreman_tasks is false for all non admin users:

```
bundle exec rake console
user = User.unscoped.where(admin: false).first
user.login
authorizer = Authorizer.new(user)
authorizer.can?(:view_foreman_tasks, { id: 1, cancellable: true })
NoMethodError: undefined method `allows_taxonomy_filtering?' for Hash:Class
from /home/vagrant/foreman/app/services/authorizer.rb:172:in `allowed_taxonomies'
```

adamruzicka: /o\ changed

adamlazik1 (approved):
LGTM, I have no idea where the test failures are coming from. Try squashing to see if re-triggering the tests fixes the issue.

adamruzicka:
They did go away after a rerun

adamlazik1 (approved):

adamlazik1:
Thanks @adamruzicka!

PR #1050, opened 2026-07-27
Refs #39253 - perf 4/4 - Fetch permissions only on initial load

adamruzicka:
Follows https://github.com/theforeman/foreman_remote_execution/pull/1049

TODOs:
- [x] make the permission checks task-specific

### DIFF
--- a/webpack/JobInvocationDetail/index.js
+++ b/webpack/JobInvocationDetail/index.js
@@ -54,6 +54,11 @@ const JobInvocationDetailPage = ({
   } = items;
   const finished = isJobFinished(statusLabel);
   const pollTimeoutRef = useRef({ timeoutId: null, cancel: () => {} });
+  const permissionsRef = useRef(null);
+  if (items.permissions && !permissionsRef.current) {
+    permissionsRef.current = items.permissions;
+  }
+  const permissions = items.permissions || permissionsRef.current;
   const jobInvocationApiStatus = useSelector(state =>
     selectAPIStatus(state, JOB_INVOCATION_KEY)
   );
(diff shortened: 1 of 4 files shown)


adamlazik1 (commented):
LGTM

webpack/JobInvocationDetail/index.js
```diff
@@ -54,6 +54,11 @@ const JobInvocationDetailPage = ({
   } = items;
   const finished = isJobFinished(statusLabel);
   const pollTimeoutRef = useRef({ timeoutId: null, cancel: () => {} });
+  const permissionsRef = useRef(null);
+  if (items.permissions && !permissionsRef.current) {
+    permissionsRef.current = items.permissions;
```
adamlazik1, line 59: ref should be reset when job id changes, because switching via breadcrumb switcher keeps the cached permissions from the first job.

adamruzicka: What I'm seeing on my end is that when you switch through breadcrumb switcher, a "first fetch" request is kicked off, which does then replace the cached permissions with the newly fetched ones, similar in principle to what happens when you navigate to a different page in the table. I mean I could go and reset it explicitly, but that doesn't seem to be necessary

adamlazik1: It replaces cached permissions on first poll, but then uses cached permissions on subsequent polls, no?

adamruzicka: > It replaces cached permissions on first poll, but then uses cached permissions on subsequent polls,

That's true, but incomplete. It replaces cached permissions on first poll, but "first" doesn't mean "first ever". A new "first poll" is fired on every page, filter or job change.

So I'd go with something like "then uses cached permissions on subsequent polls _within the same context_"

adamlazik1: Ah, nevermind me, sorry

adamlazik1 (approved):

adamlazik1:
Thanks @adamruzicka !

PR #1051, opened 2026-07-27
Fixes #39557 - replace test_setup with the foreman core test setup

andreilakatos:
Fixes #39557 - replace test_setup with the foreman core test setup

### DIFF
--- a/webpack/test_setup.js
+++ b/webpack/test_setup.js
@@ -1,13 +1,1 @@
-import 'core-js/shim';
-import 'regenerator-runtime/runtime';
-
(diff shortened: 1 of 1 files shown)


adamruzicka (approved):
Looks like the only problem are the lints, which isn't new.

andreilakatos:
> Looks like the only problem are the lints, which isn't new.

Thanks, @MariaAga has a PR here for fixing those:
https://github.com/theforeman/foreman_remote_execution/pull/1043

adamruzicka:
Thank you @andreilakatos !

PR #1052, opened 2026-07-30
Fixes #39543 - Add alert to locked Job template form

michalgritzbach:
Adds _alert.html.erb view to Job templates, bringing it on par with other Template resources. This shows a warning alert when trying to edit locked resource (same as the other templates).


adamruzicka (approved):
Not much to add here, this is in line with what we do everywhere else.

adamruzicka:
Thank you @michalgritzbach !

PR #1055, opened 2026-08-12
Fixes #39609 - Add missing ouiaId to MenuToggle components

adamruzicka:
Strict checking of ouia ids on PF5 components was introduced with [#39562](https://projects.theforeman.org/issues/39562) / theforeman/foreman#11117.


MariaAga (approved):
lgtm, thanks!

PR #1056, opened 2026-08-17
Fixes #39636 - Restrict job template revision endpoint to JobTemplate audits

anupamme:
## Summary

The `revision` endpoint is specifically for JobTemplate revision history, but the original `find(params[:version])` didn't constrain the audit's `auditable_type`. This meant a valid audit ID for a different model type (e.g. Host) would resolve successfully even though the endpoint should only serve JobTemplate audits.

This change makes that invariant explicit:

```ruby
# before
audit = Audit.authorized(:view_audit_logs).find(params[:version])

# after
audit = Audit.authorized(:view_audit_logs)
             .where(:auditable_type => 'JobTemplate', :id => params[:version])
             .first!
```

The existing `Audit.authorized(:view_audit_logs)` scope remains responsible for access control. This is a correctness fix, not a security issue.

## Changes

- `app/controllers/api/v2/job_templates_controller.rb` — scope revision lookup to `auditable_type: 'JobTemplate'`
- `test/functional/api/v2/job_templates_controller_test.rb` — add tests for the `#revision` action:
  - Positive case: authorized JobTemplate audit resolves correctly
  - Negative case: authorized audit for a different `auditable_type` returns 404

### DIFF
--- a/app/controllers/api/v2/job_templates_controller.rb
+++ b/app/controllers/api/v2/job_templates_controller.rb
@@ -88,7 +88,7 @@ module Api
       api :GET, '/job_templates/revision'
       param :version, String, :desc => N_('Template version')
       def revision
-        audit = Audit.authorized(:view_audit_logs).find(params[:version])
+        audit = Audit.authorized(:view_audit_logs).where(:auditable_type => 'JobTemplate', :id => params[:version]).first!
         render :json => audit.revision.template
       end
 
--- a/test/functional/api/v2/job_templates_controller_test.rb
+++ b/test/functional/api/v2/job_templates_controller_test.rb
@@ -133,6 +133,27 @@ module Api
         assert_response :success
         assert JobTemplate.unscoped.find_by(name: new_name)
       end
+
+      context '#revision' do
+        test 'should return revision for a job template audit' do
+          @template.update!(template: 'updated template content')
+          audit = @template.audits.last
+          get :revision, params: { :version => audit.id }
+          assert_response :success
+          assert_equal 'updated template content', ActiveSupport::JSON.decode(@response.body)
+        end
+
+        test 'should not return revision for a non-job-template audit' do
+          other_audit = Audited::Audit.create!(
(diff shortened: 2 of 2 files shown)


adamruzicka:
This is not a _high severity_ issue. Yes, it uses `find(...)` with user-supplied value, but it is called on a scope which already enforces that the user in question has to have permissions to view audit records. So even if they build and "attack", they would only be able to trick the application into showing them something they can already see.

With that being said, this could be viewed as a correctness fix to prevent the "tricking" from being possible.

On a more general note, this project uses redmine (https://projects.theforeman.org/projects/foreman_remote_execution) for issue tracking and commit messages have to be shaped like `Fixes #<issue-id> - <description>`.

anupamme:
Thanks, that's a fair point. I agree this isn't a high-severity authorization issue — \`Audit.authorized(:view_audit_logs)\` already enforces the relevant permissions.

I've reframed the PR as a correctness fix: the endpoint is specifically for JobTemplate revisions, but \`find(params[:version])\` didn't constrain \`auditable_type\`, so it could resolve an audit belonging to a different model type. The updated query makes that invariant explicit.

I've also:
- Added tests covering both the positive case (JobTemplate audit → 200) and the negative case (authorized audit for a different \`auditable_type\` → 404)
- Removed all HIGH/IDOR/vulnerability language from the PR description
- Updated the commit message — I didn't open a Redmine issue since I wasn't sure if that's expected for a change of this scope. Happy to create one if you'd prefer the standard \`Fixes #NNNNN\` format.

adamruzicka:
> I didn't open a Redmine issue since I wasn't sure if that's expected for a change of this scope. Happy to create one if you'd prefer the standard `Fixes #NNNNN` format.

Please do

anupamme:
> > I didn't open a Redmine issue since I wasn't sure if that's expected for a change of this scope. Happy to create one if you'd prefer the standard `Fixes #NNNNN` format.
> 
> Please do

done.

PR #1058, opened 2026-08-18
Fixes #39251 - Load job wizard host previews via POST

jakduch:
When rerunning jobs with many target hosts, the generated search query can exceed the maximum URI length. Both the target selection and review steps loaded host previews through `GET /api/hosts`, causing HTTP 414 before Foreman could process the request.

Send host preview searches through a POST endpoint backed by the existing hosts index. This keeps the current authorization, pagination, and response format while carrying the search query in the request body.

The final job submission already uses POST and is unchanged.

Tests:
- long host searches are sent in the POST body
- the host preview route accepts POST and targets the existing hosts index

This pull request was assisted by Codex 5.6 Sol High.


jakduch:
Could someone please move #39251 from the Foreman project to Foreman Remote Execution? The fix belongs to this plugin, and the Redmine check currently fails only because the issue is filed under the wrong project. Thanks!

PR #1059, opened 2026-08-18
Fixes #39648 - Improve live job output updates

jakduch:
The output for an expanded host is currently refreshed every five seconds by fetching the complete template invocation details. This repeatedly transfers the full accumulated output together with template, task, proxy, and permission data, then parses and renders the entire console again.

Use the existing per-host output API with the `since` parameter to request only chunks newer than the last displayed timestamp. Poll once per second while the output is open, append new chunks while preserving split lines, pause requests in hidden tabs, and refresh the complete invocation details once after the host finishes.

Keep existing output line sets stable and memoize their parsed ANSI output so long-running jobs only process newly appended content.

Tests:
- incremental requests use the latest displayed timestamp
- new output is appended and split lines are preserved
- polling stops after completion, collapse, or unmount
- hidden tabs pause output requests
- existing output rendering and ANSI colors remain unchanged

This pull request was assisted by Codex 5.6 Sol High.

### DIFF
--- a/webpack/JobInvocationDetail/useTemplateInvocationOutputPolling.js
+++ b/webpack/JobInvocationDetail/useTemplateInvocationOutputPolling.js
@@ -1,0 +1,160 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { useDispatch } from 'react-redux';
+import { APIActions } from 'foremanReact/redux/API';
+import {
+  GET_TEMPLATE_INVOCATION,
+  GET_TEMPLATE_INVOCATION_OUTPUT,
+  OUTPUT_MAX_REFRESH_INTERVAL_MS,
+  OUTPUT_REFRESH_INTERVAL_MS,
+  showTemplateInvocationUrl,
+  templateInvocationOutputUrl,
+} from './JobInvocationConstants';
+import {
+  getLastOutputTimestamp,
+  mergeOutput,
+} from './TemplateInvocationHelpers';
+
+export const useTemplateInvocationOutputPolling = ({
+  hostID,
+  jobID,
+  isExpanded,
+  response,
+}) => {
+  const dispatch = useDispatch();
+  const responseOutput = response?.output;
+  const responseFinished = response?.finished || false;
+  const hostFinishedRef = useRef(response?.finished || false);
+  const [liveOutput, setLiveOutput] = useState(responseOutput || []);
+  const lastTimestampRef = useRef(getLastOutputTimestamp(responseOutput || []));
+  const hasCachedOutput = Array.isArray(responseOutput);
+
+  const replaceOutput = useCallback(output => {
+    const nextOutput = Array.isArray(output) ? output : [];
+    lastTimestampRef.current = getLastOutputTimestamp(nextOutput);
+    setLiveOutput(nextOutput);
+  }, []);
+
+  const appendOutput = useCallback(output => {
+    const nextOutput = Array.isArray(output) ? output : [];
+    const lastTimestamp = getLastOutputTimestamp(nextOutput);
+    if (lastTimestamp !== null && lastTimestamp !== undefined) {
+      lastTimestampRef.current = lastTimestamp;
+    }
+    setLiveOutput(currentOutput => mergeOutput(currentOutput, nextOutput));
+  }, []);
+
+  useEffect(() => {
+    if (Array.isArray(responseOutput)) replaceOutput(responseOutput);
+    if (responseFinished) hostFinishedRef.current = true;
+  }, [replaceOutput, responseFinished, responseOutput]);
+
+  useEffect(() => {
+    let cancelled = false;
+    let timeoutId = null;
+    let idlePolls = 0;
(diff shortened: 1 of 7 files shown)


webpack/JobInvocationDetail/useTemplateInvocationOutputPolling.js
```diff
@@ -0,0 +1,149 @@
...
+  OUTPUT_REFRESH_INTERVAL_MS,
+  showTemplateInvocationUrl,
+  templateInvocationOutputUrl,
+} from './JobInvocationConstants';
+import {
+  getLastOutputTimestamp,
+  mergeOutput,
+} from './TemplateInvocationHelpers';
+
+export const useTemplateInvocationOutputPolling = ({
```
kmalyjur, line 17: This hook has a lot of state management: 5 refs/state vars, useEffect just to sync props - local state, two similar fetch functions (scheduleOutputPoll, fetchDetails) etc. It could be simplified. What do you think?

The suggestion:

```
export const useTemplateInvocationOutputPolling = ({
  hostID,
  jobID,
  isExpanded,
  initialOutput = [],
}) => {
  const dispatch = useDispatch();
  const [output, setOutput] = useState(initialOutput);
  const lastTimestampRef = useRef(getLastOutputTimestamp(initialOutput));

  useEffect(() => {
    if (!isExpanded) return;

    let cancelled = false;
    let timeoutId = null;
    const outputURL = templateInvocationOutputUrl(hostID, jobID);

    const poll = () => {
      if (cancelled) return;

      // Pause when tab hidden
      if (document.visibilityState === 'hidden') {
        timeoutId = setTimeout(poll, OUTPUT_REFRESH_INTERVAL_MS);
        return;
      }

      dispatch(
        APIActions.get({
          url: outputURL,
          key: `${GET_TEMPLATE_INVOCATION_OUTPUT}_${hostID}`,
          params: lastTimestampRef.current ? { since: lastTimestampRef.current } : {},
          handleSuccess: ({ data }) => {
            if (cancelled) return;

            const newChunks = data?.output || [];
            if (newChunks.length > 0) {
              setOutput(prev => {
                const merged = mergeOutput(prev, newChunks);
                lastTimestampRef.current = getLastOutputTimestamp(merged);
                return merged;
              });
            }

            // Continue polling or fetch final details
            if (data?.refresh) {
              timeoutId = setTimeout(poll, OUTPUT_REFRESH_INTERVAL_MS);
            } else {
              // Host finished - fetch complete details once
              dispatch(
                APIActions.get({
                  url: showTemplateInvocationUrl(hostID, jobID),
                  key: `${GET_TEMPLATE_INVOCATION}_${hostID}`,
                })
              );
            }
          },
          handleError: () => {
            if (cancelled) return;
          },
        })
      );
    };

    poll();

    return () => {
      cancelled = true;
      clearTimeout(timeoutId);
    };
  }, [isExpanded, hostID, jobID, dispatch]);

  return output;
};
```

kmalyjur (changes requested):
Thank you, that's a good idea and it works well. I have some comments.

Right now it polls every one sec, but there could be some kind of "adaptive polling" to reduce the load when the job is running but not producing output.

Simplified "solution" to that:

```
// In handleSuccess:
  const nextInterval = newChunks.length > 0
    ? OUTPUT_REFRESH_INTERVAL_MS      // Got output, stay fast (1s)
    : OUTPUT_REFRESH_INTERVAL_MS * 3; // Idle, slow down (3s)

  if (data?.refresh) {
    timeoutId = setTimeout(poll, nextInterval);
  }
```

webpack/JobInvocationDetail/useTemplateInvocationOutputPolling.js
jakduch: I simplified the hook while keeping the initial/final details refresh and terminal-state handling. Polling now gradually backs off from 1s to 3s while idle and returns to 1s as soon as new output arrives. I added coverage for both cases in 21a15f9.

adamruzicka:
AI-generated review.

### Critical — must fix before merge

No critical findings.

### Important — should fix before merge

**[A]** Merging a partial chunk with the next chunk can misclassify stderr as stdout (or vice versa), hiding output when filtered; only join matching types. `webpack/JobInvocationDetail/TemplateInvocationHelpers.js:12`

**[B]** When the first new chunk contains only the line completion, its remainder is empty but still rendered as an extra blank line. `webpack/JobInvocationDetail/TemplateInvocationHelpers.js:19`

**[C]** Assert that a later poll uses the timestamp of newly appended output; current coverage only checks the initial cursor. `webpack/JobInvocationDetail/__tests__/TemplateInvocationPolling.test.js:89`

**[D]** Cover failed output/details requests and verify the intended polling behavior after an error. `webpack/JobInvocationDetail/useTemplateInvocationOutputPolling.js:96`

**[E]** Copy the output array and replace only the split chunk; cloning every new chunk is unnecessary. `webpack/JobInvocationDetail/TemplateInvocationHelpers.js:8`

### Suggestions — nice to have

No suggestions.

PR #1060, opened 2026-08-19
Fixes #39651 - Fix host table not updating during job execution

kmalyjur:
The host table on the job Invocation detail page didn't update while a job was running - it stayed frozen, showing "pending" even after the hosts finished.
The server now includes "auto_refresh": true/false in the API response. The client checks this field and stops polling when it becomes false.

### DIFF
--- a/app/controllers/api/v2/job_invocations_controller.rb
+++ b/app/controllers/api/v2/job_invocations_controller.rb
@@ -118,6 +118,11 @@ module Api
       add_scoped_search_description_for(JobInvocation)
       param :id, :identifier, :required => true
       param :include_permissions, :bool, :required => false, :desc => N_('Include per-host task and permission data in the response')
+      description <<-DOC
+        Returns standard index metadata with an additional auto_refresh field.
+        The auto_refresh field indicates whether the job is still running and
+        the client should continue polling for updates.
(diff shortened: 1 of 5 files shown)


adamruzicka:
> The host table on the job Invocation detail page didn't update while a job was running - it stayed frozen, showing "pending" even after the hosts finished.

Are you sure about that? Seems to work just fine on my end.

adamruzicka:
I mean, the original issue (https://redhat.atlassian.net/browse/SAT-45960) was opened before the perf fixes landed which reworked the entirety of polling so I would expect this to not be needed anymore.

kmalyjur:
@adamruzicka try it on [deleted] and you'll see it. I think it was created after the performance fixes, but I'll double check.

adamruzicka:
That machine still has `rubygem-foreman_remote_execution-16.7.0`. Due to the polling rework being split into 4 parts, only the first two got merged in time to land in 16.7.0. Everything should be there with 16.7.1, which hit some sort of a snag in packaging and was delayed a bit.

kmalyjur:
@adamruzicka thank you for your comments. You're right that it works after the performance fixes, so I'm closing it.

PR #1067, opened 2026-09-07
Fixes #39742 - Add allowed_nested_id to ForeignInputSetsController

ofedoren:
https://github.com/theforeman/foreman/pull/11215 had a change that caused REX CI turn red. Previously we treated any `<resource>_id` in params as a parent resource, with that change we check only ones listed in `allowed_nested_id` override. Currently we don't have such override for foreign input sets controller, thus we ignore the parent id.


adamruzicka (approved):
lgtm

adamruzicka:
Thank you @ofedoren !

PR #1068, opened 2026-09-10
Fixes #39754 - Debug output in job invocation details UI is displayed in red

Thorben-D:
Before this commit, job output of type "debug" was displayed in red in the JobInvocationDetail component.

Coloring guidelines suggest reserving the color red for errors and to indicate danger.

In this commit, the color used to display debug output is changed to `--pf-v5-global--info-color--100`, which is the default color used to present informative content.

**Before:**
image

**After:**
image

### DIFF
--- a/webpack/JobInvocationDetail/JobInvocationDetail.scss
+++ b/webpack/JobInvocationDetail/JobInvocationDetail.scss
@@ -112,9 +112,12 @@ section.job-additional-info {
     }
 
     div.line.stderr,
-    div.line.error,
-    div.line.debug {
+    div.line.error {
       color: red;
+    }
+
+    div.line.debug {
+      color: var(--pf-v5-global--info-color--100)
     }
 
     div.line span.counter {


nadjaheitmann:
Thanks @Thorben-D , I think that this makes a lot of sense. @MariaAga , what is your opinion?

MariaAga (approved):
Logically this change makes sense to me, and I tested it with packit to see that it works ok.
What bothers me a bit is that errors are also shown as info (debug) it seems, and I wonder why does rex do that?

Image

adamruzicka:
> What bothers me a bit is that errors are also shown as info (debug) it seems, and I wonder why does rex do that?

Additionaly messages emitted "around the execution" are injected into the output as debug statements. `stdout` and `stderr` are strictly for things emitted by the running processes and adding to those would be wrong.

Thorben-D:
@adamruzicka 
Hmm, given that, I would probably refrain from merging this...
I guess you'd have to introduce a new `output_type` to fix this properly. That should not be impossible, but I can very much live with red debug output too, so I'd let that be your call.

adamruzicka:
Could you point me at where the messages like `START: Phase run_ansible` come from?

Thorben-D:
> Could you point me at where the messages like `START: Phase run_ansible` come from?

Certainly! They come from [here](https://github.com/ATIX-AG/smart_proxy_ansible_director/blob/0ff107b640eb83f3dd62367a4c90c238e4a3196d/lib/smart_proxy_ansible_director/runners/meta_runner.rb#L62).
This is a bit unorthodox, but basically the thing needs to call two programs in sequence (`ansible-builder` then `ansible-navigator`) and I wanted the output of both to be shown in Foreman. Perhaps there is a better way to do this, but that's what I came up with. The debug output is injected there to keep track of the two "phases" of the one job.

adamruzicka:
Looking at it, adding a new output_type sounds like the best way to go. To be honest, the current stdout/stderr/debug is a relatively narrow channel and in the past I considered adding some more to be able to inject additional "control" messages, so I wouldn't be opposed to that idea at all.

PR #1069, opened 2026-09-15
Fixes #39630 - REX Job Wizard host autofill when display_fqdn_for_host setting is disabled

snarya07:
## Summary

Fixes #39630 — REX Job Wizard shows "0 selected hosts" when triggered from Hosts Overview page with `display_fqdn_for_hosts` settings is disabled

### Root cause

When scheduling a job from host Overview with `display_fqdn_for_hosts` disabled, `autofill.js` stored the short `display_name` in the `name` property used by `buildHostQuery`. That produced a `name ^ (shortname)` search against the FQDN stored in the database, so the host preview showed 0 hosts even though the host was fetched by ID. 
- The fix keeps name as the canonical FQDN for queries and `displayName` for UI labels, and resolves non-numeric host_ids[] with `name ^ (hostname)` instead of `id = hostname`.

### Fix

- Add `buildHostSearchFromIds()` for correct API search by ID or hostname
- Store `name` (canonical FQDN) and `displayName` (UI label) separately in autofill host targets

## Files changed

- `webpack/JobWizard/autofill.js`
- `webpack/JobWizard/steps/HostsAndInputs/__tests__/HostsAndInputs.test.js`

## Test plan

- [x] `cd ~/foreman && npm run test:plugins foreman_remote_execution -- webpack/JobWizard/steps/HostsAndInputs/__tests__/HostsAndInputs.test.js`
- [x] Manual: `display_fqdn_for_hosts = No` → Host Overview → Schedule a job → 1 host selected
- [x] Manual: `display_fqdn_for_hosts = Yes` → same flow, no regression

Assisted by Cursor


webpack/JobWizard/autofill.js
```diff
@@ -11,6 +11,21 @@ import {
...
 import './JobWizard.scss';
 
+
+export const buildHostSearchFromIds = hostIds => {
+  const ids = Array.isArray(hostIds) ? hostIds : [hostIds];
+  const numericIds = ids.filter(id => /^\d+$/.test(String(id)));
+  const hostnames = ids.filter(id => !/^\d+$/.test(String(id)));
+  const parts = [];
+  if (numericIds.length) {
+    parts.push(`id = ${numericIds.join(' or id = ')}`);
```
adamruzicka: Wouldn't this work?
```suggestion
    parts.push(`id ^ (${numericIds.join(',')})`);
```

webpack/JobWizard/autofill.js
```diff
@@ -11,6 +11,21 @@ import {
 import { selectHostBookmarks } from './JobWizardSelectors';
 import './JobWizard.scss';
 
+
+export const buildHostSearchFromIds = hostIds => {
+  const ids = Array.isArray(hostIds) ? hostIds : [hostIds];
+  const numericIds = ids.filter(id => /^\d+$/.test(String(id)));
+  const hostnames = ids.filter(id => !/^\d+$/.test(String(id)));
```
adamruzicka: Could we declare the regular expression only once?

webpack/JobWizard/autofill.js
snarya07: Thanks for the suggetions @adamruzicka , pushed the changes in recent commit

adamruzicka:
And in addition to what copilot suggested, could we also get some more tests?

